PR #5, opened 2026-04-19
feat(database): define user tables

muhlemmer:
This PR introduces a high-performance **Entity-Attribute-Value (EAV)** storage model designed for Zitadel NextGen. The architecture addresses the specific challenges of multi-tenancy (B2B2C) and massive scalability by decoupling data storage from uniqueness enforcement.

## 1. The Storage Architecture
The model is split into three primary table sets to optimize for different access patterns:

* **`users` (Header):** Stores core identity metadata.
* **`user_attributes` (Data Store):** Stores flattened JSONB attributes. 
    * *Strategy:* Partitioned by `HASH (instance_id, user_id)`.
    * *Benefit:* This ensures that a user’s header and all their attributes are **physically co-located** on the same disk partitions, enabling high-speed joins.
* **`user_unique_attributes` (Uniqueness Registry):** A lean "lock" table for attributes requiring uniqueness (e.g., email, username).
    * *Strategy:* Partitioned by `HASH (instance_id, key)`.
    * *Benefit:* Enables global or organization-scoped uniqueness checks across the entire cluster without scanning all data partitions.

A relevant **ADR** has been included in this PR.

## 2. Key Technical Decisions
* **Binary Hashing (`BYTEA`):** Unique constraints are enforced on MD5 hashes (planned transition to SHA-256) of JSONB values. Storing raw bytes instead of hex strings reduces index size by 50%.
* **Composite Types & Binary Protocol:** We utilize a custom PostgreSQL type `incoming_user_attribute` and the `pgx` binary protocol to pass attribute arrays in a single round-trip.
* **Atomic CTE Writes:** User creation is handled via a single PostgreSQL statement using Common Table Expressions (CTEs), ensuring all-or-nothing atomicity for the header, attributes, and registry entries.
* **Snapshot-Safe Hydration:** The `INSERT` query returns the fully hydrated user object by aggregating from the input CTE, bypassing PostgreSQL snapshot isolation lag.

## 3. Performance Tuning
* **Partition-Wise Optimizations:** We have enabled `enable_partitionwise_join` and `enable_partitionwise_aggregate` at the database level.
* **Query Performance:** Lookup by unique attribute (Login) has been optimized to **~0.2ms - 1ms** in local tests with 1,000,000 users, utilizing index-only scans on the registry followed by co-located joins.

## 4. Data Generation / Load Testing
A script is included to generate 1M users with 6M attributes.
* **Current Performance:** ~1.5 minutes for 1M users.
* **Optimization Note:** Hashing is moved to the Go application layer to allow for parallelization and to prevent CPU bottlenecks on the database primary during bulk loads.

## 5. Examples

The users migration contains an `example` folder, containing a queries and some rough Go code, testing and explaining usage of the EAV store.

---

> [!NOTE]
> **High Risk**
> Introduces new PostgreSQL schema/migrations for core tenancy + user storage, including partitioning, constraints, and database-level settings; mistakes here could cause data loss, performance regressions, or migration failures in production.
> 
> **Overview**
> Defines new Postgres migrations for `instances` and `organizations`, and replaces the old `001_users` migration with a new `003_users` migration that creates a partitioned EAV user model (`users`, `user_attributes`, `user_unique_attributes`) with JSONB constraints, targeted indexes, and a uniqueness registry keyed by hashed values.
> 
> The user migration also enables partition-wise join/aggregate at the database level and introduces custom PG types (`uniqueness_scope`, `incoming_user_attribute`) plus an example harness (SQL + Go/pgx) to create and query users using typed attribute arrays.
> 
> Additionally updates the devcontainer VS Code extension set and adds `github.com/dmarkham/enumer` as a Go tool (with related `golang.org/x/*` bumps) to support generated enum helpers in the example code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d811bed53969023f1628eb971ae6bf133b29f2f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/storage/database/dialect/postgres/migration/003_users/README.md
muhlemmer, line 1: Please ignore this file for now, I forgot it's here and it's probably outdated. I'll fix it tomorrow.

internal/storage/database/dialect/postgres/migration/003_users/up.sql
```diff
@@ -0,0 +1,111 @@
...
+-- User attributes is EAV store containing all the properties of a user.
+-- This is the only table containing PII.
+-- It is partioned by user_id, matching the users table, and therefore optimized to hydrate a user.
+-- A couple if indexes are defined for general searching and listing of users.
+CREATE TABLE zitadel_nextgen.user_attributes (
+    instance_id TEXT COLLATE "C" NOT NULL
+    , organization_id TEXT COLLATE "C" NOT NULL
+    , user_id TEXT COLLATE "C" NOT NULL
+    , key TEXT NOT NULL COLLATE "C" CHECK ( key <> '' )
+    , value JSONB NOT NULL --JSONB allows for dynamic typing
```
IAM-marco: Is it worth checking also

```sql
CHECK (
    value != 'null'::jsonb
    AND value != '[]'::jsonb
    AND value != '{}'::jsonb
)
```

?

internal/storage/database/dialect/postgres/migration/003_users/example/get_by_attribute.sql
```diff
@@ -0,0 +1,49 @@
...
+    TEXT[]   -- $5 attributes to fetch
+) AS
+WITH target AS (
+    SELECT user_id
+    FROM zitadel_nextgen.user_unique_attributes
+    WHERE instance_id = $1
+    AND organization_id = COALESCE($2, '')
+    AND key = $3
+    AND value_hash = digest($4::text, 'md5')
+    LIMIT 2
```
IAM-marco: If `LIMIT 2` is intentional to surface constraint violations, please add a comment explaining why.

internal/storage/database/dialect/postgres/migration/003_users/example/user.go
```diff
@@ -0,0 +1,156 @@
...
+		instanceID,
+		u.SchemaURL,
+		u.ID,
+		u.OrganizationID,
+		u.Attributes,
+	}
+}
+
+func CreateUser(ctx context.Context, instanceID string, u *IncommingUser) (*User, error) {
+	rows, _ := conn.Query(ctx, insertUserStmt, u.insertArgs(instanceID)...)
```
IAM-marco, line 141: Same as above: error from `conn.Query` is discarded with `_`. Please handle it explicitly.

internal/storage/database/dialect/postgres/migration/003_users/example/user.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func buildAttributeTree(attributes []Attribute) (map[string]any, error) {
+	tree := make(map[string]any)
+	for i, a := range attributes {
+		keyNodes := strings.Split(a.Key, ".")
+
+		// empty keys are prevented in the DB schema,
+		// this is just a safety check to prevent panics in case of invalid data
+		if len(keyNodes) == 0 {
```
IAM-marco: This check is dead code. `strings.Split("", ".")` returns `[""]` (length 1, not 0), so `len(keyNodes) == 0` can never be true. The actual risk is an empty key producing a map entry with key `""`. The guard should be:
```go
if keyNodes[0] == "" {
    return nil, fmt.Errorf("illegal empty key for attribute at index %d", i)
}
```

internal/storage/database/dialect/postgres/migration/003_users/example/user.go
```diff
@@ -0,0 +1,156 @@
...
+	tree["$id"] = u.ID
+	tree["organization_id"] = u.OrganizationID
+	tree["created_at"] = u.CreatedAt
+	tree["updated_at"] = u.UpdatedAt
+	return json.Marshal(tree)
+
+}
+
+func GetUserByID(ctx context.Context, instanceID, id string) (*User, error) {
+	rows, _ := conn.Query(ctx, getUserByIdStmt, instanceID, id)
```
IAM-marco, line 74: Should handle this error

internal/storage/database/dialect/postgres/migration/003_users/up.sql
```diff
@@ -0,0 +1,111 @@
...
+
+    , PRIMARY KEY (instance_id, user_id, key)
+    , FOREIGN KEY (instance_id, user_id)
+        REFERENCES zitadel_nextgen.users(instance_id, id)
+        ON DELETE CASCADE
+) PARTITION BY HASH (instance_id, user_id);
+
+-- General Lookup Index: Only for scalar valuesI mean, w
```
IAM-marco: Fix or delete

IAM-marco (changes requested):

internal/storage/database/dialect/postgres/migration/003_users/example/user.go
muhlemmer: The documentation of this function says:

> It is allowed to ignore the error returned from Query and handle it in Rows.

`CollectExactlyOneRow` below checks for `row.Err()` and returns the original error if on was found.

internal/storage/database/dialect/postgres/migration/003_users/up.sql
muhlemmer: Yes definitely! Actually I did have that check at one point, but it got accidentally removed in 16d62a92bb39843d35d659eec6d28f3d0fa5b432. Nice catch.

PR #7, opened 2026-04-22
migration: golang-migrate + Spanner example

IAM-marco:
Based on Tim's feat-users-tables branch, here is an example of using golang-migrate to run migrations for Spanner and Postgres.


IAM-marco:
Superseeded by #65

PR #8, opened 2026-04-22
migration: goose + Spanner example

IAM-marco:
Based on Tim's feat-users-tables branch, here is an example of using goose to run migrations for Spanner and Postgres.


IAM-marco:
Superseeded by #65

PR #9, opened 2026-04-22
docs(proposal): capability handshake

grvijayan:
This PR introduces a proposal to tackle the version mismatch issue between Clients/SDKs and the Flow Engine API supported by the Server.


api/capability-handshake.md
```diff
@@ -0,0 +1,36 @@
+# Client/SDKs + API Version Compatibility
+> **Status:** Draft
+
+## Problem Statement
+When a Flow Engine dictates which UI components should be rendered, version mismatches often occur: the Flow Engine API might be on a higher version than the Client/SDK can support, or vice versa. Left unhandled, this leads to a poor user experience, such as the UI rendering a blank screen (if the API requests an unknown state) or the API returning unhelpful error messages (if the Client/SDK requests a feature the server doesn’t yet support).
+
+## Proposal - Bidirectional Capability Handshake/Exchange
```
grvijayan: I did not know what to call it, so I wrote "Bidirectional Capability Handshake/Exchange" - we can change this if there's a better name 😄

wim07101993 (approved):
Looks about like what we discussed

vitorbari:
I'm not confident with this approach, because there other attributes that will influence which flow to pick, for example, what is the intended "flow"  (sign-in, sign-up, etc), the status of the current session (for the step-up cases), etc. ([see the create flow request docs (wip)](https://github.com/zitadel/nextgen/blob/main/docs/design/flowengine/api/flow-api.yaml#L341)). We shouldn't offload this decision logic to the frontend.

I'm more inclined in the direction where the client provides its current `flow schema version` + all the other context information, and the flow engine backend decides which one to return.

grvijayan:
> I'm not confident with this approach, because there other attributes that will influence which flow to pick, for example, what is the intended "flow" (sign-in, sign-up, etc), the status of the current session (for the step-up cases), etc. ([see the create flow request docs (wip)](https://github.com/zitadel/nextgen/blob/main/docs/design/flowengine/api/flow-api.yaml#L341)). We shouldn't offload this decision logic to the frontend.
> 
> I'm more inclined in the direction where the client provides its current `flow schema version` + all the other context information, and the flow engine backend decides which one to return.

As far as I understand, the client is not picking the flow exactly, but only the `flow schema version` that it can support to avoid breaking changes in case the client cannot render certain UI elements to support the latest `flow schema version`. The client has to provide its current `flow schema version` when it chooses from the list of schema versions that the server supports, right? Please let me know if I am missing some context here.

vitorbari:
> > I'm not confident with this approach, because there other attributes that will influence which flow to pick, for example, what is the intended "flow" (sign-in, sign-up, etc), the status of the current session (for the step-up cases), etc. ([see the create flow request docs (wip)](https://github.com/zitadel/nextgen/blob/main/docs/design/flowengine/api/flow-api.yaml#L341)). We shouldn't offload this decision logic to the frontend.
> > I'm more inclined in the direction where the client provides its current `flow schema version` + all the other context information, and the flow engine backend decides which one to return.
> 
> As far as I understand, the client is not picking the flow exactly, but only the `flow schema version` that it can support to avoid breaking changes in case the client cannot render certain UI elements to support the latest `flow schema version`. The client has to provide its current `flow schema version` when it chooses from the list of schema versions that the server supports, right? Please let me know if I am missing some context here.

If frontend supports flow schema `v5`, it can only support flows based on schemas `<=v5`.
The frontend tells what is supports, the backend figures it out.

See how MCP does it: https://modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle#version-negotiation

vitorbari:
Another approach is to ignore versions and do a proper capability negotiation, which in the case of flow engine, would be base on the step types: https://github.com/zitadel/nextgen/blob/main/docs/design/flowengine/flow-engine.md#step-types

fforootd:
I was thinking about this and what about this approach.

We could publish schemas for the flows that include the capabilites. So if we for example have a flow schema 1.0 and we would add a field we would just increase the schema version to 1.1.

The client could be pinned to a specific schema version it supports. This way we could have the server have an endpoint like a ./well-known or something like that that exposes all the schemas supported by the server.

What do you two think about this? IMO this would shift the discussion to SDK server Version alignment and less into the direction of talking about capabilities.

wim07101993:
> We could publish schemas for the flows that include the capabilites. So if we for example have a flow schema 1.0 and we would add a field we would just increase the schema version to 1.1.

When we discussed this internally, that is also what we concluded. By increasing the version of the scheme it is possible to indicate changes more clearly. 

> Another approach is to ignore versions and do a proper capability negotiation

The reason we did not end up with a given list of capabilities was because we would essentially be creating a package manager. E.g. when the list of capabilities contains captcha (since we are always using that example). If the version of the captcha capability would change, this would not be translated in the list of capabilities unless you specify the version of the capability with it. Publishing those versions as well, would than again make it way harder for developers who wants to create their own login. The example of captcha is rather vague because it is frontent only but if the capability requires sending a request to the backend, this becomes more of a problem.

@grvijayan correct me if I am wrong.

livio-a:
> The client could be pinned to a specific schema version it supports. This way we could have the server have an endpoint like a ./well-known or something like that that exposes all the schemas supported by the server.

Yes I think we also discussed that, based on flow schema (versions) that would work. I guess this approach was the outcome later for a more dynamic way to allow the client / sdk to find the supported version / not pin it completely.
And  to show how you would eventually update your client to a new version, resp. make sure the flows / flow configs have a revision and point to a versioned definition and not having to handle all the logic server side.

peintnermax:
One thing to consider: With our templating, if the server sends a LiquidJS template alongside the capability payload, most version compatibility issues disappear — layout changes, new fields, and reordered elements are just template changes that don't require a client update at all. The only thing that actually bumps a client version is a new <zl-*> component, which a simple version header + hard fail covers without needing schema negotiation or .well-known discovery. so my suggestion is to skip schema negotiation entirely.

- One capability data model — fields, actions, gates as typed dictionaries (already defined in the Flow API spec). This doesn't get versioned per-flow. It evolves with the API version.
- Client sends `X-Widget-Version` on `POST /flows` — no extra round-trip, no .well-known, no schema list to parse. The backend knows what the client supports.
- Backend decides everything (agreeing with @vitorbari). It emits only capabilities the client can handle. If the client is too old for a required security gate → hard fail with widget_upgrade_required.
- Template handles presentation — the server sends a LiquidJS template alongside the capabilities. Layout changes, field reordering, copy changes → zero client updates. The schema negotiation problem disappears for our widget.
- Headless SDK users get the typed capability model via the published OpenAPI spec (I consider this an edge case). Standard API versioning (/v1/, /v2/) handles breaking changes — same as any REST API.

This avoids: the extra round-trip, maintaining parallel schema versions, the "package manager" problem, and the security risk of falling back to older schemas that skip gates.

fforootd:
> Client sends X-Widget-Version on POST /flows — no extra round-trip, no .well-known, no schema list to parse. The backend knows what the client supports.

I also considered this but the handling of HTTP headers across LBs and CDNs is usually not really nice 😁

api/capability-handshake.md
```diff
@@ -0,0 +1,58 @@
...
+participant Client
+participant Server
+    Client->>Server: GET /flow-schemas
+    Server-->>Client: Returns JSON array of schemas
+    Note over Client, Server: [<br/>  { "flow_schema_version": "1.1.0", "is_deprecated": true },<br/>  { "flow_schema_version": "1.2.0" },<br/>  { "flow_schema_version": "2.0.0" }<br/>]
+    Note left of Client: Client evaluates available schemas.<br/>Determines its UI supports ^1.1.0 (compatible with 1.x).
+    Client->>Server: POST /flows<br/>{"purpose": "login", "supported_versions": "^1.1.0"}
+    Note right of Server: Server parses SemVer constraint.<br/>Selects optimal schema (v1.2.0).
+    Server-->>Client: Returns initial step for v1.2.0 login flow
+    Note left of Client: Client includes SemVer constraint<br/>in all subsequent flow calls.
```
vitorbari: I think this is not needed. After the flow has been initialized, the flow ID being used, can be stored in the flow state.

grvijayan: Do you mean that the client's version would be stored in the flow state?

vitorbari: Client version / server version are only relevant when starting the flow. Subsequent steps of the flow just continue on the already defined flow (client version doesn't matter anymore).

docs/design/flowengine/capability-handshake.md
vitorbari, line 1: This should be moved under `/docs`.
After that, it can be merged from my perspective. We can revisit this topic later, but the current state should enough for the POC.

grvijayan: I've moved it under `docs/design/flowengine/` for now.

vitorbari (approved):

PR #12, opened 2026-04-22
docs: update session and flow interaction design

livio-a:
Clarifies how flow, auth attempts and sessions interact with each other, resp. how clients / app interact with them to get users authenticated


docs/design/api/authn-and-auth-flows.md
```diff
@@ -108,22 +108,285 @@ Exchange requirements (also documented in [`credentials.md`](credentials.md#hand
...
+    Note over Browser,CustomerBackend: Factor 2 — TOTP (policy required MFA)
+
+    Browser->>FlowEngine: POST /flows/{id}/submit { totp: { code: "123456" } }
+    FlowEngine->>AuthAttempts: POST /auth_attempts/{id}/challenges/{cid}/verify { totp: { code: "123456" } }
+    AuthAttempts-->>FlowEngine: 200 OK — factor written, assurance_levels[] updated
+    FlowEngine-->>Browser: Set-Cookie: flow=<cleared> · 200 { step: "complete" }
+
+    Note over Browser,CustomerBackend: Terminal — modern embedded handoff
+
+    Browser->>FlowEngine: POST /flows/{id}/complete
```
vitorbari: These examples assume there is a `POST /flows/{id}/complete` endpoint, which differs from the other [specs](https://github.com/zitadel/nextgen/blob/main/docs/design/flowengine/flow-engine-guide.md#flow-completion).
A client doesn't need to call `POST /flows/{id}/complete`, instead, the `POST /flows/{id}/submit` would, at some point, return a step `type=complete` with some extra behaviour (redirect, continue, etc). 

Option 1: (based on diagram)
Client receives a step response `type=complete` and makes explicit call to `POST /flows/{id}/complete`, which performs the handoff.

Option 2: (original flowengine specs)
Once the state-machine reaches the `type=complete`, it already performs the handoff, and returns it to the back to the client. No further interaction needed with the flow engine.

livio-a: Ooops yes i missed that and would go for option 2.

docs/design/api/authn-and-auth-flows.md
```diff
@@ -108,22 +108,285 @@ Exchange requirements (also documented in [`credentials.md`](credentials.md#hand
...
+
+    Browser->>FlowEngine: POST /flows/{id}/submit { password: "…" }
+    FlowEngine->>AuthAttempts: POST /auth_attempts/{id}/challenges/{cid}/verify { password: "…" }
+    AuthAttempts-->>FlowEngine: 200 OK — factor written to session
+    FlowEngine-->>Browser: Set-Cookie: flow=<encrypted_state> · 200 { step: "totp" }
+
+    Note over Browser,CustomerBackend: Factor 2 — TOTP (policy required MFA)
+
+    Browser->>FlowEngine: POST /flows/{id}/submit { totp: { code: "123456" } }
+    FlowEngine->>AuthAttempts: POST /auth_attempts/{id}/challenges/{cid}/verify { totp: { code: "123456" } }
```
grvijayan: Should there be a`POST /auth_attempts/{id}/challenges { method: "totp" }` call from `FlowEngine` to `AuthAttempts` before the submit/verify calls (similar to how it's done for `password`)?

livio-a: fixed

docs/design/api/authn-and-auth-flows.md
livio-a: fixed

livio-a:
@vitorbari and @grvijayan i fixed your comments and changed `/session_handoffs/{id}/exchange` to `/sessions/exchange` to get rid of a root resource with a single method / endpoint.

livio-a:
cleaned up conflicts after the base PR was merged and will (re)create a proper PR for the API definition of auth_attempts and sessions

docs/design/api/authn-and-auth-flows.md
```diff
@@ -129,6 +148,255 @@ requested methods and proofs.
...
+    autonumber
+    participant Browser
+    participant FlowEngine as Flow Engine
+    participant AuthAttempts as auth_attempts
+    participant SessionDB as Sessions (DB)
+    participant CustomerBackend as Customer Backend
+
+    Note over Browser,CustomerBackend: Bootstrap
+
+    Browser->>FlowEngine: POST /bootstrap/challenge { project_id, client_type: "browser" }
```
vitorbari: Is this request supposed to go to `FlowEngine`? Or is this responsibility of the auth_attempts?

docs/design/api/authn-and-auth-flows.md
```diff
@@ -129,6 +148,255 @@ requested methods and proofs.
...
+
+    Note over Browser,CustomerBackend: Optional — pre-allocate session before user is known
+
+    Browser->>SessionDB: POST /sessions { project_id, user_agent }
+    SessionDB-->>Browser: { session_id, session_token, state: "building", assurance_levels: [] }
+
+    Note over Browser,CustomerBackend: Start auth attempt (flow engine path)
+
+    Browser->>FlowEngine: POST /flows { project_id, challenge_nonce, session_id? }
+    FlowEngine->>AuthAttempts: POST /auth_attempts { project_id, challenge_nonce, session_id? }
```
vitorbari: These are internal calls instead of HTTP calls, right?

docs/design/api/authn-and-auth-flows.md
livio-a: hmm good question, it's kind of neither and just an API, but the flow engine would need to know that there's a challenge, which it needs to send when creating the auth_attempt.

Edit: made it a separate handler in the diagram

vitorbari: Ok
Clients pass the`challenge_nonce` when starting the flow, which also gets forwarded internally to the auth_attempt.
We can define the `POST /bootstrap/challenge` separately on another PR then.

docs/design/api/authn-and-auth-flows.md
livio-a: yes... missed that i only updated OIDC before. I've now changed and made a comment

vitorbari (approved):

PR #15, opened 2026-04-24
feat: schema api

wim07101993:
I added a proposal for versioning schema's and handling those versions.

There are two TODO's left on which I rfc


docs/design/schema-engine/versioning.md
```diff
@@ -0,0 +1,139 @@
...
+
+The API is fully under our control. We can deprecate a Capability which will be removed in the future. The CUSTOMER
+needs to be notified of this deprecation. Initially this can be when updating a schema; other push-based notifications
+can be implemented in the future.
+
+If a breaking change happens on a schema, a migration pattern should be provided by the CUSTOMER. A migration path can
+be: ask the user for the data. E.g.: When a CUSTOMER adds a required field to a user, users should be upgraded. As long
+as not all data is migrated, deactivation of the schema is not possible.
+
+TODO: Search for a solution on how to migrate data without asking a user **and check whether that is necessary**.
```
wim07101993: Is it necessary to have a way to auto-migrate data without user input?

docs/design/schema-engine/versioning.md
```diff
@@ -0,0 +1,141 @@
...
+the entity is created, a version is determined according to semantic versioning. The objects can then target those
+versions. But the entity itself cannot change afterward. The objects are the exception to that rule. Users for example
+need to be able to be modified from a domain perspective.
+
+## Versioning
+
+The Engine is versioned together with the Zitadel binary using semantic versioning: `Major.Minor.Patch` in which breaking changes are major upgrades, new
+features are minor upgrades and all others are patches. It could be versioned separately which would allow for a more
+fine-grained communication of what changed in the engine instead of the entire application. But that would add more
+complexity and confusion. (TODO: Not sure about this yet, separate versioning is still on the table)
```
wim07101993: Wondering what others think about this. Keep the version of the engine aligned with the binary, or create a separate version for it?

docs/design/schema-engine/versioning.md
```diff
@@ -0,0 +1,141 @@
...
+
+E.g.: A user is created from the User-schema, and a flow is created from the Flow-schema.
+
+### Schema
+
+The schema which defines how the instances should be structured.
+
+### Engine
+
+Defines the tools/features which can be used in a schema.
```
grvijayan: I thought an Engine would be the implementation of a schema, for example, the flow engine would implement the capabilities defined in a flow schema. Is that not the case?

wim07101993: As I see it, the engine provides the building-blocks for creating a schema.

E.g.: The flow engine defines a new capability `Captcha` which can be used by a CUSTOMER inside of their schema.

Maybe the wording could be better, but I think we are saying the same thing.

wim07101993: I updated some wording to align more with oder documents

docs/design/schema-engine/versioning.md
fforootd: My though is we do keep objects with the schema they have been changed with and do not migrate them. If an object gets updated we update it with the valid schema at that time.

wim07101993: do you mean denormalize the schema with the object?

fforootd: Sorry for talking in riddles 😓 

I meant like we can store the schema name and version that was used to manipulate a resource as a field on the object. For example user_machine_v1

fforootd:
Some points if found that might need some attention:

- go test ./... currently fails because the generated api.Handler includes the new schema methods, but internal/api.Handler does not implement/stub them.
- The read endpoints require schema.read, but the OAuth security scheme only declares schema.write.

A couple of API shape points are worth deciding:

- For createSchemaRevision, should the response return a revisionId or maybe the full revision resource? Right now it returns an optional id described as the created schema ID, while follow-up endpoints need revisionId.
- We should align on the release-state model. The API currently has default and archived, while the doc describes default more like a flag and uses Removed as the terminal state. My leaning would be to keep lifecycle state and “default revision” separate, but either way the API and docs should use the same model.

Overall direction looks good mostly build/contract cleanup plus a couple of design decisions to settle.

docs/design/schema-engine/README.md
```diff
@@ -0,0 +1,91 @@
...
+E.g.:
+
+- User
+- Flow
+- Auth Provider
+
+### Transformer
+
+A function which transform an object so that the resulting object fits within a schema. These can be uses as a layer
+between schema's.
```
grvijayan:
```suggestion
between schemas.
```

docs/design/schema-engine/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Capability
+
+A piece of functionality which is defined by a keyword/multiple keywords in a root-schema.
+
+E.g.:
+
+- Captcha
+- Field validation
+- Field datatype
```
grvijayan, line 54: Would `Field datatype/validation` be capabilities?

docs/design/schema-engine/README.md
grvijayan: minor: and other places where `schema's` is used to mean the plural of `schema`

docs/design/schema-engine/README.md
```diff
@@ -0,0 +1,91 @@
...
+### Root meta-schema
+
+1. ZITADEL creates a Root Schema. Let's call it `Foo`-root-schema.
+2. Customer creates a Schema which implements to the `Foo`-root-schema. Let's call it `Bar`-schema.
+3. Customer creates Objects which implement to the `Bar`-schema
+
+### Transformer
+
+1. ZITADEL creates a Root Schema. Let's call it `Foo`-root-schema.
+2. Customer creates a Schema separate schema which is similar to the `Foo`-root-schema but does not extend it. Let's
```
grvijayan: Contradicts `A schema defined by the customer. These schemas must extend a Root Schema and will be used to validate Objects against.` defined [here](https://github.com/zitadel/nextgen/pull/15/changes#diff-b1b4901c578361c8f01c12c747d696668414f615b60782377aa03194c9eddc03R28)

Also, if a customer-defined schema doesn't extend from the root schema, what do we validate it against and how do we ensure that all of the capabilities defined in their schema can be rendered/supported by the API?

docs/design/schema-engine/README.md
```diff
@@ -0,0 +1,91 @@
...
+> **Status** Draft
+> **Date:** 2026-04-28
+> **Context:** Zitadel next-generation architecture design
+
+## Introduction
+
+The new ZITADEL architecture allows for a flexible setup by allowing the customer to define the structure of their data.
+This is done using schema's. E.g.: a customer can define his own user schema which describes what properties a user has.
+
+## Terminology
```
vitorbari, line 14: Please check if the terminology used by json-schema could be reused by us: https://json-schema.org/learn/glossary

- root-schema -> https://json-schema.org/learn/glossary#meta-schema
- object -> https://json-schema.org/learn/glossary#instance

(if these make sense in our context)

api/openapi/endpoints/schemas/user-schema.yaml
```diff
@@ -0,0 +1,51 @@
+title: UserSchema
```
vitorbari, line 1: Is this the root/meta schema? Should we name it accordingly?

docs/design/schema-engine/README.md
wim07101993: I think so. The fact that you can use the uuid data-type feels to me like a capability. Same for validation. We define in the api-spec that the backend has the capability to use regex for field validation.

docs/design/schema-engine/README.md
wim07101993: good catch. As for schema validation, good question. We would not be able to validate a schema which does not implement a root schema. Objects can be validated as shown in the flowchart. But this might be an issue.

vitorbari:
> go test ./... currently fails because the generated api.Handler includes the new schema methods, but internal/api.Handler does not implement/stub them.

@fforootd , @wim07101993 ,

the following PR can fix this problem: 
https://github.com/zitadel/nextgen/pull/44

please have a look

docs/design/schema-engine/README.md
wim07101993: `incance` might be a better name indeed. As for `meta-schema`, I explicitly did not call it like that yet because in the setup described in the current version, it would be possible to create a meta-schema by a customer. The `root-schema` would be the schema which is defined by us. But since we decided this would not possible, `meta-schema` might be the appropriate name. Wil fix.

api/openapi/endpoints/schemas/user-schema.yaml
wim07101993: Well, depending on how you want to name things. In other places we name the schema as what they represent. E.g. The `issue-challenge-request.yaml` file represents the request itself whilst containing the schema.

docs/design/schema-engine/README.md
```diff
@@ -0,0 +1,91 @@
...
+Let's describe how these schema's can be used.
+
+### Single step hierarchy
+
+1. ZITADEL creates a Schema. Let's call it `Foo`-schema.
+2. Customer creates Instances which implement to the `Foo`-schema.
+
+This is the simplest use-case for a schema.
+
+### Root meta-schema
```
vitorbari, line 67: Looking at the proposed UserSchema, the user-type level isn't really a JSON Schema (https://json-schema.org/), it's a custom object with `kind: user-schema`. So JSON Schema only really shows up at the meta-schema level, not at both.

That main question for me is: do we want JSON Schema at both levels, or just one?

I'd been assuming both, meaning user types are themselves JSON Schemas, validated against a meta-schema. Something like:

1. Meta-schema — defined by us. A JSON Schema that constrains the shape of any user type a customer creates.

```json
{
    "$schema": "https://json-schema.org/draft/2020-12/schema", // the meta-schema is itself a JSON Schema
    "$id": "https://zitadel.io/nextgen/user-type.schema.json",
    "type": "object",
    "required": ["$schema", "$id", "type", "properties", "x-auth-methods"],
    "properties": {
        "$schema": {
        "const": "https://zitadel.io/nextgen/user-type.schema.json" // pinned to this $id, so user types can only derive from this meta-schema
        },
        "$id": { "type": "string", "format": "uri" },
        "type": { "const": "object" },
        "x-auth-methods": { "type": "..." },
        "properties": {
        "type": "object",
        "required": []
        }
    }
}
```

2. User type — default provided by us, could be customized by the customer. Also a JSON Schema, validated against the meta-schema above.

```json
{
    "$schema": "https://zitadel.io/nextgen/user-type.schema.json", // points at the meta-schema
    "$id": "https://example.com/human-user.schema.json",          // customer-defined
    "type": "object",
    "x-auth-methods": [
        { "enabled": true, "position": 0 }
    ],
    "required": [],
    "properties": {
        "username":    { "type": "string" },
        "given_name":  { "type": "string" },
        "family_name": { "type": "string" },
        "nationality": { "type": "string" }
    }
}
```


3. User instance — concrete data, validated against a user type.

```json
{
    // "$schema": "https://example.com/human-user.schema.json"  // implicit or explicit reference to the user type
    "username":    "john",
    "given_name":  "John",
    "family_name": "Days"
}
```

The trade-off, as I see it: the kind-discriminated approach is simpler to implement and document, but the all-JSON-Schema approach lets us reuse standard JSON Schema tooling for validation at every level and keeps the model uniform.

wim07101993: I missed to add the `$id` tags. But how you understood is is correct I think. You can see the examples for reference. The reason we need the `kind` field, is that it otherwise does not work nicely with the `oneOf` as long as the urls don't resolve properly...

api/openapi/endpoints/schemas/user-schema.yaml
vitorbari: I have started to think about this as the meta-schema for "user types".

Based on this schema, customers will be able to create user types such as human, ai agent, machine. Does it make sense to call it UserTypeSchema?

wim07101993: We could but I think it is still the same. It depends on how you want to name things. In this case the "schema"-suffix already points out the contents of this file represents the schema of a schema. 

Taking the example of the `issue-chyallenge-request.yaml` as an example again. We call it the `...-request.yaml` even though it contains the schema of a request. Not the request itself. In the same fashion the `user-schema.yaml` contains the schema which describes a `user-schema` not a user.

We could also name it `user-meta.yaml` or `user-type.yaml` but calling it `user-type-schema.yaml` would imply this file contains the schema which describes the schema which describes the type which describes the type of a user. Which is not correct I think.

api/openapi/endpoints/schemas/schema-url.yaml
```diff
@@ -0,0 +1,17 @@
+title: UserSchemaUrl
+type: object
+description: |
+  A schema definition that references an external schema via a URL. It is 
+  on creation, validated against its metaschema and cached in the database.
+required:
```
vitorbari: Is this supposed to be used only for external **user** schemas or any type of supported schema?

`UserSchemaUrl` and `const: schema-url` are not matching.

api/openapi/endpoints/schemas/schema-url.yaml
```diff
@@ -0,0 +1,17 @@
+title: UserSchemaUrl
```
wim07101993:
```suggestion
title: SchemaUrl
```

vitorbari (approved):

PR #20, opened 2026-04-25
feat(release): add goreleaser, Dockerfile, and embedded console scaffold

fforootd:
## Summary

Adds the initial release and packaging scaffold for `nextgen`: GoReleaser for the Go server binary and GHCR Docker images, a future embedded React console entry point, Changesets for npm-published packages, and ADR 002 documenting why the release lanes are split.

This PR also tightens CI around the repository tests by fixing the embedded Postgres startup helper that previously left its reserved TCP port open on CI.

## Folder Structure

The proposed layout is:

```text
apps/
  console/        # React console app; built into dist/ and embedded in the Go binary
  cli/            # future TypeScript CLI / npx entry point, released through npm
packages/
  components/     # reusable web components for downstream apps
  sdk-*/          # SDK packages for downstream apps
api/              # API contracts and generated API code
docs/             # ADRs and design docs
cmd/, internal/   # Go server/product implementation
```

Important points:

- `apps/console/` is part of the ZITADEL product. It is not a separately versioned npm package; it ships embedded with the Go server version.
- `apps/console/embed.go` exposes `//go:embed all:dist`, and `apps/console/dist/.gitkeep` keeps clean checkouts buildable until the real console build lands.
- npm-consumed surfaces live under `apps/cli/` and `packages/*` and are versioned/published by Changesets.
- Go server releases stay with GoReleaser and `v*` tags.

### Go server / embedded console

- `.goreleaser.yaml` builds `nextgen` for Linux, macOS, and Windows where supported.
- Docker images use the distroless `static-debian12:nonroot` base and publish to `ghcr.io/zitadel/nextgen`.
- GoReleaser builds the console first when `apps/console/package.json` exists, then embeds `apps/console/dist` into the binary.
- GoReleaser uses the GitHub changelog backend with its default grouping. We intentionally avoid custom regex groups here.

### npm packages

- `.changeset/config.json` and `.changeset/README.md` set up the npm release lane.
- `.github/workflows/changesets.yml` opens the Changesets version PR and publishes affected npm packages once that PR is merged.
- ADR 002 documents why GoReleaser and Changesets stay separate.

### CI / release workflows

- `.github/workflows/ci.yml` runs `go vet`, `go test`, and a GoReleaser snapshot build on PRs.
- `.github/workflows/release.yml` runs on `v*` tags and publishes GitHub releases plus GHCR images.

## Licensing

Adds `LICENSING.md` with the intended split:

[...]

PR #21, opened 2026-04-25
fix(storage): reference schema_url in users CHECK constraint

fforootd:
## Summary
- One-character fix in [internal/storage/database/dialect/postgres/migration/001_users/up.sql](internal/storage/database/dialect/postgres/migration/001_users/up.sql): the `schema_url` column's `CHECK` constraint referenced a non-existent column `schema`, causing `CREATE TABLE` to fail on Postgres 16 with `ERROR: column "schema" does not exist (SQLSTATE 42703)`.
- Introduced in #5 (commit 7a0f872, "feat(storage): define user tables"). The `internal/storage/database/repository` tests, which boot an embedded Postgres via `fergusstrange/embedded-postgres`, have been broken on `main` since that landed; reproduced from a pristine clone.

## Test plan
- [x] `go test ./internal/storage/database/repository/...` — passes (previously failed with SQLSTATE 42703)
- [x] `go test ./...` — passes

### DIFF
--- a/internal/storage/database/dialect/postgres/migration/001_users/up.sql
+++ b/internal/storage/database/dialect/postgres/migration/001_users/up.sql
@@ -1,7 +1,7 @@
 CREATE TABLE zitadel_nextgen.users(
     instance_id TEXT NOT NULL
     , id TEXT NOT NULL CHECK (id <> '')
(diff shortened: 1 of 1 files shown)


fforootd:
I will pull that into #20

PR #26, opened 2026-04-27
feat: api contract for flow engine (frontend facing)

vitorbari:


vitorbari:
closed in favor of #28

PR #28, opened 2026-04-27
feat: api contract for flow engine (frontend facing)

vitorbari:


api/openapi/components/flows/create-flow-request.yaml
```diff
@@ -0,0 +1,19 @@
+type: object
+required: [purpose]
+properties:
```
vitorbari, line 3: Missing version negotiation property

vitorbari:
There is still one small data modeling question related to flow transitions on #31. However, I will merge this PR and follow-up on that separately.

PR #6, opened 2026-04-22
docs: refine nextgen API and platform design base

fforootd:
## Summary

This PR establishes the nextgen API and platform design base. It introduces the shared glossary, API design guide, platform lifecycle docs, and draft OpenAPI sketches for claim/config plus flow/session concepts.

It keeps this branch as the slim design base for follow-up work instead of absorbing the deeper session/flow interaction PRs. The docs now reconcile review feedback and settled inputs from overlapping work while leaving implementation OpenAPI under `api/openapi/**` as the source for generated code.

## Key Decisions

- Define `project_id` as the human-readable dictionary slug shape, for example `river-8421`, and document it as the exception to the general prefixed ULID-style ID convention.
- Use header-selected API versioning via `Zitadel-Version`; public paths do not carry `/v1` segments.
- Use `PATCH /projects/{projectId}/config` for config upload and document retry-safe `Idempotency-Key` semantics for anonymous `POST /projects`.
- Standardize environment names as `development`, `preview`, and `production`.
- Standardize flow definition paths as `/flow-definitions`.
- Keep sessions as durable read models: auth attempts verify proofs and mutate factors, sessions expose `assurance_levels[]`, and SSR/direct handoff exchange happens through `POST /sessions/exchange`.
- Keep MVP scope SQL-first for the resource-scope index; distributed caching can be revisited from measurements.
- Clarify that rate limits happen at the edge/API layer and that 404s should use a useful generic message without disclosing existence vs permission failure.

## Scope Notes

- This is design documentation and draft spec work under `docs/design/**`.
- It does not edit generated code or the implementation OpenAPI source under `api/openapi/**`.
- Follow-up PRs such as the session/flow interaction work can rebase onto this base rather than being folded into it.

## Validation

- Parsed design YAML sketches for claim, config, flow, and session APIs.
- Ran `git diff --check`.
- Checked for stale/conflicting references: `/v1` path examples, `/session_handoffs`, `/flow_definitions`, `PUT /projects`, `project_id: "proj..."`, and `local|preview|production` enums.
- CI is green for `go-test`, `node-check`, `npm-pack-smoke`, and `goreleaser-snapshot`.

PR #31, opened 2026-04-27
feat: Flow definition repository

IAM-marco:
## What problems are solved

Flow definitions — customer-configured directed graphs of authentication steps (login, register, recovery, re-auth, etc.) — had no persistence layer. There was no domain model, no storage schema, and no repository implementation to create, retrieve, update, or delete them.

An initial approach using a fully normalised 5-table schema was prototyped but turned out to be a poor fit: reassembling the aggregate required multiple round-trips and in-process joins, while flow definitions are read-heavy and only mutated on status transitions. The complexity was not justified.

**Domain layer (`internal/domain`)**
- Introduces the `FlowDefinition` aggregate with all child types (`FlowDefinitionPurposeEntry`, `FlowDefinitionAudience`, `FlowDefinitionStep`, `FlowStepTransition`).
- Adds typed `uint8` enums with [`enumer`](https://github.com/dmarkham/enumer) codegen for status (`draft`, `active`, `deprecated`, `archived`), purpose (`login`, `register`, `recovery`, `profiling`, `reauth`, `link_account`), and step type. Codegen provides `String()`, `Value()`, and `Scan()` for driver/scanner compatibility.
- Defines a `FlowDefinitionRepository` interface with `Create`, `Get`, `List`, `UpdateStatus`, and `Delete` operations, plus functional options (`WithFlowDefinitionStatus`, `WithFlowDefinitionPurpose`, `WithFlowDefinitionLimit`, `WithFlowDefinitionOffset`) for list queries.
- Terminology follows the [glossary from PR #6](https://github.com/zitadel/nextgen/pull/6): `project_id` (not `instance_id`), `team_id` (not `org_id`), `IsProjectDefault` (not `IsInstanceDefault`).

**Storage schema — JSONB approach**
- A single-table design: scalar columns (`project_id`, `id`, `name`, `engine_version`, `schema_version`, `status`, `purposes`, timestamps) live as real columns; the nested aggregate (purposes with initial steps, audience, steps, transitions) is stored in a `definition JSONB` column.
- Every read is a single PK lookup with no joins, matching the access pattern: content is immutable once created; only status transitions mutate the row.
- `purposes` is also denormalised as a `flow_definition_purposes[]` array column to allow DB-side `= ANY(purposes)` filtering without touching the JSONB blob.
- Indexed on `(project_id, status)` for filtered list queries and GIN-indexed on `purposes` for containment queries.

**Migrations**
- Postgres (`002_flow_definitions`): enum types `flow_definition_states` and `flow_definition_purposes`, the `flow_definitions` table, a B-tree index on `(project_id, status)`, and a GIN index on `purposes`.
- Cloud Spanner (`002_flow_definitions`): equivalent DDL written in the [Spanner PostgreSQL dialect](https://cloud.google.com/spanner/docs/reference/postgresql/overview) — `TEXT`, `JSONB`, `TIMESTAMPTZ`, `TEXT[]` — mirroring the Postgres migration in structure and table name.

**Repository implementation (`internal/storage/database/repository`)**
- `FlowDefinitionRepository` implements `domain.FlowDefinitionRepository` against the JSONB table.
- `CreateFlowDefinition` serialises the full aggregate to JSON in a single `INSERT`; passes `purposes` as a `[]string` with an explicit `::flow_definition_purposes[]` cast and `status` with `::flow_definition_states`.
- `GetFlowDefinition` deserialises in a single PK lookup using a typed scan struct with `db` tags and a `JSON[flowDefinitionContent]` field for automatic JSONB unmarshalling.
- `ListFlowDefinitions` supports status filtering, purpose filtering (`= ANY(purposes)`), and limit/offset pagination.
- `UpdateFlowDefinitionStatus` issues a targeted `UPDATE` on the status column only.
- `DeleteFlowDefinition` removes the row (no child tables).

**Tests**
- Full integration test suite using embedded Postgres; each test runs inside a rolled-back transaction for isolation.
- Cases: create+get round-trip (JSONB config, pivot transitions, audience fields), status filter, purpose filter, pagination, status update, delete, and project isolation.


IAM-marco:
@vitorbari I added you as reviewer, mostly to get a feedback on the repository interface that I created: https://github.com/zitadel/nextgen/pull/31/changes#diff-8d98f489688512da54b463c252ac5155ee46aebb48b7349d731331f7ffb30a53
Can you let me know whether you think this fits the needs of service layer or not? If not what needs to be changed?

fforootd:
Here my unsolicited comment 😁 

I found one potential drift with the Postgres migration ordering. This adds `flow_definitions` as another `002` migration, but existing databases on `origin/main` may already have tern version `3` recorded. In that case the new migration can be skipped, and the existing users migration can be attempted again under a new sequence. **I mean this is not released but something to be aware of.**

There are also two Spanner related elements that I noticed: the dialect package no longer registers itself with `database.MustRegisterDialect`, so config-based `database: { spanner: ... }` setup fails, and the Spanner migration runner does not persist applied migration state, so a fresh process can rerun non-idempotent DDL against an already-migrated database.

One smaller repository issue: `ListFlowDefinitions` paginates with `ORDER BY created_at ASC` only. Since `NOW()` is constant within a transaction, multiple rows can tie on `created_at`, making offset pagination unstable, or a I missing something? Adding a secondary key like `id` would make paging deterministic.

internal/domain/flow_definition.go
```diff
@@ -0,0 +1,88 @@
...
+	FlowStepTypeComplete
+)
+
+// FlowDefinition is a customer-configured directed graph of authentication steps.
+// It is immutable: modifications produce a new revision with a new SchemaVersion.
+type FlowDefinition struct {
+	ProjectID     string
+	ID            string
+	Name          string
+	EngineVersion string
```
vitorbari: I think this can be dropped. There is no clear definition of EngineVersion, and the SchemaVersion will be used for versioning negotiation.

internal/domain/flow_definition_repository.go
```diff
@@ -0,0 +1,72 @@
...
+
+// WithFlowDefinitionLimit sets the maximum number of results to return.
+func WithFlowDefinitionLimit(limit uint32) FlowDefinitionListOption {
+	return func(o *FlowDefinitionListOpts) {
+		o.Limit = limit
+	}
+}
+
+// WithFlowDefinitionOffset skips the first n results.
+func WithFlowDefinitionOffset(offset uint32) FlowDefinitionListOption {
```
vitorbari, line 76: We also need `WithSchemaVersion`

internal/storage/database/dialect/postgres/migration/002_flow_definitions/up.sql
```diff
@@ -0,0 +1,36 @@
...
+    , 'profiling'
+    , 'reauth'
+    , 'link_account'
+);
+
+CREATE TABLE zitadel_nextgen.flow_definitions (
+    project_id          TEXT NOT NULL
+    , id                TEXT NOT NULL CHECK (id <> '')
+    , name              TEXT NOT NULL CHECK (name <> '')
+    , engine_version    TEXT NOT NULL CHECK (engine_version <> '')
```
vitorbari: Can be dropped.

internal/domain/flow_definition.go
```diff
@@ -0,0 +1,88 @@
...
+	FlowStepTypeInfo
+	FlowStepTypeComplete
+)
+
+// FlowDefinition is a customer-configured directed graph of authentication steps.
+// It is immutable: modifications produce a new revision with a new SchemaVersion.
+type FlowDefinition struct {
+	ProjectID     string
+	ID            string
+	Name          string
```
vitorbari, line 59: We will also need a human-readable slug, which can be used for the frontend to trigger a specific flow definition, or when building linked flows.

internal/domain/flow_definition.go
```diff
@@ -0,0 +1,88 @@
...
+// FlowDefinitionStep is a single node in the step graph.
+type FlowDefinitionStep struct {
+	Name        string
+	Type        FlowStepType
+	Config      map[string]any
+	Transitions []FlowStepTransition
+}
+
+// FlowStepTransition maps an action name to either a target step (regular) or a pivot purpose.
+type FlowStepTransition struct {
```
vitorbari, line 93: There are 3 types of transition:

```json
      "transitions": {
       // 1 - same flow step transition
        "submit": "resolve_user",
        "sso": "sso_redirect",
       
       // 2 - cross-flow step transition (switch)
       "passkey": {
          "switch": "passkey-login" // identifier of the other flow
        },
        "register": {
          "switch": "default-registration" // identifier of the other flow
        },

        // 3 - cross-flow step transition (pivot)
        "recover": {
          "pivot": "password-recovery" // identifier of the other flow
        }
      }
```

We need to find a data-model that works for all of them:

Maybe:

```json
"transitions": {
  "{key}" : {
      "target": "string", // name of the target step or identifier of linked flow
      "pivot_type": "null|string" // either switch or pivot
  },

// examples

  "submit" : {
      "target": "resolve_user", // same-flow step
      "pivot_type": null
  }

  "passkey" : {
      "target": "passkey-login",
      "pivot_type":"switch"
  },
  
  "recover" : {
      "target": "password-recovery",
      "pivot_type":"pivot"
  },
}
```

internal/domain/flow_definition.go
IAM-marco: Isn't name enough?

IAM-marco:
@fforootd 

> I found one potential drift with the Postgres migration ordering. This adds `flow_definitions` as another `002` migration, but existing databases on `origin/main` may already have tern version `3` recorded. In that case the new migration can be skipped, and the existing users migration can be attempted again under a new sequence. **I mean this is not released but something to be aware of.**

What is missing here is changing tern with goose or golang-migrate, as they seem to have compatibility with Spanner. It would allow me to get rid `splitStatements` in `runner.go` .

Also, the migration is labelled 002 just because it was copied over from postgres. To be fair, I should also add the instance migration which is missing.

> There are also two Spanner related elements that I noticed: the dialect package no longer registers itself with `database.MustRegisterDialect`, so config-based `database: { spanner: ... }` setup fails, and the Spanner migration runner does not persist applied migration state, so a fresh process can rerun non-idempotent DDL against an already-migrated database.

👍  Nice catch
 
> One smaller repository issue: `ListFlowDefinitions` paginates with `ORDER BY created_at ASC` only. Since `NOW()` is constant within a transaction, multiple rows can tie on `created_at`, making offset pagination unstable, or a I missing something? Adding a secondary key like `id` would make paging deterministic.

We should always order by ID at least. I read that spanner orders randomly if not specified, while Postgres order by PK if not specified (as far as I know). In that specific case, with postgres, in case of a tie on created_at, it would then order by PK.

internal/domain/flow_definition.go
vitorbari: We can keep only `name` for now and enforce it in the API layer to by a slug.

internal/storage/database/dialect/postgres/migration/002_flow_definitions.go
```diff
@@ -0,0 +1,14 @@
+package migration
```
adlerhurst: would currently be 004

internal/storage/database/dialect/postgres/rows.go
```diff
@@ -1,77 +1,15 @@
...
+type Rows = pgxcommon.Rows
 
-// Close implements [database.Rows].
-// Subtle: this method shadows the method (Rows).Close of Rows.Rows.
-func (r *Rows) Close() error {
-	r.Rows.Close()
-	return nil
-}
+func newRow(row pgx.Row) *Row   { return pgxcommon.NewRow(row, wrapError) }
+func newRows(rows pgx.Rows) *Rows { return pgxcommon.NewRows(rows, wrapError) }
```
adlerhurst:
```suggestion
func newRow(row pgx.Row) *Row { return pgxcommon.NewRow(row, wrapError) }

func newRows(rows pgx.Rows) *Rows { return pgxcommon.NewRows(rows, wrapError) }
```

internal/storage/database/dialect/spanner/migration/004_flow_definitions/up.sql
```diff
@@ -0,0 +1,22 @@
...
+    , name          TEXT NOT NULL
+    , engine_version TEXT NOT NULL
+    , schema_version TEXT NOT NULL
+    , status        TEXT NOT NULL DEFAULT 'draft'
+    , purposes      TEXT[] NOT NULL DEFAULT ARRAY[]::TEXT[]
+    , definition    JSONB NOT NULL
+    , created_at    TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
+    , updated_at    TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
+    , PRIMARY KEY (project_id, id)
+);
```
adlerhurst, line 15: The formatting is weird either remove the tabs or bring all type definitions to the same vertical line

internal/storage/database/dialect/spanner/rows.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"github.com/jackc/pgx/v5"
+	"github.com/zitadel/nextgen/internal/storage/database/dialect/pgxcommon"
+)
+
+type Row = pgxcommon.Row
+type Rows = pgxcommon.Rows
+
+func newRow(row pgx.Row) *Row     { return pgxcommon.NewRow(row, wrapError) }
+func newRows(rows pgx.Rows) *Rows { return pgxcommon.NewRows(rows, wrapError) }
```
adlerhurst:
```suggestion
func newRow(row pgx.Row) *Row { return pgxcommon.NewRow(row, wrapError) }

func newRows(rows pgx.Rows) *Rows { return pgxcommon.NewRows(rows, wrapError) }
```

internal/storage/database/repository/flow_definition.go
```diff
@@ -0,0 +1,334 @@
...
+		purposeElemCast: "::zitadel_nextgen.flow_definition_purposes",
+		purposeArrCast:  "::zitadel_nextgen.flow_definition_purposes[]",
+	}
+}
+
+// NewSpannerFlowDefinitionRepository returns a repository configured for the
+// Spanner PostgreSQL dialect, which uses plain TEXT columns with no casts.
+func NewSpannerFlowDefinitionRepository(client database.QueryExecutor) *FlowDefinitionRepository {
+	return &FlowDefinitionRepository{Client: client}
+}
```
adlerhurst: IMO this is something the repo has to decide not the domain logic. Can we do it by type checks? e.g. `c, ok := client.(spanner.Pool)` or add a method (`IsPostgres()`, `IsSpanner()` on all `QueryExecutor` implementation and do something like this: 
```go
switch client.(type) {
    case interface{IsSpanner()}:
        ...
    case interface{IsPostgres()}:
        ...
}
```

internal/storage/database/repository/flow_definition.go
```diff
@@ -0,0 +1,334 @@
...
+
+	purposes := make([]string, len(def.Purposes))
+	for i, p := range def.Purposes {
+		purposes[i] = p.Purpose.String()
+	}
+
+	b := database.NewStatementBuilder(
+		"INSERT INTO " + tableFlowDefinitions +
+			" (project_id, id, name, engine_version, schema_version, status, purposes, definition, created_at, updated_at)" +
+			" VALUES (")
```
adlerhurst: can you make this const? this way makes more allocations because each time it gets called it allocates new memory to create a new string. or split it into 3 parts.
```go
b := database.NewStatementBuilder("INSERT INTO ")
b.WriteString(tableFlowDefinitions)
b.WriteString(" (project_id, id, name, engine_version, schema_version, status, purposes, definition, created_at, updated_at) VALUES (")
```

internal/storage/database/repository/flow_definition.go
```diff
@@ -0,0 +1,334 @@
...
+
+	_, err = r.Client.Exec(ctx, b.String(), b.Args()...)
+	return err
+}
+
+func (r *FlowDefinitionRepository) GetFlowDefinition(ctx context.Context, projectID, id string) (*domain.FlowDefinition, error) {
+	b := database.NewStatementBuilder(
+		"SELECT project_id, id, name, engine_version, schema_version, status, definition, created_at, updated_at" +
+			" FROM " + tableFlowDefinitions +
+			" WHERE project_id = ")
```
adlerhurst: same here, concatenating strings with + is not very performant in go

adlerhurst (commented):
looks good from my side, only small comments

IAM-marco:
@fforootd 

Things that will do in a separate PR as it will likely entail a lot of refactoring:

  - Make sure Spanner dialect is registered correctly: _There are also two Spanner related elements that I noticed: the dialect package no longer registers itself with database.MustRegisterDialect, so config-based database: { spanner: ... } setup fails, and the Spanner migration runner does not persist applied migration state, so a fresh process can rerun non-idempotent DDL against an already-migrated database._
  - Migrate to golang-migrate or Goose

See https://github.com/zitadel/nextgen/pull/31#issuecomment-4339665534

internal/domain/flow_definition.go
```diff
@@ -0,0 +1,103 @@
...
+	IsProjectDefault bool
+}
+
+// FlowDefinitionStep is a single node in the step graph.
+type FlowDefinitionStep struct {
+	Name        string
+	Type        FlowStepType
+	Config      map[string]any
+	Transitions []FlowStepTransition
+}
```
vitorbari, line 90: We might still adjust some internal definitions of the Step object, for example, removing the `FlowStepType`, but it does not matter much from a storage perspective because steps go into the marshaled `content`.
reference: #50

vitorbari (approved):

PR #43, opened 2026-04-29
feat(api): start auth attempt and session API definition

livio-a:
Add the API definition (first draft) for the auth_attempts and sessions, so that we can start with a base implementation.
It also adds an ADR to define how search / list endpoints should do pagination: cursor based instead of offsets.


api/openapi/endpoints/auth_attempts/by_id/methods.yaml
```diff
@@ -0,0 +1,35 @@
+get:
+  operationId: getAuthAttempt
+  summary: Get authentication attempt state
+  description: |
+    Polls the current state of an authentication attempt.
+    
+    Returns the attempt's state, available factors for the next challenge,
+    challenges issued so far, and any errors preventing progress.
```
vitorbari, line 8: `Available factors` and `challenges issued` are currently not part of the response:

```json
{
  "attempt_id": "att_abc123def456",
  "project_id": "river-8421",
  "session_id": "sess_existing_session_123",
  "user_id": "user_id_12345",
  "expires_at": "2026-04-28T15:45:00Z",
  "created_at": "2026-04-28T15:30:00Z"
}
```

api/openapi/components/parameters/page-token.yaml
```diff
@@ -0,0 +1,11 @@
+name: page_token
+in: query
+required: false
+description: |
+  Token for fetching the next page of results.
+  Obtain this value from `next_page_token` in the previous response.
+  Omit to start from the beginning.
+  Its format is opaque and may change between releases.
+schema:
+  type: string
```
wim07101993: Shouldn't this use the schema defined at `api/openapi/schemas/page-token.yaml`?

api/openapi/endpoints/sessions/session-list-response.yaml
```diff
@@ -0,0 +1,16 @@
+type: object
+description: Paginated list of sessions.
+required:
+  - sessions
+properties:
+  sessions:
+    type: array
+    items:
+      $ref: session-response.yaml
+  next_page_token:
```
vitorbari, line 10: Should there be a previous_page_token as well?

api/openapi/endpoints/sessions/exchange/exchange-request.yaml
```diff
@@ -0,0 +1,12 @@
+type: object
+description: Request to exchange a handoff token for a session and session_token.
+required:
+  - handoff_token
+properties:
+  handoff_token:
+    type: string
+    description: |
+      The one-time handoff token minted by `POST /auth_attempts/{id}/handoff`.
+      Single-use — replaying it after a successful exchange returns `410 Gone`.
```
vitorbari, line 10: How will we enforce it to be single-use? (just curious, I think it is not critical in the context of this PR)

livio-a: The token will be stored on the auth_attempt, which will be removed / archived if needed (unsure if we really want to go with HTTP 410 or just 403/404).

api/openapi/endpoints/sessions/session-list-response.yaml
livio-a: most likely we need to provide one in combination with a corresponding parameter so you can navigate to both sides

api/openapi/endpoints/sessions/exchange/methods.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+    The `session_token` supersedes any previously issued `session_token` for the same session.
+    Clients must replace their stored token at this point.
+
+    Requires a project service key (OAuth2 client credentials).
+  tags:
+    - Sessions
+  security:
+    - oauth2:
+        - openid
```
wim07101993: Doesn't this need a special scope?

api/openapi/endpoints/auth_attempts/by_id/methods.yaml
livio-a: fixed... I also noticed the previous oneOf in issuing a challenge broke code gen so i changed it back to having optional properties if needed, e.g. passkey

api/openapi/endpoints/sessions/exchange/methods.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - name: Idempotency-Key
+      in: header
+      required: false
+      description: |
+        Optional retry key for safe retries of the handoff exchange (Category B idempotency).
+        Retries carrying the same key within the exchange window return the cached session
+        payload without consuming the one-time handoff token again.
+      schema:
+        type: string
+        example: 01HXYZ123ABC
```
wim07101993: why and idempotency-key? I can see how it could be useful but is there use-case where it is required?

api/openapi/endpoints/sessions/methods.yaml
```diff
@@ -0,0 +1,115 @@
...
+    Requires a project service key (OAuth2 client credentials).
+  tags:
+    - Sessions
+  security:
+    - oauth2:
+        - openid
+  parameters:
+    - $ref: ../../components/parameters/limit.yaml
+    - $ref: ../../components/parameters/page-token.yaml
+    - name: project_id
```
wim07101993: why not make this a component as well?

wim07101993 (commented):
1. Shouldn't the auth-attempt endpoints be secured so that they can only be used by the login-ui? Same as how now the loginV2 requires a pat.

2. In #15 other conventions are used to structure the openapi spec. I think we should be consistent and use one. Which one does not really matter to me. Current discrepancies:
   - parameters which are defined for a single resource reside in parameters directory/next to the requests in which they are defined
   - to get a resource the patttern `resource/{id}` is used. Both prs put those in separate directories. But one used a `by_id` directory for every path, the other uses a plural/singular directory structure

FYI: when I run the redocly linter against this, it gives some warnings, errors (`redocly-cli`: `redocly-cli lint openapi-spec.yaml`)

api/openapi/endpoints/auth_attempts/by_id/methods.yaml
wim07101993: @livio-a fyi: if you would want to keep the types instead of the enum:
```yaml
description: |
  Request to issue a factor challenge.

  Use the `method` field as the discriminator. Each method variant enforces its
  own required fields at the schema level.
discriminator:
  propertyName: method
  mapping:
    identifier: issue-challenge-by-identifier.yaml
    password: issue-challenge-by-password.yaml
    passkey: issue-challenge-by-passkey.yaml
    totp: issue-challenge-by-totp.yaml
    otp_sms: issue-challenge-by-otp-sms.yaml
    otp_email: issue-challenge-by-otp-email.yaml
    recovery_code: issue-challenge-by-recovery-code.yaml
    idp: issue-challenge-by-idp.yaml
    captcha: issue-challenge-by-captcha.yaml
oneOf:
  - $ref: issue-challenge-by-identifier.yaml
  - $ref: issue-challenge-by-password.yaml
  - $ref: issue-challenge-by-passkey.yaml
  - $ref: issue-challenge-by-totp.yaml
  - $ref: issue-challenge-by-otp-sms.yaml
  - $ref: issue-challenge-by-otp-email.yaml
  - $ref: issue-challenge-by-recovery-code.yaml
  - $ref: issue-challenge-by-idp.yaml
  - $ref: issue-challenge-by-captcha.yaml
```

api/openapi/endpoints/sessions/exchange/methods.yaml
livio-a: I've added it based on the api design docs: https://github.com/zitadel/nextgen/blob/af7a9e88f8e9e6cd16a687b68ace37cb9c0067bd/docs/design/api/conventions.md#category-b-one-time-auth-operations-with-replay-safe-semantics
It's not required to be used, but we could still offer it.

PR #46, opened 2026-04-30
feat: JSON Schema repository and resolver

muhlemmer:
Adds persistent storage for JSON Schemas (per instance, keyed by URL), a domain repository API, and a resolver that loads schemas from the database and optionally over HTTP, resolves $ref chains with depth limits, and caches fully resolved schemas in memory.

- Introduces `domain.JSONSchema` and J`SONSchemaRepository` (get/list/create/delete; schemas treated as *immutable* for validation workflows).
- Implements `repository.JSONSchemaRepository` against zitadel_nextgen.json_schemas (instance_id, url, created_at, payload), with migration `003_json_schemas` (and FK from user-related migration where applicable).
- Wires in `github.com/ianlancetaylor/jsonschema` and supporting pieces (e.g. httputil.GetJSON for JSON-over-HTTP fetch with content-type handling).
- Adds ADR 009 proposing how schemas are used for user validation and upgrade paths.

## Rough edges

- Lacking an internal error library, errors are wrapped `fmt.Errorf`. This needs to be improved in the futute, as schema resolution is recursive errors can become deeply nested.
- Current data model stores schemas on instance level only. Allthough schema validation isn't implemented yet, this basically means "any schema defined on instance level can be used to validate users"


docs/adrs/009-user-json-schema-validation.md
```diff
@@ -0,0 +1,81 @@
...
+
+### 2. Immutability and URL-based versioning
+
+Schemas stored at a given URL are **immutable**. To ship breaking or additive changes, register a **new** URL (for example `https://acme.com/schemas/user/v2/users.json`). Customers define their own URL layout and manage how they roll out breaking changes.
+
+This matches domain storage semantics: schemas are created and not updated in place; multiple versions coexist as distinct URLs.
+
+### 3. Trust boundary for resolution
+
+**Runtime validation** uses schemas loaded for the instance from storage (for example via `JSONSchemaResolver` without treating remote URLs as an authority for validation). Fetching a schema over HTTP is an **ingestion** path to populate the repository, not a substitute for trusting only instance-stored definitions at validation time.
```
wim07101993, line 31: Wasn't it a requirement that the customer would have the option to keep their schemas in a git repository, and reference them via url?

docs/adrs/009-user-json-schema-validation.md
```diff
@@ -0,0 +1,81 @@
...
+
+### 1. `$schema` on create and update
+
+When users are created or updated over the API, the request body includes a **`$schema`** property: a URL string that references a JSON Schema **already registered** for the instance (a row in `json_schemas`, same URL as the `url` column).
+
+The API value maps to the persisted user header: `users.schema_url` references `json_schemas (instance_id, url)` (see database migration). It is the same identifier, not a parallel field.
+
+### 2. Immutability and URL-based versioning
+
+Schemas stored at a given URL are **immutable**. To ship breaking or additive changes, register a **new** URL (for example `https://acme.com/schemas/user/v2/users.json`). Customers define their own URL layout and manage how they roll out breaking changes.
```
wim07101993, line 25: With registering a new URL do you mean that the customer can register a url at which they host their own schema? Or that they need to register a url as an "identifier" of the schema?

Letting the customer define their own schema url seems "dangerous". Since the schema needs to be fetched from that url, we would also need to host the schema at that url. Which is very prone to collisions unless we require some format in which the url has the be submitted.

wim07101993 (commented):
This a completely different approach than the approach in https://github.com/zitadel/nextgen/pull/15. I wonder how this could happen. The other pr has been open for about a week and has been discussed heavily with multiple team members. I wonder whether it has something to do with the communication issues that were mentioned in the last meeting

internal/domain/json_schema.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	if o.MaxSize == 0 {
+		o.MaxSize = DefaultMaxJSONSchemaSize
+	}
+}
+
+// Resolve retrieves a JSON schema by its URL recursively,
+// starting from the given schema URL and following all references in the schema.
+// It first tried to find the fully resolved schema in the cache.
+// If not found in the cache, it looks for the schema and all its references by URL from the database.
```
grvijayan, line 180: @muhlemmer, I am looking into validating customer-defined schemas against meta-schemas. Per my understanding, the meta-schemas would be embedded. Using this `Resolve` method defined here would mean storing meta-schemas also in the DB. Also, this feels more tailored to customer-defined schemas (meta-schemas aren't bound to a particular instanceID, right?). Should we separate resolving/compiling meta-schemas from this implementation? What do you think?

go.mod
```diff
@@ -8,6 +8,8 @@ require (
 	github.com/georgysavva/scany/v2 v2.1.4
 	github.com/go-faster/errors v0.7.1
 	github.com/go-faster/jx v1.2.0
+	github.com/hashicorp/golang-lru/v2 v2.0.7
+	github.com/ianlancetaylor/jsonschema v0.0.0-20251021232724-46ecbf32a0a5
```
wim07101993, line 11: Is there a reason why you used this json-schema package and not e.g. the one from google? https://github.com/google/jsonschema-go

As far as I can see, the current one does not support reading the schema, only validating it making it very hard to map from a `jsonschema.Schema` to the open-api generated one.

muhlemmer: No real reasons. Both libs were published in 2025 and the one I picked was the one that caught my eye at the time. The publisher is also a Go upstream maintainer, so I took it as a serious enough implementation.

Feel free to switch it out if you find that the google implementation is better.

docs/adrs/009-user-json-schema-validation.md
muhlemmer: We fetch by URL, so yes the schema needs to be available at that location. We do not need to re-host the schema on the URL. The URL is an absolute location, where all tools should still be able to retrieve the schema. Zitadel just caches the schema for internal computation. We demand immutability because there's no way of knowing if the schema at the source changed. It is better and safer to just declare that we never expect the source to change.

docs/adrs/009-user-json-schema-validation.md
muhlemmer: Yes, and that is what we do:

1. When the schema is "created" over the API using a ref URL we store that schema once in our DB. It is then considered immutable.
2. During validation of an object (like a user) we only use the schema stored (and validated) during creation. We will not re-fetch any refs during validation, they all need to be locally available.

internal/domain/json_schema.go
muhlemmer: Latest commit embeds and resolves builtin json schemas.

muhlemmer:
> This a completely different approach than the approach in #15. I wonder how this could happen. The other pr has been open for about a week and has been discussed heavily with multiple team members. I wonder whether it has something to do with the communication issues that were mentioned in the last meeting

@wim07101993  I'm not sure where this went wrong then. I've explicitly mentioned in the early all-hands meetings that I'll be implementing the Schema repository and resolver, because the Resolver tightly talks to storage. I've also mentioned in those meetings I'll use the work I've already done in flexcore as a base and when this is merged I will hand it over to the service layer.

I'm only really concerned about the storage access pattern. I'm not married to any of the other code, so feel free to rip it out and replace it. As long as there is reasonable caching and normalization of the data, it's fine by me.

PR #47, opened 2026-04-30
refactor: unify flow transition data model

vitorbari:
Aligns the flow transition data model across OpenAPI schemas, design docs, and the visualizer to use a single unified object format:

```json
{
  "target": "<step-name or flow-slug>",
  "action": "switch|pivot"  // nullable — omit for same-flow transitions
}
```

Previously, transitions used three different formats:
- String shorthand: `"submit": "resolve_user"`
- Pivot object: `"register": { "pivot": "register" }`
- Switch object: `"login": { "switch": "login" }`


vitorbari:
relates to #31

PR #50, opened 2026-04-30
refactor: simplify flow engine to schema-driven step model

vitorbari:
## Summary

Replace the type-driven step model with a schema-driven model where steps reference user schema properties by name and the engine resolves field metadata at runtime.

- Remove step `type` enum — steps are now differentiated by their properties (`fields`, `on_success`, `complete`) instead of an explicit type
- Change `fields` from a pre-resolved object dict to a string array referencing user schema properties (e.g. `["email", "password"]`)
- Add `on_success` (server-side mutation) and `complete` (terminal step behavior) as step-level properties
- Add `user_schema` to flow definitions, linking steps to a user schema
- Remove `behavior` from runtime step payload, replaced by `complete`
- Policy evaluation is implicit — the engine checks session assurance after every submit, no explicit `policy_check` step type needed

### Auth attempts / credential checks

The engine knows a field requires credential verification through the `x-credential` schema annotation. When a step submits a field marked `x-credential: "password"`, the engine performs the auth attempt challenge automatically. Failed attempts produce an `invalid_credential` implicit outcome. This follows the same pattern as `x-identifier` producing `user_not_found`.

### Out of MVP scope

Verification (email + phone) is not part of the MVP and has been removed from this PR:

- The `verify` step property was dropped.
- The `x-verify` schema annotation was removed from docs and the visualizer.
- Email/phone verification examples were removed from the engine guide and end-to-end examples.
- The visualizer's standalone email-verify-based "Recovery" flow was removed.

`recovery` is still listed as a flow purpose, and `reset_credential` is still a valid `on_success` mutation — these are forward-looking and do not strictly depend on verification.

### Visualizer

- Add "User Schema" tab showing the resolved schema for the current flow
- Add "Schema Resolution" panel in the simulator showing definition→runtime field resolution, schema annotations, and implied transition outcomes
- Flow definitions in the visualizer now use the definition format (field arrays, simplified actions) with runtime resolution happening in JS

---

### Before (login)

```json
{
  "steps": [
    {
      "name": "login",
      "type": "form",
      "fields": {
        // field metadata is duplicated from the schema into every step
        "email": { "type": "email", "required": true, "text_key": "login.field.email" },
        "password": { "type": "password", "required": true, "text_key": "login.field.password" }
      },
      "actions": {
        "submit": { "primary": true, "text_key": "login.action.submit" }
      },
      "transitions": { "submit": "check_policy" }
    },
    {
      // explicit policy check — author must wire it into every flow
      "name": "check_policy",
      "type": "policy_check",
      "transitions": { "satisfied": "done", "step_up": "mfa" }
    },
    { "name": "done", "type": "complete", "behavior": "redirect" }
  ]
}
```

### After (login)

```json
{
  "user_schema": "human_user",
  "steps": [
    {
      "name": "login",
      // just reference schema properties — engine resolves type, required, validation, text_keys
      "fields": ["email", "password"],
      // x-identifier on email → implicit user_not_found outcome
      // x-credential on password → implicit auth attempt + invalid_credential outcome
      // policy evaluation happens automatically after every submit
      "actions": { "submit": { "primary": true } },
      "transitions": { "submit": { "target": "done" } }
    },
    { "name": "done", "complete": "redirect" }
  ]
}
```

### Before (registration)

```json
{
  "steps": [
    {
      "name": "profile",
      "type": "form",
      "fields": {
        "email": { "type": "email", "required": true, "text_key": "profile.field.email" },
        "given_name": { "type": "text", "required": true, "text_key": "profile.field.given_name" },
        "family_name": { "type": "text", "required": true, "text_key": "profile.field.family_name" }
      },
      "transitions": { "submit": "set_password" }
    },
    {
      "name": "set_password",
      "type": "form",
      "fields": {
        "password": { "type": "password", "required": true, "text_key": "set_password.field.password" }
      },
      "transitions": { "submit": "create_user" }
    },
    {
      // separate step type for user creation
      "name": "create_user",
      "type": "action",
      "config": { "action": "create_user" },
      "transitions": { "success": "check_policy" }
    },
    {
      "name": "check_policy",
      "type": "policy_check",
      "transitions": { "satisfied": "done" }
    },
    { "name": "done", "type": "complete", "behavior": "show" }
  ]
}
```

### After (registration)

```json
{
  "user_schema": "human_user",
  "steps": [
    {
      "name": "profile",
      "fields": ["email", "given_name", "family_name"],
      "transitions": { "submit": { "target": "set_password" } }
    },
    {
      "name": "set_password",
      "fields": ["password"],
      "on_success": "create_user",   // creates user after password is set
      "transitions": { "submit": { "target": "done" } }
    },
    { "name": "done", "complete": "show" }
  ]
}
```

Field types, required, validation rules, and text_keys are resolved at runtime from the `human_user` schema annotations (`x-identifier`, `x-credential`, etc.).


docs/design/flowengine/flow-engine-guide.md
```diff
@@ -245,111 +240,152 @@ See [Template Security Model](template-security.md) for XSS attack vectors, trus
...
+{
+  "name": "verify_email",
+  "verify": "email",
+  "transitions": {
+    "submit": { "target": "done" }
+  }
+}
+```
+
+When the engine reaches this step, it sends a verification code to the user's email (from collected data). The step renders a code input field. On successful verification, the engine follows the `submit` transition.
```
livio-a: there already some documented: https://github.com/zitadel/nextgen/blob/main/docs/design/flowengine/user-schema.md#annotations-consumed-by-flow--policy

would `x-verify: "email"` still be required for the verify action?
And "x-credential` for any factor verification incl. OTP sms? I noted there's currently also `x-mfa: "sms"` in the list.

docs/design/flowengine/flow-engine.md
```diff
@@ -57,7 +57,7 @@ The server resolves which flow definition to use:
 
 ## Flow Definitions
 
-A flow definition is a directed graph of steps, managed as an API resource.
+A flow definition is a directed graph of steps, managed as an API resource. Each definition references a **user schema** via `user_schema` — step fields are property names from that schema, and the engine resolves field metadata (type, validation, implicit outcomes) from schema annotations at runtime.
```
livio-a, line 60: I'm unsure if a single schema is enough. I clearly see that for registration it should be ok, but for login i might have different schemas for local and federated users. Would we then split the flow and have a separate one for the local part and transition into a specific idp login flow which targets another schema?
But that would mean i need to configure sso in the flow even if the corresponding schema does not allow sso in teh x-auth-methods?

livio-a (commented):
Mostly LGTM, but I'm still not entirely clear on the relationship between the flow engine and the schema.

docs/design/flowengine/flow-engine-guide.md
vitorbari: I dropped the `x-verify` because it is out-of-scope for the MVP. Otherwise, we would have to build the underlying functionality which does not really belong to the flow engine.

docs/design/flowengine/flow-engine.md
vitorbari: good point. I would say we can keep one-to-one (flow->user-schema) for now until we have more clarity on this topic. I will take note of this.

docs/design/flowengine/api/flow-api.yaml
vitorbari, line 1: removing these files to avoid maintaining outdated api docs inside the docs directory.

livio-a (approved):

PR #52, opened 2026-05-01
feat: define and session repositories

adlerhurst:
Introduce the basic structure for authentication attempts and sessions, including repository definitions and necessary migrations. Adjust methods and fields in the `AuthAttemptRepository` and define domain objects for sessions and authentication checks. Ensure proper foreign key relationships and clean up domain definitions.


adlerhurst:
implemented in https://github.com/zitadel/nextgen/pull/130

PR #53, opened 2026-05-04
feat: add ULID based id generator

livio-a:
Introduces the internal/domain/idgen package for ID generation with a consistent, safe, and testable approach across all domain resources.

Resource IDs in this codebase are defined as prefixed, opaque, and time-sortable (docs/design/api/conventions.md). 

- Defines the Generator interface with a single New(prefix string) (string, error) method
- Provides a ULID implementation via [github.com/oklog/ulid/v2](https://github.com/oklog/ulid/) (Apache 2):
  - Uses ulid.Monotonic entropy to guarantee lexicographic ordering within the same millisecond
  - Guards the monotonic entropy source with a sync.Mutex (required — MonotonicEntropy is not goroutine-safe)
  - Injects now func() time.Time for deterministic testing
  - Validates prefix: rejects empty, strips trailing _ leniently 
- Generated typed mock via go.uber.org/mock/mockgen (already a project dependency) is added


internal/domain/idgen/generator.go
```diff
@@ -0,0 +1,15 @@
+// Package idgen provides ID generation for domain resources.
+// IDs are prefixed, opaque, and time-sortable (e.g. "att_01J0Z9KX7Y0Q2Y7JX5M9K2YF3C").
+// The project_id is explicitly excluded from this package — it uses a dictionary slug strategy.
+package idgen
+
+//go:generate mockgen -typed -package idgenmock -destination ./idgenmock/idgen.mock.go github.com/zitadel/nextgen/internal/domain/idgen Generator
```
wim07101993:
```suggestion
//go:generate go tool mockgen -typed -package idgenmock -destination ./idgenmock/idgen.mock.go github.com/zitadel/nextgen/internal/domain/idgen Generator
```
if https://github.com/zitadel/nextgen/pull/57 is merged first

wim07101993 (approved):

internal/domain/idgen/generator.go
livio-a: added it already incl. the go mod change

PR #54, opened 2026-05-04
feat: added initial user endpoints

wim07101993:


api/openapi/endpoints/users/methods.yaml
```diff
@@ -1,32 +1,83 @@
 get:
```
vitorbari, line 1: We need to provide some filtering/searching options, such as retrieving users of a specific type (human,ai agent, etc).

wim07101993: I was waiting on the user-repo merge to add those. But yes, those need to be added.

PR #60, opened 2026-05-11
feat: add flow definition schema

vitorbari:
Path and `$id` convention are based on #15


api/openapi/endpoints/schemas/examples/login.yaml
grvijayan, line 1: Should this also contain a `schema_uri`/`schema_version`  field to indicate which schema version to validate against? We could make it optional, and validate against the latest version if not provided.

vitorbari: The native json-schema way is to include a `$schema`, which is optional. 

If not that, we have to specify the schema via header or another convention where we can map flow definitions to flow schema versions.

grvijayan: In case of `$schema`, ogen generates the field in Go with the `$` symbol (`json:"$schema,omitempty"`), which looks weird. For the flow definition management APIs, I just added a `schema_uri` field to be included in the request body (we can also make it optional). We can decide what to go ahead with, and I can update accordingly.

grvijayan (approved):

PR #61, opened 2026-05-11
feat: validation rules for flow definition

vitorbari:
This doc contains the validation rules that can't be expressed via json-schema (#60)


docs/design/flowengine/implementation-validation.md
```diff
@@ -0,0 +1,86 @@
...
+  `sso_providers`, `gates`, `transitions.callback`.
+- A terminal step (`complete` set) has nothing else.
+- Every key in `actions` has a matching key in `transitions`.
+- Every key in `transitions` is either an action name declared in this
+  step's `actions`, or a reserved outcome: `user_not_found`, `callback`.
+- When `sso_providers` is non-empty, `transitions.callback` is defined.
+  The `sso` action itself is engine-handled and never appears in
+  `transitions`.
+- Every entry in `fields` resolves to a property in the referenced
+  `user_schema`. Requires the schema to be live-fetched.
```
grvijayan: Could you please elaborate on `live-fetched`?

Per the implementation [here](https://github.com/zitadel/nextgen/blob/main/internal/domain/json_schema.go#L177-L199), the understanding is that a user object is validated against a customer-defined schema that has already been resolved/validated/persisted in the cache/DB. So, is it fair to assume that the fetching here is also from the cache/DB, i.e., it was already resolved and validated before being used in the flow definition?

docs/design/flowengine/implementation-validation.md
```diff
@@ -0,0 +1,86 @@
...
+  that exists and is runtime-eligible in the same project.
+- Every step in `steps` is reachable from some entry point in
+  `initial_steps` by walking transitions. Catches dead steps — typos in
+  transition `target`s or steps left unwired.
+- Every non-terminal step has at least one outgoing transition.
+- Every cycle of steps has at least one exit transition to a terminal step
+  or another flow (`switch` / `pivot`). Prevents infinite loops. Pragmatic
+  check; full liveness is undecidable.
+- A step with an `x-identifier` field defines a `user_not_found` transition,
+  unless its `submit` target handles "found and not found" identically.
```
grvijayan: how would we validate this target: ```unless its `submit` target handles "found and not found" identically.```?

docs/design/flowengine/implementation-validation.md
vitorbari: Even remote schemas will have to be registered via API and pre-fetched by the schemas API, which means live-fetching does not make sense for flow-definition.

We can say:
> - Every entry in `fields` resolves to a property in the referenced `user_schema` which should have been pre-registered via the schemas API.

wdyt?

grvijayan: sounds good

docs/design/flowengine/implementation-validation.md
vitorbari: A `user_not_found` might be defined when the step contains a `x-identifier` field, otherwise, the `submit` transition will be used for both cases (found and not found).

From a validation perspective, `user_not_found` can only be used together with `x-identifier` fields

docs/design/flowengine/implementation-validation.md
```diff
@@ -0,0 +1,86 @@
...
+> **Future optimization.** Audience filtering currently scans every row's
+> JSONB. When the selector outgrows in-memory filtering, promote
+> `team_ids` / `app_ids` / `user_schema` to top-level columns with indexes —
+> `purposes` is the existing precedent.
+
+## Engine-checked rules
+
+### Definition
+
+- `name` unique within `(project_id, schema_version)`. **[DDL]**
```
grvijayan, line 41: minor clarification: does the `schema_version` here refer to the flow-definition schema version defined by nextgen?

docs/design/flowengine/implementation-validation.md
grvijayan: My question is more about the case where `user_not_found` transition is not defined for an `x-identifier` field. 

From what I understand, the `submit` would have a `target` that handles ```"found and not found" identically``` - how would we validate this handling (look up the target step and validate how it handles user found/not found), or is this validation not necessary?

vitorbari: Actually, we can do it like this:
- user found:
  - follows submit target

- user not found:
  - if "user_not_found" defined -> follows
  - else -> returns error / don't follow any transition

docs/design/flowengine/implementation-validation.md
vitorbari: Yes, it refers to the flow definition schema.
The problem is, there is no clear versioning at the moment: https://github.com/zitadel/nextgen/pull/60/changes#diff-7dc46bba60265742959079d4bb3c0f3fabaa2bc598be846cb6d8f5e21f4afea4
This is following https://github.com/zitadel/nextgen/pull/15/changes#diff-fb38c9ccea54cd59a6f42b1f975afe538e6307f4b75b2b48ac889474b22c6c2c

docs/design/flowengine/implementation-validation.md
vitorbari:
```suggestion
- Every entry in `fields` resolves to a property in the referenced `user_schema` which should have been pre-registered via the schemas API.
```

docs/design/flowengine/implementation-validation.md
```diff
@@ -0,0 +1,85 @@
...
+  that exists and is runtime-eligible in the same project.
+- Every step in `steps` is reachable from some entry point in
+  `initial_steps` by walking transitions. Catches dead steps — typos in
+  transition `target`s or steps left unwired.
+- Every non-terminal step has at least one outgoing transition.
+- Every cycle of steps has at least one exit transition to a terminal step
+  or another flow (`switch` / `pivot`). Prevents infinite loops. Pragmatic
+  check; full liveness is undecidable.
+- A step with an `x-identifier` field defines a `user_not_found` transition,
+  unless its `submit` target handles "found and not found" identically.
```
vitorbari:
```suggestion
- A step with an `x-identifier` field might define a `user_not_found` transition,
  otherwise an error will be return for "not found".
```

grvijayan (approved):

PR #63, opened 2026-05-11
docs: design project layout and responsibilities

livio-a:
Proposes to use a hexagonal architecture to have clear responsibilities of packages and make them easily testable.
Additionally adds a linter to enforce the design pattern.

Open question:
Can we get rid of the database dependency in the domain package?


docs/adrs/README.md
```diff
@@ -4,14 +4,16 @@ This directory contains architecture decision records (ADRs) for nextgen.
...
+| [010](010-cursor-based-pagination.md)            | Cursor-Based Pagination                   | All list endpoints in this API use **cursor-based pagination** via `page_token` (request) and `next_page_token` (response) instead of offset/limit.                                                                                                                                                                                         |
+| [011](011-project-layout-and-architure.md)       | Project Layout and Responsibilities       | We adopt a **Modular Hexagonal Architecture** using a flat layout within the `internal/` directory. All dependencies must flow strictly **inward** toward the Domain. Infrastructure details, specifically database executors and transaction types, must be abstracted or propagated via `context.Context` to keep Domain interfaces pure. |
```
vitorbari, line 19:
```suggestion
| [011](011-project-layout-and-architecture.md)    | Project Layout and Responsibilities       | We adopt a **Modular Hexagonal Architecture** using a flat layout within the `internal/` directory. All dependencies must flow strictly **inward** toward the Domain. Infrastructure details, specifically database executors and transaction types, must be abstracted or propagated via `context.Context` to keep Domain interfaces pure. |
```

.go-arch-lint.yml
```diff
@@ -0,0 +1,54 @@
...
+allow:
+  depOnAnyVendor: true # TODO: maybe restrict in the future
+
+components:
+  main:
+    in:
+      - "/"
+      - "cmd/**"
+  domain:
+    in: "internal/domain/**"
```
vitorbari, line 11: Should we enforce a flat domain package or allow for sub packages such as `internal/domain/flow`?

docs/adrs/011-project-layout-and-architecture.md
```diff
@@ -0,0 +1,44 @@
...
+│       └── database/      # INFRASTRUCTURE: DB Implementations
+│           └── repository # Postgres/SQL logic
+```
+
+### Layer Responsibility Table
+
+| Layer       | Logical Role   | Responsibility                                                      | Allowed Imports              |
+|:------------|:---------------|:--------------------------------------------------------------------|:-----------------------------|
+| **api**     | Adapters       | Translates JSON/XML into Service calls; handles HTTP/SAML statuses. | `service`, `domain`          |
+| **service** | Orchestration  | Manages business workflows and transaction boundaries.              | `domain`                     |
```
wim07101993, line 32: shouldn't services also allow for storage imports?

docs/adrs/011-project-layout-and-architecture.md
```diff
@@ -0,0 +1,44 @@
...
+| **domain**  | Core           | Owns entities, business invariants, and repository interfaces.      | utilities                    |
```
wim07101993, line 33: Why would the repo interface be declared in the domain layer? Since they won't interact with them. Wouldn't it make sense to make the service layer own the repo interfaces? Since the service layer will use them.

.go-arch-lint.yml
```diff
@@ -0,0 +1,54 @@
...
+  service:
+    in: "internal/service/**"
+  api:
+    in: "internal/api/**"
+  stubs:
+    in: "api/**"
+  storage:
+    in: "internal/storage/database/**"
+  httputil:
+    in: "internal/httputil/**"
```
vitorbari, line 21: This would usually be placed under `/pkg/httputil`. I have now a similar case for cookie encryption logic. It is not domain logic, but rather a reusable pkg. (there is already /packages in the root used for the frontend stuff, which can be confusing)

livio-a:
let's look after the MVP again

PR #65, opened 2026-05-11
feat: Goose as migration framework

IAM-marco:
## What problems are solved

The project had three separate infrastructure gaps that blocked further feature work:

1. **Postgres migrations** were driven by `jackc/tern/v2`, which requires a `*pgx.Conn` and has no support for Go-function migrations. This made it impossible to share a `*sql.DB`-based migration runner across both dialects.

2. **Spanner** was connected through PGAdapter (PostgreSQL wire protocol over pgx). PGAdapter already forces Spanner-specific SQL workarounds (no ENUMs, no GIN, no schema-scoped CHECKs), so the "write once, run on both" promise was already broken — while also forfeiting native GoogleSQL features. The migration system was a hand-rolled runner with no version tracking.

3. **Repository inconsistency** — `FlowDefinitionRepository` stored a client on the struct (preventing reuse across transactions); `AuthAttemptRepository` had `isSpannerClient()` guards in every method body; `jsonPayloadScanner` duplicated functionality already present in the generic `JSON[T]` scanner.

**Postgres: replace tern with goose**
- Migrations move from tern-format Go embed files to `pressly/goose/v3` SQL files under `migration/sql/` with `-- +goose Up/Down` annotations.
- Version tracking moves from `zitadel_nextgen.migrations` to `zitadel_nextgen.goose_db_version`.
- `Pool` and `Conn` bridge to `*sql.DB` via `stdlib.OpenDBFromPool` for goose compatibility.

**Spanner: full switch to GoogleSQL via `go-sql-spanner`**
- Drop PGAdapter (pgx); replace with `github.com/googleapis/go-sql-spanner` behind the `spanner_integration` build tag.
- Pool/Conn/Tx/Rows/Error reimplemented over `*sql.DB` and `*sql.Conn` from `database/sql` with `sqlscan` for struct collection.
- `$N` → `@pN` placeholder translation in the pool/conn layer so existing `StatementBuilder` output is accepted unchanged.
- Goose uses `DialectSpanner` + `-- +goose NO TRANSACTION` / `StatementBegin`/`End` guards for DDL.
- Embedded test container switches from the PGAdapter image to the GoogleSQL emulator; instance and database are created via the Spanner admin API on startup.

**Repository: consistent client-per-method pattern**
- `FlowDefinitionRepository`: removes stored `Client` field; all five methods now accept `client database.QueryExecutor` as an explicit parameter. `domain.FlowDefinitionRepository` interface updated to match.
- `AuthAttemptRepository`: split into `pgAuthAttempt` and `spannerAuthAttempt`. `NewAuthAttemptRepository(pool)` factory dispatches at construction time. All `isSpannerClient()` guards and `jsonPayloadScanner` are deleted; Spanner scan path uses `JSON[json.RawMessage]`.
- `JSONSchemaRepository`: `NewJSONSchemaRepository(client)` detects dialect and configures table name, `NOW` vs `CURRENT_TIMESTAMP()` instruction, and payload encoder (`[]byte` for Postgres, `string` for Spanner JSON columns). `jsonSchemaRow.Payload` switches to `JSON[json.RawMessage]`.

**json_schemas migration**
- Converts the tern-format `003_json_schemas/` to goose for Postgres; adds an equivalent GoogleSQL migration for Spanner.
- Renumbers `json_schemas` to position 3 (ahead of `users`, which FKs to it).

**Glossary renames (LOCKED)**
Per the [design glossary](../docs/design/glossary.md):
- DB tables: `instances` → `projects`, `organizations` → `teams`
- DB columns: `instance_id` → `project_id`, `organization_id` → `team_id`
- Go types and fields: `JSONSchema.InstanceID` → `ProjectID`, `InstanceID()` → `ProjectID()`, `InstanceIDCondition()` → `ProjectIDCondition()`
- Applied to all migration SQL files, repository code, domain types, and tests.


internal/domain/json_schema.go
```diff
@@ -129,7 +129,8 @@ func jsonSchemaResolverCacheKey(instanceID, schemaURL string) string {
 type JSONSchemaResolver struct {
 	repository JSONSchemaRepository
 	// cache of fully resolved JSON schemas,
-	// keyed by instanceID and schemaURL
+	// keyed by projectID and schemaURL
+
```
adlerhurst:
```suggestion
	// keyed by projectID and schemaURL
```

internal/storage/database/dialect/postgres/conn.go
```diff
@@ -4,13 +4,15 @@ import (
...
 	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/jackc/pgx/v5/stdlib"
 
 	"github.com/zitadel/nextgen/internal/storage/database"
 	"github.com/zitadel/nextgen/internal/storage/database/dialect/postgres/migration"
 )
 
 type pgxConn struct {
 	*pgxpool.Conn
+	pool *pgxpool.Pool
```
adlerhurst: wouldn't it make sense to just pass `Pool` instead of `pgxpool.Pool` and call the migrate method of `Pool`?

adlerhurst (commented):
I think for the moment the implementation is fine. But later we should think of splitting the repos into dialects analog to the migrations because the code feels quite hard to read at the moment.

IAM-marco:
> I think for the moment the implementation is fine. But later we should think of splitting the repos into dialects analog to the migrations because the code feels quite hard to read at the moment.

Agreed. I kinda did that for auth attempts as the queries are very different from each other. The driver specific implementations of the auth attempt repository are in the same file though, but can be split.

adlerhurst (approved):

PR #67, opened 2026-05-12
feat(api): auth attempt

livio-a:
## Auth Attempt API

This PR implements the auth attempt API — the interactive sign-in flow surface for nextgen. An auth attempt is the ephemeral bundle that collects challenge/proof pairs (identifier lookup, password, passkey) before handing off to a session.

The work follows the model defined in ADR 010 and ADR 013.

---

### What's included

**Storage layer**
- `AuthAttemptRepository` with full DB schema for attempts, challenges, and factors
- `SetChallenge`, `ChallengeSucceeded`, `ChallengeFailed`, `Handoff` repository methods

**Domain layer** (domain)
- `AuthAttempt` aggregate with state machine: in-progress → completed → handed-off / expired
- `AuthChallenge` and `AuthFactor` discriminated unions for identifier, password, and passkey checks
- WebAuthn challenge creation and assertion verification via `go-webauthn/webauthn` (user_passkey.go)

**Service layer** (auth_attempt.go)
- `AuthAttemptService` interface with five operations:
  - `Create` — new attempt, optionally seeded from an existing session (step-up auth)
  - `GetByID` — load attempt with all challenges and completed factors
  - `IssueChallenge` — generate a challenge for identifier / password / passkey
  - `VerifyProof` — verify the submitted proof; records failure counts; marks attempt complete when all required checks pass
  - `Handoff` — mint a single-use handoff token on a completed attempt

**API layer** (api)
- REST handlers wired to the above service: `POST /auth_attempts`, `GET /auth_attempts/{id}`, `POST /auth_attempts/{id}/challenges`, `POST /auth_attempts/{id}/challenges/{challengeId}/verify`, `POST /auth_attempts/{id}/handoff`
- Error mapping with structured domain error codes
- `auth_check.go` — helpers for mapping between domain checks and the OpenAPI `CompletedFactor` / `ChallengeResponse` types

**OpenAPI spec**
- Added `user-password-invalid` error schema
- Corrected `expires_at` field to nullable (`OptNilDateTime`)
- Minor cleanups (trailing newlines) across check component files

---

### What is not yet done / known gaps

| Area | Status |
|---|---|
| **Passkey challenge payload** | `CreatePasskeyChallenge` builds the WebAuthn session data but does not yet serialize and return the full assertion options to the client (`TODO: implement` stub). The challenge ID and RPID are returned; the `publicKey.challenge` bytes block is not yet wired back to the API response. |
| **Passkey RP origins on verify** | `VerifyPasskeyChallenge` configures only a single RP origin (`challenge.RPOrigin`) instead of the full list. Multi-origin setups will fail verification until this is fixed. |
| **Discoverable passkey (no prior identifier)** | Issuing a passkey challenge without a preceding identifier check is intentionally blocked (`domain/auth_attempt.go`). Discoverable/conditional UI login is deferred until a policy gate is in place to prevent user-enumeration. |
| **Project default required checks** | `Create` does not yet load the project's `DefaultRequiredChecks` (commented-out TODO in the service). Callers must pass explicit `RequiredChecks`; falling back to project policy is not wired. |
| ~~**Session exchange endpoint**~~ | ~~`POST /sessions/exchange` — the endpoint that converts a completed attempt's handoff token into a live session — is not part of this PR. The full sign-in round-trip cannot complete without it.~~ |
| **Challenge nonce** | The `challengeNonce` field on `CreateAuthAttempt` is accepted in the request but silently ignored. |
| **Integration tests for auth attempt endpoints** | No `auth_attempt_test.go` exists yet. The existing integration test helpers were refactored (`EnsureAPIClient` now takes a `projectID`) but no new end-to-end coverage was added. |
| **Security / token validation** | security.go still carries `TODO: add proper token validation`. Auth attempt endpoints are project-scoped but token validation is incomplete. |

---

### Testing done

- Existing integration tests (project, schema, team) pass after the `EnsureAPIClient` signature change
- `go vet zitadel.` passes

---

## original description

- auth attempt API implementation
- added domain driven / defined errors
  - added a generator to build openapi schemas for the errors
  - added a generator to build a list of expected errors for each endpoint (to be used in the default response of such)
  -added interceptors to map the domain errors to HTTP errors 
- ...


api/openapi/endpoints/auth_attempts/auth-attempt-response.yaml
```diff
@@ -76,10 +76,12 @@ properties:
     items:
       $ref: by_id/challenges/challenge-response.yaml
   expires_at:
-    type: string
-    format: date-time
-    description: When this attempt expires (15-minute TTL from creation)
-    example: '2026-04-28T15:45:00Z'
+    oneOf:
+      - type: 'null'
```
vitorbari: When can it be null?

internal/api/auth_attempt.go
```diff
@@ -1,12 +1,231 @@
...
+
+	attempt, err := h.authAttemptService.Create(ctx, input)
+	if err != nil {
+		return nil, err
+	}
+	return authAttemptToAPI(attempt), nil
+}
+
+func (h Handler) GetAuthAttempt(ctx context.Context, params api.GetAuthAttemptParams) (api.GetAuthAttemptRes, error) {
+	scopeCtx, _ := GetScopeContext(ctx)
```
vitorbari, line 29: This is the same we had for sessions, and later we changed to project_id as parameter.

internal/api/auth_attempt.go
```diff
@@ -1,12 +1,231 @@
...
+	case api.FactorMethodIdentifier:
+		return domain.AuthCheckTypeUser, nil
+	case api.FactorMethodPassword:
+		return domain.AuthCheckTypePassword, nil
+	case api.FactorMethodPasskey:
+		return domain.AuthCheckTypePasskey, nil
+	default:
+		return domain.AuthCheckTypeUnspecified, domain.ErrAuthAttemptInvalidRequest()
+	}
+}
```
vitorbari, line 166: 💅 this could be a static map

api/openapi/endpoints/auth_attempts/auth-attempt-response.yaml
adlerhurst: it was nullable before and therefore i thought there might be a case but i cannot think of one so i remove it.

internal/api/auth_attempt.go
adlerhurst: the [HandleOAuth](https://github.com/zitadel/nextgen/blob/08cc77179d043d7b024b1d8fc5bb37136279888a/internal/api/security.go#L24)-method got adjusted. should i still add the project_id parameter?

vitorbari: I think it is fine then. We can also remove it from the other ones later.

vitorbari (approved):

PR #71, opened 2026-05-12
feat: schema api implementation 

wim07101993:


internal/service/schema_service.go
```diff
@@ -0,0 +1,136 @@
...
+
+	err = s.schemaValidator.ValidateAgainstMetaSchema(input.Schema)
+	if err != nil {
+		return nil, domain.ErrJSONSchemaInvalid().WithParent(err)
+	}
+
+	err = tx.Commit(ctx)
+	if err != nil {
+		return nil, domain.ErrInternal(err).WithMessage("failed to commit transaction")
+	}
```
IAM-marco, line 96: maybe you should use `tx.End()`

IAM-marco: And maybe it should be in the defer at the top

internal/service/schema_service.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}()
+
+	strUri := input.URL.String()
+	_, err = s.schemaResolver.Resolve(ctx, tx, input.ProjectID, strUri, nil)
+	if err != nil {
+		return nil, domain.ErrInternal(err).WithMessage("failed to resolve schema when creating")
+	}
+
+	err = tx.Commit(ctx)
```
IAM-marco, line 118: As before

internal/domain/schema_validator.go
```diff
@@ -108,7 +108,7 @@ func (v *SchemaValidator) ValidateAgainstMetaSchema(tenantSchemaBytes []byte) er
 		return fmt.Errorf("%w: %w", ErrSchemaParseFailed, err)
 	}
 
-	schemaID, _ := tenantSchema["$schema"].(string)
+	schemaID, _ := tenantSchema["metaSchema"].(string)
```
IAM-marco: Should `tenant` be removed ?

internal/api/integration_test/schema_test.go
```diff
@@ -0,0 +1,172 @@
...
+			schema string
+		}{
+			{
+				name:   "user-schema in body",
+				schema: harness.Schemas.CreateSchemaRequestUserSchema,
+			},
+			//{
+			//	name: "user-schema by url",
+			//	req:  harness.Schemas.CreateSchemaRequestUserSchemaByUrl,
+			//},
```
IAM-marco, line 37: Is this needed?

internal/api/integration_test/schema_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+			httpClient := harness.EnsureHttpClient(t)
+			resp, err := httpClient.Do(req)
+			assert.NoError(t, err)
+			defer func() { _ = resp.Body.Close() }()
+
+			if !assert.Equal(t, http.StatusBadRequest, resp.StatusCode) {
+				bs, err := io.ReadAll(resp.Body)
+				require.NoError(t, err)
+				t.Fatal(string(bs))
```
IAM-marco: I would have used assert/require

internal/api/integration_test/helpers/auth_attempt.go
```diff
@@ -0,0 +1,37 @@
+//go:build integration
```
IAM-marco: this file is mispelled

internal/api/integration_test/helpers/harness.go
```diff
@@ -0,0 +1,45 @@
...
+	DBPool     database.Pool
+	HttpClient *http.Client
+	TestServer *httptest.Server
+
+	GeneratedServer *generated.Server
+	Handler         *api.Handler
+	SecurityHandler *api.SecurityHandler
+
+	APIClient          *generated.Client
+	FakeSecuritySource *FakeSecuritySource
```
IAM-marco, line 27: I think this should be `test_data.FakeSecuritySource`

internal/service/schema_service.go
```diff
@@ -0,0 +1,136 @@
...
+
+	_, err = s.schemaResolver.Resolve(ctx, tx, input.ProjectID, input.SchemaID, nil)
+	if err != nil {
+		return nil, domain.ErrInternal(err).WithMessage("failed to resolve schema when creating")
+	}
+
+	err = s.schemaValidator.ValidateAgainstMetaSchema(input.Schema)
+	if err != nil {
+		return nil, domain.ErrJSONSchemaInvalid().WithParent(err)
+	}
```
IAM-marco: **Claude:** The current order is: insert to DB → resolve into LRU cache → validate against meta-schema. If validation fails, the DB transaction is rolled back — but the compiled schema is already in the resolver's LRU cache. Two consequences:

1. A subsequent `CreateSchemaByUrl` call for the same URL gets a cache hit, writes nothing to DB in the new transaction, then `GetByID` finds no row and fails.
2. A retry `CreateSchema` with corrected bytes silently reuses the stale cached compiled schema for reference resolution.

Suggested fix: run `ValidateAgainstMetaSchema` **before** `schemaRepo.Create` and `Resolve`. Invalid schemas then never touch the cache or the DB.

internal/service/schema_service.go
```diff
@@ -0,0 +1,136 @@
...
+	if err := json.Unmarshal(input.Schema, &tenantSchema); err != nil {
+		return nil, domain.ErrJSONSchemaInvalid().WithParent(err)
+	}
+
+	model := &domain.JSONSchema{
+		ProjectID: input.ProjectID,
+		URL:       input.SchemaID,
+		CreatedAt: time.Now().UTC(),
+		Schema:    input.Schema,
+	}
```
IAM-marco, line 74: **Claude:** `CreateSchemaInput.TeamID` is assigned from `params.TeamID` in the handler but is never stored in `domain.JSONSchema` (which has no `TeamID` field). Neither `schemaRepo.Create` nor `schemaRepo.GetByID` apply team scope, so all schemas are shared across teams within a project. Either add `TeamID` to the domain model + repository, or remove the field from the input structs and document that team scoping is deferred.

internal/service/schema_service.go
```diff
@@ -0,0 +1,136 @@
...
+	err = tx.Commit(ctx)
+	if err != nil {
+		return nil, domain.ErrInternal(err).WithMessage("failed to commit transaction")
+	}
+
+	return s.schemaRepo.GetByID(ctx, s.pool, input.ProjectID, strUri)
+}
+
+func (s *SchemaService) GetSchema(ctx context.Context, projectID string, teamID string, schemaID string) (*domain.JSONSchema, error) {
+	schema, err := s.schemaRepo.GetByID(ctx, s.pool, projectID, schemaID)
```
IAM-marco, line 127: **Claude:** `teamID` is accepted by `GetSchema` but never forwarded to `schemaRepo.GetByID`. Any caller with the correct `projectID` and `schemaID` can retrieve a schema regardless of team, bypassing team-level access control.

internal/service/schema_service.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return nil, domain.ErrInternal(err).WithMessage("failed to resolve schema when creating")
+	}
+
+	err = tx.Commit(ctx)
+	if err != nil {
+		return nil, domain.ErrInternal(err).WithMessage("failed to commit transaction")
+	}
+
+	return s.schemaRepo.GetByID(ctx, s.pool, input.ProjectID, strUri)
```
IAM-marco, line 123: **Claude:** `Resolve` checks the LRU cache before touching the database. If a previous `CreateSchema` call for the same URL populated the cache but then rolled back (e.g., meta-schema validation failure), `Resolve` returns a cache hit and writes nothing to the DB inside this new transaction. The empty transaction commits successfully, but the following `GetByID` finds no row and returns an error. The caller receives an unexpected failure for what appeared to be a successful URL resolution.

internal/api/schema.go
```diff
@@ -0,0 +1,82 @@
...
+	case domain.ErrJSONSchemaAlreadyExists().Code:
+		return errorResponseWithStatusCode(http.StatusConflict, err)
+	case domain.ErrJSONSchemaInvalid().Code:
+		return errorResponseWithStatusCode(http.StatusBadRequest, err)
+	default:
+		return internalErrorResponse(err)
+	}
+}
+
+var UnknownSchemaKindError = errors.New("unknown kind of schema")
```
IAM-marco: Can be unexported

internal/api/schema.go
```diff
@@ -0,0 +1,82 @@
...
+	default:
+		return nil, UnknownSchemaKindError
+	}
+}
+
+func (h *Handler) GetSchemaById(ctx context.Context, params api.GetSchemaByIdParams) (api.GetSchemaByIdRes, error) {
+	schema, err := h.schemaService.GetSchema(ctx, string(params.ProjectID.Value), string(params.TeamID.Value), params.ID)
+	if err != nil {
+		return nil, err
+	}
```
IAM-marco, line 53: **Claude:** Returning `nil, err` on a not-found error routes through `NewError` → `errorResponse` → `*api.ErrorDetailsStatusCode` (generic). The integration test at line 165 of `schema_test.go` asserts `*api.GetSchemaByIdNotFound`. Even if the client-side decoder maps the 404 correctly today, returning the typed response makes the handler's contract explicit and keeps it consistent with how ogen handlers are expected to work:

```go
if err != nil {
    var domErr domain.Error
    if errors.As(err, &domErr) && domErr.Code == domain.ErrJSONSchemaNotFound().Code {
        return &api.GetSchemaByIdNotFound{...}, nil
    }
    return nil, err
}
```

internal/api/integration_test/helpers/schema.go
```diff
@@ -0,0 +1,91 @@
...
+	if h.SchemaResolver == nil {
+		cache, err := lru.New2Q[string, *jsonschema.Schema](100)
+		require.NoError(t, err)
+
+		h.SchemaResolver = domain.NewJSONSchemaResolver(
+			h.SchemaRepo,
+			cache,
+			0,
+			0,
+			h.EnsureHttpClient(t),
```
IAM-marco, line 52: **Claude:** `builtinPublicBase` is `nil` here, disabling builtin schema serving in the resolver. The validator is initialized with `BuiltinSchemaBaseURL`, so a schema that references a builtin URL will pass validation but fail during `Resolve` with a confusing internal error. Consider passing `BuiltinSchemaBaseURL` (parsed as a `*url.URL`) to the resolver to keep the two components consistent.

internal/service/schema_service.go
IAM-marco: Disregard if team was scrapped from the PoC

IAM-marco (changes requested):
I prefixed the comments done by Claude with `Claude:`. Some may not be relevant, although I went through them.

internal/service/schema_service.go
wim07101993: wouldn't `End` swallow the error because it would be in a `defer` block?

internal/api/integration_test/schema_test.go
wim07101993: it is a placeholder for when we have a public repo from which to fetch a schema. Will add a todo

internal/api/integration_test/schema_test.go
wim07101993: for what? I swapped out `t.Fatal` for `t.Log` since it more reflects what the statement is for

internal/api/integration_test/helpers/auth_attempt.go
wim07101993: good catch

internal/api/integration_test/helpers/harness.go
wim07101993: what do you mean? `FakeSecuritySource` does not exist in the test_data package as it is not data

internal/service/schema_service.go
wim07101993: as I understand, currently schemas are scoped by project, not team. Might change in the future.

internal/service/schema_service.go
wim07101993: as I understand, currently schemas are scoped by project, not team. Might change in the future.

internal/api/integration_test/schema_test.go
IAM-marco: Preference 😆 . I just don't like using t.Error/Fatal/Log 

I used assert and require extensively in the past.

internal/api/integration_test/helpers/harness.go
IAM-marco: Uhm, I was getting a compile error 🤔

internal/service/schema_service.go
IAM-marco: Not sure what you mean.

If you do:

```go
func (s *SchemaService) CreateSchema(ctx context.Context, input CreateSchemaInput) (_ *domain.JSONSchema, err error) {
	tx, txErr := s.pool.Begin(ctx, nil)
	defer func ()  {
        // (2)
		err = tx.End(ctx, err)
	}()

    // (1)
 	err = s.schemaRepo.Create(ctx, tx, model)
    if err != nil {
        return nil, err
    }
}
```

In case (1) returns an error:
  - `tx.End()` is called with `err != nil` -> Rollbacks the tx and sets `err = err` . `err` is returned since its a named returned value

In case (1) returns no error:
  - `tx.End()` is called with `err == nil` -> Attempts commit:
    - If commit returns an error -> `tx.End()` returns that error, so `err = err`. `err` is returned since its a named returned value
    - If commit returns NO error -> `tx.End()` returns `nil`, so `err = nil`. Function returns `nil`

IAM-marco (approved):

internal/service/schema_service.go
wim07101993: Hadn't thought of that yet, changing the response value using named responses from within the defer block. Will fix

wim07101993: After giving it a quick try, I decided not to fix it because the err handling in all the `if err != nil` blocks would get bigger. One line, but times 7 is quiet a lot imho.

PR #72, opened 2026-05-12
feat(api): open api spec for flow definition management

grvijayan:
This PR API definition to manage “flow definitions” and updates the ogen-generated Go server/client stubs accordingly.

Changes:
- Adds /flow_definitions endpoints to create, list flow definitions
- Adds /flow_definitions/{id} endpoints to get, update, delete flow definitions
- Adds OpenAPI component schemas for flow definition request/response shapes
- Regenerates ogen output


api/openapi/components/flows/flow-definition-create-request.yaml
```diff
@@ -0,0 +1,8 @@
+allOf:
```
vitorbari: With this approach, both `project_id` and the flow definition will share the same object, which makes it  feel like the `project_id` is part of the schema.

Maybe we can move the flow definition one level deeper. The payload would be like this:
```json
{
"flow_definition": {}
"project_id": ""
}
```

wdyt?

api/openapi/components/flows/flow-definition.yaml
```diff
@@ -10,34 +10,41 @@ description: |
...
-      reference other flows. Must be unique within the project. Renaming
-      the flow does not change the slug.
-    example: default-login
+      Stable identifier for this flow, used as the target of cross-flow
+      `switch` and `pivot` transitions. Renaming is not supported — the
+      `name` is part of the public contract another definition may
+      reference. Acts as the human display label as well; no separate slug.
 
-  name:
+  schema_uri:
```
vitorbari: This isn't part of the flow-schema, maybe it can go outside together with the project id?

api/openapi/endpoints/flow_definitions/by_id/methods.yaml
```diff
@@ -0,0 +1,100 @@
...
+      application/json:
+        schema:
+          $ref: ../../../components/flows/flow-definition-update-request.yaml
+  responses:
+    '200':
+      description: Flow definition updated successfully.
+      content:
+        application/json:
+          schema:
+            $ref: ../../../components/flows/flow-definition-detail-response.yaml
```
vitorbari, line 55: Would it be possible to use same response shape for the POST endpoint?

grvijayan: yes, i wasn't sure if we wanted to include steps, purposes, etc in the POST response.

grvijayan: we could also use the same in the list response as well if that's what we want

vitorbari: That is fine. Less flow representations for us to maintain. In the long-term, we can revisit this and try to optimize it, probably, listing endpoint does not need to full representation.

grvijayan: updated

vitorbari (approved):

PR #82, opened 2026-05-13
feat: flow field resolver

vitorbari:
## Summary

This PR introduces the `FlowFieldResolver` interface and a schema-driven implementation. The flow state machine will call this to turn `step.fields[]` (an array of property names) into fully-resolved field payloads at render time, and to validate submitted values at submit time.

The resolver loads the user schema through `domain.JSONSchemaResolver` (which handles caching, `$ref` resolution, and embedded built-ins) and derives every `FlowField` trait from meta-schema keywords. 

## Alignment with the user meta-schema

| Resolver surface | Meta-schema source |
|---|---|
| `FlowField.Required` | schema-level `required[]` array |
| `FlowField.Type` | property `type` + `format`; `x-password: true` forces password input |
| `FlowField.Challenge = identifier` | property `x-identifier: true` |
| `FlowField.Challenge = password` | property `x-password: true` + schema-level `x-auth-methods.password.enabled` |
| `FlowField.Unique` | property `x-unique` (enum: `instance` / `organization`) |
| `FlowFieldValidation.Format` / `MinLength` / `MaxLength` | property `format` / `minLength` / `maxLength` |
| Implicit `user_not_found` outcome | any property with `x-identifier: true` |

**Annotations defined by the meta-schema but not yet surfaced** (`x-claim`, `x-editable`, `x-sensitive`, `x-mfa`, `x-verify`) will land as the consumers that need them do. JSON Schema's `pattern` keyword is intentionally not surfaced — the user meta-schema does not expose it. Other credential kinds (passkey, magic_link, sso, otp) have no user-property-shaped proof and are produced by the state machine as challenge steps, not by this resolver.


api/openapi/endpoints/schemas/user-property.yaml
```diff
@@ -29,6 +29,11 @@ properties:
     type: boolean
     default: false
     description: Whether this property is an identifier for the user or not
+  x-password:
```
vitorbari: We need this to map a user property with a auth-attempt challenge `"method": "password"`, regardless of how it is called.

api/openapi/endpoints/schemas/user-property.json
```diff
@@ -63,6 +63,11 @@
       "default": false,
       "description": "Whether this property is an identifier for the user or not"
     },
+    "x-password": {
```
vitorbari, line 67: We need this to map a user property with a auth-attempt challenge "method": "password", regardless of how it is called.

wim07101993: @muhlemmer mentioned to keep auth related fields (passwords, otp secrets,...) out of the schema.

internal/domain/flow_field_resolver_schema.go
```diff
@@ -0,0 +1,338 @@
...
+		ImplicitOutcomes: implicit,
+	}, nil
+}
+
+// Validate runs the schema-derived rules over the submitted values.
+// Required-on-empty-string and unknown-property checks are layered on
+// top of the library's per-keyword validators (the library treats
+// `required` as a presence check, not an emptiness check, and treats
+// `format` as an annotation unless an external validator is registered).
+func (r *SchemaFieldResolver) Validate(
```
wim07101993: It feels weird to have to manually validate the object. Gut feeling says that since we are using schema's, it should be possible to do this automatically. I do see why it isn't and don't have a solution to fix it, but it feels weird. And a bit dirty since it will be both manual validation and automatic validation which will validate the object in the end, I guess.

vitorbari: Agreed.
We need a mechanism that can actually validate and enforce the UserProperty, for example, the `format (email,uuid,etc)`, `minLength`, `maxLength`, etc.

From the point of view of the Flow, it is not always the whole user object, but the validation of individual fields against the semantics defined in the user-schema.

api/openapi/endpoints/schemas/user-property.json
vitorbari: Can you please provide more context?

My undesrtanding was that we would a use these annotations (`x-identifier`, `x-password`, etc) to understand the semantics of each field, and handle them accordingly in the different layers of the applications. 

For example, in the storage layer:
```yaml
properties:
  email:
    type: string
    x-identifier: true
    format: email
    description: The user's email address.
  password:
    type: string
    x-password: true # don't save it as an attribute
    minLength: 8
    description: The user's password, which must be at least 8 characters long.
  firstName: # normal user attribute
    type: string
    description: The user's first name.
```

From the flow engine perspective, it has to tie the properties with auth-attempt challenge methods (e.g. `x-password` triggers a proper auth-attempt password challenge)

wim07101993: @muhlemmer could you elaborate on the idea?

internal/domain/flow_field_resolver_schema.go
```diff
@@ -0,0 +1,263 @@
...
+
+var _ FlowFieldResolver = (*SchemaFieldResolver)(nil)
+
+func (r *SchemaFieldResolver) Resolve(
+	ctx context.Context,
+	client database.QueryExecutor,
+	projectID, userSchemaURL string,
+	fieldNames []string,
+) (FlowResolvedFields, error) {
+	schema, err := r.schemas.Resolve(ctx, client, projectID, userSchemaURL, nil)
```
grvijayan, line 56: At the moment, `JSONSchemaResolver.Resolve` first tries to get the tenant-defined schema from DB. And if it's not available, it falls back to HTTP fetch if the `httpClient` is set in the resolver.

In this case though, we want to get the user schema that's already persisted in the DB (which would have been validated during creation), right? I have the same dilemma in flow definition as well, and was wondering if we should add a `JSONSchemaResolver.ResolveFromStorage` which doesn't allow HTTP-fetch. What do you think?

cc: @muhlemmer as you know the Resolver better :D

PR #87, opened 2026-05-14
feat: minimal Project API implementation

IAM-marco:
## What problems are solved

`POST /projects` and `GET /projects/{project_id}` were declared in the OpenAPI spec and had ogen-generated stubs, but no implementation. Any call returned 501 Not Implemented. There was also no service layer for project operations and the projects table lacked the secret columns required by the design.

**Database migration** (`000001_projects.sql` — Postgres & Spanner)
- Added `project_secret`, `preview_secret` (bearer tokens, `NOT NULL`, `CHECK (<> '')`) and `preview_origins` (JSON-encoded `[]string`, `NOT NULL`) columns to the projects table.
- `preview_origins` is stored as JSON text rather than a native array for cross-dialect portability (the existing `JSONArray[string]` scanner handles round-tripping).
- Both dialects use `CREATE TABLE IF NOT EXISTS` for idempotency.

**Domain model** (`internal/domain/project.go`)
- Added `ProjectSecret`, `PreviewSecret`, and `PreviewOrigins` fields to `Project`.
- Updated the `Create` contract: callers must pre-populate the three secret fields; the repository continues to own timestamps.

**Repository** (`internal/storage/database/repository/project.go`)
- `Create` now inserts the three new columns.
- `Get` selects and scans them back via `JSONArray[string]` for `preview_origins`.
- Nil `PreviewOrigins` is normalised to `[]string{}` before marshalling so the DB never stores `null`.

**Service** (`internal/service/project.go`)
- New `ProjectService` interface with `Create` and `Get`.
- `Create` generates a `proj_<ulid>` project ID and two `sk_proj_<ulid>` bearer secrets via `idgen.Generator`, then delegates to the repository. Returns the stored record (with DB-set timestamps) via a subsequent `Get`.

**Handler** (`internal/api/handler.go`, `internal/api/project.go`)
- `Handler` gains a `projectService` field; `NewHandler` updated accordingly.
- `CreateProject`: anonymous (security block removed from spec, client regenerated), returns 201 with `id`, `projectSecret`, `previewSecret`, `previewOrigins`, `createdAt`.
- `GetProject`: maps `NoRowFoundError` → 404; returns `id`, `createdAt`, `updatedAt`.
- Also fixes a stale field reference in `CreateFlow` (`req.Slug` → `req.FlowDefinitionName`) that was preventing compilation after the last codegen run.

**Unit tests** (`internal/service/project_test.go`)
- Table-driven tests for `ProjectService.Create` and `Get` using gomock for `ProjectRepository` and `idgen.Generator`.
- Uses `testify/require` and `testify/assert` throughout.

**Integration tests** (`internal/api/integration_test/`)
- Spins up a real ogen HTTP server against an embedded database and exercises `POST /projects` and `GET /projects/{id}` end to end.
- Shared test logic under `//go:build integration || spanner_integration`; dialect setup in separate files (`setup_pg_test.go` / `setup_spanner_test.go`).
- Covers: happy path create, create with preview origins, get existing, get not found, get unauthorized.

**CI** (`.github/workflows/ci.yml`)
- Adds `go-integration-test` (Postgres, embedded) and `go-integration-test-spanner` as parallel jobs alongside the existing unit-test and Spanner repository jobs.
- `goreleaser-snapshot` waits on all four.


internal/api/project.go
```diff
@@ -0,0 +1,53 @@
...
+	"errors"
+
+	api "github.com/zitadel/nextgen/api/generated"
+	"github.com/zitadel/nextgen/internal/storage/database"
+)
+
+func (h Handler) CreateProject(ctx context.Context, req *api.CreateProjectRequest) (api.CreateProjectRes, error) {
+	project, err := h.projectService.Create(ctx, req.PreviewOrigins)
+	if err != nil {
+		return nil, &api.ErrorDetailsStatusCode{
```
wim07101993: if we want to return a generic error, you could use the `NewError` function on the `Handler`. If a method returns an `error`, the `NewError` method is called to transform the error to an `*api.ErrorDetailsStatusCode`. I think the default behavior of that function can be to return a 500. 

Maybe all generic errors can be handled this way, e.g. the `database.NoRowFoundError` can also result in a 404.

internal/service/project_test.go
```diff
@@ -0,0 +1,190 @@
...
+		setupIDs       func(*idgenmock.MockGenerator)
+		setupRepo      func(*domainmock.MockProjectRepository)
+		wantErr        bool
+		check          func(t *testing.T, got *domain.Project)
+	}{
+		{
+			name:           "ok — no preview origins",
+			previewOrigins: nil,
+			setupIDs: func(g *idgenmock.MockGenerator) {
+				gomock.InOrder(
```
wim07101993: does it matter the calls happen in that order?

internal/service/project_test.go
```diff
@@ -0,0 +1,190 @@
...
+		{
+			name:           "ok — with preview origins",
+			previewOrigins: []string{"*.vercel.app", "*.netlify.app"},
+			setupIDs: func(g *idgenmock.MockGenerator) {
+				gomock.InOrder(
+					g.EXPECT().New("proj").Return("proj_xxx", nil),
+					g.EXPECT().New("sk_proj").Return("sk_proj_yyy", nil),
+					g.EXPECT().New("sk_proj").Return("sk_proj_zzz", nil),
+				)
+			},
```
wim07101993: why put this in a separate function? it is the same for all test cases.

wim07101993 (commented):
I did not look into the sql to deeply since it is not my forte

internal/api/project.go
livio-a: IMO the service / domain is responsible for error handling incl. hiding any database error (this might potentially even leak information otherwise).
The API should just translate it to HTTP (incl status code). That's what I did now in https://github.com/zitadel/nextgen/pull/93

wim07101993: Fair enough, but than generic error handling in the Handler is still not really necessary. If it is a specific error, it can still be done here, but generics can just be handled in the  `NewError` method which should translate that to a 500 error.

internal/service/project_test.go
IAM-marco: Nope

PR #90, opened 2026-05-15
feat: add functionality to validate tenant schema

grvijayan:
This PR adds functionality to validate tenant-defined schemas.

Changes:
- Implemented `TenantSchemaValidator` to compile/resolve meta-schemas in-memory and validate tenant schema payloads by declared kind
- Extended JSONSchemaResolver to optionally validate HTTP-fetched schemas before persisting them
- Unit tests


internal/domain/schema/meta/user-schema.json
```diff
@@ -0,0 +1,35 @@
...
+  "required": ["$schema", "$id", "kind", "title", "x-auth-methods"],
+  "properties": {
+    "$schema": { "type": "string"},
+    "$id":     { "type": "string", "format": "uri" },
+    "kind":    { "type": "string", "const": "user-schema" },
+    "title":   { "type": "string", "minLength": 1, "maxLength": 255 },
+    "description": { "type": "string", "maxLength": 1024 },
+    "x-auth-methods": {
+      "type": "object",
+      "minProperties": 1,
```
grvijayan: I set `"minProperties": 1` as we want to have at least one auth-method and `"additionalProperties": false` to ensure only one of the allowed auth-methods are set. Please let me know if you have a different opinion on this. cc: @wim07101993 / @muhlemmer

wim07101993:
Why define the meta schema's again? Can't we use the ones from the openapi spec to avoid discrepancies?

grvijayan:
> Why define the meta schema's again? Can't we use the ones from the openapi spec to avoid discrepancies?

I wanted to adapt this to the json_schema implementation, which currently embeds and walks the `schema/` directory for builtin templates. Also, the library seems to intrinsically deal with JSON rather than YAML (only exposes `SchemaFromJSON`), and I just wanted to keep it simple. 

But yes, there could be discrepancies. We could add a generator to convert `user-schema.yaml` (and other schemas) to the meta schema JSON files, and they will all be available in one place for the validator. What do you think?

wim07101993:
The openapi spec can also be json. If we could point the walker to the root of the openapi directory and make it walk there, then we would have all schema's at once.

If that is no option, we could maybe use symlinks or something like that?

internal/domain/json_schema.go
```diff
@@ -3,24 +3,26 @@ package domain
...
 	"time"
 
 	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/ianlancetaylor/jsonschema"
+	apischemas "github.com/zitadel/nextgen/api/openapi/endpoints/schemas"
 	"github.com/zitadel/nextgen/internal/httputil"
 	"github.com/zitadel/nextgen/internal/storage/database"
 )
 
+var absoluteScheme = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9+\-.]*:`)
```
wim07101993: do we allow anything else than `http://` or `https://`? In that case a simpler regex like `^https?:\/\/` could be used.

internal/domain/json_schema.go
```diff
@@ -69,53 +71,29 @@ const (
...
-			return nil
-		}
-		if !strings.HasSuffix(fsPath, ".json") {
-			return nil
-		}
-		raw, err := schemaFS.ReadFile(fsPath)
-		if err != nil {
-			return fmt.Errorf("domain: read builtin schema %q: %w", fsPath, err)
+	m := make(map[string][]byte)
+	entries, err := fs.ReadDir(apischemas.FS, ".")
```
wim07101993, line 79: what is apischema.FS? I don't think it exists yet

internal/domain/schema_validator.go
```diff
@@ -0,0 +1,164 @@
...
+		metaSchemas[kind] = metaSchema
+	}
+
+	return &TenantSchemaValidator{
+		metaSchemas: metaSchemas,
+	}, nil
+}
+
+// ValidateAgainstMetaSchema safely checks byte payloads entirely in memory.
+func (v *TenantSchemaValidator) ValidateAgainstMetaSchema(tenantSchemaBytes []byte) error {
```
wim07101993, line 90: this function unmarshals the schema multiple times, can't we reduce that?

internal/domain/schema_validator.go
```diff
@@ -0,0 +1,164 @@
...
+		if s, ok := cache[schemaURL]; ok {
+			return s, nil
+		}
+		schema, err := unmarshalJSONSchema(schemaURL, data)
+		if err != nil {
+			return nil, err
+		}
+		cache[schemaURL] = schema
+		err = schema.Resolve(&jsonschema.ResolveOpts{
+			Loader: func(schemaID string, uri *url.URL) (*jsonschema.Schema, error) {
```
wim07101993: shouldn't this loader also get docs from the database?

internal/domain/json_schema.go
grvijayan: this was supposed to be general, but if we don't prefer it, i can modify this to check only for http/https.

internal/domain/json_schema.go
grvijayan: sorry, missed adding this file in my commit.

internal/domain/schema_validator.go
grvijayan: this is loading only the embedded meta schemas - we don't need to load anything from the database for validation against the metaschema.

internal/domain/json_schema.go
wim07101993: I think we can use https only.

grvijayan: done

wim07101993 (approved):

PR #93, opened 2026-05-18
feat(api): update auth attempt api and error handling

livio-a:
This PR updates the auth attempt API with latest changes for an actual service implementation. It also adds to generators to build OpenAPI error schemas for all domain errors and reflect the possible codes returned by each endpoint (currently they're mostly empty since there's no implementation of the service yet).
The server is updated to run db migrations and initialize repositories for the services.
The auth attempt service interface and possible (domain) errors are defined alongside the domain error structure itself.


internal/service/auth_attempt.go
```diff
@@ -0,0 +1,233 @@
...
+	Proof Proof
+}
+
+type HandoffInput struct {
+	ProjectID      string
+	AttemptID      string
+	IdempotencyKey *string
+}
+
+// ---- Challenge types (discriminated union) --------------------------------------
```
wim07101993, line 99: Should these unions exist here, or in the domain layer?

livio-a: as just discussed, since they're used as input for the domain logic, we need to keep them here right now

wim07101993 (approved):

PR #105, opened 2026-05-19
feat: zitadel-login self-completes sign-in

bastionstack:
## Summary

When a flow's terminal step is `complete: "show"` and the consumer has set `post-sign-in-url`, `<zitadel-login>` now exchanges the handoff token at `POST /sessions/exchange` and performs the full-page navigation itself — addressing the TODO from #86 about framework consumers reimplementing exchange + redirect glue. The `zitadel-flow-complete` event remains for callers that want to override behaviour.

## Why

Both `apps/demo-next/` and `apps/demo-nuxt/` were carrying near-identical \`useEffect\` / \`onMounted\` glue around the \`zitadel-flow-complete\` event to call \`/sessions/exchange\` and redirect. Pulling that into the orchestrator collapses an integration step out of every framework consumer and makes the cookie-set + navigation boundary a single, testable place.

## Coverage

| Layer | Where |
|---|---|
| Exchange call shape, \`location.assign\` wiring | \`packages/components\` Vitest unit (MSW) |
| Real proxy → cookie origin → middleware redirect (Next) | **new** \`apps/demo-next-e2e\` (Playwright) |
| Real proxy → cookie origin → middleware redirect (Nuxt) | **new** \`apps/demo-nuxt-e2e\` (Playwright) |

Both demo SDKs need their own e2e suite because \`@zitadel-nextgen/sdk-next\`'s edge middleware and \`@nextgen/sdk-nuxt\`'s Nitro middleware are independent implementations — a regression in one wouldn't surface in the other.

## CI

New blocking \`node-e2e\` job in \`.github/workflows/ci.yml\` runs both Playwright suites with cached browsers (~1 min cached, ~3 min cold). Console-e2e is excluded until it gains specs. Failure artifacts (traces, error contexts) are uploaded on failed runs.

## Docs

- Root \`AGENTS.md\` gains a **Testing Layers** section codifying the unit → browser → e2e ladder and the \"lowest layer that proves the property; do not duplicate upward\" rule.
- \`packages/components/AGENTS.md\` cross-links to the e2e suites and notes the rebuild caveat (\`dist/\` is loaded by demos, not source).
- Each new e2e project ships a scoped \`AGENTS.md\` with explicit non-goals so future contributors don't duplicate Vitest territory.


peintnermax:
LGTM 👍  The only small piece I'd maintain is to ensure that the exact endpoint path it fetches `/sessions/exchange` is configurable via an attribute on <zitadel-login>, just in case a user sets up their rewrite (especially for SPAs) under a specific prefix like /api/auth/exchange.

mridang:
Looks good to me.

PR #112, opened 2026-05-20
feat: draft session service

livio-a:
WIP: define session service and repository and implement api and service handling


internal/domain/session.go
```diff
@@ -17,13 +30,81 @@ type Session struct {
...
-	// AuthAttempts are deleted as soon as they are handed off to a session and are therefore not accessible on a session.
+	// UserAgent contains information about the user's device and browser.
+	UserAgent *UserAgent
+
+	// State is computed at runtime and therefore not stored in the database.
+	State SessionState
+	// AssuranceLevels are computed at runtime and therefore not stored in the database.
+	AssuranceLevels []string
+	// Factors contain the authentication factors that were used to verify the session.
+	// There is one factor per [AuthCheckType].
```
adlerhurst, line 68: This means the repo does not have to store the factors by their id and use the type instead correct? E.g. passkeys do not matter if it was faceid or a physical yubikey.

internal/domain/session.go
```diff
@@ -17,13 +30,81 @@ type Session struct {
...
 
 type SessionRepository interface {
-	GetByID(ctx context.Context, q database.QueryExecutor, projectID, sessionID string) (*Session, error)
+	// Create creates a new (anonymous) session.
+	// The storage must set the ID and the read-only fields (CreatedAt, UpdatedAt), create a token, and set its id to the session.
+	Create(ctx context.Context, q database.QueryExecutor, session *Session) error
+
+	// Exchange is used to exchange a handoff token issued on an auth attempt for a session.
+	// In case the handoff token is valid, either the session associated with the auth attempt is updated or a new session is created if no session was linked.
+	// All the auth attempts factors are merged into the session.
```
adlerhurst: This means all the factors which were verified or does this include incomplete / failed ones?

internal/domain/session.go
livio-a: I think so. The necessary information should be part of the (factor) payload

internal/domain/session.go
livio-a: Maybe i need to make this clearer. By factor i'm referring to verified checks (so unsolved challenges are not copied / merged over)

adlerhurst (approved):

api/openapi/endpoints/schemas/user-property.json
```diff
@@ -19,6 +19,11 @@
       "default": false,
       "description": "Whether this property carries the user's password proof."
     },
+    "x-password": {
+      "type": "boolean",
+      "default": false,
+      "description": "Whether this property carries the user's password proof."
+    },
```
vitorbari: duplicated with the property above

adlerhurst: fixed

PR #113, opened 2026-05-20
docs(api): accept no-prefix ephemeral IDs and align OpenAPI schemas

livio-a:
Accept ADR 012 decision to remove resource-type prefixes from ephemeral IDs (session, attempt, challenge) and update OpenAPI ID schemas/examples accordingly.

Note: Domain and Service logic might need to be changed on additional PRs to reflect that ids are no longer generated by domain logic.


api/openapi/components/schemas/attempt-id.yaml
```diff
@@ -1,4 +1,4 @@
 type: string
-pattern: '^att_[a-zA-Z0-9_-]+$'
```
wim07101993, line 2: shouldn't this be a uuid? 
```
"format": "uuid"
```

api/openapi/components/schemas/challenge-id.yaml
```diff
@@ -1,4 +1,4 @@
 type: string
-pattern: '^ch_[a-zA-Z0-9_-]+$'
```
wim07101993, line 2: shouldn't this be a uuid? 
```
"format": "uuid"
```

api/openapi/components/schemas/attempt-id.yaml
livio-a: Depends on the database. Also i don't want to enforce this on the API, but rather just make clear they're alphanumeric with potentially dashes or underscores

wim07101993 (approved):

PR #122, opened 2026-05-21
fix: split up oas schema and internal schema

wim07101993:

PR #128, opened 2026-05-21
feat(api): create flow definition

grvijayan:
This PR contains the implementation for the CreateFlowDefinition API. 

- Added `CreateFlowDefinition` in the api-layer to map incoming requests to service layer, and to map back service/domain responses to api responses
- Added `FlowDefinitionService` implementation in the service layer to orchestrate the validation and creation of flow definitions
- Added `ValidateFlowDefinition` implementation in the domain layer to validate a flow definition based on business rules defined in https://github.com/zitadel/nextgen/blob/main/docs/design/flowengine/implementation-validation.md

todos:
1. The following validation is not in place atm as this field could be removed from the user metaschema ([PR](https://github.com/zitadel/nextgen/pull/122))
> A step with an `x-identifier` field might define a `user_not_found` transition, otherwise an error will be return for "not found".


internal/service/flow_definition.go
```diff
@@ -0,0 +1,460 @@
...
+}
+
+// Validate validates the flow definition steps and transitions
+func (fd *flowDefinitionService) Validate(ctx context.Context, flowDefinition domain.FlowDefinition, flowSchemaURI string, flowDefinitionRaw []byte) error {
+	// 1. structural validation against the flow definition schema
+	if err := fd.validateAgainstSchema(flowSchemaURI, flowDefinitionRaw); err != nil {
+		return err
+	}
+
+	// 2. validate purpose and initial steps
```
vitorbari: This feels to me more like domain logic (or a standalone validator). We could pass the flow definition and the resolved `userSchema` and check the graph and other constraints there.

internal/api/flow_definition.go
```diff
@@ -0,0 +1,195 @@
...
+		purpose, err := domain.FlowDefinitionPurposeString(p)
+		if err != nil {
+			return nil, domain.ErrFlowDefinitionInvalid("invalid purpose", nil)
+		}
+		purposes[purpose] = entryStep
+	}
+	return purposes, nil
+}
+
+func flowDefinitionSuccessResponse(flowDefinition *domain.FlowDefinition, schemaURI string) api.CreateFlowDefinitionRes {
```
vitorbari: (for later) to define a pattern how to transform `domain` to `api` and vice-versa.

internal/service/flow_definition.go
grvijayan: yes, agreed
i realize all the validation is in service layer atm, and service layer is supposed to only orchestrate. 
i'll move them to the domain layer. 

> or a standalone validator

but this is still specific to flow definition

cmd/server/config.go
```diff
@@ -7,8 +7,14 @@ import (
...
+	Schema   SchemaConfig    `mapstructure:"schema"`
 }
 
 type ServerConfig struct {
 	Address string `mapstructure:"address"`
 }
+
+type SchemaConfig struct {
+	BuiltinPublicBase string `mapstructure:"builtin_public_base"`
+	LRUCacheSize      int    `mapstructure:"lru_cache_size"`
```
vitorbari, line 21: You have to `v.BindEnv()` or `v.SetDefault()` these

grvijayan: i've set temporary default values for these - we can review the exact default values to be set

PR #131, opened 2026-05-22
feat: implement minimal auth attempt service

livio-a:
This PR implements the minimal AuthAttemptService including all methods:
`Create`, `GetByID`, `IssueChallenge` (only user and password challenges), `VerifyProof` (of the same challenge types) and `Handoff`.

It also adds all auth attempt error types as generated schemas and improves some error handling.


internal/service/auth_attempt.go
```diff
@@ -195,39 +204,193 @@ func NewAuthAttemptService(
...
-func (a *authAttemptService) Create(ctx context.Context, input CreateAuthAttemptInput) (*domain.AuthAttempt, error) {
-	//TODO implement me
-	panic("implement me")
+// Create creates a new auth attempt.
+// If input.SessionID is set, the existing session's verified checks are copied
+// into the attempt for step-up auth — no new session is created.
+func (s *authAttemptService) Create(ctx context.Context, input CreateAuthAttemptInput) (res *domain.AuthAttempt, err error) {
+	requiredChecks := input.RequiredChecks
+	if requiredChecks == nil {
+		// TODO: implement this
```
wim07101993: is this still suposed to be here?

internal/service/auth_attempt.go
```diff
@@ -195,39 +204,193 @@ func NewAuthAttemptService(
...
+	}
+	return attempt, nil
 }
 
-func (a *authAttemptService) GetByID(ctx context.Context, projectID, attemptID string) (*domain.AuthAttempt, error) {
-	//TODO implement me
-	panic("implement me")
+// GetByID retrieves an auth attempt by its ID and all its factors and challenges.
+func (s *authAttemptService) GetByID(ctx context.Context, projectID, attemptID string) (*domain.AuthAttempt, error) {
+	return s.attempts.GetByID(ctx, s.pool, projectID, attemptID)
```
wim07101993: Shouldn't the error be checked and a "not-found" be returned?

wim07101993: also valid for all future `return nil, err` statements.

internal/service/auth_attempt.go
livio-a: At one point we'll need to configure this, but for the MVP it's anyway set by the flow. I'll remove it for now

PR #139, opened 2026-05-26
feat: wire flow engine through HTTP handler

vitorbari:
## Summary

Wires the flow engine end-to-end through the HTTP handler.

- `POST /flow`, `POST /flow/{id}/submit`, `GET /flow/{id}` (event endpoint deferred)
- `_zflow` encrypted cookie via `cookie.Sealer` — handler owns I/O, state machine stays cookie-free
- `FlowService` use-case layer (resolve / start / submit / get-step) over the auth-attempt–backed state machine
- Default branding liquid template

### DIFF
--- a/internal/domain/flow_field_resolver.go
+++ b/internal/domain/flow_field_resolver.go
@@ -18,8 +18,10 @@ import (
 // api/openapi/endpoints/schemas/user-schema.yaml.
 type FlowFieldResolver interface {
 	// Resolve returns the per-field metadata for fieldNames sourced
-	// from the user schema at userSchemaURL.
-	Resolve(ctx context.Context, client database.QueryExecutor, projectID, userSchemaURL string, fieldNames []string) (FlowResolvedFields, error)
+	// from the user schema at userSchemaURL. stepName is the flow
+	// step the fields are being resolved for; it prefixes each
+	// field's text_key (`<stepName>.field.<name>`).
+	Resolve(ctx context.Context, client database.QueryExecutor, projectID, userSchemaURL, stepName string, fieldNames []string) (FlowResolvedFields, error)
 
 	// Validate checks submitted values against the rules carried by a
 	// previously resolved field set.
--- a/internal/domain/flow_field_resolver.go
+++ b/internal/domain/flow_field_resolver.go
@@ -111,9 +111,9 @@ const (
 type FlowFieldUniqueScope string
 
 const (
-	FlowFieldUniqueScopeNone         FlowFieldUniqueScope = ""
-	FlowFieldUniqueScopeInstance     FlowFieldUniqueScope = "instance"
-	FlowFieldUniqueScopeOrganization FlowFieldUniqueScope = "organization"
+	FlowFieldUniqueScopeNone    FlowFieldUniqueScope = ""
+	FlowFieldUniqueScopeProject FlowFieldUniqueScope = "project"
+	FlowFieldUniqueScopeTeam    FlowFieldUniqueScope = "team"
 )
 
 // FlowFieldValidation carries the validation rules the resolver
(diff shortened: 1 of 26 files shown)


internal/cookie/sealer.go
```diff
@@ -14,6 +14,7 @@
 package cookie
```
muhlemmer: I think sealing and verifying shouldn't be cookie only. We should have a sealer / verifier for all token types and reuse that code. A sealed token then can be set to a cookie, but that is more a protocol concern of how we send a token of the wire.

Ideally we have a single implementation which signs + encrypts (seals) and then decrypt + verify all token types, so we reduce the maintenance surface. (lessons learned from zitadel, where we have 3/4 different places and we had to refactor lots of code when we had to change the algorithm to a more secure one).

internal/domain/flow_field_resolver.go
```diff
@@ -111,9 +113,9 @@ const (
 type FlowFieldUniqueScope string
 
 const (
-	FlowFieldUniqueScopeNone         FlowFieldUniqueScope = ""
-	FlowFieldUniqueScopeInstance     FlowFieldUniqueScope = "instance"
-	FlowFieldUniqueScopeOrganization FlowFieldUniqueScope = "organization"
+	FlowFieldUniqueScopeNone    FlowFieldUniqueScope = ""
+	FlowFieldUniqueScopeProject FlowFieldUniqueScope = "project"
+	FlowFieldUniqueScopeTeam    FlowFieldUniqueScope = "team"
```
muhlemmer: Let's use a single definition of constants.

https://github.com/zitadel/nextgen/blob/acb57e047e34763c4da47a1f983a58d3be9b7c59/internal/domain/attribute.go#L17-L23

internal/domain/flow_on_success_create_user.go
```diff
@@ -104,14 +104,14 @@ func findPasswordField(resolved map[string]FlowField, submitted map[string]any)
 }
 
 // attributeUniquenessFor maps a [FlowFieldUniqueScope] to the
-// [AttributeUniqueness] the user repository understands. The
-// identifier field defaults to team-level uniqueness so two users
+// [AttributeUniqueness] the user repository understands. Both Project
+// and Team meta-schema scopes currently land on team-level storage
+// uniqueness; storage doesn't yet expose a project-wide scope short of
+// Global. The identifier field defaults to team-level so two users
```
muhlemmer: Yeah my bad, I named that constant Global before the defined the concept of a Project. Global was meant in the ctonext of an instance, where an instance is a container (see Zitadel concepts). Hence, for customer facing Global == Instance uniqueness. We can just rename the repository constant to Project, because that is what it does.

Follow up question: do we need Global as a uniqueness which bridges all Projects? Eg: should it be possible to interact with multiple projects at once? Like granting persissions and such?

internal/cookie/sealer.go
vitorbari: Agreed. There is nothing cookie specific in this sealer, it is just a bad path. This takes a slice of bytes, so there is not even the concept of token types.

I would ideally put this under `/pkg` (reusable components), but based on the current layout, it can just go under `/internal`.

wim07101993: Created a pr which fixes this: https://github.com/zitadel/nextgen/pull/166

vitorbari: Thanks @wim07101993. Then, I won't touch the `cookie.Sealer` in this PR apart from instantiating it.

internal/domain/flow_on_success_create_user.go
vitorbari: Ok. Let's rename Global to Project, I think that would make more sense.

> Follow up question: do we need Global as a uniqueness which bridges all Projects? Eg: should it be possible to interact with multiple projects at once? Like granting persissions and such?

I am not sure to be honest if we have use-cases for cross-project uniqueness.

vitorbari: Opened a PR: https://github.com/zitadel/nextgen/pull/167

PR #145, opened 2026-05-26
feat: Team repository

IAM-marco:
## What problems are solved

The `Team` domain type was a minimal stub with only `ProjectID` and `ID` — no timestamps, no repository interface, no persistence layer, and no mocks. Any feature work that needs to create or look up teams had nowhere to go.

**Domain model** (`internal/domain/team.go`)
- Added `CreatedAt` and `UpdatedAt` fields to `Team`, matching the columns already present in the migration tables for both Postgres and Spanner.
- Defined `TeamRepository` interface with two methods: `Create` and `Get`.
- Documented error contracts: `UniqueError` on duplicate `(project_id, id)` for `Create`; `NoRowFoundError` for a missing key on `Get`.
- Added `//go:generate` annotation so mocks can be regenerated with `go generate`.

**Repository implementation** (`internal/storage/database/repository/team.go`)
- `NewTeamRepository(client)` detects the dialect at construction time (`SpannerPooler` vs `PostgresPooler`) and configures the table name (`zitadel_nextgen.teams` / `teams`) and the timestamp instruction (`NOW()` / `CURRENT_TIMESTAMP()`).
- `Create` builds the `INSERT` via `StatementBuilder`, writing the dialect-specific timestamp instruction directly into the SQL string (not as a bound parameter).
- `Get` uses the shared `getOne[teamRow]` helper, filtering on both `project_id` and `id` to respect the composite primary key. Returns `NoRowFoundError` when no row matches.
- `teamMeta` satisfies both the `updatable` and `deletable` interfaces (compile-time assertions included) so the struct is ready for future update/delete operations without further scaffolding.

**Integration tests** (`internal/storage/database/repository/team_test.go`)
- `TestTeamRepository_Create`: verifies timestamps are set by the database, and that a duplicate `(project_id, id)` pair produces an error (uses `savepointForRollback` for Postgres, same transaction for Spanner).
- `TestTeamRepository_Get`: verifies round-trip of all fields and that a missing key returns `NoRowFoundError`.
- Both test functions run against whichever dialect is selected by the test environment (embedded Postgres by default, Spanner emulator under the `spanner_integration` build tag).

**Mocks** (`internal/domain/mock/team.mock.go`)
- Generated via `go generate ./internal/domain/...`; typed mock following the same pattern as `ProjectRepository`.


wim07101993 (approved):

internal/storage/database/repository/team.go
```diff
@@ -0,0 +1,99 @@
...
+		return &TeamRepository{
+			meta: teamMeta{tableName: pgTableTeams},
+			now:  database.NowInstruction,
+		}
+	}
+	panic("NewTeamRepository: unsupported client type")
+}
+
+func (r *TeamRepository) Create(ctx context.Context, client database.QueryExecutor, team *domain.Team) error {
+	b := database.NewStatementBuilder("INSERT INTO ")
```
wim07101993, line 71: Shouldn't the team id be generated by the database?

IAM-marco: No

Per [ADR 011](https://github.com/zitadel/nextgen/blob/5761ad2a2914d328203f5863b120e95300c60a22/docs/adrs/011-resource-identifiers.md), `team_id` is a **managed resource**: its ID is a caller-provided string (optionally supplied by the API client, otherwise generated by the service layer via `idgen.Generator` with the `team_` prefix). Database-generated integer identities are reserved for **ephemeral** resources like `session_id`, `token_id`, and `check_id`.

I've added a `CHECK (id <> '')` constraint to the Postgres migration (matching `projects` and `users`) and an early-return guard in `Create` for Spanner (which doesn't use that constraint pattern), so an empty ID is rejected at the storage boundary in both dialects.

Issue #148, opened 2026-05-27
refactor repository package


adlerhurst:
Currently the repository package provides one implementation for all dialects and uses many if statements to build the correct statement. This leads to hard to manage code and makes adding an additional dialects exponentially difficult.

We should decide on a strategy which keeps maintenance overhead low and adding new dialects easy.

I could think of using one of the following patterns:

- adapter pattern duplicate the whole logic (might get hard to maintain)
- outsource the statements into files and use the same logic, adapter pattern on statement layer instead of repository layer
- ...?

IAM-marco:
> Currently the repository package provides one implementation for all dialects and uses many if statements

By using a constructor for the repository, that takes in input the db client, we can instantiate a repository that works for the right dialect.

The implementations can then be split into separate files.

This is the cleanest solution to me.

IAM-marco:
Other things refactor wise:

May want to re-work the parameters in the interfaces, like:
  - re-introduce the query options on the get and list methods, 
  - add `conditions`and `changes` to the update methods (I didn't on the flow definition repository) 

**To discuss:**
Repository interfaces:

  - Option A (as done on relational Zitadel): one big interface that embeds condition and columns interfaces (and other interfaces if needed)
  - Option B (smaller interfaces): As done in the `userLookup` interface on nextgen, 1 DB method (Get) + its conditions (ProjectID, ID, Attributes)

PR #153, opened 2026-05-27
fix: combine user/password lookup conditions with database.And

vitorbari:
`WithCondition` overwrites instead of combining (`opts.Condition = condition`), so only the last call took effect, the `project_id` filter was silently dropped.

Wrap each pair with `database.And` so both conditions apply.


internal/service/auth_attempt.go
```diff
@@ -356,11 +356,13 @@ func (s *authAttemptService) verify(ctx context.Context, attempt *domain.AuthAtt
...
-			database.WithCondition(s.users.AttributesCondition([]domain.Attribute{{
-				Key:   p.AttributeName,
-				Value: p.LoginName,
-			}})),
+			database.WithCondition(database.And(
+				s.users.ProjectIDCondition(attempt.ProjectID),
+				s.users.AttributesCondition([]domain.Attribute{{
+					Key:   p.AttributeName,
+					Value: p.LoginName,
+				}}),
```
IAM-marco, line 364: The structure of the interface is up for debate IMO. I would keep the format we had in Zitadel:

  - Repository interface that includes all condition and column interfaces: https://github.com/zitadel/zitadel/blob/c73e58489ba0ad7590b2a66c9f0d77de23b2298c/backend/v3/domain/session.go#L37
  - An unexported struct that implements the repository interface (and so all other embedded interfaces): https://github.com/zitadel/zitadel/blob/c73e58489ba0ad7590b2a66c9f0d77de23b2298c/backend/v3/storage/database/repository/session.go#L450

For the moment, maybe we can add a TODO/Discussion point here? https://github.com/zitadel/nextgen/issues/148

internal/service/auth_attempt.go
```diff
@@ -356,11 +356,13 @@ func (s *authAttemptService) verify(ctx context.Context, attempt *domain.AuthAtt
...
 			ctx,
 			s.pool,
-			database.WithCondition(s.users.ProjectIDCondition(attempt.ProjectID)),
-			database.WithCondition(s.users.AttributesCondition([]domain.Attribute{{
-				Key:   p.AttributeName,
-				Value: p.LoginName,
-			}})),
+			database.WithCondition(database.And(
+				s.users.ProjectIDCondition(attempt.ProjectID),
+				s.users.AttributesCondition([]domain.Attribute{{
```
IAM-marco, line 361: Is it intended to not have a real implementation of this?

internal/service/auth_attempt.go
IAM-marco: (unless you know this was discussed and the decision is final)

internal/service/auth_attempt.go
vitorbari: No, it is not. We should have it properly implemented.

IAM-marco: Ok I see the problem. I might ask Tim how to implement this

internal/service/auth_attempt.go
vitorbari: I don't know. We have to ask the other people involved in the storage layer. I have little context.

IAM-marco: Added the point for discussion to the mentioned ticket

IAM-marco (approved):

PR #155, opened 2026-05-27
fix: allow project_id as a query parameter in session API

adlerhurst:
Introduce the ability to filter sessions by project_id in the session API. This change enhances the API's functionality by making project_id a required query parameter for relevant requests.


internal/api/session.go
```diff
@@ -31,10 +32,8 @@ func (h Handler) CreateSession(ctx context.Context, req *api.CreateSessionReques
 }
 
 func (h Handler) ExchangeHandoff(ctx context.Context, req *api.ExchangeRequest, params api.ExchangeHandoffParams) (api.ExchangeHandoffRes, error) {
-	scopeCtx, _ := GetScopeContext(ctx)
-
 	input := service.ExchangeInput{
-		ProjectID:    scopeCtx.ProjectID,
+		ProjectID:    string(params.ProjectID.Value),
```
IAM-marco: ProjectID is optional from the APIs, as far as I can see.

So, you should be using `params.ProjectID.Get()` and check the `ok` bool value before setting `input.ProjectID = projectID`. (or make project ID required at API level)

internal/api/session.go
```diff
@@ -48,10 +47,8 @@ func (h Handler) ExchangeHandoff(ctx context.Context, req *api.ExchangeRequest,
 }
 
 func (h Handler) GetSession(ctx context.Context, params api.GetSessionParams) (api.GetSessionRes, error) {
-	scopeCtx, _ := GetScopeContext(ctx)
-
 	input := service.GetSessionInput{
-		ProjectID: scopeCtx.ProjectID,
+		ProjectID: string(params.ProjectID.Value),
```
IAM-marco: As before, should use ProjectID.Get()

internal/api/session.go
```diff
@@ -83,10 +80,8 @@ func (h Handler) GetMySession(ctx context.Context, params api.GetMySessionParams
 }
 
 func (h Handler) ListSessions(ctx context.Context, params api.ListSessionsParams) (api.ListSessionsRes, error) {
-	scopeCtx, _ := GetScopeContext(ctx)
-
 	input := service.ListSessionInput{
-		ProjectID: scopeCtx.ProjectID,
+		ProjectID: string(params.ProjectID.Value),
```
IAM-marco: As before, should use ProjectID.Get()

internal/api/session.go
```diff
@@ -97,10 +92,8 @@ func (h Handler) ListSessions(ctx context.Context, params api.ListSessionsParams
 }
 
 func (h Handler) RevokeSession(ctx context.Context, params api.RevokeSessionParams) (api.RevokeSessionRes, error) {
-	scopeCtx, _ := GetScopeContext(ctx)
-
 	input := service.DeleteSessionInput{
-		ProjectID: scopeCtx.ProjectID,
+		ProjectID: string(params.ProjectID.Value),
```
IAM-marco: As before, should use ProjectID.Get()

IAM-marco (changes requested):

internal/api/session.go
adlerhurst: I added a temporary error to get a clear error message.

IAM-marco (approved):

PR #156, opened 2026-05-27
feat: implemented create/get team api

wim07101993:


internal/service/team.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+func (s *TeamService) CreateTeam(ctx context.Context, input CreateTeamInput) (*domain.Team, error) {
+	tx, txErr := s.pool.Begin(ctx, nil)
+	if txErr != nil {
+		return nil, domain.ErrInternal(txErr).WithMessage("failed to start transaction")
+	}
+	defer func() {
+		if txErr == nil {
```
IAM-marco: This is misleading because:

  - `txErr`is always `nil` after line 38
  - It will trigger every time 
  - Despite triggering, if `tx.Commit()` happened already, it won't do anything

internal/api/team.go
```diff
@@ -0,0 +1,52 @@
...
+	"net/http"
+
+	api "github.com/zitadel/nextgen/api/generated"
+	"github.com/zitadel/nextgen/internal/domain"
+	"github.com/zitadel/nextgen/internal/service"
+)
+
+func (h *Handler) CreateTeam(ctx context.Context, req *api.CreateTeamRequest, params api.CreateTeamParams) (r api.CreateTeamRes, _ error) {
+	if !params.ProjectID.IsSet() {
+		return nil, domain.ErrTeamProjectNotFound()
```
IAM-marco: A missing `project_id` parameter is a malformed request → HTTP **400**, not 404. 404 implies the resource was looked up and not found.

Consider marking `project_id` as `required: true` in the OpenAPI spec so the generated layer rejects the request before it reaches the handler, or return a distinct 400-class domain error here.

internal/api/team.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	return &api.CreateTeamResponse{
+		ID:        team.ID,
+		CreatedAt: team.CreatedAt,
+	}, nil
+}
+
+func (h *Handler) GetTeam(ctx context.Context, params api.GetTeamParams) (api.GetTeamRes, error) {
+	if !params.ProjectID.IsSet() {
+		return nil, domain.ErrTeamProjectNotFound()
```
IAM-marco: Same issue as `CreateTeam` above: missing `project_id` should return 400, not 404.

internal/api/team.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/zitadel/nextgen/internal/service"
+)
+
+func (h *Handler) CreateTeam(ctx context.Context, req *api.CreateTeamRequest, params api.CreateTeamParams) (r api.CreateTeamRes, _ error) {
+	if !params.ProjectID.IsSet() {
+		return nil, domain.ErrTeamProjectNotFound()
+	}
+	team, err := h.teamService.CreateTeam(ctx, service.CreateTeamInput{ProjectID: string(params.ProjectID.Value)})
+	if err != nil {
+		return nil, err
```
IAM-marco, line 15: Inconsistent with the rest of the package: `project.go` handlers use `return h.NewError(ctx, err), nil` rather than `return nil, err`. Both are handled by the OAS middleware, but the explicit form is the established convention in this package. Same applies to line 32.

api/openapi/endpoints/teams/create-team-request.yaml
```diff
@@ -0,0 +1 @@
+type: object
```
IAM-marco: Also: missing newline at end of file.

api/openapi/endpoints/teams/create-team-response.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - id
+  - createdAt
+properties:
+  id:
+    type: string
+    description: The unique identifier of the team.
+  createdAt:
+    type: string
+    format: date-time
+    description: The time when the team was created.
```
IAM-marco: Missing newline at end of file.

api/openapi/endpoints/teams/methods.yaml
```diff
@@ -0,0 +1,46 @@
...
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/error-details.yaml
+    default:
+      description: Unexpected error response.
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/error-details.yaml
```
IAM-marco: Missing newline at end of file.

internal/api/integration_test/schema_test.go
```diff
@@ -95,37 +93,34 @@ func TestCreateSchema(t *testing.T) {
 			defer func() { _ = resp.Body.Close() }()
 
 			if !assert.Equal(t, http.StatusBadRequest, resp.StatusCode) {
-				bs, err := io.ReadAll(resp.Body)
-				require.NoError(t, err)
-				t.Log(string(bs))
+				helpers.LogInvalidResponse(t, resp)
```
IAM-marco: The original code was correct:
```go
bs, err := io.ReadAll(resp.Body)
require.NoError(t, err)
t.Log(string(bs))
```
I think you need to put io.ReadAll() back.

api/openapi/endpoints/teams/methods.yaml
```diff
@@ -0,0 +1,46 @@
...
+          schema:
+            $ref: create-team-response.yaml
+    '400':
+      description: The provide parameters were not acceptable
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/error-details.yaml
+    '404':
+      description: The project could not be found
```
IAM-marco: Should this be team?

api/openapi/endpoints/teams/methods.yaml
```diff
@@ -0,0 +1,47 @@
...
+          $ref: create-team-request.yaml
+  responses:
+    '201':
+      description: The team was created successfully
+      content:
+        application/json:
+          schema:
+            $ref: create-team-response.yaml
+    '400':
+      description: The provide parameters were not acceptable
```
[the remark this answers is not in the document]
IAM-marco: Valid

internal/api/integration_test/helpers/harness.go
```diff
@@ -26,12 +26,13 @@ type Harness struct {
...
-	FlowService        service.FlowService
-	AuthAttemptService service.AuthAttemptService
-	ProjectService     service.ProjectService
+	SchemaService         *service.SchemaService
+	SessionService        service.SessionService
+	FlowService           service.FlowService
+	AuthAttemptService    service.AuthAttemptService
+	ProjectService        service.ProjectService
 	FlowDefinitionService service.FlowDefinitionService
+	TeamService           *service.TeamService
```
IAM-marco, line 36: I'd make interfaces for schema and team service TBH

internal/api/integration_test/helpers/logging.go
```diff
@@ -0,0 +1,15 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func LogInvalidResponse(t *testing.T, resp any) {
+	bs, err := json.Marshal(resp)
+	require.NoError(t, err)
+	log.Println(string(bs))
+}
```
IAM-marco: What is the point of this? The assert/require functions can take variadic arguments for a customised message

internal/api/integration_test/helpers/team.go
```diff
@@ -0,0 +1,47 @@
...
+		)
+	}
+	return h.TeamService
+}
+
+func (h *Harness) CreateTeam(t *testing.T, projectID string) string {
+	t.Helper()
+	team := &domain.Team{
+		ProjectID: projectID,
+	}
```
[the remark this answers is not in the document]
IAM-marco: Valid, should pass an ID

internal/api/integration_test/schema_test.go
```diff
@@ -95,37 +93,34 @@ func TestCreateSchema(t *testing.T) {
 			defer func() { _ = resp.Body.Close() }()
 
 			if !assert.Equal(t, http.StatusBadRequest, resp.StatusCode) {
```
IAM-marco, line 93: Why not:

```go
func requestParser(t *testing.T, body any) string {
    bs, err := io.ReadAll(body) //resp.body
    require.NoError(t, err)
    marshalled, err := json.Marshal(resp)  
    require.NoError(t, err)
    return fmt.Printf("%s", string(marshalled))
}

assert.Equal(t, http.StatusBadRequest, resp.StatusCode, requestParser(t, resp.Body))
```

internal/api/team.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// ------------------ Errors ---------------
+
+func teamErrorResponse(err domain.Error) *api.ErrorDetailsStatusCode {
+	switch err.Code {
+	case domain.ErrTeamNotFound().Code:
+		return errorResponseWithStatusCode(http.StatusNotFound, err)
+	case domain.ErrTeamProjectNotFound().Code:
+		return errorResponseWithStatusCode(http.StatusNotFound, err)
```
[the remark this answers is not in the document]
IAM-marco: Valid

internal/service/team.go
```diff
@@ -0,0 +1,73 @@
...
+func (s *TeamService) CreateTeam(ctx context.Context, input CreateTeamInput) (*domain.Team, error) {
+	tx, txErr := s.pool.Begin(ctx, nil)
+	if txErr != nil {
+		return nil, domain.ErrInternal(txErr).WithMessage("failed to start transaction")
+	}
+	defer func() {
+		if txErr != nil {
+			_ = tx.Rollback(ctx)
+		}
+	}()
```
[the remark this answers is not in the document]
IAM-marco: Valid and it endorses the use of `tx.End()`

IAM-marco (changes requested):

api/openapi/endpoints/teams/methods.yaml
wim07101993: it was here because the project yielded a 404. Removed it

internal/api/integration_test/helpers/harness.go
wim07101993: Why? Then were back again with large interfaces declared at the implementor.

internal/api/integration_test/helpers/team.go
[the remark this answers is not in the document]
wim07101993: removed the func since it was not used, and duplicates the code from the services

internal/api/team.go
wim07101993: Not really. The `project.go` file is an outlier here I think. Because error handling is done by the middleware, we decided to let it do that. Examples: `session.go`, `schema.go`

internal/service/team.go
wim07101993: was checking the wrong error. fixed it

internal/api/team.go
IAM-marco: I'm confused, considering also [this](https://github.com/zitadel/nextgen/pull/87#discussion_r3247133955)

But fine

wim07101993: I know, that was while Livio was figuring out the mechanics of error handling. Afterwards he created the entire mechanism which maps the error codes properly and generates openapi types for them.

IAM-marco (approved):

PR #160, opened 2026-05-27
feat: step-prefixed localization keys + drop cookie path

vitorbari:
Stacked on #139. Four small tweaks to the flow HTTP surface:

- **`internal/api/branding/default.liquid`** — replace the MVP skeleton with the orchestrator's auth-form template (mirrors `packages/components/src/orchestrator/templates/auth-form.liquid.ts`).
- **`internal/api/flow.go`** — drop `Path=/flow` from the `_zflow` cookie so it's readable on whichever path the client mounts the flow under.
- **`FlowStep.Texts`** — `buildStep` emits `<step>.title` / `<step>.description`, matching the locale convention in `packages/components/src/orchestrator/locales/en.ts`.
- **Resolver `stepName` parameter** — `FlowFieldResolver.Resolve` builds each field's `text_key` as `<step>.field.<name>` directly.

## Test plan

- [x] `go build ./internal/... ./api/...`
- [x] `go test ./internal/api/... ./internal/domain/... ./internal/service/...`
- [ ] Manual: `POST /flow` returns `step.texts.title_key == "<step>.title"`, `fields.<name>.text_key == "<step>.field.<name>"`, and `Set-Cookie: _zflow` omits a `Path` attribute.

### DIFF
--- a/internal/api/branding/default.liquid
+++ b/internal/api/branding/default.liquid
@@ -1,52 +1,103 @@
-{% comment %}
-  Default LiquidJS template for the flow engine MVP. Rendered by the
-  orchestrator into Shadow DOM. Customers can replace it via the
-  Team/App Branding API once that endpoint ships.
-
-  Available context:
-    step.name             - step identifier
-    step.error            - optional error key
-    step.fields           - dictionary of {name -> field}
-    step.actions          - dictionary of {name -> action}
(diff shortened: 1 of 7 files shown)


vitorbari:
Screenshot 2026-05-27 at 17 00 09

PR #166, opened 2026-05-28
refactor: use simple interfaces for en/decryption and hashing

wim07101993:
- Used the same interfaces as defined in the `zitadel/oidc` package for encrypting/decrypting so that the implementations used there are compatible
- Abstracted interface layer for hashing


internal/crypto/encryption.go
```diff
@@ -0,0 +1,13 @@
+package crypto
+
+type Encrypter interface {
+	Encrypt(string) (string, error)
```
vitorbari, line 4: For a crypto package, `[]byte` is usually the more idiomatic and flexible choice, examples:
https://pkg.go.dev/hash#Hash
https://pkg.go.dev/crypto/cipher#Block

wim07101993: I know, I thought about it, but the `oidc` package currently uses `string` and I wanted to keep compatibility with it. Also https://github.com/zitadel/nextgen/pull/67 in which we decided this. It is still not merged yet though

cmd/server/server.go
```diff
@@ -82,7 +82,7 @@ func run(ctx context.Context, cfg Config, pool database.Pool, userFiles []string
 		}
 	}()
 
-	sealer, err := buildCookieSealer(cfg.Server.CookieSealerKey)
+	crypter, err := buildCrypter(cfg.Server.CookieSealerKey)
```
vitorbari: Can you please rename the config `cfg.Server.CookieSealerKey`? The current name does not make sense anymore.

vitorbari (changes requested):

vitorbari (approved):

PR #167, opened 2026-05-28
refactor: rename AttributeUniqueness Global to Project

vitorbari:
- Rename `AttributeUniquenessGlobal` → `AttributeUniquenessProject` (We are still unsure if cross-project uniqueness is also needed)
- Rename the matching `'global'` value in the `uniqueness_scope` Postgres enum to `'project'` (safe, no data yet)

### DIFF
--- a/internal/domain/flow_on_success_create_user.go
+++ b/internal/domain/flow_on_success_create_user.go
@@ -110,7 +110,7 @@ func findPasswordField(resolved map[string]FlowField, submitted map[string]any) 
 func attributeUniquenessFor(name, identifierName string, scope FlowFieldUniqueScope) AttributeUniqueness {
 	switch scope {
 	case FlowFieldUniqueScopeInstance:
-		return AttributeUniquenessGlobal
+		return AttributeUniquenessProject
 	case FlowFieldUniqueScopeOrganization:
 		return AttributeUniquenessTeam
 	}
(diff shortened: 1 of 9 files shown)


internal/storage/database/dialect/postgres/migration/sql/000004_users.sql
```diff
@@ -172,7 +172,7 @@ CREATE EXTENSION IF NOT EXISTS pgcrypto;
 CREATE TYPE zitadel_nextgen.uniqueness_scope AS ENUM (
     'unspecified',
     'team',
-    'global'
+    'project'
```
vitorbari, line 175: I'm assuming this is fine since there is no real data yet.

internal/domain/flow_on_success_create_user.go
```diff
@@ -110,7 +110,7 @@ func findPasswordField(resolved map[string]FlowField, submitted map[string]any)
 func attributeUniquenessFor(name, identifierName string, scope FlowFieldUniqueScope) AttributeUniqueness {
 	switch scope {
 	case FlowFieldUniqueScopeInstance:
-		return AttributeUniquenessGlobal
+		return AttributeUniquenessProject
```
vitorbari, line 113: I'm going to try to remove this duplication in a follow-up PR.

IAM-marco (approved):

PR #170, opened 2026-05-28
feat: added create and get user endpoints

wim07101993:
Adds a POST /user and GET /user/{user_id} API surface backed by a new UserService, plus the supporting OpenAPI definitions, generated code, domain helpers (nested-map flattening, JSON-schema-driven attribute uniqueness), error wiring, and integration tests.

Spanner support is not yet finished for the repository, so it does not work yet


internal/api/user.go
```diff
@@ -1 +1,64 @@
...
+
+	api "github.com/zitadel/nextgen/api/generated"
+	"github.com/zitadel/nextgen/internal/domain"
+	"github.com/zitadel/nextgen/internal/service"
+)
+
+func (h *Handler) CreateUser(ctx context.Context, req *api.User, params api.CreateUserParams) (api.CreateUserRes, error) {
+	var teamID *string
+	if params.TeamID.IsSet() {
+		teamID = new(string(params.TeamID.Value))
```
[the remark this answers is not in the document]
wim07101993: it does work, proven by the integration tests

internal/api/user.go
```diff
@@ -1 +1,64 @@
...
+		return nil, err
+	}
+
+	return convertUsingJson[api.CreateUserResponse](u)
+}
+
+func (h *Handler) GetUserByID(ctx context.Context, params api.GetUserByIDParams) (api.GetUserByIDRes, error) {
+	var teamID *string
+	if params.TeamID.IsSet() {
+		teamID = new(string(params.TeamID.Value))
```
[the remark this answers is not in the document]
wim07101993: it does work, proven by the integration tests

internal/domain/attribute.go
```diff
@@ -76,3 +78,44 @@ func setAttribute(keyNodes []string, value any, tree map[string]any) {
...
+
+		default:
+			var unique AttributeUniqueness
+			strUnique, _ := maputil.GetNested[string](schema, "properties."+key+".x-unique")
+			switch strUnique {
+			case "project":
+				unique = AttributeUniquenessProject
+			case "team":
+				unique = AttributeUniquenessTeam
+			}
```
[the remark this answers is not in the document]
wim07101993: @copilot could you fix this and write a test for it?

internal/storage/database/repository/user.go
```diff
@@ -223,6 +223,14 @@ func userHydrationExpressions(rowQualifier, attrKeysPlaceholder, authPlaceholder
     CASE WHEN ` + authPlaceholder + ` THEN EXISTS(SELECT 1 FROM zitadel_nextgen.user_passkeys p WHERE p.project_id = ` + rowQualifier + `.project_id AND p.user_id = ` + rowQualifier + `.id) ELSE FALSE END AS has_pk`
 }
 
+func (r *UserRepository) GetByID(ctx context.Context, client database.QueryExecutor, projectID string, teamID *string, userID string) (*domain.User, error) {
+	condition := r.PrimaryKeyCondition(projectID, userID)
+	if teamID != nil {
+		condition = database.And(condition, r.PrimaryKeyCondition(projectID, userID))
```
grvijayan: I think we need the `TeamIDCondition` here instead of `PrimaryKeyCondition`, which is already set in line 227.

internal/service/user.go
```diff
@@ -0,0 +1,134 @@
...
+) *UserService {
+	return &UserService{
+		pool:       pool,
+		userRepo:   userRepo,
+		schemaRepo: schemaRepo,
+	}
+}
+
+func (s *UserService) CreateUser(ctx context.Context, input CreateUserInput) (_ map[string]any, err error) {
+	tx, err := s.pool.Begin(ctx, nil)
```
grvijayan: do we need the explicit transaction handling here?

api/openapi/endpoints/users/user.yaml
```diff
@@ -0,0 +1,17 @@
...
+type: object
+additionalProperties: true
+required:
+  - $schema
+properties:
+  $schema:
+    type: string
+    description: |
+      The schema that defines the content of the user. This is determined by
+      the configuration of the system and can be used to validate the content
```
grvijayan: > This is determined by the configuration of the system

Sounds a bit vague. Could we elaborate?

api/openapi/security/oauth2.yaml
```diff
@@ -9,10 +9,14 @@ flows:
...
+      user.write: Permission to create users
+      user.read: Permission to read users
   clientCredentials: 
     tokenUrl: https://acc.nextgen.zitadel.com/oauth/token
     scopes: 
       openid: OpenID Connect scope
       schema.read: Permission to read schemas
       schema.write: Permission to create and manage schemas
+      user.create: Permission to create users
+      user.list: Permission to list users
```
grvijayan: It would probably be less confusing to use just `user.write` and `user.read` instead of `write`, `create`, `read`, and `list`

api/openapi/endpoints/users/methods.yaml
```diff
@@ -1,3 +1,52 @@
+post:
+  operationId: createUser
+  summary: Create user
+  tags:
+    - User
+  security:
+    - oauth2:
+        - users.write
```
grvijayan:
```suggestion
        - user.write
```

api/openapi/endpoints/users/methods.yaml
```diff
@@ -1,3 +1,52 @@
...
+        application/json:
+          schema:
+            $ref: ../../components/error-details.yaml
+    '401':
+      description: Missing or invalid credentials.
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/error-details.yaml
+    '403':
```
grvijayan, line 37: perhaps also a `409` response if the user already exists.

internal/service/user.go
```diff
@@ -0,0 +1,134 @@
...
+	err = json.Unmarshal(schemaEntity.Schema, &schema)
+	if err != nil {
+		return nil, domain.ErrInternal(err).WithMessage("failed to unmarshal json schema")
+	}
+
+	err = schema.Validate(input.User)
+	if err != nil {
+		return nil, domain.ErrUserInvalid().
+			WithParent(err).
+			WithMessage("user does is not valid according to schema")
```
grvijayan:
```suggestion
			WithMessage("user is not valid according to schema")
```

internal/service/user.go
wim07101993: I thought so because we need to fetch and validate against the schema. However, since schema's are immutable, it might not be required. Should I remove?

grvijayan: we do need to validate against the schema, but I don't think the explicit transaction handling is adding any benefit here, so, yeah, we can remove that.

grvijayan (approved):

grvijayan:
please add a proper PR description

PR #175, opened 2026-05-28
refactor(ci): run integration tests on testcontainers, smoke-test the embedded binary

fforootd:
## Summary

Restructures the Go test jobs and the database test bring-up. **Postgres and Spanner integration tests now run on Docker testcontainers**, the **embedded-postgres single-binary path is covered by a server smoke test**, the Go CI jobs are **split by backend** behind backend-explicit build tags, and a shared **`dbtest` helper** lets every integration suite run against a database you manage (`ZITADEL_TEST_*_URL`) with no Docker.

This PR began as a fix for a flaky CI failure (`panic: ... no version found matching 18.3.0`); the root-cause analysis below still motivates the design, but the fix evolved from "retry the Maven download" into "don't depend on the Maven download for integration tests at all."

## Root cause (original flake)

All three Go jobs ran `go test ./...`. Because `-tags X ./...` **adds** tagged files to the always-compiled untagged ones (it does not filter), the untagged DB-backed tests ran in *every* job — and on a cold cache `fergusstrange/embedded-postgres` downloads the PG 18.3 binary from Maven at test time. A transient non-200 there is misreported as `no version found matching 18.3.0`, panicking `TestMain`.

## What changed

**1. Integration tests run on testcontainers.** A new `StartContainer(ctx)` (build-tagged `postgres_integration || spanner_integration`) starts `public.ecr.aws/docker/library/postgres:18.3` from the rate-limit-free public ECR mirror and flows its mapped DSN through the existing `postgres.DecodeConfig` connector seam — mirroring how Spanner already uses a testcontainer emulator. The four postgres test entry points now call `StartContainer` instead of the embedded library. `testcontainers-go` is build-tagged out of the production binary (`go list -deps ./cmd/server | grep testcontainers` → empty).

**2. The embedded path is covered by a server-binary smoke test.** The embedded-postgres default connector stays — it's the zero-config way to run zitadel with no DB setup (sqlite-style). Instead of an integration test, a new `go-smoke-test-embedded-postgres` job builds `nextgen`, boots it on embedded PG (migrations run during startup), waits for `server listening`, sends `SIGTERM`, and asserts a clean exit 0 — *we test the start, not the integration suites*. The prior `go test -tags embedded_integration` test is removed (superseded; it exercised the test-only `MigrateTest`, the smoke test exercises the real production `Migrate`).

**3. One backend per job — build tags + `ci.yml`.** The `integration` tag (which always meant postgres) is renamed to `postgres_integration` for symmetry with `spanner_integration`; backend-agnostic suites use `postgres_integration || spanner_integration`. Go CI jobs are split and named for what they exercise.

**4. Resilience retained where it still matters.** `StartEmbedded()`'s bounded retry + the `~/.embedded-postgres-go` binary cache (`actions/cache@v5`, keyed on `go.sum`) now live on the embedded smoke job — the only job that still downloads the PG binary from Maven.

**5. One source of truth for test-DB bring-up.** A new test-only `internal/storage/database/dbtest` package owns the override decision — `ZITADEL_TEST_POSTGRES_URL` / `ZITADEL_TEST_SPANNER_URL` ? connect to that DB : start a testcontainer — and every suite routes through `dbtest.Postgres(ctx)` / `dbtest.Spanner(ctx)`. So a full `go test -tags … ./...` honors the override everywhere and needs no Docker when a var is set (previously `internal/bootstrap/users` and the Spanner API setup always started their own container, regardless of the var). The container code stays build-tagged, so testcontainers never enters production.

## Result — Go CI jobs

| Job | Build tags | Backend |
| --- | --- | --- |
| `go-unit-test` | none | none — pure unit, DB-free |
| `go-integration-test-postgres` | `postgres_integration` | Docker testcontainer (`postgres:18.3`) |
| `go-integration-test-spanner` | `spanner_integration` | Docker testcontainer (Spanner emulator) |
| `go-smoke-test-embedded-postgres` | none (built binary) | embedded postgres — build → boot → migrate → exit 0 |

## Test plan

- [x] `go build .`; `go vet` under no-tags, `-tags postgres_integration`, `-tags spanner_integration`
- [x] Integration suites pass on testcontainers (PG + Spanner)
- [x] Embedded smoke validated locally: fallback → embedded PG 18.3 → `server listening` → SIGTERM → exit 0
- [x] `embedded_integration` tag fully removed; workflow YAML parses; all `needs:` resolve
- [x] Override end-to-end: with `ZITADEL_TEST_POSTGRES_URL` set, `internal/bootstrap/users` connects to that DB and starts no testcontainer
- [ ] CI green on the renamed/added jobs
- [ ] Switch the required status check from `go-test` to `go-unit-test` and also remove `go-integration-test`


.github/workflows/ci.yml
```diff
@@ -52,7 +53,15 @@ jobs:
           cache: true
       - name: Install embedded-postgres runtime deps
         run: sudo apt-get update -qq && sudo apt-get install -y -qq libicu-dev libssl-dev
-      - run: go test -v -tags integration -timeout=10m ./...
+      # Persist the embedded-postgres binary (downloaded from Maven Central on a
+      # cold cache) so test runs skip the flaky network fetch. Keyed on go.sum so
+      # an embedded-postgres bump invalidates it automatically.
+      - name: Cache embedded-postgres binary
+        uses: actions/cache@v4
```
adlerhurst: Use latest cache version (v5)

fforootd: Done in 4bb684d7 — `actions/cache@v4` → `@v5`. Also bumped the pre-existing Playwright cache step in `node-e2e` for consistency, so the workflow isn't left on mixed cache-action versions.

fforootd: Follow-up: the testcontainers migration (4753b94d) removed this embedded-postgres cache step entirely — postgres integration tests now run against a Docker testcontainer, so there's no Maven binary to cache. The only remaining `actions/cache` (Playwright, in `node-e2e`) is already on `@v5`. Thread is outdated now; leaving it resolved.

README.md
```diff
@@ -103,18 +103,22 @@ The devcontainer at [.devcontainer/](.devcontainer/) pins Go 1.26 and a PostgreS
...
 
-Run Spanner repository tests (same command as CI):
+Run the integration tests (same commands as CI):
 
 ```sh
+# Postgres (all packages)
+go test -v -tags postgres_integration -timeout=10m ./...
+
+# Spanner (repository suite)
 go test -v -tags spanner_integration -timeout=10m ./internal/storage/database/repository/...
```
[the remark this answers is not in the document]
fforootd: Fixed in 120e23f6 — the Spanner example now runs `./...` to match the CI `spanner_integration` job; it previously ran only the repository package, skipping the API integration tests.

internal/service/postgres_testmain_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build integration
+//go:build postgres_integration
```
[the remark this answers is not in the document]
fforootd: Fixed in 120e23f6 — updated the run hint and the failure message from `-tags=integration` to `-tags=postgres_integration`.

internal/api/integration_test/setup_pg_test.go
```diff
@@ -34,7 +34,7 @@ func runTests(m *testing.M) int {
 		connector, err = postgres.DecodeConfig(url)
 		stop = func() {}
 	} else {
-		connector, stop, err = embedded.StartEmbedded()
+		connector, stop, err = embedded.StartContainer(ctx)
 		helpers.Connector = connector
```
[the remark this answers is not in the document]
fforootd: Fixed in 120e23f6 — `helpers.Connector` is now assigned after both branches, so the `ZITADEL_TEST_POSTGRES_URL` path also wires up the connector that `EnsureDBPool` dials (it was nil on that path before).

internal/storage/database/repository/repository_test.go
```diff
@@ -81,9 +83,9 @@ func newEmbeddedDB(ctx context.Context) (pool database.PoolTest, stop func(), er
 		}
 		stop = func() {}
 	} else {
-		connector, stop, err = embedded.StartEmbedded()
+		connector, stop, err = embedded.StartContainer(ctx)
 		if err != nil {
-			return nil, nil, fmt.Errorf("unable to start embedded postgres: %w", err)
+			return nil, nil, fmt.Errorf("unable to start postgres container: %w", err)
```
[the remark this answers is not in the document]
fforootd: Fixed in 120e23f6 — `newEmbeddedDB` now returns a no-op stop func on both error paths, so a container-start failure surfaces the setup error instead of a nil-func panic in the deferred `stop()`.

README.md
```diff
@@ -103,18 +103,22 @@ The devcontainer at [.devcontainer/](.devcontainer/) pins Go 1.26 and a PostgreS
...
+go test -v -tags postgres_integration -timeout=10m ./...
+
+# Spanner
+go test -v -tags spanner_integration -timeout=10m ./...
 ```
 
 If `docker info` fails and the host uses **rootless Docker**, override the socket mount in [`.devcontainer/devcontainer.json`](.devcontainer/devcontainer.json) per the [docker-outside-of-docker feature docs](https://github.com/devcontainers/features/tree/main/src/docker-outside-of-docker#rootless-docker-support), for example bind `/run/user/<uid>/docker.sock` to `/var/run/docker-host.sock` (use `id -u` on the host for `<uid>`).
 
-To use an emulator you start yourself instead of testcontainers, set `ZITADEL_TEST_SPANNER_URL` to the Spanner DSN; tests skip container startup when that variable is set.
+To use a database you start yourself instead of testcontainers, set `ZITADEL_TEST_POSTGRES_URL` (Postgres DSN) or `ZITADEL_TEST_SPANNER_URL` (Spanner DSN); tests skip container startup when the matching variable is set.
```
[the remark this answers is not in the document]
fforootd: Fixed in 801846d3: reworded to scope the override to the suites that read `ZITADEL_TEST_*_URL`, and noted that `internal/bootstrap/users` (Postgres) and the Spanner API integration setup always start a container, so a Docker daemon is still required.

internal/storage/database/dbtest/postgres.go
```diff
@@ -0,0 +1,28 @@
+//go:build postgres_integration || spanner_integration
```
IAM-marco, line 1: **Question / clarity:** The `spanner_integration` build tag on this Postgres helper is not immediately obvious. The reason: `repository_test.go` is tagged `postgres_integration || spanner_integration` and calls `newEmbeddedDB`, which uses `dbtest.Postgres` — so the Spanner integration job also builds the shared repository suite and needs a Postgres connector for it.

Please add one-line doc comment here exaplining the this cross-tag dependency.

internal/storage/database/dialect/postgres/embedded/container.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+		ContainerRequest: req,
+		Started:          true,
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to start postgres container: %w", err)
+	}
+
+	stop := func() { _ = container.Terminate(ctx) }
```
IAM-marco: **Minor (correctness):** `stop` captures the startup context — if that context is already canceled by the time the test suite calls `stop()`, `container.Terminate` will return immediately with a context error and the container leaks.

Use a fresh context for teardown:

```go
stop := func() { _ = container.Terminate(context.Background()) }
```

Can you also fix this in internal/storage/database/dialect/spanner/embedded/start.go ?

internal/storage/database/dialect/postgres/embedded/start.go
IAM-marco, line 1: Can you remove the comment at the top that says: `// embedded is used for testing purposes`

This is not true anymore

IAM-marco (changes requested):

IAM-marco:
@fforootd I think the required steps of the pipeline need to be updated: it seems like github is still waiting on the old steps that you renamed.

Please see if you can fix that, I don't have permission

fforootd:
> @fforootd I think the required steps of the pipeline need to be updated: it seems like github is still waiting on the old steps that you renamed.
> 
> 
> 
> Please see if you can fix that, I don't have permission

That is kind of a GitHub quirk. We can only change the required fields for all branches merging into main.

So we should do this at the time we want to merge this one. (That will require all others btw. to merge main as well) 😆

internal/storage/database/dbtest/postgres.go
fforootd: fixed in last push

internal/storage/database/dialect/postgres/embedded/container.go
fforootd: fixed in last push

internal/storage/database/dialect/postgres/embedded/start.go
fforootd: fixed in last push

IAM-marco (approved):

PR #183, opened 2026-05-29
feat: added default user schema and flow definition

wim07101993:
feat: added default user schema and flow definition

refactor: removed redundant schema validation for flow-definitions since api already does that


api/openapi/endpoints/flow_definitions/embed.go
```diff
@@ -0,0 +1,222 @@
...
+		return nil, err
+	}
+
+	steps, err := convertSteps(req.FlowDefinition.GetSteps())
+
+	return domain.NewFlowDefinition(
+		projectID,
+		req.FlowDefinition.GetName(),
+		new(url.URL(req.GetSchemaURI().Value)).String(),
+		new(req.FlowDefinition.GetUserSchema()).String(),
```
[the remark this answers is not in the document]
wim07101993: this does compile and the suggestion doesn't because `String()` requires a pointer receiver

grvijayan:
> removed redundant schema validation for flow-definitions since api already does that

which API?

api/openapi/endpoints/flow_definitions/embed.go
```diff
@@ -0,0 +1,228 @@
...
+		purposes,
+		domain.FlowDefinitionAudience{
+			AppIDs:  req.FlowDefinition.GetAudience().Value.AppIds,
+			TeamIDs: req.FlowDefinition.GetAudience().Value.TeamIds,
+		},
+		steps,
+	)
+}
+
+func convertPurposes(purposes api.FlowDefinitionPurposes) (map[domain.FlowDefinitionPurpose]string, error) {
```
vitorbari, line 64: Can we reuse more of the conversion logic already exists in the flow-definition APIs?

api/openapi/endpoints/schemas/embed.go
```diff
@@ -1,9 +1,22 @@
...
+	"strings"
+)
 
 // FS holds the embedded JSON schema files from this directory.
 //
 //go:embed *.json
 var FS embed.FS
+
+//go:embed examples/default-human-user-schema.json
+var defaultHumanUserSchema []byte
```
vitorbari, line 15: Ideally, we need to support multiple default user types.

api/openapi/endpoints/flow_definitions/embed.go
```diff
@@ -0,0 +1,228 @@
...
+	"encoding/json"
+	"net/url"
+	"strings"
+
+	api "github.com/zitadel/nextgen/api/generated"
+	"github.com/zitadel/nextgen/internal/domain"
+)
+
+//go:embed examples/default-login-flow-definition.json
+var defaultLoginFlowDefinition []byte
```
vitorbari, line 15: I did not thing much about it, but what if we use a DB migration to insert them in the flow definition table?

wim07101993:
> > removed redundant schema validation for flow-definitions since api already does that
> 
> which API?

@grvijayan the api with which the flow definitions are set.https://github.com/zitadel/nextgen/blob/bc9ee02cebe4ec50f5b90f353f11f293152c5f2b/internal/api/flow_definition.go#L15

Ogen validates that the request body is valid according to the OAS, which describes a valid flow-definition.

api/openapi/endpoints/flow_definitions/embed.go
wim07101993: Not really because it creates circular dependencies if we do. 

I wanted to use the json representation for the default schemas, since that way we can also provide them as examples in the OAS. That means we need to convert an generated api model to a domain model.

That would result in a cyclic import:
```
github.com/zitadel/nextgen/api/openapi/endpoints/flow_definitions (in file embed.go) ->
github.com/zitadel/nextgen/internal/api (in file flow.go) ->
github.com/zitadel/nextgen/internal/service (in file project.go) ->
github.com/zitadel/nextgen/api/openapi/endpoints/flow_definitions 
```

I could move more code around so that the converting method exists outside the current packages but than we would have a `CreateFlowDefinitionRequest` instead of a `FlowDefinition`. So I opted to just convert it directly.

api/openapi/endpoints/schemas/embed.go
wim07101993: yes that is why I called this one `defaultHumanUserSchema`. Than we can call a second one `defaultServiceUserSchema` or something like that.

api/openapi/endpoints/flow_definitions/embed.go
wim07101993: that would not be possible because it requires a project, which we don't have a migration time.

Also that would not embed them in the OAS. Where it is useful as an example I think.

internal/service/project.go
```diff
@@ -20,48 +24,112 @@ type ProjectService interface {
...
 	if err != nil {
 		return nil, err
 	}
-	projectSecret, err := s.ids.New("sk_proj")
+
+	if err := s.projectRepo.Create(ctx, tx, project); err != nil {
+		return nil, domain.ErrInternal(err).WithMessage("failed to create project in the database")
+	}
+
+	userschema, err := s.createDefaultUserSchema(ctx, tx, project.ID)
```
vitorbari: `userschema` is always the one used for the flow definitions, that is correct. But inside this transaction, we should also retrieve the other default user-schemas such as machine-user, ai-agent-user, etc

internal/service/project.go
```diff
@@ -20,48 +24,112 @@ type ProjectService interface {
...
 	if err != nil {
 		return nil, err
 	}
-	previewSecret, err := s.ids.New("sk_proj")
+	var userSchema *jsonschema.Schema
+	err = json.Unmarshal(userschema.Schema, &userSchema)
 	if err != nil {
+		return nil, domain.ErrInternal(err).WithMessage("failed to parse default user schema")
+	}
+	if err := s.createDefaultLoginFlowDefinition(ctx, tx, project.ID, userSchema); err != nil {
```
vitorbari: We should make it plural to support multiple flow definitions

internal/service/project.go
wim07101993: to create the login flow, we need a user schema to link to it. If we would use an array, we do not know which one to pick for which flow. So I think we better keep it as is for now, until we see an actual need for it.

vitorbari (approved):

PR #187, opened 2026-05-29
feat: implement passkey (WebAuthn) challenge build and verification

IAM-marco:
## What problems are solved

The auth-attempt flow already had `user` and `password` challenges wired end-to-end, but the **passkey** challenge was only scaffolded. The core WebAuthn logic in `internal/domain/user_passkey.go` was a stub: `CreatePasskeyChallenge` ran the go-webauthn login ceremony but discarded most of the resulting session data (returning with a `// TODO: implement`). Because those fields were never persisted, `VerifyPasskeyChallenge` could not reconstruct a valid `webauthn.SessionData`, so a passkey assertion could never be verified. There were also no tests covering this path.

As a result it was impossible to authenticate with a passkey through the auth-attempt API, even though the API handler, service dispatch, domain aggregate state machine, and storage serialization were all in place.

**Domain — complete the WebAuthn build/verify** (`internal/domain/user_passkey.go`)
- `CreatePasskeyChallenge` now persists the full session data: challenge, allowed credential IDs, user verification requirement, RP id/origins, user handle, expiry, and extensions. User-verification options are now actually passed to `BeginLogin`/`BeginDiscoverableLogin`.
- `sessionDataFromChallenge` restores those fields so verification matches what was issued.
- `VerifyPasskeyChallenge` validates against the persisted RP origins and resolves the user for both identified and discoverable (usernameless) logins.
- `PasskeyVerification`/`AuthFactorPasskey` carry the credential id, sign count and backup state.

**Service & API** (`internal/service/auth_attempt.go`, `internal/api/auth_attempt.go`, `internal/api/auth_check.go`)
- Passkey challenge issuing and proof verification are dispatched through the domain functions.
- After a successful assertion the authenticator's advanced sign count, backup state and last-used time are persisted (best-effort: a write failure must not reject an otherwise valid proof — the stored sign count is a clone-detection signal, not an auth gate).
- The challenge response returns the `allowed_credentials` and `user_verification` the browser needs for an identified login; the completed passkey factor payload returns the real credential id and backup state instead of empty placeholders.
- The repository gains `SetSignCount` and `Update`; the service secondary-port interfaces are exported so they can be mocked.

**Tests** (`internal/domain/user_passkey_test.go`, `internal/service/auth_attempt_test.go`)
- New domain and service tests drive a real challenge → signed assertion → verify round-trip using `github.com/descope/virtualwebauthn` (test-only dependency), covering identified and discoverable logins, sign-count persistence, and rejection paths.
- The auth-attempt service tests were migrated to gomock + testify with generated mocks.

# Out of scope (follow-ups)

- passkey registration API
- link this logic with the flow engine


internal/service/auth_attempt.go
```diff
@@ -444,13 +464,32 @@ func (s *authAttemptService) verify(ctx context.Context, attempt *domain.AuthAtt
 		if err != nil {
 			return passkeyChallenge, nil, domain.ErrAuthAttemptProofRejected(err)
 		}
+		// Use the verified user as the source of truth: it is set for both identified and
+		// discoverable logins, whereas userFactor is nil in the discoverable case.
+		s.recordPasskeyUsage(ctx, attempt.ProjectID, verification)
```
vitorbari, line 468: Just to note that this is not being done for `PasswordProof`

internal/domain/user_passkey.go
```diff
@@ -10,12 +10,11 @@ import (
...
 
 type PasskeyChallenge struct {
-	data                 []byte
 	Challenge            string
 	AllowedCredentialIDs [][]byte
-	UserVerification     uint8 // domain.UserVerificationRequirement
+	UserVerification     protocol.UserVerificationRequirement
 	RPID                 string
-	RPOrigin             url.URL
+	RPOrigins            []url.URL
```
vitorbari: Are these supposed to be the `previewOrigins` from the project?

internal/domain/user_passkey.go
```diff
@@ -10,12 +10,11 @@ import (
 )
 
 type PasskeyChallenge struct {
-	data                 []byte
 	Challenge            string
 	AllowedCredentialIDs [][]byte
-	UserVerification     uint8 // domain.UserVerificationRequirement
+	UserVerification     protocol.UserVerificationRequirement
```
livio-a: Do we want a direct dependency on an external library in our domain model?
IMO we should either convert it into domain enums / types or change the passkey challenge to just hold necessary values and pass session data (and credential assertion) as []byte / any. For the former we also need to ensure we pass all information to the client, which we currently don't. For the latter, we could just use the blob in the api as well and don't explicitly document the fields in the api:
```yaml
title: PasskeyChallengePayload
type: object
description: |
  WebAuthn-specific challenge data for passkey authentication including the publicKey options for assertion generation (dictionary PublicKeyCredentialRequestOptions).
  See also: https://www.w3.org/TR/webauthn/#dictdef-publickeycredentialrequestoptions
additionalProperties: true
```
@peintnermax would this be an issue?

internal/domain/user_passkey.go
IAM-marco: I believe there is an overlap between the 2. The RPOrigins are the URLs of the resource you want to authenticate for.

livio-a: It kind of fulfills the same requirement yes, but IIRC the previewOrigins are the allowed origins for preview envs, right? For prod non-preview envs we currently don't have a configuration afaik. In any case the RPOrigins must be a subset of them.

IAM-marco: I tried to look this up in the docs. There is some mention of preview origins in the context of vercel deployment, but I couldn't get a clear answer on what they are used for.

internal/domain/user_passkey.go
IAM-marco: > Do we want a direct dependency on an external library in our domain model?

I changed this myself, I don't have strong feelings about this

internal/domain/auth_attempt.go
livio-a: IMO we should remove this to allow discoverable credentials

livio-a (changes requested):
Also just noted that the passkey verification still requires a user to be verified first, this prevents discoverable credentials.

internal/domain/auth_attempt.go
IAM-marco: Good call — fixed. `PreparePasskeyVerification` now returns the `AuthFactorUser` as nullable (nil for discoverable logins) and the service path (`if userFactor != nil { ... } else { discoverable }`) already handles both cases. `PreparePasskeyChallenge` was already permissive (returns an empty user ID when none is identified). The `// TODO(adlerhurst)` note is removed.

internal/domain/user_passkey.go
IAM-marco: Aligning on what surfaced in the thread: RPOrigins should be a subset of the project's allowed origins. `previewOrigins` (Vercel preview envs) is one such allow-list slice; we do not have a prod-equivalent origin config on the project yet. For now origins are passed in by the caller and validated against whatever the project exposes; tying this to a unified project-level origin config (incl. prod) is a follow-up — I will leave a comment in the code pointing at it.

internal/domain/user_passkey.go
IAM-marco: Update: pushed the change. The domain `PasskeyChallenge` struct + `CreatePasskeyChallenge` signature now use plain Go types (`string` for `UserVerification`/`Mediation`, `map[string]any` for `Extensions`, a local `PasskeyCredentialParameter` for `CredParams`). All conversions to/from `protocol.*` happen inside `CreatePasskeyChallenge` and `sessionDataFromChallenge`. The cascade dropped `go-webauthn/protocol` from `internal/service/auth_attempt.go` and `internal/api/auth_attempt.go`. JSON shape of the persisted payload is unchanged. Tests green (`go test ./...`).

livio-a (approved):

PR #188, opened 2026-05-30
feat(infra): tofu migration

conblem:
Moves over most of the infra parts of oxidel.
Adds a postgres database and moves secrets management out of 1password.
For now I only moved over the infra pipeline as we have no releases yet in this repo.


fforootd:
Since we changed a lot of things i am going to do this a new once we have the authz parts ready

PR #192, opened 2026-06-01
feat(webauthn): wire passkey auth into the flow engine 

IAM-marco:
⚠️  **Please review https://github.com/zitadel/nextgen/pull/187 first** ⚠️ 

## Summary

- Adds a two-phase WebAuthn passkey ceremony to the flow state machine: the `passkey` step action issues a `PublicKeyCredentialRequestOptions` challenge, halts on the current step so the browser can sign, and on the next submit verifies the assertion and records the resolved user.
- Supports both identified (identifier step → passkey step) and discoverable/usernameless (passkey as entry step; user resolved from the assertion's `userHandle`) login flows.
- Relaxes `PreparePasskeyVerification` to allow a nil user factor for genuine usernameless entry while preventing subject-switching in multi-factor attempts.
- Threads the `Origin` header through the submit endpoint to derive WebAuthn RP params (RPID = hostname, allowed origin = full origin) without project-level config.

## Changes

| Layer | What changed |
|---|---|
| `internal/domain/user_passkey.go` | `BuildPasskeyRequestOptions` — reusable `PublicKeyCredentialRequestOptions` JSON builder |
| `internal/domain/auth_attempt.go` | Relax `PreparePasskeyVerification` for discoverable; bind resolved user after verify |
| `internal/domain/flow_auth_attempt.go` | `IssuePasskeyChallenge` / `SubmitPasskey` on `FlowAuthAttemptService` + input/output types |
| `internal/domain/flow_state_machine.go` | `processPasskey` issue/verify dispatch; `FlowSubmitInput.ChallengeResponse` / `PasskeyRP`; `FlowStep.Challenge`; `FlowPendingChallenge.Options` |
| `internal/service/flow_auth_attempt.go` | Adapter implementations |
| `internal/service/flow.go` | `SubmitFlowRequest` extensions |
| `internal/api/flow.go` | Origin→RP derivation; challenge/challenge_response mapping |
| `api/openapi/…` + `api/generated/…` | `Origin` header param on submit; regenerated |

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./internal/domain/... ./internal/service/... ./internal/api/...` — all green
- [x] State-machine unit tests: issue/verify/handoff cycle, rejected proof, discoverable entry
- [x] Service adapter unit tests: `IssuePasskeyChallenge` and `SubmitPasskey`
[...]

PR #195, opened 2026-06-01
feat(webauthn): passkey registration ceremony (flow engine + standalone REST)

IAM-marco:
**⚠️ Please review https://github.com/zitadel/nextgen/pull/192 first ⚠️** 

## Summary

- Implements the full WebAuthn registration ceremony on two tracks:
  - **Flow engine** (`passkey_register` action in the state machine) — a user already identified in a flow can be prompted to enroll a new passkey
  - **Standalone REST** (`POST /passkeys` → `POST /passkeys/{id}/finish`) — a session-authenticated user can enroll a passkey outside of a flow
- All three Gemini security review findings addressed:
  1. **IDOR** — `ProjectID` is always derived from the authenticated token scope, never from the request body
  2. **Error mapping** — `pkreg.*` errors now route correctly; `ErrPasskeyRegistrationNotFound` → HTTP 404
  3. **Origin validation** — `BeginPasskeyRegistration` and `SubmitFlowStep` validate the request `Origin` against `Project.PreviewOrigins` before issuing a challenge (empty allowlist = allow all, for dev/test)

## Commits

| Commit | What |
|--------|------|
| `feat(domain)` | `PasskeyRegistrationChallenge`, `FlowPasskeyRegistrationService` interface, `passkey_register` in state machine, credential-ID base64url fix |
| `feat(storage)` | `passkey_registrations` migration + repository |
| `feat(service)` | `PasskeyRegistrationService` (flow engine + standalone) |
| `feat(api)` | OpenAPI spec + regenerated ogen code for `/passkeys` endpoints |
| `feat(api)` | `BeginPasskeyRegistration` / `FinishPasskeyRegistration` handlers, server wiring |
| `fix(security)` | IDOR, error mapping, origin validation |
| `test` | Standalone, flow engine, and origin-rejection integration tests |

## Test plan

- [ ] `go build ./... && go vet ./...` — clean
- [ ] `go test ./...` — all unit tests pass
- [ ] `go test -tags integration -timeout=10m ./...` — all integration tests pass (confirmed locally)
- [ ] `TestStandalonePasskeyRegistration` — full begin→finish round trip via session auth
- [ ] `TestPasskeyRegistrationFlow` — begin→finish through the flow engine
- [ ] `TestBeginPasskeyRegistrationRejectsUnauthorizedOrigin` — origin not in allowlist returns error


api/openapi/openapi-spec.yaml
```diff
@@ -86,6 +86,10 @@ paths:
     $ref: endpoints/teams/methods.yaml
   /teams/{team_id}:
     $ref: endpoints/teams/by_id/methods.yaml
+  /passkeys:
+    $ref: endpoints/passkeys/methods.yaml
+  /passkeys/{registration_id}:
+    $ref: endpoints/passkeys/by_id/methods.yaml
```
livio-a: I'd not expose these endpoints yet as they're not required for the MVP and would need some details to clarify first.
IMO we should remove all the changes in the openapi folder on this PR.

api/openapi/components/flows/flow-step.yaml
```diff
@@ -110,7 +110,7 @@ properties:
     properties:
       method:
         type: string
-        enum: [passkey]
+        enum: [passkey, passkey_register]
```
vitorbari: No need to expose the split to the client. The server already controls the `PendingChallenge` which is kept in the flow state. For the clients, there is no difference really. We can keep `FlowChallengeMethodPasskeyRegister` as an internal-only concept.

internal/api/flow.go
```diff
@@ -132,9 +132,17 @@ func (h *Handler) SubmitFlowStep(ctx context.Context, req *api.FlowSubmitRequest
...
 	if origin, ok := params.Origin.Get(); ok {
 		if rp := passkeyRPFromOrigin(origin); rp != nil {
+			project, err := h.projectService.Get(ctx, state.ProjectID)
+			if err != nil {
+				return internalErrorResponse(err), nil
+			}
+			if err := validateOriginAgainstProject(origin.String(), project); err != nil {
+				return errorResponseWithStatusCode(http.StatusBadRequest,
+					domain.ErrRequestInvalid().WithMessage(err.Error())), nil
+			}
```
vitorbari, line 145: Should this be a middleware logic for all endpoints?

internal/domain/flow_state_machine.go
```diff
@@ -577,6 +613,40 @@ func (r *FlowStateMachineRuntime) processPasskey(ctx context.Context, state *Flo
...
+		}
+		if in.PasskeyRP == nil {
+			return passkeyPhaseResult{}, fmt.Errorf("%w: passkey relying-party params missing", ErrIntegrity)
+		}
+		if r.passkeyRegistration == nil {
+			return passkeyPhaseResult{}, fmt.Errorf("%w: passkey registration service not wired", ErrIntegrity)
+		}
+		userID, ok := state.CollectedData[FlowCollectedUserIDKey].(string)
+		if !ok || userID == "" {
+			return passkeyPhaseResult{}, fmt.Errorf("%w: passkey_register requires an identified user", ErrIntegrity)
```
vitorbari, line 629: The [flow definition validator](https://github.com/zitadel/nextgen/blob/main/internal/domain/flow_definition_validator.go) should enforce some constraints around passkey_register — e.g. an upstream identifier is required, and passkey must be enabled in the user schema. Today both fall through to runtime. Could be a follow-up, since PR #206 will need a similar traversal for `create_user_with_passkey`.

internal/service/passkey_registration.go
```diff
@@ -0,0 +1,207 @@
+package service
```
vitorbari, line 1: Better to align the layering with how sign-in on top of auth-attempt is organized: a canonical PasskeyRegistrationService (Begin/Finish) plus a thin flow adapter that implements domain.FlowPasskeyRegistrationService on top of it.

Today the REST `BeginPasskeyRegistration` refers to `domain.Flow*`, which is backwards.
The flow isn't authoritative for passkeys; it's a consumer.

vitorbari: Authoritative: https://github.com/zitadel/nextgen/blob/6f9abc7cf633a38a1b5e38d804d24cbbe639749e/internal/service/auth_attempt.go
Thin adapter: https://github.com/zitadel/nextgen/blob/f2ac8058355f7824baec1e0a3037a9b1ba20fbe4/internal/service/flow_auth_attempt.go

internal/api/flow.go
livio-a: IMO yes either a middleware for all browser based calls, or in case of this specific case where we'll use the origin header to set the RP ID, we could also test it in the service rather than the API layer but generally prefer a middleware

internal/api/passkey.go
livio-a, line 1: as commented on the openapi spec, i would not yet implement an API for the passkey part

internal/domain/passkey_registration_repository.go
livio-a, line 1: IMO we should either use auth attempts (yes wrong name) to handle these challenges as well or the passkey repo should be enabled to handle the registration process as well... but OK for me to continue like this and fix it later

internal/api/flow.go
IAM-marco: The validation only fires when `passkeyRPFromOrigin` returns non-nil (Origin header present + passkey step), so making it middleware would run it on all requests unnecessarily. Additionally, the semantics of `PreviewOrigins` on the project are not yet fully clarified: it's not certain that enforcing an origin match against preview origins is the right check here.

internal/domain/passkey_registration_repository.go
IAM-marco: Made an issue: https://github.com/zitadel/nextgen/issues/220

vitorbari (approved):

PR #201, opened 2026-06-02
feat: added password endpoints

wim07101993:


internal/service/user.go
```diff
@@ -121,3 +136,76 @@ func (s *UserService) GetUserByID(ctx context.Context, input GetUserInput) (map[
...
+
+func (s *UserService) createPassword(ctx context.Context, client database.QueryExecutor,
+	projectID string, teamID *string, userID string, pwdHash string, isPasswordChangeRequired bool) error {
+	err := s.passwordRepo.Create(ctx, client, &domain.CreateUserPassword{
+		ProjectID:      projectID,
+		TeamID:         teamID,
+		UserID:         userID,
+		EncodedHash:    pwdHash,
+		ChangeRequired: isPasswordChangeRequired,
+		VerificationID: nil, // TODO what should I do with this?
```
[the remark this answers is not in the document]
wim07101993: we leave it as is for now

api/openapi/openapi-spec.yaml
```diff
@@ -40,10 +40,12 @@ paths:
     $ref: endpoints/livez/methods.yaml
   /readyz:
     $ref: endpoints/readyz/methods.yaml
-  /user:
+  /users:
```
muhlemmer, line 43: Why are we renaming to the plural form? I see no justification in the PR description. Please revert to singular use or link appropiate ADR that resources are now in the plural naming form. (Eg, is the flow API also being renamed to flows?)

internal/service/user.go
```diff
@@ -121,3 +135,58 @@ func (s *UserService) GetUserByID(ctx context.Context, input GetUserInput) (map[
...
+	}
+	defer func() {
+		if err != nil {
+			_ = tx.Rollback(ctx)
+		}
+	}()
+
+	// REMOVE OLD PASSWORD IF ONE EXISTS
+
+	pwd, err := s.passwordRepo.GetByUserID(ctx, tx, input.ProjectID, input.UserID)
```
muhlemmer: No need to check up front, just DeleteByID directly, if the row was not found by the DB nothing gets deleted.

api/openapi/openapi-spec.yaml
wim07101993: flow is indeed the only other resource which is singular. `auth_attempts`, `projects`, `sessions`, `schemas`, `flow_definitions` and `teams` all are plural. So this was a correction.

I do no know about any ADR talking about whether it is a rule or not, but there are ADRs which use the other resources in plural form.

internal/service/user.go
```diff
@@ -121,3 +135,45 @@ func (s *UserService) GetUserByID(ctx context.Context, input GetUserInput) (map[
...
+		ProjectID:      input.ProjectID,
+		UserID:         input.UserID,
+		EncodedHash:    hash,
+		ChangeRequired: input.IsPasswordChangeRequired,
+		VerificationID: nil, // TODO what should I do with this?
+	})
+	if err != nil {
+		if _, ok := errors.AsType[*database.ForeignKeyError](err); ok {
+			return domain.ErrUserNotFound()
+		}
```
vitorbari, line 170: Can we be sure `UserID` is the only FK in this table? What about ProjectID?

api/openapi/openapi-spec.yaml
vitorbari: Most of the examples under `docs/design/api` are using the plural form: https://github.com/zitadel/nextgen/blob/main/docs/design/api/resource-map.md

internal/service/user.go
wim07101993: The idea was that there is no user found to set the password on given both of the id's. Not specifying which table is missing the resource.

PR #206, opened 2026-06-02
feat: end-to-end passkey registration and login flow 

IAM-marco:
**⚠️  Please review https://github.com/zitadel/nextgen/pull/195 first ⚠️** 

### Flow & backend

- **Atomic user + passkey creation**: user creation is deferred to Phase 2 (verify), sharing the same DB transaction as the passkey save. Removed the FK on `passkey_registrations.user_id` so the Phase 1 challenge can be stored before the user row exists.
- **Passkey login `allowCredentials` fix**: `dispatchChallenges` (which runs `SubmitIdentifier`) now runs before `processPasskey` in Phase 1, so `PreparePasskeyChallenge` finds the user and populates `allowCredentials` with the correct credential IDs.
- **Registration ceremony rendering**: templates distinguish `passkey` vs `passkey_register` challenge methods, assign the correct WebAuthn ceremony, and pass `method=` to `<zl-passkey>`. The orchestrator submits using `method` from the event.
- **Secondary action buttons**: generic loop for non-primary actions so `register-with-passkey` renders on `choose-register`.
- **Provisional user ID**: `GenerateUserID()` mints a user ID at ISSUE time; `HandleProvisional()` creates the user at VERIFY time from `CollectedData["email"]`.
- **Field pre-fill**: `prefillFromCollected` propagates `CollectedData` into resolved fields before every `buildStep`.
- **`on_success` API mapping**: serialise/deserialise `on_success` in the flow definition API.
- **Session cookie path**: set `Path=/` on the flow session cookie.
- **`user_id` after registration**: `RegisterCreatedUser` registers a verified user factor on the auth attempt after `on_success: create_user` (password registration) and after `SubmitPasskeyRegistration` (passkey registration). Previously neither path set a user check, so the exchanged session had no `user_id` and server components saw the user as unauthenticated.

### SDK / frontend

- **Proxy body fix** (`sdk-next`): buffer `req.body` via `req.arrayBuffer()` before forwarding to fix WinterCG/undici incompatibility.
- **Exchange auth** (`sdk-next`): inject `Authorization: Bearer sk_<project_id>` when proxying `POST /sessions/exchange`.
- **Opaque session token support** (`sdk-next`): the backend issues AES-256-GCM encrypted tokens, not JWTs. The middleware now falls back to `GET /sessions/me` when JWT verification returns null, and `auth()` in server components does the same to resolve `user_id`. `verifyJwt` silently returns null on `SyntaxError` (opaque token) instead of logging noise.
- **Logout fix** (`zitadel-logout`): use `DELETE /sessions/me` instead of `GET /auth/end-session`. The end-session endpoint requires `usernamePassword` auth which is unimplemented; `DELETE /sessions/me` is the correct endpoint for cookie-based sessions.
- **Admin page**: fall back to `userId` when `email` is null (passkey-only users have no email factor on the session).

## Test plan

- [ ] Start backend, run Bruno bootstrap (01 → 02 → 03), start frontend
- [ ] Unknown email → `choose-register` shows both password and passkey buttons
- [ ] **Password registration**: fill credentials → passkey-upsell skip → done → `/admin` shows the user ID
- [ ] **Passkey registration**: "Register with passkey" → `collect-passkey-email` (email pre-filled) → `passkey-enroll` → WebAuthn `credentials.create()` → user + passkey created atomically → done → `/admin` shows the user ID
- [ ] **Passkey login**: same email → "Sign in with passkey" → `credentials.get()` → done → `/admin`
- [ ] **Logout**: click the avatar → "Sign out" → redirected to `/login`, session cookie cleared
- [ ] Abandonment: start passkey registration, close at ceremony step, retry same email → no conflict
- [ ] `go test ./internal/...` green


fforootd:
Yay, CI passed. Codex had this to say in a review

[P2] Provisional passkey users can persist after rejected attestation in flow_state_machine.go (line 596). The verify leg creates the user before SubmitPasskeyRegistration; when verification rejects the proof, the code returns a normal step error, and FlowService.Submit is passing the pool, not a transaction, so the partial user can remain without a password or passkey. For MVP, this is still worth fixing because one bad passkey_register response can reserve an email and break retry state.

[P2] auth() validates opaque tokens against a different env var than the middleware/scaffold uses in auth.ts (line 54). Fresh Next apps get ZITADEL_URL, and createProxy also uses that URL, but auth() falls back to NEXTGEN_ISSUER_URL ?? localhost:4000. Any app using auth() with only the documented/scaffolded ZITADEL_URL can report unauthenticated after middleware already accepted the session.

[P3] The added HAR and SQL files look like local debugging artifacts: harFail.har (line 1) includes local request traces/paths and query-demo.sql (line 1) has exploratory DB queries plus trailing whitespace. I’d remove them from the branch.

internal/domain/flow_on_success_create_user.go
```diff
@@ -86,7 +92,37 @@ func (h *FlowCreateUserHandler) Handle(ctx context.Context, client database.Quer
...
+}
+
+// HandleProvisional creates a passwordless user with a pre-assigned ID from
+// state.CollectedData. If the user already exists (UniqueError from a prior
+// on_success handler), the call succeeds silently. Intended to be called
+// within the passkey verify phase, sharing the same client transaction as
+// the passkey save for atomicity.
+func (h *FlowCreateUserHandler) HandleProvisional(ctx context.Context, client database.QueryExecutor, userID string, state *FlowState) error {
+	email, _ := state.CollectedData["email"].(string)
+	attr, err := NewCreateAttribute("email", email, AttributeUniquenessTeam)
```
livio-a: shouldn't this be based on the schema, e.g. it might not define an `email` property

apps/demo-next/query-demo.sql
livio-a, line 1: is this intended to be commited?

internal/api/flow.go
```diff
@@ -132,18 +132,29 @@ func (h *Handler) SubmitFlowStep(ctx context.Context, req *api.FlowSubmitRequest
...
+			if rp := passkeyRPFromOrigin(*originURL); rp != nil {
+				project, err := h.projectService.Get(ctx, state.ProjectID)
+				if err != nil {
+					return internalErrorResponse(err), nil
+				}
+				if err := validateOriginAgainstProject(originStr, project); err != nil {
+					return errorResponseWithStatusCode(http.StatusBadRequest,
+						domain.ErrRequestInvalid().WithMessage(err.Error())), nil
+				}
+				submitReq.PasskeyRP = rp
```
livio-a, line 156: we definitely need to improve this in the future, but ok for now

internal/domain/project.go
```diff
@@ -35,11 +36,11 @@ func NewProject(previewOrigins []string, secretGenerator secrets.Generator) (*Pr
 	if err != nil {
 		return nil, ErrInternal(err).WithMessage("failed to create project id")
 	}
-	projectSecret, err := secretGenerator.New()
+	projectSecret, err := idgen.NewULID().New("sk_proj")
 	if err != nil {
 		return nil, ErrInternal(err).WithMessage("failed to generate project secret")
 	}
-	previewSecret, err := secretGenerator.New()
+	previewSecret, err := idgen.NewULID().New("sk_proj")
```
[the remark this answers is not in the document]
livio-a: there's a separate PR to change token usage, let's fix it there

internal/storage/database/dialect/postgres/migration/sql/000010_passkey_registrations.sql
```diff
@@ -6,10 +6,6 @@ CREATE TABLE zitadel_nextgen.passkey_registrations (
     , challenge   JSONB       NOT NULL
     , expires_at  TIMESTAMPTZ NOT NULL
     , created_at  TIMESTAMPTZ NOT NULL DEFAULT now()
-
-    , FOREIGN KEY (project_id, user_id)
-        REFERENCES zitadel_nextgen.users(project_id, id)
-        ON DELETE CASCADE
```
livio-a, line 12: why would we remove this?

internal/api/security.go
```diff
@@ -39,6 +40,42 @@ func NewSecurityHandler() *SecurityHandler {
...
+		if host == "" {
+			host = r.Host
+		}
+		if host != "" {
+			ctx := context.WithValue(r.Context(), requestHostKey{}, proto+"://"+host)
+			next.ServeHTTP(w, r.WithContext(ctx))
+			return
+		}
+		next.ServeHTTP(w, r)
+	})
```
livio-a, line 71: IMO we should already match against the project origins, but let's have a look at this later
also we should handle `forwarded`

api/openapi/components/flows/flow-step.yaml
```diff
@@ -110,7 +110,7 @@ properties:
     properties:
       method:
         type: string
-        enum: [passkey]
+        enum: [passkey, passkey_register]
```
vitorbari: https://github.com/zitadel/nextgen/pull/195#discussion_r3355842121

internal/api/passkey.go
```diff
@@ -0,0 +1 @@
+package api
```
vitorbari: Can we drop this completely?

internal/domain/project.go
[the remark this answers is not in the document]
IAM-marco: What should we do then? This is going to break the demo

Should I merge the demo broken?

internal/storage/database/dialect/postgres/migration/sql/000010_passkey_registrations.sql
IAM-marco: The FK is removed so Phase 1 of passkey registration (challenge issue) can store the challenge before the user row exists. The user is created only on Phase 2 to enforce atomicity of user and passkey creation.

PR #207, opened 2026-06-03
feat: added token generators (no database) and validation

wim07101993:


internal/domain/project.go
```diff
@@ -30,16 +30,16 @@ type Project struct {
...
 	if err != nil {
 		return nil, ErrInternal(err).WithMessage("failed to create project id")
 	}
-	projectSecret, err := secretGenerator.New()
+	projectSecret, err := tokenGenerator.Generate(map[string]any{"projectID": id})
 	if err != nil {
 		return nil, ErrInternal(err).WithMessage("failed to generate project secret")
 	}
-	previewSecret, err := secretGenerator.New()
+	previewSecret, err := tokenGenerator.Generate(map[string]any{"projectID": id})
```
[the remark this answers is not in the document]
wim07101993: authorization is not yet implemented. Once it is, this implementation will probably change as well so that the token contains some form of claim which indicate it can be used for readonly access or also has write permissions

internal/api/security.go
```diff
@@ -31,18 +33,36 @@ func (s SecurityHandler) HandleOAuth2(ctx context.Context, operationName api.Ope
...
+// payload. This is the path API clients (and the integration tests) use.
+//
+// Browser-originated flows cannot hold that secret. The SDK request boundary
+// proxies project-scoped calls such as POST /sessions/exchange with an
+// "sk_<project_id>" bearer that only *identifies* the project: the request's
+// handoff token is the actual credential, and the project secret never leaves
+// the CLI's .zitadel/secret. For those, accept the well-formed project-key
+// form and use the embedded project id.
+func (s SecurityHandler) resolveProjectID(token string) (string, bool) {
+	if payload, err := s.tokenVerifier.Verify(token); err == nil {
```
wim07101993: @bastionstack the idea is that this is the correct verification. The token needs to be valid. 

When creating a project, the api [generates a secret (token)](https://github.com/zitadel/nextgen/blob/d5aacdb4084aedd7a980d13501bbfc27af6416e6/internal/domain/project.go#L37-L38). This token should be the only valid one. It contains both the writing scope and the project id. That token is returned, and should also be used when performing actions. Same for the [session token](https://github.com/zitadel/nextgen/blob/d5aacdb4084aedd7a980d13501bbfc27af6416e6/internal/domain/session.go#L121)

So if there is an err, the security handler should fail. Where does the playwrite test get an invalid token from?

wim07101993: @mridang

PR #215, opened 2026-06-04
feat: render auth widgets in Nuxt via auto-imported Vue components

mridang:
Adds Nuxt rendering for the Zitadel auth widgets. The `@zitadel/sdk-nuxt` module now auto-imports `<ZitadelLogin>` / `<ZitadelLogout>` Vue components — a page drops them in with no imports and no `vue.compilerOptions.isCustomElement` config. They wrap the Lit web components with a render function (string tag) so SSR emits the inert element and the browser upgrades it after hydration; `@zitadel/components` is imported in `onMounted` (client-only). The module also wires `runtimeConfig.public.zitadelProjectId` (from a `projectId` option or `NUXT_PUBLIC_ZITADEL_PROJECT_ID`) so the runtime plugin's `configureZitadel()` fires. Widgets read config from the SDK `globalThis` singleton, so no per-instance `project` prop is needed.

Self-contained and independent of the Next.js PR (#211): it includes the same `@zitadel/api` `globalThis` config-singleton fix as its own commit, so the bundled-config copy inside `@zitadel/components` reads the handle `configureZitadel()` writes.


mridang:
Merged into `feat/meta-framework-sdks` (commit e8892038) rather than landing separately, so the Nuxt auth widgets ship alongside the rest of the meta-framework SDK work in #373. The globalThis config-singleton fix from this PR is already on that branch (a more refined readSlot/writeSlot version), so only the Vue `<ZitadelLogin>`/`<ZitadelLogout>` components and the `zitadelProjectId` runtimeConfig wiring carried over. Closing as superseded.

PR #232, opened 2026-06-09
feat: added logging and otel

### LINKED ISSUE #49: improve telemetry
The storage package currently does not follow a structure to log errors. Mostly an error is printed using the core `log` package.

We should port the [`instrumentation`](https://github.com/zitadel/zitadel/tree/main/backend/v3/instrumentation) package of Zitadel to have a clear API for logging, tracing and metrics.

wim07101993:
This pull request adds structured logging and OpenTelemetry-based instrumentation. The configuration system is extended to support detailed instrumentation settings, and several new dependencies are added to support these features.

Parts of this have been taken from the current otel implementation in zitadel.


internal/api/middleware/logging.go
```diff
@@ -0,0 +1,104 @@
...
+		operationID, _ := GetOperationIDContext(ctx)
+		logger := zlog.GetLoggingContext(ctx)
+
+		if lrw.statusCode >= 400 {
+			logger.Error("error while handling request",
+				slog.Int("status_code", lrw.statusCode),
+				slog.Duration("duration", time.Since(start)),
+				slog.String("response", lrw.body.String()),
+				slog.String("operation_id", operationID),
+			)
```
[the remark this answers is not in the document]
wim07101993: this is only done when the call fails. In the error response there should not be any sensitive data

cmd/server/server.go
```diff
@@ -375,3 +439,64 @@ func buildCrypter(hexKey string) (crypto.Crypter, error) {
...
+}
+
+func createResource(serviceName string) (*resource.Resource, error) {
+	attributes := []attribute.KeyValue{
+		semconv.ServiceNameKey.String(serviceName),
+	}
+	if build.Version() != "" {
+		attributes = append(attributes, semconv.ServiceVersionKey.String(build.Version()))
+	}
+	return resource.Merge(
```
mridang: Looks like we are missing some of the detectors here. I would recommend addinf https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/detectors/autodetect so that we also collect all the necessary GCP metrics. If you couple if to just the GCP detector, it will not work in the future for K8. But do note. that the K8 detector is WIP rn.

mridang:
I noticed we aren't wiring up https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation if we want to add host, runtime (GC etc) and net instrumentation, they would need to be wired up in both the setup-metrics and setup-traces parts.

wim07101993:
@mridang tracing and other metrics are currently not implemented indeed. I thought to first fix logging, than move on to the other metrics.

cmd/server/server.go
wim07101993: This code was taken straight from the current Zitadel implementation. (https://github.com/zitadel/zitadel/blob/b27fdfc28c627dbb833cf78ab65f5767ddd0af44/backend/v3/instrumentation/otel.go#L125-L141)

mridang (approved):

PR #233, opened 2026-06-09
docs: add flow-definition examples

vitorbari:
Adds six worked examples under api/openapi/endpoints/flow_definitions/examples, ordered from minimal (single-purpose password login) to full (combined login+register with password, passkey, and post-signup upsell). Each example ships a flow-definition.json plus a README with a Mermaid graph, capabilities exercised, walk-through, and gotchas. The root README indexes the ladder and cross-links to the engine's capabilities and definition-rules docs.

Scope is deliberately what the engine runs today; cross-flow transitions, SSO, gates, OTP, and recovery are called out as not shown.

The flow definitions were also added as examples in the OpenAPI spec:

Screenshot 2026-06-10 at 12 34 42


api/openapi/endpoints/flow_definitions/examples/01-password-login/flow-definition.json
```diff
@@ -0,0 +1,37 @@
+{
```
vitorbari, line 1: One problem, this is not the actual flow-definition, it is the payload to call `POST /flow_definitions`. The actual flow is nested under `flow_definition` (line 4).

wim07101993: yeah, the example describes the request, not the actual model. But for me that is not really a problem. Since for the user of the api, the only thing that matters is the request.

vitorbari: the body is composed of `project_id`, `schema_uri` and `flow_definition`. 
Maybe in the future we can move the `project_id` and `schema_uri` to query params or headers, so that the body becomes only the actual flow-definition.
cc @grvijayan

grvijayan (approved):

PR #253, opened 2026-06-10
feat: collect profile fields in default register flow

vitorbari:
## Summary

Expand the default login flow's `register` step to collect the user's profile attributes (`givenName`, `familyName`, `dateOfBirth`) alongside their email. Newly provisioned projects now get a realistic registration flow out of the box instead of an email-only signup.

The default human user schema is updated in step so the flow has matching properties to read:

| Field | Type | Required | Notes |
|---|---|---|---|
| `givenName` | string, maxLength 50 | no | Replaces `name` — there is currently a bug in the frontend when using `name` as an attribute, so the schema can't expose a top-level `name` field |
| `familyName` | string, maxLength 50 | no | Pairs with `givenName` |
| `dateOfBirth` | string, format `date` | no | Replaces `phoneNumber` — picked to demonstrate a string field with a non-default `format` |


wim07101993:
I don't know whether we should expand the default schema. We should think about what the default should look like, but I do not think we should just add fields to it. The entire thing was to let the customer determine their schema. Choosing their own wording for first/given last/family name etc. Currently the default is a bare minimum to get the system up and running, what more do we need?

vitorbari:
> I don't know whether we should expand the default schema. We should think about what the default should look like, but I do not think we should just add fields to it. The entire thing was to let the customer determine their schema. Choosing their own wording for first/given last/family name etc. Currently the default is a bare minimum to get the system up and running, what more do we need?

Hi @wim07101993 , thanks for the review.

The main idea was to expose the fields in the default flow-definition to show the flexibility of the user-schema. `name` and `phoneNumber` are not there, and the registration step does not collect them.

We can keep `name` and `phoneNumber`, the reason to change them was:
- there is a bug with fields called `name`, there are not being rendered properly by the frontend (I already reported it), so I replaced it with given and family name
- `phoneNumber` was replaced with `dateOfBirth` to show/test a different format - phone is a simple text, and `dateOfBirth` is a date.

wim07101993 (approved):

vitorbari:
Screenshot 2026-06-12 at 14 46 16

PR #258, opened 2026-06-11
docs: clarify user team lifecycle ownership

fforootd:
## Summary
- Add ADR 022 to define project-owned, team-owned, and externally managed users.
- Clarify that memberships are access relationships, while FGA only answers authorization.
- Update hierarchy, resource map, glossary, and platform docs so `DELETE` reads as lifecycle/tombstone behavior, not blind cascade deletion.

## Testing
- Not run (not requested)
- Validated markdown links and checked the edited docs for conflicting containment/cascade wording.


docs/adrs/024-user-team-lifecycle-ownership.md
```diff
@@ -0,0 +1,108 @@
...
+FGA is the **authorization decision layer**. It answers whether a principal can
+perform an action on a resource by evaluating memberships, grants, roles,
+credential class, and resolved resource scope. FGA does not decide whether a
+user identity should be deprovisioned or purged.
+
+Every user has exactly one lifecycle owner inside the project:
+
+| Lifecycle owner | Meaning | Typical source |
+|---|---|---|
+| `self` | The user owns its own lifecycle inside the project and may have zero or many team memberships. | Self-serve signup, user-created workspaces. |
```
vitorbari, line 45: Shouldn't signup auto-create a personal team so they're always in at least 1 team, like Vercel's personal account? That gives a simpler UX, you always operate inside a team and can toggle between them.

elinashoko: Yes, agree, that's the best UX, otherwise you have to request access to a team and you're either blocked or forced to create a new team anyway to access the portal.

adlerhurst: IMO this depends on the use case. For self hosted no, for zitadel.cloud yes, for B2B it depends(?)

fforootd: I agree this is probably the right default UX for Zitadel platform accounts: a self-owned signup can auto-create a personal team/workspace so users always operate in a team context. I’d keep that as provisioning/product policy rather than lifecycle ownership, though. The user remains self-owned; the personal team owns workspace resources. Deleting/deprovisioning the user removes their access, but should not transitively delete teams/resources they created without an explicit transfer/delete policy.

docs/adrs/024-user-team-lifecycle-ownership.md
```diff
@@ -0,0 +1,112 @@
...
+  proof that the team owns the user identity.
+- External IdPs and directories do not become a third lifecycle-owner target in
+  the database. They are source/provisioning metadata attached to a local
+  self-owned or team-owned user.
+- Team/resource ownership is not inherited from the creator's lifecycle owner.
+  A user owned by Team A can create Team B; deprovisioning the user through Team
+  A removes that user's access to Team B but does not delete Team B.
+- Current or transitional storage artifacts such as `users.team_id` or
+  team-to-user `ON DELETE CASCADE` constraints are not canonical lifecycle
+  semantics. Schema follow-up should align storage with this ADR instead of
```
vitorbari, line 99: I created a placeholder issue for the follow-up schema changes: https://github.com/zitadel/nextgen/issues/267

vitorbari (approved):

PR #259, opened 2026-06-11
feat: flow back-navigation via back actions and the browser back gesture (ADR 022)

### LINKED ISSUE #467: back navigation: browser history API integration
## Context

Flow engine now supports server-driven back navigation ([ADR 022], PR #453). On rendered step responses, the engine injects a `back` action whenever the current step has a non-empty back stack and is non-terminal:

```json
{
  "actions": [
    { "name": "submit", "kind": "submit", "text_key": "password.action.submit", "primary": true },
    { "name": "back", "kind": "back", "text_key": "password.action.back" }
  ]
}
```

**Contract details from the ADR:**

- **Identify the back action by `kind`, never by name.** The name is conventionally `"back"` but not load-bearing.
- **Presence of the action is the single source of truth** for whether back is available on the current step. The engine already handles the rules (non-empty history AND non-terminal AND not past an irreversible mutation).

## Sync browser history with flow steps

Use **fragment-based `pushState`** so the browser gets a history stack that mirrors the flow's step progression. Fragment-only URL changes are same-document navigation — no reload, no unmount, no state loss — the widget stays connected and only `popstate` fires.

### On each submit response (after `applyResponse`)

- If the new step has a `kind: "back"` action AND the step name changed:
  ```ts
  history.pushState(null, '', '#s' + this.stepSeq++);
  ```
- Otherwise (same step re-render for validation errors / captcha / passkey re-issue, or the new step has no back action): **no history call**. Do not `replaceState`.

State payload is `null` — fragment increment is the only signal. Server is authoritative for flow state; the browser stack just mirrors traversal.

### On `popstate`

- If the current step has a `kind: "back"` action: look up the action's name by kind and submit `{ action: <that name> }`. Apply the response.
- If the current step does **not** have a back action (initial step, terminal step, or past an irreversible mutation): call `history.forward()` to restore the consumed entry (keeps the stack fixed) and surface a brief visual indicator that back is unavailable. This avoids trapping the user in an ever-growing back loop — once the flow's entries are exhausted, the host page becomes reachable again.

### On `disconnectedCallback`

- Remove the `popstate` listener.

### Forward gesture

- On popstate that would move forward (user pressed browser Forward after going back): call `history.back()` to undo the traversal. Flow state is server-authoritative — the client cannot skip ahead.

### Render the back action

- The `default.liquid` template already iterates all non-primary actions, so a `kind: "back"` action renders automatically as a secondary button — **no template changes required**.
- Optional visual polish: render back as a left-arrow link above the card (common auth UI pattern) by selecting on kind:

  ```liquid
  {% assign back = actions | where: "kind", "back" | first %}
  {% if back %}
    <a class="zl-card-nav__link" data-action="{{ back.name }}">
      {{ back.text_key | t }}
    </a>
  {% endif %}
  ```

### Translations

Add `<step>.action.back` per step that renders it. Already covered for the three interactive steps in the current graph (`password`, `register`, `register-password`); add more when new steps join.

## Acceptance

- Continue on a step that returns a new step with `kind: "back"` → new fragment pushed (`#s1`, `#s2`, …).
- Validation error re-render of the same step → **no** new history entry.
- Browser Back on a step with `kind: "back"` action → engine re-renders the previous step (prefill + purpose restored).
- Browser Back on the initial step → no `kind: "back"` action → browser navigates the host page (leaves the flow).
- After an irreversible mutation (`create_user` / passkey register) → no `kind: "back"` action rendered → browser Back leaves the flow rather than rewinding past the mutation.
- Browser Forward after going back → no forward navigation in the flow (`history.back()` restores alignment).
- Widget removed → `popstate` listener detached.

[ADR 022]: docs/adrs/022-flow-back-navigation.md

peintnermax:
## Summary

Client-side implementation of [ADR 022 – Flow Back-Navigation](docs/adrs/022-flow-back-navigation.md), building on the engine-side support already on `main`. Back arrives through the existing action contract (`kind: "back"` — identified by kind, never by name) and is surfaced two ways, sharing a single submit path:

- **Browser back gesture** — while the current step supports back-navigation, the orchestrator keeps exactly one same-document history entry (a *sentinel*, pushed with the current URL — no fragments, no visible URL change, no interaction with hash-router hosts) so the native back gesture fires `popstate` instead of leaving the page. The handler re-arms the sentinel and submits the step's back action; when a step without a back action renders, the sentinel is retired and the next back press navigates the host page. Arming only on the unarmed → armed transition means consecutive back-capable steps and same-step re-renders never grow the stack. Forward traversal onto a retired sentinel is bounced (flow state is server-authoritative), a traversal landing *on* the sentinel (e.g. backing out of a host-pushed `#anchor` entry) is a deliberate no-op, and the listener is added/removed in `connectedCallback`/`disconnectedCallback`. Two SPA-host robustness guards: the sentinel *spreads* the host's `history.state` (vue-router/Nuxt keep `position`/`back`/`forward` there and read them on popstate), and retirement only traverses while the sentinel is still the top entry — if the host pushed its own entry after arming, the sentinel leaks (skipped in one hop by the handler) instead of popping a host entry.
- **No visible control** — back-navigation is gesture-only. Templates render nothing for the action; the `kind: "back"` action and its `text_key` stay on the wire, so tenant templates that want an explicit control can still render one from the contract.

Supporting changes:

- **Templates** — `kind: "back"` is excluded from the generic secondary-action loop in all six templates (default + the five branding designs), so an engine-injected back can never render as an off-design secondary button. Templates forward wire data only and encode no flow topology.
- **`@zitadel/api-mock`** — `password` and `register-password` advertise the back action with guarded machine transitions (token rotation included); initial steps advertise none. The unreachable legacy `passkey-upsell` state no longer advertises back or carries a dead guard encoding a predecessor no path produces (review feedback).
- **ADR 022** — updated to match this implementation and flipped to *Accepted*: sentinel lifecycle and edge cases replace the earlier per-step fragment/`stepSeq` draft (kept as a documented, reviewed-away alternative), §Template rendering documents the gesture-only decision and the kind-based exclusion, §Mock server changes reflects the explicit mock edges.
- **Docs** — `demo-next` / `demo-nuxt` READMEs gain a "Running against the Go server" section, with review fixes applied: `curl` sends the required `name`, the response sample uses `projectSecret` (camelCase), and the unused `ZITADEL_PROJECT_SECRET` env var is dropped in favor of a note on what the secret is actually for. `README-branding.md` documents how tenant templates can render their own back control from the wire action.

## Validation

- `packages/components` — orchestrator suite incl. the `back-navigation (ADR 022)` tests: sentinel pushed once without touching the URL, gesture submits the back action then re-arms and retires the sentinel, popstate on a step without back leaves history alone, forward-onto-retired-sentinel bounces, sentinel-landing popstate is ignored, retiring never pops a host-pushed entry, no visible control is rendered for the action, listener removed on disconnect. Per-design catalog tests assert every shipped design renders nothing for the back action and never a secondary button. `tsc --noEmit` clean.
- `packages/api-mock` — back transitions and token rotation covered (`register-password → back → register`, no back on initial steps); `tsc --noEmit` clean.
- `packages/config` — template validation suite passes against the migrated designs.
- [x] Manual: browser back/forward gestures in `demo-next` / `demo-nuxt` against the mock and the Go server.

## Release notes / changeset

`.changeset/components-flow-back-navigation.md` — `@zitadel/components: patch` (sentinel gesture handling, default template exclusion) and `@zitadel/config: patch` (branding design catalog exclusion). Patch matches the engine-side `flow-back-navigation` changeset already shipped and keeps the fixed release group on the 0.1.0 alpha train.

## Notes

- The generic "Back" control was deliberately removed rather than restyled: back splits into *data correction* (served contextually where needed) and *step traversal* (served by the browser gesture); a generic control serves both poorly and adds card noise. Alternatives considered and rejected: an email-chip affordance (encodes flow topology only the engine knows) and a hover-revealed title chevron (prototyped, then dropped in favor of gesture-only — the wire contract still lets tenant templates render an explicit control).
- The sentinel design replaced an earlier per-step fragment approach whose stack drifted from the flow position (unreachable forward detection, a back-button trap once stale entries accumulated, growth on failed-submit re-renders). ADR 022 §Why a single re-armed entry records the reasoning.
- A sentinel left on the stack when the widget disconnects is a harmless no-op entry — deliberately not cleaned up, since Lit disconnects also fire on DOM moves and touching history there could fight a host router.
- Part of the diff in `zitadel-login.spec.ts` and `locales/it.ts` is formatting-only churn (prettier re-wrapping).


fforootd:
Is this still relevant?

packages/components/src/orchestrator/zitadel-login.ts
```diff
@@ -523,6 +548,18 @@ export class ZitadelLogin extends LitElement {
...
+    // History API: push a new entry when the step supports back-navigation
+    // so the browser's back gesture fires `popstate`. Steps without a
+    // `kind: "back"` action get no new entry — the browser's back button
+    // then navigates the host page (leaves the flow), which is correct.
+    if (typeof window !== "undefined") {
+      const hasBack = Boolean(wire.step.actions?.some((a) => a.kind === "back"));
+      if (hasBack) {
+        history.pushState({ zl: true, seq: this.stepSeq }, "", `#s${this.stepSeq++}`);
+      }
+    }
```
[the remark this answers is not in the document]
peintnermax: this is fixed now

packages/components/src/orchestrator/zitadel-login.ts
```diff
@@ -1008,6 +1045,38 @@ export class ZitadelLogin extends LitElement {
...
+
+    const backAction = this.response.step.actions.find((a) => a.kind === "back");
+    if (backAction) {
+      void this.submit(backAction.name);
+    } else {
+      // No back action — restore the consumed history entry so the
+      // browser's back button eventually reaches the host page.
+      history.forward();
+    }
```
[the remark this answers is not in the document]
peintnermax: outdated

packages/components/src/orchestrator/zitadel-login.ts
```diff
@@ -523,6 +559,28 @@ export class ZitadelLogin extends LitElement {
...
+      const hasBack = Boolean(wire.step.actions?.some((a) => a.kind === "back"));
+      if (hasBack && !this.armed) {
+        history.pushState({ zl: true }, "");
+        this.armed = true;
+      } else if (!hasBack && this.armed) {
+        this.armed = false;
+        this.ignoreNextPop = true;
+        history.back();
+      }
+    }
```
fforootd, line 559: Two robustness issues in the arm/retire block.

**Replacing `history.state` wholesale.** vue-router (so, Nuxt) keeps `{ back, current, forward, position, scroll }` in `history.state` and reads `position` on `popstate` to infer traversal direction. Overwriting it with `{ zl: true }` leaves the sentinel entry opaque to the host router — and `demo-nuxt` is one of the two apps this PR documents. Spreading costs nothing and the `zl` marker still identifies the sentinel.

**The retire assumes the sentinel is still the top entry.** If the host app pushed its own entry after we armed, the bare `history.back()` pops *that* one, and `ignoreNextPop` then swallows the resulting `popstate` — a silently undone host navigation. Guarding on `history.state` means the worst case degrades to a leaked sentinel, which is the same tradeoff already accepted for `disconnectedCallback`.

```suggestion
    if (typeof window !== "undefined") {
      const hasBack = Boolean(wire.step.actions?.some((a) => a.kind === "back"));
      if (hasBack && !this.armed) {
        // Spread the host's state: vue-router (Nuxt) keeps `position` /
        // `back` / `forward` here and reads them on popstate. Replacing it
        // wholesale leaves the sentinel opaque to the host router.
        history.pushState({ ...history.state, zl: true }, "");
        this.armed = true;
      } else if (!hasBack && this.armed) {
        this.armed = false;
        // Only traverse while we still own the current entry. If the host
        // pushed its own entry after we armed, `history.back()` would pop
        // *that* one and `ignoreNextPop` would swallow the popstate,
        // silently undoing a host navigation. Leaving a stale sentinel
        // behind is the lesser evil — same tradeoff as disconnect.
        if ((history.state as { zl?: boolean } | null)?.zl === true) {
          this.ignoreNextPop = true;
          history.back();
        }
      }
    }
```

Worth re-running the suite to confirm, but I traced both the link-click and gesture paths and the existing assertions should still hold — `history.state.zl` is true at every retire point they exercise.

packages/components/src/orchestrator/zitadel-login.ts
```diff
@@ -1008,6 +1066,46 @@ export class ZitadelLogin extends LitElement {
...
+        void this.submit(backAction.name);
+      }
+      return;
+    }
+
+    if ((event.state as { zl?: boolean } | null)?.zl === true) {
+      this.ignoreNextPop = true;
+      history.back();
+    }
+  }
```
fforootd, line 1094: `void this.submit(...)` here is the only submit path in the class without a `loading` guard — compare `handleAtomSubmit` (:864), `handleDelegatedAction` (:882), `handleFormSubmit` (:913).

ADR 022 lists "multiple rapid back presses" as an edge case and says token rotation prevents races. It doesn't *prevent* anything — it makes the second request **fail**: two pops inside one round-trip fire two submits carrying the same not-yet-rotated session token, and the second comes back a token mismatch that surfaces as a transport error. Two swipe-backs on mobile is an entirely ordinary gesture, so this is reachable in normal use rather than under stress.

Re-arming unconditionally also drops the transient `armed = false`, so the stack shape stays identical whether or not we act on the gesture:

```suggestion
  private onPopState(event: PopStateEvent): void {
    if (this.ignoreNextPop) {
      this.ignoreNextPop = false;
      return;
    }

    if (this.armed) {
      // Back press: the browser popped the sentinel. Re-arm unconditionally
      // so the stack shape is identical at every flow depth — including when
      // we decline to act on the gesture below.
      history.pushState({ ...history.state, zl: true }, "");

      // Decline while a submit is in flight. Every other submit path guards
      // on `loading`; without it two quick back gestures (an ordinary mobile
      // swipe repeat) fire two submits sharing one session token, and the
      // second fails the rotation check as a transport error.
      if (this.loading) return;

      const backAction = this.response?.step?.actions?.find((a) => a.kind === "back");
      if (backAction) void this.submit(backAction.name);
      return;
    }

    if ((event.state as { zl?: boolean } | null)?.zl === true) {
      this.ignoreNextPop = true;
      history.back();
    }
  }
```

The docblock just above wants a line about the in-flight case too.

packages/api-mock/src/flow-machine.ts
```diff
@@ -227,6 +234,11 @@ export const flowMachine = createMachine({
     "passkey-upsell": {
       on: {
         SUBMIT: [
+          {
+            guard: ({ event }) => event.action === "back",
+            target: "identifier",
+            actions: [rotateToken],
+          },
```
fforootd: `passkey-upsell` is unreachable — the machine's own header says so: "legacy upsell pair; the default flow no longer routes through them (passkey registration is offered up front instead). Kept so tests can target them directly via actor injection." I checked, and no state targets it.

So this guard never fires, and the target encodes a predecessor relationship (`passkey-upsell` → `identifier`) that no path in the machine actually produces — which is the kind of thing that reads as ground truth to whoever revives this state later. The `register-password` → `register` edge is the one that earns its keep; I'd drop this one.

```suggestion
```

packages/api-mock/src/fixtures/login.ts
```diff
@@ -281,6 +282,7 @@ export function passkeyUpsellStep(input: StepFixtureInput): CreateFlow201 {
         primary: true,
       },
       { name: "skip", kind: "navigate", text_key: "passkey-upsell.action.skip" },
+      { name: "back", kind: "back", text_key: "action.back" },
```
fforootd: Same as the machine edge — `passkey-upsell` is unreachable in the default flow, so this action is never rendered by anything. Worth dropping alongside the guard, otherwise the fixture asserts a back capability that the mock's own flow can't reach.

```suggestion
```

apps/demo-next/README.md
```diff
@@ -45,6 +45,64 @@ moon run demo-next:dev
...
+```
+
+The response contains the `id` and `projectSecret`:
+
+```json
+{
+  "id": "proj_01JXXXXXXXXXXXXXXXX",
+  "project_secret": "eyJhbGci..."
+}
+```
```
fforootd, line 82: Two problems in this block.

`-d '{}'` returns 400 — `name` is required (`api/openapi/endpoints/projects/create-project-request.yaml:2-3`). And the JSON sample says `project_secret`, but the field is `projectSecret` (`create-project-response.yaml:16`); the demo-nuxt README has this one right. Copilot flagged the casing and suppressed it as low-confidence — it's correct.

````suggestion
```bash
curl -s -X POST http://localhost:8080/projects \
  -H 'Content-Type: application/json' \
  -d '{"name":"demo"}' | jq .
```

The response contains the `id` and `projectSecret`:

```json
{
  "id": "proj_01JXXXXXXXXXXXXXXXX",
  "projectSecret": "eyJhbGci..."
}
```
````

File: apps/demo-next/README.md [hunk 8]
```diff
@@ -45,6 +45,64 @@ moon run demo-next:dev
...
+
+#### 3. Configure `.env.local`
+
+Update `apps/demo-next/.env.local` with the values from step 2:
+
+```env
+ZITADEL_URL=http://localhost:8080
+NEXT_PUBLIC_ZITADEL_PROJECT_ID=proj_01JXXXXXXXXXXXXXXXX
+ZITADEL_PROJECT_SECRET=eyJhbGci...
+```
```
fforootd, line 91: `ZITADEL_PROJECT_SECRET` isn't read by anything — not by demo-next, not by demo-nuxt, not by `@zitadel/api`. The demos only read `ZITADEL_URL` and `NEXT_PUBLIC_ZITADEL_PROJECT_ID` (`apps/demo-next/src/zitadel.ts:14-15`).

Instructing people to paste a project credential into a file where it has no effect is worse than omitting it: it teaches a habit of spreading secrets around for no gain, and it silently misleads anyone debugging an auth failure. Either drop it or wire it up.

`.env.example` also carries `NEXTGEN_ISSUER_URL`, which this section skips — worth a line if it matters for the Go-server path.

````suggestion
```env
ZITADEL_URL=http://localhost:8080
NEXT_PUBLIC_ZITADEL_PROJECT_ID=proj_01JXXXXXXXXXXXXXXXX
```
````

File: apps/demo-nuxt/README.md [hunk 9]
```diff
@@ -45,6 +45,64 @@ moon run demo-nuxt:dev
...
+`nextgen-data/`). The server is ready when you see
+`Local Zitadel server is ready.`
+
+#### 2. Create a project
+
+```bash
+curl -s -X POST http://localhost:8080/projects \
+  -H 'Content-Type: application/json' \
+  -d '{}' | jq .
+```
```
fforootd, line 73: `-d '{}'` returns 400 — `name` is required on this endpoint (`api/openapi/endpoints/projects/create-project-request.yaml:2-3`). The JSON sample below is correct here (unlike demo-next's).

````suggestion
```bash
curl -s -X POST http://localhost:8080/projects \
  -H 'Content-Type: application/json' \
  -d '{"name":"demo"}' | jq .
```
````

File: apps/demo-nuxt/README.md [hunk 10]
```diff
@@ -45,6 +45,64 @@ moon run demo-nuxt:dev
...
+
+#### 3. Configure `.env`
+
+Update `apps/demo-nuxt/.env` with the values from step 2:
+
+```env
+ZITADEL_URL=http://localhost:8080
+NUXT_PUBLIC_ZITADEL_PROJECT_ID=proj_01JXXXXXXXXXXXXXXXX
+ZITADEL_PROJECT_SECRET=eyJhbGci...
+```
```
fforootd, line 91: Same as demo-next: `ZITADEL_PROJECT_SECRET` is read by nothing. `nuxt.config.ts:32-35` only picks up `ZITADEL_URL` and `NUXT_PUBLIC_ZITADEL_PROJECT_ID`.

````suggestion
```env
ZITADEL_URL=http://localhost:8080
NUXT_PUBLIC_ZITADEL_PROJECT_ID=proj_01JXXXXXXXXXXXXXXXX
```
````

**Review by fforootd — CHANGES_REQUESTED** [review]
Reviewed the client half of ADR 022 against current `main`. The sentinel design is the right call and an improvement on what the ADR specified — no URL writes, no `stepSeq`, the stack never grows regardless of flow depth. Most of what follows is about surfaces the PR didn't reach; inline suggestions are attached where they're mechanical.

**Still relevant?** Yes. `main` has no `popstate` handling at all, so the browser-gesture half is genuinely un-landed. The api-mock half has been partly overtaken — `main` already ships `kind: "back"` on `passwordStep` (`fixtures/login.ts:243`) and a back guard on the `password` state (`flow-machine.ts:211`).

---

## The one that would ship broken: the five branding presets weren't updated

`default.liquid` is only the bundled fallback. When a project publishes branding, the CLI inlines `login.liquid` into `branding.liquid_template` (`packages/config/src/defaults.ts:116`) and the orchestrator renders **that** instead — there's already a spec for exactly that precedence ("renders a server-sent branding.liquid_template instead of the bundled default").

All five shipped presets still carry the un-patched loop:

| File | Line |
|---|---|
| `packages/config/defaults/branding/centered/login.liquid` | 101 |
| `packages/config/defaults/branding/hero/login.liquid` | 157 |
| `packages/config/defaults/branding/minimal/login.liquid` | 92 |
| `packages/config/defaults/branding/split/login.liquid` | 113 |
| `packages/config/defaults/branding/split-right/login.liquid` | 113 |

So on any branded deployment the injected back action falls through the secondary loop and renders as a full-width secondary **"Back" button**, with no nav link — the opposite of the intended treatment, and inconsistent with the browser gesture, which works everywhere. Each file needs the same two hunks the default got:

```diff
-      {% unless a.primary or a.name == 'passkey' or a.name == 'register' or a.name == 'sign_in' or a.name == 'recover' %}
+      {% unless a.primary or a.kind == 'back' or a.name == 'passkey' or a.name == 'register' or a.name == 'sign_in' or a.name == 'recover' %}
```

plus the nav block, alongside the `zl-card-nav` rows these templates already have for `register` / `sign_in`:

```liquid
{% assign back = actions | where: "kind", "back" | first %}
{% if back %}
  <p class="zl-card-nav">
    <a href="#" class="zl-card-nav__link" data-action="{{ back.name }}" data-testid="zitadel-action-back-link">{{ back.text_key | t }}</a>
  </p>
{% endif %}
```

Two assertions in `liquid.spec.ts` — a `kind: "back"` action produces `zl-card-nav__link` and does *not* produce a `<zl-button>` — would have caught this, and would keep catching it as presets get added. That file already renders templates directly and asserts on `zl-card-nav` (line 599).

---

## ADR 022 now contradicts what shipped

The ADR is still `Status: Proposed` and specifies fragment entries (`#s1`/`#s2`), a `stepSeq` counter, `history.forward()` on unavailable-back plus a visual indicator, and — in §Template rendering — "no template changes required". The implementation deliberately does none of that, for good reasons that the PR description already lays out well. Since the ADR is the authority in this repo, leaving a doc describing a design nobody shipped is a trap for the next reader. Suggest porting the PR description's prose into §Browser History API integration, dropping the §Why fragments rationale, fixing §Template rendering, and flipping to Accepted.

## Changeset

`@zitadel/components` is publishable and CI flags it. The PR body already drafts the changeset — just commit it.

## Formatting churn (~250 lines)

I checked rather than assumed: with the repo config (`printWidth: 100`) and prettier 3.8.3, **both** `main`'s version and this branch's version of `zitadel-login.spec.ts` and `locales/it.ts` come back clean — zero changes either way. Prettier's default `objectWrap: "preserve"` makes both forms stable, so this isn't a formatting fix, it's noise. It's roughly 40% of the PR's additions, it buries the substantive `back-navigation (ADR 022)` block, and it will conflict with anything else touching that spec. Please revert the formatting-only hunks.

## Test coverage

The new suites are well-targeted and the assertions carry real weight — `expect(zlCalls[0]?.[2]).toBeUndefined()` pinning "no URL argument" is exactly the right thing to lock down. Gaps:

- **`settleHistory()` is a flake vector.** A fixed five-macrotask drain to absorb popstates bleeding in from *previous tests* is a symptom, not a fix — history state is never reset between tests. Either reset the stack in `afterEach` or stub the History API instead of driving jsdom's real one.
- **The disconnect test is too weak.** `popstateCalls.length >= 1` passes even if the removed listener were a different function reference. Assert the same bound handler reached both `addEventListener` and `removeEventListener` — or, functionally: dispatch a `popstate` after `.remove()` and assert no submit.
- **Every history test is a synthetic `new PopStateEvent(...)`.** Nothing performs a real traversal, so arm/retire/bounce sequencing against an actual stack is covered only by the manual checkbox. `zitadel-login.browser.spec.ts` exists for this class of thing; one real back-gesture case there would be worth more than the five synthetic ones.

## Nits

- Fixtures use `text_key: "action.back"`, but the engine emits `<step>.action.back` (`internal/domain/flow_state_machine.go:1160`). Both resolve — the locales carry a generic fallback — but it means mock-driven tests never exercise the key production actually sends. Pre-existing on `main`, so out of scope unless you want it swept here.

## Security & performance

Nothing concerning. Back rides the normal submit path with token rotation and server-authoritative state; the forward-bounce correctly refuses to let the browser skip ahead. No URL writes means nothing leaks into history entries or referrers, and `pushState` stays bounded to one entry regardless of flow depth.

apps/demo-nuxt/README.md
peintnermax: done

apps/demo-nuxt/README.md
peintnermax: done

apps/demo-next/README.md
peintnermax: done

apps/demo-next/README.md
peintnermax: done

packages/api-mock/src/fixtures/login.ts
peintnermax: done

packages/api-mock/src/flow-machine.ts
peintnermax: done

peintnermax:
I wasn't fully happy with the back nav design and since we do not have figmadesigns providing a generic back navigation, I removed the link from all templates so it can no longer render as a generic secondary button nor link.

I did implement a zl-title component instead which detects the kind back action and renders a hidden chevron on hover. 
This allows us to have a clean, generic back nav which is always rendered the same way not polluting other secondary actions most of the time within the main content of the template.

regular appearance:
Screenshot 2026-07-29 at 11 09 03
appearance when hovering/focusing title:
Screenshot 2026-07-29 at 11 09 09

Let me know if you like that approach. I'll review it with @jjjssswww 

I've also regenerated the PR description to match, and revised ADR 022 accordingly (sentinel replaces the fragment/`stepSeq` draft, §Template rendering fixed, status → **Accepted**).

packages/components/src/orchestrator/zitadel-login.ts
fforootd: Withdrawing this — I was wrong, please don't implement it.

`submit()` already guards on `loading` itself (`zitadel-login.ts:985`):

```ts
private async submit(action, challengeResponse?): Promise<void> {
  if (!this.response || this.loading) return;
```

So the guards in `handleAtomSubmit` / `handleDelegatedAction` / `handleFormSubmit` are belt-and-braces at the call site, not the only protection — this call site isn't missing anything. Two pops inside one round-trip do **not** produce two requests: the second `submit()` returns before touching the network, so there's no shared-token mismatch and no transport error. The ADR's edge-case row ("submits are serialized by the loading guard, so presses during an in-flight submit are absorbed") is accurate as written.

The unconditional-re-arm half doesn't earn its keep either: `armed === true` implies the last applied step carried a back action, so `backAction` is always found and the re-arm already runs before every submit. `if (backAction)` is defensive, not a branch that fires.

Sorry for the noise. The sibling thread on the arm/retire block (`:582`) still stands.

---

packages/components/src/atoms/zl-title.ts
```diff
@@ -0,0 +1,143 @@
...
+        border-radius: var(--zl-radius-sm, 0.375rem);
+        background: none;
+        color: inherit;
+        cursor: pointer;
+        opacity: 0;
+        transform: translate(-0.5rem, -50%);
+        transition:
+          opacity 160ms ease,
+          transform 160ms ease;
+      }
```
fforootd: **The `<h1>` sits on top of the chevron, so the first click/tap is swallowed.**

`transform: translateX(0)` on `.zr-title__heading` (:51) makes the heading a stacking context, which paints it with the positioned descendants — same z-index as the absolutely positioned `.zr-title__back`, later in DOM order, so the heading wins hit-testing across the whole overlap. The chevron only becomes hittable once the heading has translated far enough out from under it.

I checked in Chromium rather than reasoning from the spec — `shadowRoot.elementsFromPoint()` at the chevron's centre:

```
["H1.zr-title__heading", "BUTTON.zr-title__back", "DIV.zr-title zr-title--backable", ...]
```

and driving it for real:

```
touch tap #1: chevron opacity=1  fired=""                  ← nothing happens
touch tap #2: chevron opacity=1  fired="BACK-SUBMIT-FIRED"
desktop hover+  0ms -> topElement=H1     click fired=false
desktop hover+ 40ms -> topElement=BUTTON click fired=true
```

On desktop it's a sub-40ms dead window — real but rarely hit, since the chevron is still fading in. On touch it's the whole interaction: tap #1 lands on the heading and only sets sticky `:hover`, tap #2 submits. Two taps, the first indistinguishable from a broken control.

One line fixes both — verified, first click and first tap then fire:

```suggestion
      .zr-title__heading {
        margin: 0;
        font: inherit;
        color: inherit;
        letter-spacing: inherit;
        text-align: inherit;
        transform: translateX(0);
        transition: transform 160ms ease;
      }
      .zr-title__back {
        position: absolute;
        left: 0;
        top: 50%;
        /* The heading's `transform` makes it a stacking context painted with
           the positioned descendants — same z-index, later in DOM order, so
           it would occlude this button until it slides clear. */
        z-index: 1;
        display: flex;
        align-items: center;
        justify-content: center;
        width: 2rem;
        height: 2rem;
        padding: 0;
        border: 0;
        border-radius: var(--zl-radius-sm, 0.375rem);
        background: none;
        color: inherit;
        cursor: pointer;
        opacity: 0;
        transform: translate(-0.5rem, -50%);
        transition:
          opacity 160ms ease,
          transform 160ms ease;
      }
```

---

packages/components/src/atoms/zl-title.ts
```diff
@@ -0,0 +1,143 @@
...
+      :host(:hover) .zr-title__back,
+      .zr-title__back:focus-visible {
+        opacity: 1;
+        transform: translate(0, -50%);
+      }
+      :host(:hover) .zr-title--backable .zr-title__heading,
+      .zr-title--backable:has(.zr-title__back:focus-visible) .zr-title__heading {
+        transform: translateX(2.5rem);
+      }
```
fforootd: **The affordance has no touch story.** The only reveal triggers are `:host(:hover)` and `:focus-visible`. On a `hover: none` device neither resolves without interaction, so the chevron sits at `opacity: 0` indefinitely — I confirmed `(hover: none)` matches and the resting opacity stays `0` under both Pixel 7 and iPhone 14 emulation.

That leaves phones with no visible back control anywhere in the UI. The browser gesture still works, which is why this isn't a blocker — but it's invisible chrome in a standalone PWA or an embedded WebView, and the PR positions the chevron as *the* visible affordance. Right now it's a desktop-only one.

Pinning it visible where hover doesn't exist covers it (media queries resolve fine inside the shadow root — I checked):

```suggestion
      :host(:hover) .zr-title__back,
      .zr-title__back:focus-visible {
        opacity: 1;
        transform: translate(0, -50%);
      }
      :host(:hover) .zr-title--backable .zr-title__heading,
      .zr-title--backable:has(.zr-title__back:focus-visible) .zr-title__heading {
        transform: translateX(2.5rem);
      }
      /* No hover to reveal on: show the chevron outright rather than leaving
         touch users with the browser gesture as the only way back. */
      @media (hover: none) {
        .zr-title--backable .zr-title__back {
          opacity: 1;
          transform: translate(0, -50%);
        }
        .zr-title--backable .zr-title__heading {
          transform: translateX(2.5rem);
        }
      }
```

Worth a design call rather than taking this verbatim — a permanently-visible chevron on mobile is a different look from the hover reveal, and you may prefer to solve it another way. But leaving it hover-only means shipping it as desktop-only.

---

packages/components/src/atoms/zl-title.ts
```diff
@@ -0,0 +1,143 @@
...
+  override render() {
+    const backable = Boolean(this.backAction);
+    const classes = classMap({ "zr-title": true, "zr-title--backable": backable });
+    return html`<div class=${classes} part="title">
+      ${backable
+        ? html`<button
+            class="zr-title__back"
+            part="back"
+            type="button"
+            aria-label=${this.backLabel ?? "Back"}
```
fforootd: Nit: `??` only catches `null`/`undefined`, not `""`. A step that ships a back action without `text_key` renders `back-label=""` (the `t` filter returns `""` for an empty key rather than a fallback), and the button ends up with an empty accessible name instead of "Back" — worse than no `aria-label` at all, since it suppresses the icon's own naming path.

`||` covers both:

```suggestion
            aria-label=${this.backLabel || "Back"}
```

Low priority — the engine always emits `<step>.action.back` and `t` degrades to the raw key, so this needs a malformed step to surface.

---

packages/components/src/atoms/zl-title.ts
```diff
@@ -0,0 +1,143 @@
...
+    css`
+      :host {
+        display: block;
+      }
+      .zr-title {
+        position: relative;
+        display: flex;
+        align-items: center;
+      }
+      .zr-title__heading {
```
fforootd: Nit, latent: the heading is now a shrink-to-fit flex item, so `text-align` set on the host stops having a visible effect — a box sized to its own text can't align text within itself.

`.zl-card-title` is a customization surface (`customization.browser.spec.ts` asserts against it), so a tenant stylesheet doing `.zl-card-title { text-align: center }` used to centre the heading and now silently won't. No shipped design centres the title — chrome CSS is `text-align: left` throughout — so nothing regresses today.

```suggestion
      .zr-title__heading {
        flex: 1 1 auto;
        min-width: 0;
        margin: 0;
```

---

fforootd (commented):
Re-reviewed at `9bc36a7`. The blocking item from last round is gone — all five branding presets render `<zl-title>` and exclude `kind: "back"` from the secondary loop, with a per-design catalog test that will keep catching it as presets get added. ADR 022 now describes the sentinel that actually shipped, the changeset is in, and the unreachable `passkey-upsell` edge and fixture action are reverted. CI is green.

Everything below is about `<zl-title>`, which didn't exist when I last looked.

## The chevron is effectively desktop-only

Two separate causes, both inline:

1. The heading's `transform: translateX(0)` makes it a stacking context that paints **over** the absolutely positioned chevron, so the first click/tap lands on the `<h1>` and does nothing. On desktop that's a sub-40ms window; on touch it's the entire interaction — tap #1 is a no-op that only sets sticky `:hover`, tap #2 submits.
2. `:host(:hover)` / `:focus-visible` are the only reveal triggers, so on `hover: none` devices the chevron never appears on its own.

Together: on a phone the only visible back control in the product is invisible *and* unresponsive to the first tap. The browser gesture still covers the functionality, which is why I'm not blocking — but it does mean the visible affordance this PR is built around doesn't reach mobile users. I drove both cases in Chromium rather than reading the cascade; the traces are in the inline comments, and `z-index: 1` on the chevron fixes cause (1) outright.

## Two nits

Empty-string `aria-label` from `??` vs `||`, and a latent flex-item change that neutralises `text-align` on `.zl-card-title` for tenant CSS. Both inline, both low.

## Still open from last round

- The **arm/retire** thread (`zitadel-login.ts:582`) is unaddressed. The retire half is the one I'd press: `onPopState` already defends against "the host pushed an entry above the sentinel," and ADR 022 §Edge cases now documents that scenario explicitly — but the retire path's bare `history.back()` still assumes it owns the top entry, so in exactly that case it pops the host's entry and `ignoreNextPop` swallows the `popstate`, silently undoing a host navigation.
- **Test gaps** carried over: `settleHistory()`'s five-macrotask drain (`zitadel-login.spec.ts:1517`) is still papering over history state that's never reset between tests; the disconnect assertion is still `toBeGreaterThanOrEqual(1)` (:1671), which passes even if a different function reference were removed; and every history case is still a synthetic `new PopStateEvent(...)`, so no test performs a real traversal. `zitadel-login.browser.spec.ts` is the natural home for one real back-gesture case — and per the above, a browser-level test is what would have caught the hit-testing bug, since jsdom has no layout.
- **Formatting churn** is much smaller now (~26 lines in the spec, 9 in `it.ts`, all line-joins). I re-checked with the repo config and the installed prettier 3.8.1: both `main`'s and this branch's versions of both files come back clean, so it's still noise rather than a fix — but at this size it's not worth another round trip.

## Withdrawn

My `onPopState` loading-guard comment was wrong — `submit()` guards internally at `:985`, so rapid back presses are already absorbed and the ADR's edge-case row is accurate. Resolved with an explanation on the thread.

---

peintnermax:
> I wasn't fully happy with the back nav design and since we do not have figmadesigns providing a generic back navigation, I removed the link from all templates so it can no longer render as a generic secondary button nor link.
> 
> I did implement a zl-title component instead which detects the kind back action and renders a hidden chevron on hover. This allows us to have a clean, generic back nav which is always rendered the same way not polluting other secondary actions most of the time within the main content of the template.
> 
> regular appearance: Screenshot 2026-07-29 at 11 09 03 appearance when hovering/focusing title: Screenshot 2026-07-29 at 11 09 09
> 
> Let me know if you like that approach. I'll review it with @jjjssswww
> 
> I've also regenerated the PR description to match, and revised ADR 022 accordingly (sentinel replaces the fragment/`stepSeq` draft, §Template rendering fixed, status → **Accepted**).

after further discussion, we remove the default back navigation UI from the default templates. Custom templates remain free to render an explicit control.

packages/components/src/atoms/zl-title.ts
peintnermax: no relevant anymore

packages/components/src/atoms/zl-title.ts
peintnermax: no relevant anymore

packages/components/src/orchestrator/zitadel-login.ts
peintnermax: fixed with https://github.com/zitadel/nextgen/pull/259/commits/a75c18ae101aad3ffc8deeed89be20d6f4ab7b32

fforootd (approved):

PR #263, opened 2026-06-11
feat: user service uses database.QueryExecutor

IAM-marco:
# Problem

userService.SetPassword() needs to start a transaction or re-use an existing one. However, `database.Pool`, which is what is used by userService to access DB, has no idea of whether a transaction is started or not. This is the DB client job. So, when trying to create a password for a user in an already open transaction, the procedure fails, because the system attempts to open another transaction, instead of re-using the existing one.

# Solution

- Switch `database.Pool` paramter to `database.QueryExecutor`
- Type assert db client on `database.Beginner` to make sure the client can handle transaction
- Use `beginner.Begin()` to gracefully start a new transaction or re-use the existing one

# Extras

- Create missing mocks
- Add tests for SetPassword


internal/service/user.go
```diff
@@ -150,15 +150,16 @@ func (s *UserService) SetPassword(ctx context.Context, input SetPasswordInput) (
 		return err
 	}
 
-	tx, err := s.pool.Begin(ctx, nil)
+	beginner, ok := s.dbClient.(database.Beginner)
+	if !ok {
+		return domain.ErrInternal(database.ErrNotABeginner)
+	}
+
+	tx, err := beginner.Begin(ctx, nil)
```
vitorbari, line 158: One of the goals, is to be able to call `CreateUser` and `SetPassword` within the same transaction. How will this nested transaction behave?

IAM-marco:
Closing as we are migrating to a new approach to handle expensive cryptographic ops outside of DB tx

PR #269, opened 2026-06-12
refactor: make create user compositable

wim07101993:
- Extracted logic from service layer to domain layer
- Created `UserAction` which can be used to combine multiple action on user spanning on transaction


internal/service/user.go
```diff
@@ -67,63 +70,53 @@ func NewUserService(
...
-	if !ok {
-		return nil, domain.ErrUserInvalid().
-			WithDetails("No $schema provided for the user. A schema must be provided when creating a new user. Against this schema, the user will be validated")
-	}
-
-	schemaEntity, err := s.schemaRepo.GetByID(ctx, s.pool, input.ProjectID, schemaURL)
-	if err != nil {
-		if _, ok := errors.AsType[*database.NoRowFoundError](err); ok {
-			return nil, domain.ErrUserInvalid().WithDetails("$schema is not known to the system. First create a schema, then create users.")
+func (s *UserService) ApplyActions(ctx context.Context, actions ...UserAction) (err error) {
```
vitorbari, line 73: So, the idea would be to use this function if both `CreateUser` and `SetPassword` are needed?
Something like:
```go
svc.ApplyActions(
  ctx, 
  WithCreateUser(input, s.userRepo, s.schemaRepo, s.pool), 
  WithSetUserPassword(input, s.hasher, s.passwordRepo)
)
```

It works, but there is a chance the order of the actions is invalid, for example `WithSetUserPassword` before `WithCreateUser`.

This pattern is usually used to construct an object, in this case, the User. So the idiomatic way would be to accept the options/actions as part of the `CreateUser`

internal/domain/user.go
```diff
@@ -54,13 +57,44 @@ type CreateUser struct {
...
+func SchemaFromUserMap(user map[string]any) (string, error) {
+	schemaURL, ok := maputil.Get[string](user, "$schema")
+	if !ok {
+		return "", ErrUserInvalid().
+			WithDetails("No $schema provided for the user. A schema must be provided when creating a new user. Against this schema, the user will be validated")
+	}
+	return schemaURL, nil
+}
+
+func NewCreateUser(projectID string, teamID *string, schemabs []byte, muser map[string]any) (*CreateUser, error) {
```
vitorbari, line 60: 💅 keep the constructor next to the struct `CreateUser`

internal/service/user.go
```diff
@@ -213,3 +169,100 @@ func (s *UserService) GetMyUser(ctx context.Context, input GetMyUserInput) ([]by
...
+
+	createUser *domain.CreateUser
+}
+
+func WithCreateUser(input CreateUserInput, userRepo domain.UserRepository, schemaRepo domain.JSONSchemaRepository, db database.Pool) *CreateUserAction {
+	return &CreateUserAction{
+		CreateUserInput: input,
+		userRepo:        userRepo,
+		schemaRepo:      schemaRepo,
+		db:              db,
```
vitorbari: This `db` is not needed if it is already passed to `Apply`.

internal/service/user.go
```diff
@@ -213,3 +169,100 @@ func (s *UserService) GetMyUser(ctx context.Context, input GetMyUserInput) ([]by
...
+	CreateUserInput
+
+	userRepo   domain.UserRepository
+	schemaRepo domain.JSONSchemaRepository
+	db         database.Pool
+
+	createUser *domain.CreateUser
+}
+
+func WithCreateUser(input CreateUserInput, userRepo domain.UserRepository, schemaRepo domain.JSONSchemaRepository, db database.Pool) *CreateUserAction {
```
vitorbari: Because these are not traditional functional options, we can call them `NewCreateUserAction`.

internal/service/user.go
wim07101993: Yes, I had to pick between a pattern which can be generalized and used with other combinations but make the order important, make it implementation specific for the create. I choose the first. If this turns out to be the wrong choice, we can still refactor later I think

internal/service/user.go
wim07101993: it is, because the prepare step needs it to query the schema repo. This is done outside of a transaction since schemas are immutable.

internal/service/user.go
wim07101993: Was doubting whether I should call them With* or New*

internal/service/user.go
vitorbari: wouldn't it make sense for both `Prepare` and `Apply` to receive `db database.QueryExecutor` as an argument?

vitorbari (approved):

PR #287, opened 2026-06-15
fix(database): separate user lifecycle ownership from team membership

### LINKED ISSUE #267: feat: align user/team lifecycle with ADR 023
**Goal:**

Make the storage layer reflect ADR 023: lifecycle ownership, team participation, and authorization are separate concerns. Today's schema collapses them into users.team_id + ON DELETE CASCADE.

**Scope:**

- Replace users.team_id with an explicit lifecycle-owner concept (self or team).
- Introduce team_memberships as an N:N resource with its own status.
- Add first-class user_status (active/suspended/deactivated/pending_purge).

**Acceptance signal:**

Self-owned user survives team deletion. Team-owned user is deactivated, not hard-deleted, when their owning team is deleted. Deleting a user does not cascade-delete teams they created. No ON DELETE CASCADE remains on the user/team graph.

adlerhurst:
## Summary

Fixes #267

Align user/team storage with [ADR 024](docs/adrs/024-user-team-lifecycle-ownership.md):

- Replace `users.team_id` with nullable `lifecycle_owner_team_id` (NULL = self-owned, set = team-owned)
- Add `users.status` (active/suspended/deactivated/pending_purge)
- Add `team_memberships` N:N table with membership status
- Add `teams.status` and lifecycle `Deactivate` methods on team/user repositories
- Remove `ON DELETE CASCADE` on the user/team graph; lifecycle deletes via repository policy
- Backfill existing `team_id` data in migration `000011_user_team_lifecycle.sql` (Postgres + Spanner)

## Validation

- `go test -tags=postgres_integration ./internal/storage/database/repository/... -count=1`
- `corepack pnpm run check -- --only go`

## Release notes / changeset

`.changeset/user-team-lifecycle-storage.md` — `@zitadel/server` patch: align user/team lifecycle storage with ADR 024.

## Notes

Closes https://github.com/zitadel/nextgen/issues/267

Branch merged with `main` before review fixes. Harness mutex fix landed via #289.

Migration backfill maps legacy `users.team_id` to `lifecycle_owner_team_id` as a conservative alpha default (old schema required a team context, so self-serve vs enterprise provenance cannot be distinguished retroactively). New creates stay self-owned unless `LifecycleOwnerTeamID` is set.

Acceptance checklist:
- [x] Self-owned user survives team deletion
- [x] Team-owned user deactivated when owning team deleted
- [x] Deleting a user does not cascade-delete teams they created
- [x] No ON DELETE CASCADE on user/team graph
- [x] Team-owned users lose all memberships when owning team is deactivated

<div><a href="https://cursor.com/agents/bc-f52f2c82-7c40-4962-bf78-39a8915c3d00"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f52f2c82-7c40-4962-bf78-39a8915c3d00"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


internal/storage/database/repository/user.go
```diff
@@ -119,7 +149,18 @@ func WithUserScalarList(project string, filters []UserListScalarFilter, teamScop
 }
 
 func (r *UserRepository) Delete(ctx context.Context, client database.QueryExecutor, condition database.Condition) error {
```
fforootd, line 158: Nice direction on making memberships explicit. One thing I think we should tighten here: the membership cleanup runs before `deleteOne` gets a chance to validate that the condition is constrained by the user primary key. If a broader condition ever reaches this method, we would remove matching memberships first and only then fail the user delete. Could we validate the condition before this cleanup, or run both behind the same already-validated helper/transaction?

fforootd:
One clarification from my side before we settle this PR: is this really intended to be a `feat`?

From how I read the change, this is important storage/model hygiene and lifecycle correctness work, but not necessarily a user-facing feature. Something like `refactor(database): align user/team lifecycle storage with ADR 024` or possibly `fix(database): separate user lifecycle ownership from team membership` may describe it better.

I think its worth mentioning that for [Go we also use changesets now](https://zitadel.slack.com/archives/C0ATLEUKPAN/p1781595079445769). Since we started publishing the Go server through `@zitadel/server` / the platform server packages, meaningful server runtime changes should declare release intent through Changesets too. **That feeds into the single generated GitHub release post**, so even changes under `internal/**` can matter when they change the shipped server behavior.

I noticed our docs are a bit inconsistent on this 😅 : some places still imply that `internal/` changes never need a changeset, while the newer Moon + Changesets model treats the server as a published release surface. I’m going to open a follow-up PR to adjust those instructions so the rule is clearer.

Edit: I created a follow up pr to clarify the logic and approach https://github.com/zitadel/nextgen/pull/361

internal/storage/database/repository/user.go
```diff
@@ -192,10 +233,23 @@ func (r *UserRepository) Create(ctx context.Context, client database.QueryExecut
...
 	)
-	return err
+	if err != nil {
+		return err
+	}
+	if user.InitialMembershipTeamID == nil || *user.InitialMembershipTeamID == "" {
+		return nil
+	}
+	membershipRepo := NewTeamMembershipRepository(client)
+	return membershipRepo.Create(ctx, client, &domain.TeamMembership{
```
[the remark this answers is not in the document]
adlerhurst: out of scope

internal/storage/database/repository/user.go
adlerhurst: changed that both runs in one transaction.

internal/storage/database/repository/user.go
```diff
@@ -119,7 +152,31 @@ func WithUserScalarList(project string, filters []UserListScalarFilter, teamScop
 }
 
 func (r *UserRepository) Delete(ctx context.Context, client database.QueryExecutor, condition database.Condition) error {
-	_, err := deleteOne(ctx, client, r, condition)
+	if beginner, ok := condition.(database.Beginner); ok {
```
fforootd: Nice improvement moving the PK guard before membership cleanup. I think the transaction wrapper still does not run, though: `Beginner` is implemented by the pool/tx, not by `database.Condition`, so this assertion should be against `client`. One extra wrinkle: after switching that, the function also needs to carry the later error into `tx.End` (for example with a named `err` return), otherwise the deferred closure can commit even when the membership delete or `deleteOne` fails. Could we tighten this so the membership cleanup and user delete are actually all-or-nothing?

internal/storage/database/repository/user.go
```diff
@@ -192,10 +249,23 @@ func (r *UserRepository) Create(ctx context.Context, client database.QueryExecut
...
 	)
-	return err
+	if err != nil {
+		return err
+	}
+	if user.InitialMembershipTeamID == nil || *user.InitialMembershipTeamID == "" {
+		return nil
+	}
+	membershipRepo := NewTeamMembershipRepository(client)
+	return membershipRepo.Create(ctx, client, &domain.TeamMembership{
```
fforootd: Could we keep this create path atomic too? The user/EAV rows are inserted first, and the initial membership is inserted in a second statement. If the membership insert fails, for example because the team id is stale, we can leave a persisted user plus scoped attributes but no membership; retries then look like duplicate-user/unique-attribute failures instead of a clean failed create. Wrapping both writes in a transaction when `client` supports `database.Beginner`, or folding the membership into the create SQL like the example CTE does, would make this all-or-nothing.

internal/storage/database/repository/team.go
```diff
@@ -97,7 +110,56 @@ func (r *TeamRepository) Get(ctx context.Context, client database.QueryExecutor,
 	return &domain.Team{
 		ProjectID: row.ProjectID,
 		ID:        row.ID,
+		Status:    domain.TeamStatus(row.Status),
 		CreatedAt: row.CreatedAt,
 		UpdatedAt: row.UpdatedAt,
 	}, nil
 }
+
+func (r *TeamRepository) Deactivate(ctx context.Context, client database.QueryExecutor, projectID, id string) error {
```
fforootd, line 119: The lifecycle policy here spans three writes: deactivate the team, remove memberships, then deactivate team-owned users. The tests pass a transaction, but the repository method also accepts a pool, and a pool-based caller could persist only the earlier steps if a later update fails. Since this method is the policy application point, could it open a transaction when `client` implements `database.Beginner` so the policy is applied atomically regardless of caller?

fforootd (commented):
Thanks for the update here. I left a few focused comments around making the new multi-statement lifecycle paths atomic; the ADR shape looks good, this is mostly about avoiding partial writes when a later statement fails.

internal/storage/database/repository/user.go
```diff
@@ -224,10 +292,27 @@ func userHydrationExpressions(rowQualifier, attrKeysPlaceholder, authPlaceholder
     CASE WHEN ` + authPlaceholder + ` THEN EXISTS(SELECT 1 FROM zitadel_nextgen.user_passkeys p WHERE p.project_id = ` + rowQualifier + `.project_id AND p.user_id = ` + rowQualifier + `.id) ELSE FALSE END AS has_pk`
 }
 
-func (r *UserRepository) GetByID(ctx context.Context, client database.QueryExecutor, projectID string, teamID *string, userID string) (*domain.User, error) {
+func (r *UserRepository) Deactivate(ctx context.Context, client database.QueryExecutor, projectID, userID string) error {
```
fforootd, line 299: This looks like the one multi-statement path that missed the `withTransaction` sweep: the status update and the membership sweep still run directly on `client`. A pool caller that fails the second statement leaves a deactivated user with active roster rows — the same partial-write shape we just closed in `Create`, `Delete`, and `TeamRepository.Deactivate`. Could we wrap this one too?

internal/storage/database/repository/team.go
```diff
@@ -97,7 +110,57 @@ func (r *TeamRepository) Get(ctx context.Context, client database.QueryExecutor,
...
+		mb.WriteArg(projectID)
+		mb.WriteString(" AND team_id = ")
+		mb.WriteArg(id)
+		mb.WriteString(" AND status <> ")
+		mb.WriteArg(membershipRemoved)
+		if _, err := tx.Exec(ctx, mb.String(), mb.Args()...); err != nil {
+			return err
+		}
+
+		ub := database.NewStatementBuilder("UPDATE ")
```
fforootd, line 153: Policy question on the users we deactivate here: we remove this team's memberships, but the team-owned users we deactivate keep `active` membership rows in their *other* teams. `UserRepository.Deactivate` removes all of a user's memberships, and ADR 024 reads the same way ("deactivates that user and removes their memberships/access") — so the same event, a user becoming deactivated, currently applies different roster policy depending on the entry point. Since nothing filters reads on `users.status` yet, those users still resolve via `MembershipTeamCondition` in their other teams.

If the intent is that a service-layer flow does the cross-team sweep later, a short comment here saying so is enough; otherwise I'd fold the sweep into this same transaction.

internal/storage/database/dialect/postgres/migration/sql/000011_user_team_lifecycle.sql
```diff
@@ -0,0 +1,94 @@
...
+    ADD COLUMN status TEXT NOT NULL DEFAULT 'active'
+        CHECK (status IN ('active', 'deactivated', 'pending_purge'));
+
+ALTER TABLE zitadel_nextgen.users
+    ADD COLUMN lifecycle_owner_team_id TEXT COLLATE "C",
+    ADD COLUMN status TEXT NOT NULL DEFAULT 'active'
+        CHECK (status IN ('active', 'suspended', 'deactivated', 'pending_purge'));
+
+UPDATE zitadel_nextgen.users
+SET lifecycle_owner_team_id = team_id
```
fforootd, line 16: Just making sure this is a conscious choice: the backfill makes every existing user with a `team_id` team-owned. That faithfully preserves the old cascade semantic, but it creates an asymmetry — an existing self-serve signup user (who only had `team_id` because the old schema required a team context) now gets deactivated with their workspace team, while the identical signup made after this PR is self-owned (`NewCreateUser` maps the team to `InitialMembershipTeamID` only). There's no way to distinguish provisioning source retroactively, so I think this is the right conservative default for alpha data — a sentence here or in the PR body documenting the choice would do.

internal/storage/database/dialect/spanner/migration/sql/000011_user_team_lifecycle.sql
```diff
@@ -0,0 +1,105 @@
...
+DROP INDEX idx_users_lifecycle_owner_team_id
+-- +goose StatementEnd
+-- +goose StatementBegin
+ALTER TABLE users DROP CONSTRAINT fk_users_lifecycle_owner_team
+-- +goose StatementEnd
+-- +goose StatementBegin
+ALTER TABLE users ADD COLUMN team_id STRING(MAX)
+-- +goose StatementEnd
+-- +goose StatementBegin
+UPDATE users u
```
fforootd, line 82: Spanner DML requires a `WHERE` clause on `UPDATE`, so this down statement will fail to parse — `WHERE TRUE` suffices. Low stakes since it's the down path and the lifecycle tests all skip Spanner, but it's a one-liner while you're here.

fforootd (commented):
Re-reviewed at f58e4f9. The atomicity feedback from the last round is addressed nicely — the shared `withTransaction` on Create/Delete/Team.Deactivate plus the rollback tests is exactly what I was hoping for. I also ran the `postgres_integration` suites for the repository and API packages locally against a fresh embedded PG 18: all green, including migration 000011 from scratch.

Four focused comments inline: one real gap (`UserRepository.Deactivate` is the last non-atomic multi-statement path), the rest are judgment calls I'd like us to make consciously.

One thing to track separately rather than in this PR: reads don't filter `users.status`/`teams.status` yet, so deactivation doesn't block anything (login included). That's consistent with the storage-only scope here — let's just make sure service-layer enforcement has its own issue.

Issue #240, opened 2026-06-09
feature: flow-definition lifecycle


vitorbari:
**Context**

Flow definitions today expose CRUD only — create, list, get, patch, delete ([`api/openapi/endpoints/flow_definitions/methods.yaml`](https://github.com/zitadel/nextgen/blob/main/api/openapi/endpoints/flow_definitions/methods.yaml), [`api/openapi/endpoints/flow_definitions/by_id/methods.yaml`](https://github.com/zitadel/nextgen/blob/main/api/openapi/endpoints/flow_definitions/by_id/methods.yaml)). The `status` enum is already modelled as `draft / active / deprecated / archived` ([`internal/domain/flowdefinitionstatus_enumer.go`](https://github.com/zitadel/nextgen/blob/main/internal/domain/flowdefinitionstatus_enumer.go), [`docs/design/flowengine/flow-definition-rules.md:21`](https://github.com/zitadel/nextgen/blob/main/docs/design/flowengine/flow-definition-rules.md#L21)) and only `active` is resolvable, but no endpoints exist to drive transitions between states or to validate / simulate a definition before activating it. Goal: add the lifecycle endpoints (`/flow_definitions/{id}/activate`, `/archive`, `/validate`, `/simulate`) that the design implies, so authors can ship a definition without flipping status through a raw PATCH.

**Open design questions**

- **Lifecycle rules.** `flow-definition-rules.md` explicitly flags this as undecided ([line 102](https://github.com/zitadel/nextgen/blob/main/docs/design/flowengine/flow-definition-rules.md#L102)): which transitions are legal (e.g. `draft → active`, `active → deprecated`, `→ archived`), whether activating one version auto-deprecates others sharing `(name, purpose)`, and whether `archived` is reversible.
- **Validate vs. simulate.** `validate` runs the static validator only; `simulate` walks the graph with synthetic submissions to flag unreachable steps and dead transitions. Decide whether both ship together or `simulate` is a follow-up.
- **Schema-resolution timing.** Same doc, same line: when `user_schema` is resolved (write-time / activate-time / runtime) is tied to this lifecycle decision.

**References**

- [`api/openapi/endpoints/flow_definitions/`](https://github.com/zitadel/nextgen/tree/main/api/openapi/endpoints/flow_definitions) — current CRUD surface
[...]

PR #308, opened 2026-06-16
docs: flow definition lifecycle

grvijayan:
## Summary

Define flow definition states and lifecycle. 

## Validation
N/A

## Release notes / changeset
- No changeset required — no public npm package files changed.

Related to https://github.com/zitadel/nextgen/issues/240


docs/design/flowengine/flow-definition-lifecycle.md
```diff
@@ -0,0 +1,48 @@
...
+
+**Note:** There must be at least one flow definition in the `active` state for a given `purpose` at all times to ensure that new flows can be started.
+
+## Endpoints to Support Lifecycle
+To support the lifecycle defined above, the following endpoints can be utilized:
+- `POST /flow-definitions`: Create a new flow definition in the `draft` state by default. The payload can include a `state` field to specify the initial state to support creating a flow definition via CLI.
+- `PUT /flow-definitions/{id}`: Update a flow definition. Allowed only in the `draft` state to prevent breaking changes to active flow definitions.
+- `POST /flow-definitions/{id}/activate`: Activate a flow definition in the `draft` or `deprecated` state.
+- `POST /flow-definitions/{id}/deprecate`: Deprecate a flow definition in the `active` state.
+- `POST /flow-definitions/{id}/archive`: Archive a flow definition in the `active` or `deprecated` state.
```
vitorbari: If the state is defined as an attribute of the flow-definition, we can make state transitions go through `PUT /flow-definitions/{id}` and drop these dedicated state change endpoints.

`PUT /flow-definitions/{id}` would control which state transitions are allowed, and when definition change is allowed.

I think this might be easier for the CLI to always call the same endpoint.

cc @mridang

grvijayan: Yes, I thought we could do both to keep it flexible. For clients that want to use the API and not send the whole payload again, explicit endpoints would be useful. For the CLI, it makes sense to include the `state` field in the flow definition itself.

vitorbari (approved):

mridang (approved):

PR #336, opened 2026-06-17
fix: flip CurrentPurpose in passkey early-dispatch path

vitorbari:
## Summary
- This PR fixes the issue where a failed passkey that hit a `user_not_found` does not change the `CurrentPurpose` from `login` to `register` in the flow state.
- Adds a regression test: passkey + unknown email on the identifier step → routes to register, `CurrentPurpose` flips to register, no WebAuthn challenge issued.


IAM-marco (approved):
Can't really test. Looks good otherwise

PR #357, opened 2026-06-18
feat(api): update flow def

grvijayan:
## Summary

Adds support for updating flow definitions (including optional status transitions) across the domain/service/repository layers and exposes it via the `PUT /flow_definitions/{id}` endpoint.

Changes:

- Add repository + service support for updating an existing FlowDefinition, and add a new API handler for UpdateFlowDefinition.
- Extend domain contracts and update domain.NewFlowDefinition to accept an explicit ID + status.
- Update OpenAPI docs and regenerate API stubs
- Add/extend unit + integration tests for update behavior.
- Additional: Remove setting of created_at/updated_at in the domain layer as they're anyway set to `now()` in the repo layer 

## Validation
Ran the following commands from the root
- `go generate ./... `
- `go test -v ./...`

## Release notes / changeset
- - No changeset required — no public npm package files changed.

## Notes
- Related to https://github.com/zitadel/nextgen/issues/240


internal/service/flow_definition_test.go
```diff
@@ -143,7 +142,7 @@ func Test_flowDefinitionService_Create(t *testing.T) {
 						},
 						{
 							Name:     "step_2",
-							Complete: gu.Ptr(domain.FlowStepCompleteRedirect),
+							Complete: new(domain.FlowStepCompleteRedirect),
 						},
```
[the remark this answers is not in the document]
grvijayan: compiled fine locally

internal/service/flow_definition.go
```diff
@@ -112,6 +117,86 @@ func (fd *flowDefinitionService) Create(ctx context.Context, req CreateFlowDefin
 	return flowDefinition, nil
 }
 
+func (fd *flowDefinitionService) Update(ctx context.Context, req FlowDefinitionRequest) (*domain.FlowDefinition, error) {
+	retrievedFlowDef, err := fd.Get(ctx, req.ProjectID, req.FlowDefinitionID)
+	if err != nil {
+		return nil, err
+	}
+
+	// if the status is not set, use the existing status
```
grvijayan, line 127: @vitorbari, would like your opinion on this?
This endpoint is supposed to replace the flow definition completely, but `status` is an optional parameter.
I didn't feel like failing the request because `status` is optional. Didn't feel right to set to a default value. Please let me know if you have a preference.

internal/service/flow_definition.go
```diff
@@ -112,6 +117,110 @@ func (fd *flowDefinitionService) Create(ctx context.Context, req CreateFlowDefin
...
+	err = fd.Validate(ctx, flowDefinition)
+	if err != nil {
+		return nil, err
+	}
+	err = fd.flowDefinitionRepo.UpdateFlowDefinition(ctx, fd.db, flowDefinition)
+	if err != nil {
+		return nil, err
+	}
+	return flowDefinition, nil
+}
```
IAM-marco, line 167: Claude:

`NewFlowDefinition` always sets `CreatedAt = time.Now().UTC()`, so the object returned here carries a fresh timestamp. The DB UPDATE does not touch `created_at`, so a subsequent GET will return a different `created_at` than the PUT response body.

The fix `flowDefinition.CreatedAt = retrievedFlowDef.CreatedAt` was in commit `03b785fa` but was dropped in `7f1b9f17`. Add it back between the `NewFlowDefinition` error check and `Validate`.

Ref: https://github.com/zitadel/nextgen/blob/d5a047a12c9f75d14ce153122feb7df40b9a0967/internal/service/flow_definition.go#L142-L165

internal/domain/flow_definition_repository.go
```diff
@@ -22,6 +22,9 @@ type FlowDefinitionRepository interface {
 	// belonging to the given instance. Child records are not populated.
 	ListFlowDefinitions(ctx context.Context, client database.QueryExecutor, projectID string, opts ...FlowDefinitionListOption) ([]*FlowDefinition, error)
 
+	// UpdateFlowDefinition updates a definition and all its child records.
+	UpdateFlowDefinition(ctx context.Context, client database.QueryExecutor, def *FlowDefinition) error
+
 	// UpdateFlowDefinitionStatus transitions a definition to the given status.
 	UpdateFlowDefinitionStatus(ctx context.Context, client database.QueryExecutor, projectID, id string, status FlowDefinitionStatus) error
```
IAM-marco, line 29: Is it still relevant to have this?

internal/service/flow_definition.go
```diff
@@ -112,6 +117,110 @@ func (fd *flowDefinitionService) Create(ctx context.Context, req CreateFlowDefin
...
+				purposesToCheck[p] = struct{}{}
+			}
+		}
+	}
+
+	if len(purposesToCheck) == 0 {
+		return nil
+	}
+
+	for purpose := range purposesToCheck {
```
IAM-marco, line 202: Please a todo for when we're going to refactor ListFlowDefinitions. It should be possible to pass a list of purposes, in a `database.Or()` (for example) and just run 1 sql query.

This loop is expensive

internal/service/flow_definition.go
```diff
@@ -112,6 +117,110 @@ func (fd *flowDefinitionService) Create(ctx context.Context, req CreateFlowDefin
...
+			projectID,
+			domain.WithFlowDefinitionStatus(domain.FlowDefinitionStatusActive),
+			domain.WithFlowDefinitionPurpose(purpose),
+		)
+		if err != nil {
+			return domain.ErrInternal(err).WithMessage(fmt.Sprintf("failed to list flow definitions for old purpose %q", purpose))
+		}
+		hasActive := false
+		for _, fd := range fds {
+			if fd.ID != flowDefID {
```
IAM-marco: Shouldn't this be `fd.ID == flowDefID`?

IAM-marco: And in any case, isn't it easier to create an option `domain.WithFlowDefinitionID(flowDefID)`?

And if `flowDefID` is unique (?), can't you just Get?

internal/service/flow_definition.go
```diff
@@ -173,6 +282,18 @@ func mapPurposesToDomain(reqPurposes map[string]string) (map[domain.FlowDefiniti
 	return purposes, nil
 }
 
+func samePurposes(current, req map[domain.FlowDefinitionPurpose]string) bool {
```
IAM-marco: Not sure, as I don't entirely understand this, but can't you use `maps.Equal()`?

internal/service/flow_definition.go
grvijayan: Missed this.
We should perhaps let the repository layer handle `created_at`, `updated_at` (which it already [does](https://github.com/zitadel/nextgen/blob/main/internal/storage/database/repository/flow_definition.go#L159)), and not set it in the domain layer. What do you think?

internal/domain/flow_definition_repository.go
grvijayan: There's a discussion to introduce `/activate` and `/deactivate` endpoints. It'd be nice to have it for these endpoints. If we decide not to introduce them, we can clean this repo implementation later.

internal/service/flow_definition.go
grvijayan: Thanks for the catch.
I was thinking the same as well and later forgot. Will check what we can go about this, or add a todo.

internal/service/flow_definition.go
grvijayan: > Shouldn't this be fd.ID == flowDefID?

The check here is to see if there are flow definitions besides the one being updated to support the `purpose` being modified / being deactivated.

internal/service/flow_definition.go
grvijayan: Currently, `ListFlowDefinitions` does an `AND` on all the purposes ([implementation](https://github.com/zitadel/nextgen/blob/main/internal/storage/database/repository/flow_definition.go#L201)). 
Would be nice to have an `OR` based fetch as you suggested. I'll add a todo for now.

internal/service/flow_definition.go
grvijayan: We concluded that `status` will be a required field in flow definition. 
Will make this change in a different PR (and also update default examples, etc.,) to keep the scope of this PR small.

internal/service/flow_definition.go
grvijayan: removed from the domain layer

internal/service/flow_definition.go
```diff
@@ -112,6 +118,105 @@ func (fd *flowDefinitionService) Create(ctx context.Context, req CreateFlowDefin
...
+	if err != nil {
+		return nil, err
+	}
+	return flowDefinition, nil
+}
+
+func (fd *flowDefinitionService) isUpdateAllowed(
+	ctx context.Context,
+	projectID,
+	flowDefID string,
```
IAM-marco, line 172: This param is unused

IAM-marco (approved):

PR #363, opened 2026-06-19
fix: re-identify user on every dispatch instead of trusting stored _user_id

vitorbari:
## Issue

Two existing users.
1. User 1 (password only) types their email, clicks "Login with passkey" → server rejects (no credential matches).
2. User 2 (passkey only) then types their email and clicks the same action → also fails. The only way to recover is to refresh the page.

## Fix

The flow dispatch loop was skipping `SubmitIdentifier` whenever a previous `user_id` was already stored, so user 2's email was never looked up and the new ceremony stayed scoped to user 1.


vitorbari:
Reopened as #365 with a `patch/` prefixed branch.

PR #365, opened 2026-06-19
fix: re-identify user on every dispatch instead of trusting stored _user_id

vitorbari:
## Issue

Two existing users.
1. User 1 (password only) types their email, clicks "Login with passkey" → server rejects (no credential matches).
2. User 2 (passkey only) then types their email and clicks the same action → also fails. The only way to recover is to refresh the page.

## Fix

The flow dispatch loop was skipping `SubmitIdentifier` whenever a previous `user_id` was already stored, so user 2's email was never looked up and the new ceremony stayed scoped to user 1.

### DIFF
--- a/internal/domain/flow_state_machine.go
+++ b/internal/domain/flow_state_machine.go
@@ -483,9 +483,6 @@ func (r *FlowStateMachineRuntime) dispatchChallenges(ctx context.Context, def *F
 		}
 		switch challenge {
 		case FlowFieldChallengeIdentifier:
-			if _, pinned := state.CollectedData[FlowCollectedUserIDKey]; pinned {
-				continue
-			}
 			userID, err := r.authAttempts.SubmitIdentifier(ctx, FlowSubmitIdentifierInput{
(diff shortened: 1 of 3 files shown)


internal/domain/flow_state_machine.go
```diff
@@ -483,9 +483,10 @@ func (r *FlowStateMachineRuntime) dispatchChallenges(ctx context.Context, def *F
 		}
 		switch challenge {
 		case FlowFieldChallengeIdentifier:
-			if _, pinned := state.CollectedData[FlowCollectedUserIDKey]; pinned {
-				continue
-			}
+			// Always look the user up. If the result differs from the
+			// previously resolved id, drop any ceremony that was scoped
+			// to the old user.
+			prevUserID, _ := state.CollectedData[FlowCollectedUserIDKey].(string)
```
IAM-marco: Shouldn't you be checking if the type assertion was successful?

IAM-marco (changes requested):

IAM-marco (approved):

PR #369, opened 2026-06-19
docs: back-navigation ADR changes

### LINKED ISSUE #238: feature: back-navigation (ADR 022)
Implement the back-navigation system in the flow-engine as proposed by https://github.com/zitadel/nextgen/pull/124

vitorbari:
## Summary

Restructures ADR 022 (Flow Back-Navigation). Docs-only — no code or schema changes; the listed contract additions land in a follow-up implementation PR.

- **Engine injects `back`** from runtime state. Flow definitions no longer declare per-step `back` transitions or reversibility metadata.
- **Two-rule injection contract**: (1) `history` non-empty, (2) not terminal. Reversibility is folded into rule 1 — the engine pushes onto `history` only on reversible transitions and clears it on irreversible ones. Action semantics (which only the engine knows) classify reversibility.
- **Identify by `kind`, not name**: the ADR specifies a new `kind: "back"` value on `step-action.yaml` so clients match on kind rather than the literal action name. The schema change itself is **not** in this PR — it will land with the engine implementation.
- **Browser History API lifecycle**: `pushState` iff a `kind: "back"` action is present and the step name changed; otherwise no-op. Single signal keeps the browser and engine in lockstep.
- **Heading fix**: `# ADR 016` → `# ADR 022`.

## Test plan

Docs-only — no code paths or schemas touched.

- [ ] Engine contract reads cleanly (no reversibility leaks to flow authors).
- [ ] Browser lifecycle and engine rules align on `kind: "back"` presence.
- [ ] "Mock server" and "Consequences" sections agree with the new model.
- [ ] Heading matches filename.

### DIFF
--- a/api/openapi/components/flows/step-action.yaml
+++ b/api/openapi/components/flows/step-action.yaml
@@ -11,7 +11,7 @@ properties:
     example: submit
   kind:
     type: string
-    enum: [submit, passkey, passkey_register, navigate]
+    enum: [submit, passkey, passkey_register, navigate, back]
     example: submit
     description: |
       Classifies how the engine handles this action:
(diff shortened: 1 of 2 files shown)


peintnermax (approved):
LGTM 👍

Issue #243, opened 2026-06-09
feat(flow-engine): implement ADR 020 — credentials out of the user schema


vitorbari:
**Context**

The user schema currently doubles as the user-attribute contract and the credential dictionary: a property with `x-password: true` plus `x-auth-methods.password.enabled = true` becomes a password challenge ([`internal/domain/flow_field_resolver_schema.go:104,122,159`](https://github.com/zitadel/nextgen/blob/main/internal/domain/flow_field_resolver_schema.go#L104)). This blocks the schema from describing the `GET /users/{id}` response shape without leaking credentials, and credential kinds beyond password (passkey, OTP, magic link, SSO) don't fit JSON-Schema vocabulary anyway.

[ADR 020](https://github.com/zitadel/nextgen/blob/main/docs/adrs/020-credentials-out-of-user-schema.md) decides the schema describes attributes only. Flow steps reference credentials through one of two mechanisms, both resolving back to schema-root `x-auth-methods.<kind>`: reserved token `x-auth-methods-<kind>` inside `step.fields[]` (field-shaped), or `auth_method: "<kind>"` on the action descriptor (action-shaped). Per the ADR appendix, `FlowField.Validation` becomes `nil` for credential tokens, and `FlowProgress` gains a `CollectedCredentials` bucket so writers like `FlowCreateUserHandler` stop reading credential proofs out of `CollectedData` under a schema property name. Composes with ADR 017 — 020 is *how* a step references a credential; 017 is *whether* it's verified or established.

**Open design questions**

- **Token spelling.** ADR keeps `x-auth-methods-password` but flags it for revisit (reuses the JSON-Schema `x-` convention outside its native context). Alternatives: `auth-method:password`, `credential:password`, bare reserved names. Decide before touching meta-schema prose — the spelling shows up in every flow definition.
- **`sso_providers[]` overlap.** SSO buttons live on `step.sso_providers[]` today, not under `actions{}`. Folding them into `auth_method: "sso"` removes a second mechanism but breaks the existing contract. Tracked, not blocking.
- **Credential policy home.** Removing `x-password` removes the only customer-facing place for password rules. Likely belongs on project/team config; follow-up ADR. This issue accepts `Validation: nil` in the interim.

**References**

- [`docs/adrs/020-credentials-out-of-user-schema.md`](https://github.com/zitadel/nextgen/blob/main/docs/adrs/020-credentials-out-of-user-schema.md) — decision, before/after, internal-impact appendix
[...]

PR #371, opened 2026-06-19
refactor: use userservice in flow statemachine

wim07101993:
## Summary

- User/Password creation is now handled by the `UserService`
- Some test cleanup using generated mocks instead of fakes
- Expanded errors for spanner
- Made redirect URI a URI instead of string

## Validation

- `go test ./...`

## Release notes / changeset

- No changeset required

## Notes

- because of circular imports, the statemachine had to move to the service layer.


internal/api/flow.go
```diff
@@ -63,8 +63,7 @@ func (h *Handler) CreateFlow(ctx context.Context, req *api.CreateFlowRequest) (a
 	}
 	if uri, ok := req.RedirectURI.Get(); ok {
 		// TODO: validate against RP's redirect_uri allowlist when OIDC lands.
-		s := uri.String()
-		startReq.RedirectURI = &s
+		startReq.RedirectURI = new(uri)
 	}
```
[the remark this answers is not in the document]
wim07101993: not true

internal/api/error_handler.go
```diff
@@ -16,13 +17,38 @@ import (
...
+		if j, err := json.Marshal(domErr.Details); err == nil {
+			details["details"] = j
+		}
+	}
+	if domErr.Parent != nil {
+		strParent := domErr.Parent.Error()
+		if j, err := json.Marshal(strParent); err == nil {
+			details["parent"] = j
+		}
+	}
```
vitorbari, line 35: Can this leak some internal/sensitive info?

internal/domain/user.go
```diff
@@ -23,6 +23,11 @@ const (
 	PrefixUser ResourcePrefix = "user"
 )
 
+const (
+	UserIDFieldName     = "id"
```
vitorbari, line 27: Does user-schema influence on this? Or is this the id we generate internally? Should we call it something else if the user wants to add their own `id` as one of the user properties?

internal/service/user.go
```diff
@@ -261,3 +263,43 @@ func (o *SetPasswordUserAction) Apply(ctx context.Context, db database.QueryExec
...
+type LazyUserAction struct {
+	factory UserActionFactory
+	action  UserAction
+}
+
+func NewLazyUserAction(factory UserActionFactory) *LazyUserAction {
+	return &LazyUserAction{
+		factory: factory,
+	}
+}
```
vitorbari, line 280: Please add an explanation when lazy action should be used, how it works, etc.

internal/api/error_handler.go
wim07101993: Good point. I was assuming errors should not contain sensitive data. But that is not enforced I guess. I thought about internal info, but I don't know whether it matters whether internal stuff is exposed. It just makes stuff easier to debut. Don't see any downsides. If you want, I can take it out.

internal/domain/user.go
wim07101993: This only abstracts the value in a const because it was used at multiple locations. It seemed brittle, so I centralized the value. I thought of changing the value but that broke stuff, so I reverted it.

internal/service/flow_on_success_create_user.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+}
+
+var _ domain.FlowOnSuccessHandler = (*FlowCreateUserHandler)(nil)
+
+func (h *FlowCreateUserHandler) Handle(ctx context.Context, in domain.FlowOnSuccessInput) (domain.FlowOnSuccessResult, error) {
+	createUserAction := NewCreateUserAction(
+		CreateUserInput{
+			ProjectID: in.ProjectID,
+			User:      collectedDataToUserData(in.State.CollectedData, in.State.UserSchemaURL),
```
vitorbari, line 45: This will return all collected data, including passwords!!
Options I see: 
1. Don't add credentials to `State.CollectedData` (might be safer long-term)
2. Skip credentials when collecting user attributes (simpler in-place logic) - reference: https://github.com/zitadel/nextgen/blob/fb075538cad22d9b9a67729e1fa02f394964e9e1/internal/domain/flow_on_success_create_user.go#L61

vitorbari (changes requested):

internal/api/error_handler.go
vitorbari: We can review this later as part of the observability improvements

internal/service/flow_on_success_create_user.go
wim07101993: Maybe the `CollectedData` is currently to broad. It contains a bunch of stuff which does not actually need to be stored with the user. Maybe we should refine that structure to contain the data of the user which is defined in the schema, and thus needs to be stored in the db users table, and other fields like the password which need to be stored separately or is used as working data like `"_user_id"` and `"_passkey_provisional"`

vitorbari: this will naturally happen when we implement https://github.com/zitadel/nextgen/issues/243 - credentials won't be part of the attributes

For now, we can have some simpler logic that skips credentials/password. wdyt?

vitorbari: I see what you mean, to split user-schema data from the "flow" related data. Currently, everything gets stored together.

wim07101993: There is possibly no way around having the password in the database since we require it by schema. To create a user, we validate the user by schema so either we remove the password from the schema, or it has to be in there

wim07101993: we need to solve https://github.com/zitadel/nextgen/issues/243 first

wim07101993:
Closing due to merge conflicts. See #415 for new pr

PR #376, opened 2026-06-22
fix: proxy API requests to Go backend in dev mode

IAM-marco:
## Summary

- Keep login-ui Vite env knobs development-only: `VITE_PROJECT_ID`, `VITE_PROXY_PATH`, and `VITE_BACKEND_URL` are only used by the local dev server path.
- Use a dev-only Vite proxy with default path `/__nextgen` and default backend `http://localhost:8080`; `vite preview` keeps the embedded `/ui/login/` base and does not apply the dev proxy.
- Set the embedded hosted-login shell to `proxy-path="/"` in production so `<zitadel-login>` calls the same-origin Flow API at `/flow` instead of the framework proxy default.
- Update first-run docs to use the API-only backend, `.env.development.local`, default project provisioning, and a curl example that does not require `jq`.

## Validation

- `git diff --check`
- `moon run login-ui:typecheck`
- `moon run login-ui:build` with a temporary `apps/login-ui/.env.local` containing fake `VITE_PROJECT_ID`, `VITE_PROXY_PATH`, and `VITE_BACKEND_URL`; verified `internal/staticui/login/dist` did not contain the fake values.
- `moon run login-ui:preview`; verified `http://localhost:4173/ui/login/` emits `/ui/login/assets/...` URLs and both JS/CSS assets return `200`.
- `NEXTGEN_SERVER_CONSOLE_ENABLED=false NEXTGEN_SERVER_LOGIN_ENABLED=false go run . server`; created a project with `POST /projects`, started `moon run login-ui:dev` with a temporary `apps/login-ui/.env.development.local`, and verified `POST /__nextgen/flow` returned a flow with the `identifier` step.
- `corepack pnpm exec changeset status --since origin/main` after staging the changeset.

## Release notes / changeset

Changeset required: `.changeset/quiet-buckets-compete.md` patches `@zitadel/server` because the embedded hosted-login shell now sets the production API base to same-origin `/flow`.

## Notes

This does not implement the final hosted-login/OIDC request contract; the local `project_id` URL parameter remains a temporary bootstrap/debug path.


apps/login-ui/vite.config.mts
```diff
@@ -11,6 +11,13 @@ export default defineConfig(({ command }) => ({
   server: {
     port: 5175,
     strictPort: true,
+    proxy: {
+      "/__nextgen": {
+        target: process.env.VITE_BACKEND_URL ?? "http://localhost:8080",
```
fforootd: `VITE_BACKEND_URL` is documented as an override to put in `apps/login-ui/.env.local`, but Vite config files do not automatically load `.env*` values into `process.env`. With this code, a contributor who follows the new docs still proxies to `http://localhost:8080` unless they exported the variable in the shell before starting Vite. Please load it in the config with `loadEnv(...)` or change the docs/example to make this a shell-only override.

fforootd (commented):
Left one inline comment on the dev proxy env handling.

PR #378, opened 2026-06-22
feat: split step-action schema for back-navigation contract

vitorbari:
## Summary

Splits the shared `step-action.yaml` into two schemas so the wire contract enforces which kinds are valid where:

- `flow-definition-action.yaml` — authored actions. Kinds: `submit`, `passkey`, `passkey_register`, `navigate`.
- `flow-step-action.yaml` — rendered response actions. Same plus `back`.

Back-navigation (ADR 022) introduces a `back` kind the engine injects but authors must never declare. Splitting the schema enforces that at the contract layer.

Contract-only — no engine behavior changes. Regenerated bindings split `api.StepAction` into `api.FlowDefinitionAction` and `api.FlowStepAction`; mappers and integration-test fixtures updated.

## Test plan

- [x] `go build ./...` / `go test ./...`
- [x] `go test -tags postgres_integration ./internal/api/integration_test/...` (one pre-existing flake on `main`)
- [ ] CI


vitorbari:
Closing — going with a single `step-action.yaml` and rejecting `kind: back` on inbound definitions via Go validation instead. The schema split duplicated too much.

PR #388, opened 2026-06-23
chore(api): make status required in flow definition

grvijayan:
## Summary

This PR includes changes to make `status` a required field in `flow_definition` in the API spec used when creating/updating a flow definition to simplify DevX. 

Changes:
* Made `status` required in the API spec
* Removed the default behaviour in api/service-layer implementation when status is not set
* Updated default and example flow definitions 
* Update relevant unit/integration tests

## Validation
- `go test ./...`

## Release notes / changeset

- No changeset required — no public npm package files changed.

## Notes

- Follow-up for https://github.com/zitadel/nextgen/pull/357
- Related to https://github.com/zitadel/nextgen/issues/240


api/openapi/components/flows/flow-definition-detail-response.yaml
```diff
@@ -1,5 +1,5 @@
 type: object
-required: [id, project_id, status, flow_definition, created_at, updated_at]
+required: [id, project_id, flow_definition, created_at, updated_at]
```
grvijayan, line 2: @mridang this change might be relevant to you - `status` is removed from the create/update/get flow definition response. The `status` should now be available within the flow definition itself.

internal/service/flow_definition.go
```diff
@@ -168,8 +166,7 @@ func (fd *flowDefinitionService) Update(ctx context.Context, req FlowDefinitionR
 
 func (fd *flowDefinitionService) isUpdateAllowed(
 	ctx context.Context,
-	projectID,
-	flowDefID string,
+	projectID string,
 	currentStatus, reqStatus domain.FlowDefinitionStatus,
 	currentPurposes, reqPurposes map[domain.FlowDefinitionPurpose]string) error {
```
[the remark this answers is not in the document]
grvijayan: fixed in https://github.com/zitadel/nextgen/pull/388/commits/6128a1541438ff719702669f42e5ce50aebe7a64

api/openapi/components/flows/flow-definition-detail-response.yaml
grvijayan: I use Copilot to fix it in the cli: https://github.com/zitadel/nextgen/pull/388/changes/1b98b9b2909e9cc19cefd508564d386a27579844
Please review.

api/openapi/components/flows/flow-definition-detail-response.yaml
```diff
@@ -1,5 +1,5 @@
...
 properties:
```
IAM-marco, line 3: This creates an asymmetry with the list response: `GET /flow_definitions` returns `status` as a top-level field (see [flow-definition-response.yaml](https://github.com/zitadel/nextgen/blob/1b98b9b2909e9cc19cefd508564d386a27579844/api/openapi/components/flows/flow-definition-response.yaml#L30-L35)), but after this PR `GET /flow_definitions/{id}` only exposes it inside `flow_definition.status`. Clients reading both endpoints need different paths to access the same field. If intentional, document as a breaking change.

api/openapi/components/flows/flow-definition.yaml
```diff
@@ -1,5 +1,5 @@
 type: object
-required: [ name, user_schema, purposes, steps ]
+required: [ name, status, user_schema, purposes, steps ]
 description: |
```
IAM-marco, line 3: The companion schema `api/openapi/components/flows/flow-definition-update-request.yaml` (not changed in this PR) still says on line 6: "If `flow_definition.status` is omitted, the current status is preserved." That description directly contradicts `status` being required here and is now misleading to API consumers. It should be updated to reflect that `status` is now mandatory.

api/openapi/components/flows/flow-definition-status.yaml
```diff
@@ -1,6 +1,5 @@
 type: string
 enum: [ active, draft ]
```
IAM-marco, line 2: Shouldn't these states match the ones we have in https://github.com/zitadel/nextgen/blob/1b98b9b2909e9cc19cefd508564d386a27579844/internal/domain/flow_definition.go#L15 ?

internal/service/flow_definition_test.go
```diff
@@ -121,6 +121,7 @@ func Test_flowDefinitionService_Create(t *testing.T) {
 				req: service.FlowDefinitionRequest{
 					ProjectID:     "project1",
 					Name:          "login",
+					Status:        "Active",
```
IAM-marco, line 124: Is it supposed to be capitalized?

api/openapi/components/flows/flow-definition-detail-response.yaml
grvijayan: The list response doesn't return a detailed response, i.e., it does not include the flow_definition in the response, so, it's not possible to expose it inside `flow_definition.status` like in `GET` 

> If intentional, document as a breaking change

Removing this field is a breaking change - I've updated the client (only CLI at the moment) accordingly. We're allowed to do breaking changes at this point, right?

api/openapi/components/flows/flow-definition-status.yaml
grvijayan: Had a conversation about this with @vitorbari, we decided to keep it simple for now and maintain only two states: `active` and `draft`. As we are anyway allowing editing of the flow definitions in `active` state, the other states (`deprecated` and `archived`) are irrelevant.

internal/service/flow_definition_test.go
grvijayan: not necessarily - I added it capitalized here to show that this is accepted as well. 
`active`, `ACTIVE` are also accepted because of the lower-case conversion in the enumer
`internal/domain/flowdefinitionstatus_enumer.go:61`

api/openapi/components/flows/flow-definition.yaml
grvijayan: fixed this

api/openapi/components/flows/flow-definition-status.yaml
IAM-marco: But should we remove them from the domain side? I put them, but not against removing them if not needed

api/openapi/components/flows/flow-definition-detail-response.yaml
IAM-marco: > We're allowed to do breaking changes at this point, right?

I would say so, sorry this is a comment from Claude. I missed that

api/openapi/components/flows/flow-definition-status.yaml
grvijayan: > But should we remove them from the domain side? 

I am not sure either. We can leave it as is for now, and remove once we're sure. What do you think?

api/openapi/components/flows/flow-definition-detail-response.yaml
grvijayan: No worries. It nice to double-check anyway.

api/openapi/components/flows/flow-definition-status.yaml
IAM-marco: Fine

IAM-marco (approved):

PR #390, opened 2026-06-23
feat: add checkbox and select atoms and a unified Storybook workbench

bastionstack:
## Summary

- Add form-associated **Checkbox** and **Select** atoms in both renderers: `<zl-checkbox>`/`<Checkbox>` and `<zl-select>`/`<Select>` (select-only WAI-ARIA combobox with keyboard nav). Both share their `.zr-*` surface CSS in `@zitadel/shared-component-styles`; new `chevron-down` icon.
- Render the `select` and `checkbox` **field types in the orchestrator**: `default.liquid` now emits `<zl-select>`/`<zl-checkbox>`, select options come from a field's `validation.enum` via a new `selectOptions` Liquid filter, and `<zl-select>` accepts options as a JS array or a JSON `options` attribute.
- Replace the Lit and console dev playgrounds with a single **Storybook workbench** (`apps/storybook`) that hosts both renderers + the orchestrator (MSW), and gates render + a11y + a computed-style **cross-renderer parity** check in CI via `@storybook/addon-vitest`.

## Validation

Run locally (all green):

- `moon run components:test` — 177/177
- `moon run api-mock:test` — 36/36
- `moon run components:lint ui-react:lint shared-component-styles:lint api-mock:lint` — 0 errors
- `moon run components:typecheck ui-react:typecheck shared-component-styles:typecheck storybook:typecheck` — clean
- Manually verified the Select/Checkbox stories and the SignUp orchestrator flow in a real browser (Lit ↔ React parity, options + form participation).

Not run locally: `storybook:test` browser suite and the opt-in demo/journey e2e (left to CI).

## Release notes / changeset

Added `.changeset/checkbox-select-atoms.md` — **minor** bump for `@zitadel/components` and `@zitadel/ui-react` (new public atoms + orchestrator field rendering). Per the changeset decision table, new public component surface warrants a minor.

## Notes

- This changes the **shared** `default.liquid`, so any flow (including the demo apps) that returns `select`/`checkbox` field types now renders these atoms — not Storybook-only.
- The new orchestrator fields on the register step are intentionally **optional** so the registration e2e/journey flows still submit without touching them.
- Storybook is `private` (no `dist`, `build` is `runInCI: false`), but `storybook:test` runs in CI.

PR #404, opened 2026-06-24
feat: support boolean and enum user-schema fields in the login flow

vitorbari:
Make `boolean` (`checkbox`) and `enum` (`select`) user-schema properties work end-to-end — render → submit → validate. Previously they always failed `create_user` with `user.invalid`.

**Frontend**
- Checkbox submits a real JSON `true`/`false` (not a string); select submits a real `enum` member, with optional selects clearing to omitted instead of `""`.
- `<zl-select>` is agent- and screen-reader-accessible via a native `<select>`.
- Required fields are enforced client-side, and errors render inline under each control (select/checkbox included), styled like server errors.
- The mandatory-gates safety net no longer duplicates required select/checkbox fields.

**Flow engine**
- Validates typed values at the step: booleans for checkboxes, `const` (e.g. a must-accept `const: true`), and required fields the client omits — on both submit and passkey-register — so errors surface per-field instead of failing late at `create_user`.


internal/domain/flow_state_machine.go
```diff
@@ -403,18 +403,26 @@ func (r *FlowStateMachineRuntime) resolveInputs(pc *processCtx) (FlowResolvedFie
 // them into CollectedData. Returns a rendered halt step on validation
 // failure.
 //
-// TODO: this rejects passkey submissions on steps that declare required
-// fields (e.g. "sign in with passkey" on the password step) because
-// in.Fields is empty. Validation should scope to fields the client
-// actually submitted, or to what the current action kind needs.
-func (r *FlowStateMachineRuntime) validateAndMerge(pc *processCtx, resolved FlowResolvedFields) (*FlowStepResult, error) {
+// Every action validates the values it sent; field-collecting actions
```
fforootd, line 406: This scoping fixes the *omission* half nicely (and the passkey-register test pinning "no challenge minted" is 👌), but I don't think we can retire the old TODO just yet: `Validate` still flags a *submitted* `""` required field on every action kind, and our web client submits all declared step fields with `""` defaults on every action (secondary actions bypass the client-side gate by design).

I reproduced with the default login flow's `password` step shape — `fields: ["x-auth-methods#password"]` (which resolves `Required: true`) plus a secondary `passkey` action:

```go
result, _ := w.sm.Process(t.Context(), nil, def, start.State, domain.FlowSubmitInput{
	Action:    domain.FlowActionPasskey,
	Fields:    map[string]any{"x-auth-methods#password": ""},
	PasskeyRP: &domain.FlowPasskeyRP{RPID: "example.com", Origins: []string{"https://example.com"}},
})
// halts with error.x-auth-methods#password_required — no challenge issued
```

Same result on main, so it's pre-existing rather than a regression — but with the TODO gone we'd lose the only documentation of it. The clean fix also isn't trivial: on the *identifier* step, an empty email on the passkey action is arguably a correct rejection (the issue leg needs it for `SubmitIdentifier`), so the old TODO's "scope to what the action kind needs" is the real shape of it.

Could we keep a narrowed TODO here (or spin it into an issue), and add the password-step companion to your passkey-register test when we fix it?

packages/components/src/orchestrator/zitadel-login.ts
```diff
@@ -590,52 +606,106 @@ export class ZitadelLogin extends LitElement {
...
-        next[name] = value;
-        changed = true;
-      }
+  /**
+   * Names of the current step's `required` fields whose captured value is
+   * empty. Reads each atom's live `formValue` (the getter reflects the native
+   * control, so autofill that skipped `input` events is still seen), so this
+   * is browser-independent and does not rely on native constraint validation.
+   */
+  private missingRequiredFields(): string[] {
```
fforootd, line 633: An unchecked checkbox's `formValue` is `""`, so a schema-`required` boolean gets blocked here with `error.<field>_required` — but the server accepts `false`: this client always submits checkboxes as a real bool, so `MissingRequired` never fires, and `Validate` is happy (JSON-Schema `required` = property present). Net effect: a browser user must tick a required checkbox, while an agent driving the API can register with `false`.

Since `const: true` is the sanctioned must-accept mechanism (per your PR notes — which I like), I'd exclude `type === "checkbox"` from this gate (a rendered checkbox always has a value — `false`) and drop the `{% if f.required %}required{% endif %}` render in the checkbox branch of `default.liquid`. Alternatively, if must-tick for required booleans *is* the intended semantics, it should be enforced server-side too so API clients match. WDYT?

packages/components/src/atoms/zl-button.ts
```diff
@@ -203,11 +203,19 @@ export class ZlButton extends LitElement {
     }
     const form = this.internals.form;
     if (type === "submit" && form) {
-      // Mirror native <button type="submit"> behaviour through the
-      // shadow boundary so the host <form>'s validation runs.
+      // Mirror native <button type="submit"> through the shadow boundary:
+      // delegate to the host <form> so its constraint validation runs and an
```
fforootd: Tiny wording nit: inside `<zitadel-login>` the form is `novalidate`, so constraint validation doesn't actually run there — the orchestrator's `missingRequiredFields` gate is the enforcement. The comment is true for generic host forms; maybe add a clause so future readers don't go looking for native validation in the login flow.

packages/components/src/atoms/zl-select.ts
```diff
@@ -218,186 +274,83 @@ export class ZlSelect extends LitElement {
...
   private handleTriggerClick = (): void => {
     if (this.disabled) {
       return;
     }
-    if (this.open) {
-      this.close();
-    } else {
-      this.openMenu();
-    }
+    this.open = !this.open;
```
fforootd, line 308: Two small pointer-popup observations (apply to both the Lit and React implementations):

- The old combobox closed on `Escape`; the styled popup now only closes on outside-pointerdown or selection. Worth wiring `Escape` back up for mouse users.
- `pointer-events: none` on the native `<select>` means touch users get the custom popup instead of the native mobile picker. Fine if deliberate (Figma parity) — just flagging the trade-off.

pnpm-workspace.yaml
```diff
@@ -97,7 +97,7 @@ catalog:
   vite: ^8.0.14
   vite-plugin-dts: ^4.5.0
   vite-plugin-web-components-hmr: ^0.1.3
-  vitest: ^4.1.7
+  vitest: ^4.1.9
```
fforootd, line 100: Curious: was `^4.1.7 → ^4.1.9` needed to get browser mode stable in CI? A line in the PR description would help future archaeology, since it churns the whole lockfile.

fforootd (commented):
Really nice work — this makes checkbox/select feel properly first-class end-to-end. Things I especially like: the client-side required gate reusing the server's own `error.<field>_required` dialect (so localisation and inline/banner routing stay single-pathed), the agent-first native `<select>` rewrite with the AGENTS.md write-up, and that the Go validator got the subtle bits right (`""` short-circuits before the `const` check, type-mismatched `const` defers to `create_user`).

I did some digging before commenting: ran a state-machine test shaped like the default login flow's `password` step against both this branch and main (details inline on `validateAndMerge`), checked that `fields` is `additionalProperties: true` on the wire so the boolean payloads are spec-legal, that the new `error="{{ field_err }}"` bindings are safe (`outputEscape: "escape"` + DOMPurify), and that the full `internal/domain` suite passes on this branch.

Two substantive comments inline (the TODO removal and required-checkbox semantics) plus a few nits — none of them need to grow the PR's scope; a breadcrumb and a one-function tweak would do.

packages/components/src/orchestrator/zitadel-login.ts
bastionstack: Done in a7534e53: checkboxes are now excluded from the client-side required gate (a rendered checkbox always submits real `false`), and the `required` render is dropped from the checkbox branch. Must-accept stays a schema concern via `const: true`, so browser and API clients match. Added a browser regression test.

packages/components/src/atoms/zl-button.ts
bastionstack: Done in a7534e53: reworded to note the orchestrator form is `novalidate`, so `missingRequiredFields` is the enforcement inside `<zitadel-login>` (native constraint validation only applies on a generic host form).

packages/components/src/atoms/zl-select.ts
bastionstack: Escape is now wired up in a7534e53 for both Lit and React (with a unit test). Leaving `pointer-events: none` on the native `<select>` as-is: deliberate for Figma parity so mouse users get the styled popup, while keyboard/AT/automation drive the native control. Touch getting the custom popup is the accepted trade-off there.

pnpm-workspace.yaml
bastionstack: It keeps `vitest` in lockstep with `@vitest/browser` / `@vitest/browser-playwright` (both `^4.1.9`), which the real-Chromium browser project needs at a matching core version — it landed with #390 (browser test project + Storybook workbench). Ill add a line to the PR description.

fforootd (approved):

PR #339, opened 2026-06-17
fix: optimize passkey registration defaults and user labeling

peintnermax:
## Summary

Optimizes the WebAuthn `PublicKeyCredentialCreationOptions` sent to the browser during passkey registration. Before this change, passkeys were created with browser defaults (non-discoverable) and labeled with the internal UserID — both hurting discoverability and password manager UX.

### Changes

**1. Require discoverable credentials via `authenticatorSelection`**

`BuildPasskeyCreationOptions` previously omitted `authenticatorSelection`, causing browsers to default to `residentKey: "discouraged"`. Passkeys created this way are not discoverable — breaking autofill, conditional UI, and passwordless login. Now explicitly sets `residentKey: "required"` and `userVerification: "preferred"`.

**2. Derive human-readable `user.name` from collected flow data**

The WebAuthn `user.name` field is shown in the browser's passkey dialog and stored as the credential label in password managers (iCloud Keychain, 1Password, Google Password Manager). Previously this was the opaque internal UserID (e.g. `user_2kF9xBq...`).

Now derives `user.name` using a priority chain:
1. First collected field with `format: email` — always human-readable
2. Identifier field value (whichever property has `x-unique`)
3. UserID fallback (service layer)

**3. Use `resolveVisitedFields` for cross-step field resolution**

The username lookup now uses `resolveVisitedFields` (the union of all fields from visited steps) instead of the current step's resolved fields. This fixes the **passkey-upsell pattern** (example `06-combined-password-passkey`) where `passkey_register` appears on a step with no fields, but the email was collected on an earlier step.

**4. Thread `Username` through the registration chain**

Added `Username` to `FlowIssuePasskeyRegistrationChallengeInput` → `BeginRegistrationInput`. The service layer falls back to UserID when empty.

### Files changed

| File | Change |
|---|---|
| `internal/domain/user_passkey.go` | Add `AuthenticatorSelection` to `BuildPasskeyCreationOptions` |
| `internal/domain/flow_passkey_registration.go` | Add `Username` to input struct |
| `internal/domain/flow_state_machine.go` | Email-first heuristic via `resolveVisitedFields`, add `findCollectedEmailField` helper, thread `def`/`userSchemaURL` into `processPasskey` |
| `internal/service/passkey_registration.go` | Add `Username` to `BeginRegistrationInput`, fallback logic in `Begin()` |
| `internal/service/flow_passkey_registration.go` | Thread `Username` through adapter  |
[...]

PR #417, opened 2026-06-25
fix: repair passkey registration flow

fforootd:
## Summary

- Label passkey registration challenges with the collected identifier when available, while keeping the generated user id as the WebAuthn user handle.
- Preserve `passkey_register` through the flow API/templates and request discoverable passkey credentials for registration.
- Prevent the login orchestrator from reconnecting `<zl-passkey>` and starting a second WebAuthn create ceremony while the registration proof submit is still loading.
- Make local CLI package artifacts deterministic by forcing release package rebuilds and cleaning package `dist` output before packing; keep private Storybook out of changeset bump planning.

## Validation

- `go test ./internal/domain ./internal/service ./internal/api`
- `moon run --force components:test`
- `moon run release:pack`
- `moon run workspace:cli -- setup --cwd /private/tmp/zitadel-fresh-hodor-codex --framework next --server local --non-interactive`
- `moon run release:changesets`
- `git diff --check`

## Release notes / changeset

- Added `.changeset/passkey-registration-labels.md` with patch bumps for `@zitadel/server`, `@zitadel/api`, and `@zitadel/components`.

## Notes

- Verified a fresh generated Next app installs the rebuilt component chunk with the loading guard instead of the stale orchestrator chunk.

### DIFF
--- a/internal/domain/flow_state_machine.go
+++ b/internal/domain/flow_state_machine.go
@@ -765,16 +765,6 @@ func needsPasskeyRegistrationVisitedFields(state *FlowState, in FlowSubmitInput,
 }
 
 func passkeyRegistrationDisplay(resolved FlowResolvedFields, collected map[string]any) (string, string) {
-	for _, f := range resolved.Fields {
-		if f.Type != FlowFieldTypeEmail {
-			continue
-		}
-		label := strings.TrimSpace(asString(collected[f.Name]))
-		if label != "" {
-			return label, label
-		}
-	}
-
 	_, _, value, ok := findCollectedFieldByChallenge(resolved.Fields, collected, FlowFieldChallengeIdentifier)
 	if !ok {
 		return "", ""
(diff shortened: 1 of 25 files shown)


fforootd:
I think this PR should replace #339

internal/domain/flow_state_machine.go
```diff
@@ -738,6 +751,41 @@ func (r *FlowStateMachineRuntime) processPasskey(ctx context.Context, client dat
...
+func passkeyRegistrationDisplay(resolved FlowResolvedFields, collected map[string]any) (string, string) {
+	for _, f := range resolved.Fields {
+		if f.Type != FlowFieldTypeEmail {
+			continue
+		}
+		label := strings.TrimSpace(asString(collected[f.Name]))
+		if label != "" {
+			return label, label
+		}
+	}
```
livio-a: IMO this is unneeded and potentially even wrong. Don't we always want to display / use the identifier. In case of a form where there's a username and an email, I'd want to use the username field no matter the field type.

livio-a (approved):

Issue #352, opened 2026-06-18
[ARCH-ADR] Define Architecture for Credential & Factor Migration and Recovery


muhlemmer:
**Context & Risk:**

Epic #249 focuses on UX and the platform model but does not specify a migration and lifecycle strategy for credentials and authentication factors. For a public MVP we need defined procedures for password/passkey migration, credential rotation, account recovery flows, and emergency key compromise handling. This ADR will define safe upgrade paths, rehashing strategies, passkey and MFA recovery processes, and emergency revocation/incident response playbooks. Without this, migrations can lock out users or create irreversible security exposures.

**Reference Material:**

- passwap hasher implementation (permalink): https://github.com/zitadel/nextgen/blob/a92b8997ab82226561be62a2b1e2ab8f3fec7833/internal/crypto/passwap.go
- session/auth ADR: https://github.com/zitadel/nextgen/blob/main/docs/adrs/010-session-auth-attempt-check-model.md

Suggested oxidel ADRs to adapt:
- https://github.com/zitadel/oxidel/blob/main/docs/adr/028-secrets-hashers-key-lifecycle.md

**Acceptance Criteria:**
- [ ] Inventory current credential storage and migration code paths.
- [ ] Define rehashing and migration triggers and safe rollout procedures.
- [ ] Define recovery flows that balance security and usability.
- [ ] Draft ADR and review with Security and Product.
- [ ] Merge ADR into `nextgen` before implementing migrations.

PR #429, opened 2026-06-26
docs: added ADR for cryptography, secrets and key lifecycle

### LINKED ISSUE #334: [ARCH-ADR] Define Architecture for Cryptography, Secrets, and Key Lifecycle
**Context & Risk:**

Epic [#249](https://github.com/zitadel/nextgen/issues/249) expects a production-ready IAM platform. The `nextgen` POC includes project secrets and crypto abstractions, but lacks a unified architecture for generation, storage, rotation, and publication of cryptographic materials. This ADR will define: project/preview secret handling, KMS/envelope encryption boundaries, signing key (JWKS) lifecycle and rotation, password hashing policy and rehash/migration strategy, and crypto-agility/FIPS constraints. Without this, key rotation mistakes, secret sprawl, and inconsistent hashing will threaten production security and availability.

**Code & File Links (permalinks):**

- passwap hasher implementation (permalink to specific blob): https://github.com/zitadel/nextgen/blob/a92b8997ab82226561be62a2b1e2ab8f3fec7833/internal/crypto/passwap.go
- Project secret lifecycle doc (main): https://github.com/zitadel/nextgen/blob/main/docs/design/platform/secret.md

**Why change the presentation:**
- Code examples now link directly to the canonical source commit (permalink) rather than duplicating raw code in the issue body. This keeps the ADR concise and guarantees readers see the exact line context in the repo.

**Reference Material:**
- Project secret doc: https://github.com/zitadel/nextgen/blob/main/docs/design/platform/secret.md
- passwap implementation (see permalink above)

Suggested `oxidel` ADRs to adapt:
- https://github.com/zitadel/oxidel/blob/main/docs/adr/028-secrets-hashers-key-lifecycle.md
- https://github.com/zitadel/oxidel/blob/main/docs/adr/027-fips-compliance.md

**Acceptance Criteria:**
- [ ] Review relevant `oxidel` ADRs.
- [ ] Draft an ADR specifying data models, KMS boundaries, rotation/retire flows, JWKS publication cadence, and hashing/rehash policy.
- [ ] Validate the ADR satisfies Epic #249 MVP goals.
- [ ] Review and merge ADR into `nextgen`.

wim07101993:
## Summary

- Added the ADR for cryptography, secrets and key lifecycle

## Release notes / changeset

- No changeset required

## Notes

Issue: https://github.com/zitadel/nextgen/issues/334


docs/adrs/028-cryptography-secrets-and-key-lifecycle.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Password/secret hashing
+
+For password hashing we use the [`passwap`](https://github.com/zitadel/passwap) 
+package. It provides all functionality to hash and migrate hashes of passwords.
+
+A project admin should be able to configure which hashing method should be used.
+This so that a tenant can have specific requirements like FIPS.
+
+By default, we should use `bcrypt` for password/secret hashing.
```
grvijayan: This [oxidel doc](https://github.com/zitadel/oxidel/blob/main/docs/guides/secrets-and-crypto.md#algorithm-selection) suggests `argon2id` as the default hashing algorithm.

wim07101993: I saw, `argon2` is considered ok according to FIPS. However `bcrypt` is actually better if I am not mistaking. But because it uses the `blowfish` hashing algorithm which is not approved by NIST, it is not approved by them.

grvijayan: I am not sure this really answers my question 🤔 
Using `argon2id` as the default seems to be decided already.
As far as FIPS compliancy is concerned, this [ADR](https://github.com/zitadel/oxidel/blob/main/docs/guides/secrets-and-crypto.md#fips-compliance) specifically mentions `Swaps Argon2id for PBKDF2 (FIPS-approved)`

wim07101993: I was mistaking it is argon2 which is better one. I was mistaking it for PBKDF2. Will fix

muhlemmer (commented):
I think this ADR needs to expanded a bit further. It currently describes a couple of use-cases of crypto, but some of the fundamentals are missing.

## Crypto targets

We should better document what types of crypto will be used for which high-level usecase:

1. **Data at rest**: Symmetric encryption of secret data that we need to retrieve and read later. Unlike passwords, which a verify only, we also have a type of secrets that can't be stored clear-text, but yet need to be decrypted when consumed. Think about IdP secrets, 3rd party API keys etc.
2. **Passwords / secrets**: already included
3. **Tokens**: The current ADR focuses more on separate types of tokens. Instead, lets define high-level token crypto target which sub-divides in:
    1. Opaque tokens: encrypted + signed for integrity. Do we want to use a derived key from the maskerkey for these (like zitadel), or can we use (asymmetric) keys managed inside a project?
    2. JWT tokens: signed only, uses asymetric crypto.

## Key management

How are keys created, stored and rotated? Differentiate between risk levels. For example, data at rest crypto is there in case we have a data leak. It's a "just in case" protection. Tokens represent identities and are continuously sent to 3rd party systems. A compromised token means a compromised identity. A compromised token key means all identities are compromised. 

1. **Masterkey and derived keys**: Both Zitadel and Nextgen use a key provided from the commandline.
    1. How do we derive keys? Can we perhaps use key-stretching methods so we can drop the fixed-length requirements, which is fragile IMO.
    2. How do we rotate the derived keys? Once a new masterkey is provided, currently it would break all data at rest. So do need to store previous revisions? What are the security risk of storing an encryption key in the same DB as the data it protects? Please investigate methods that can be employed to enable secure rotation without data loss.
    3. If we can solve the rotation problem, is it perhaps even possible for the runtime to obtain random entropy data on startup, instead of relying on the maskerkey? So continuous rotation.
2. **Asymmetric keys**: the current ADR proposes to use the single masterkey as a source. However, this will not allow customers to rotate their own keys inside a project, like we now provide for Zitadel instances. This is/was a common requested feature. Using the masterkey for tokens gives a big blast radius if the key is leaked. It will affect all identities of all projects / customers. (and may intantly put us out of business). Using per-project key pairs reduces the blast radius. Please have a look at the current [webkeys feature](https://zitadel.com/docs/guides/integrate/login/oidc/webkeys) of zitadel. I'm open to a better solution, but not a lesser one.

## Nist

The linked issue also asks about NIST constraints. I don't see any mention of of we wish to handle NIST. Zidadel recently rolled this out as a new feature, doing seperate NIST and non-NIST container builds. However, as NIST is already included by Go by default, we can also see if we can provide NIST compliance on project level. It simply means avoiding certain crypto. For example only pbkdf2 is allowed as password hash algorithm, others should remain disabled. I think the easiest way to explain what it entails is this PR: https://github.com/zitadel/zitadel/pull/12233 

## Zitadel

Please have a look at Zitadel's `internal/crypto` package and how it is used to store sensitive data in the database. However, that implementation isn't flawless and there are some existing problem I hope we can fix in nextgen:

- The oidc and saml specific keys are used for both encryption at rest and tokens. These are separate concerns and should use different derived keys.

docs/adrs/029-cryptography-secrets-and-key-lifecycle.md
```diff
@@ -0,0 +1,118 @@
...
+
+> **Status:** Proposed
+> **Date:** 2026-06-24
+> **Context:** cryptography, secrets and key lifecycle
+
+## Context
+
+Currently, there is not a unified architecture for generation, storage,
+rotation, and publication of cryptographic materials. This ADR will define:
+project/preview secret handling, KMS/envelope encryption boundaries, signing key
```
grvijayan, line 11: Should we elaborate more on `KMS/envelope encryption` for data-at-rest in the ADR?
I see envelope encryption specifically mentioned in [oxidel docs](https://github.com/zitadel/oxidel/blob/main/docs/guides/secrets-and-crypto.md#envelope-encryption) as well. Would help to have more specifics.

docs/adrs/028-cryptography-secrets-and-key-lifecycle.md
```diff
@@ -0,0 +1,118 @@
...
+
+By keeping creating key sets per project, they can be rotated in case of a leak
+or incident. Since all keys are asymmetric, data encrytped/signed with previous
+keys can still be decrypted/verified. The private keys never leave our system.
+The public keys can be requested in the form of a [JWK](https://datatracker.ietf.org/doc/html/rfc7517)
+
+Since these encryption keys need to be stored in the database and storage needs 
+to be according to the [data at rest decision](#data-at-rest), they also need
+to be encrypted. To do this, a master-key has to be provided by config. This
+can be done via a cli argument or a config parameter/environment variable. If
```
grvijayan: > If no master-key is provided, the application will create one and write it to the host's file system.

This isn't advisable for production setups, right?
Maybe make it explicit that this is for local/dev environments?

docs/adrs/029-cryptography-secrets-and-key-lifecycle.md
```diff
@@ -0,0 +1,118 @@
...
+(JWKS) lifecycle and rotation, password hashing policy and rehash/migration
```
grvijayan, line 12: The context mentions key rotation, but I couldn't find specifics about this in the ADR. 
Also, there's this other [issue](https://github.com/zitadel/nextgen/issues/352) that also mentions addressing `emergency key compromise handling`. We could also decide where to define this, either elaborate in this doc (makes more sense) or handle it as part of the other ADR.

docs/adrs/029-cryptography-secrets-and-key-lifecycle.md
```diff
@@ -0,0 +1,118 @@
...
+Since these encryption keys need to be stored in the database and storage needs 
+to be according to the [data at rest decision](#data-at-rest), they also need
+to be encrypted. To do this, a master-key has to be provided by config. This
+can be done via a cli argument or a config parameter/environment variable. If 
+no master-key is provided, the application will create one and write it to the
+host's file system. 
+
+To ensure master-key rotation:
+
+- A master-key needs to be asymmetric like all other encryption keys
```
grvijayan: > A master-key needs to be asymmetric

according to Gemini, AES-GCM is symmetric - this statement seems to be conflicting with the [statement here](https://github.com/zitadel/nextgen/pull/429/changes#diff-3e7dc1b7fa00d68f01b4687ff4db773b1d1413a2e939708027d6504105297d82R57)

docs/adrs/029-cryptography-secrets-and-key-lifecycle.md
wim07101993: The ADR does talk about it, just not in these words. By having the master-key which encrypts the other keys, we do envelope encryption. But I can mention it explicitly

docs/adrs/029-cryptography-secrets-and-key-lifecycle.md
wim07101993: Good catch, forgot to replace that. We use RSA for encryption, it is assymetric

muhlemmer (approved):

docs/adrs/029-cryptography-secrets-and-key-lifecycle.md
```diff
@@ -0,0 +1,153 @@
...
+according to the [data at rest decision](#data-at-rest), they also need to be
+encrypted. To do this, a master-key has to be provided by config. This can be
+done via a cli argument or a config parameter/environment variable. If no
+master-key is provided, the application will create one and write it to the
+host's file system. This auto-creation is mainly for testing and developing the
+system. A production system should specify its own master-key.
+
+To ensure master-key rotation:
+
+- A master-key needs to be asymmetric
```
grvijayan, line 134: hi, @wim07101993 
i am going over this ADR to inform the one I am working on (key rotation)

this statement contradicts `AES-GCM will be used.` mentioned in the [data-at-rest](https://github.com/zitadel/nextgen/blob/main/docs/adrs/029-cryptography-secrets-and-key-lifecycle.md#data-at-rest) and [third-party-secrets](https://github.com/zitadel/nextgen/blob/main/docs/adrs/029-cryptography-secrets-and-key-lifecycle.md#third-party-secrets) sections as `AES-GCM` is symmetric. 

could you please clarify/update this?

wim07101993: Yes, this is what the envelop encryption is. The diagram at [Signing/encryption keys](https://github.com/zitadel/nextgen/pull/429/changes#diff-abfa4574f0d703ee3646d83c03b37524ba45f0ba9bde5c3b1a69611757eae802R77) describes this.

We indeed have data-at-rest and third-party-secrets which need to be encrypted. To do this, we use the DEK (Data Encryption Key). 

Ideally we would use asymmetric keys as much as possible because of the ease of rotation. But those keys have a drawback. They should not be used to encrypt large chunks of data and are rather compute heavy. Solution: use a symmetric key instead and encrypt that key using an asymmetric key, a KEK (key encryption key).

Concrete:
- Encrypt data at rest using the DEK
- Encrypt the DEK with the KEK

As for generation the process is:
1. Outside process defines KEK (RSA key pair), preferrably in x.509 form because of built in lifecycle
2. Nextgen generates a DEK (AES-GCM) and stores it in the database, encrypted with the KEK
3. All other secrets are afterwards store in the database, encrypted using the DEK

grvijayan: thanks for the clarification

PR #415, opened 2026-06-25
refactor: user service to create user in flow

wim07101993:
## Summary

- Use the `UserService` to create users when registering using a password

## Validation

- `go test ./...`

## Release notes / changeset

- No changeset required

## Notes

- To also use the `UserService` when registering using as passkey, more refactoring is needed.

PR #433, opened 2026-06-26
fix: persist all collected attributes in passkey registration

vitorbari:
## Summary

Two changes in one PR, both against the same handler.

### 1. Fix: persist all collected attributes in passkey registration

The provisional-user write on the passkey-register verify leg (`FlowCreateUserForPasskeyHandler.CreateProvisionalUser`) only extracted the identifier from `state.CollectedData.UserData` and silently dropped every other schema field the user filled in on the registration step. With the default flow definition (`api/openapi/endpoints/flow_definitions/examples/default-login-flow-definition.json`), the `register` step collects `email, givenName, familyName, dateOfBirth, maritalStatus, newsletterOptIn`. Pre-fix, only `email` survived the passkey branch; the password branch (via `FlowCreateUserWithPasswordHandler` → `CreateUserAction`) correctly persisted all of them.

The fix mirrors the loop the password `on_success` handler used to use before #415: look up the identifier name once, then iterate every collected schema-known field and build a `CreateAttribute` with the right uniqueness scope.

Three behavior changes inside that one function:

1. Every schema-known collected field is now written as an attribute (the bug fix).
2. `attributeUniquenessFor(name, name, …)` is now `attributeUniquenessFor(name, identifierName, …)` — the prior call would have flipped every collected field to team-scoped uniqueness once the loop expanded beyond the identifier.
3. Missing identifier returns `domain.ErrIntegrity` (was a silent skip), matching the password handler's invariant. The flow-definition validator's `on_success` manifest cross-check already guarantees an identifier is collected on a reachable step.

The `// TODO: streamline the passkey registration process like passwords` and `// TODO: use UserService to create users.` markers from #415 are preserved — that follow-up refactor is tracked separately and has open design calls (pre-assigned user IDs in `domain.NewCreateUser`, a tx-borrowing variant of `UserService.ApplyActions`, identifier-uniqueness fallback semantics). This fix decouples the live correctness bug from those decisions; the new tests act as behavior-pinning regression armor when the larger refactor lands.

### 2. Refactor: replace `FindCollectedFieldByChallenge` with `FlowResolvedFields.IdentifierField`

The exported helper returned a 4-tuple `(name, field, value, ok)` and fanned out to two callers that each discarded most of the tuple:

| Caller | What it actually wanted |
|---|---|
| `flow_create_user_for_passkey_handler.go` | identifier field **name** |
| `flow_state_machine.go:passkeyRegistrationDisplay` | identifier **value** |

Different intents threaded through a single generic shape.

Replace it with a focused `FlowResolvedFields.IdentifierField()` method that returns just the `FlowField`. Each caller does its own collected-data lookup with intent visible. The provisional-user path now distinguishes "no identifier field declared" from "identifier not collected" — both still surface as `ErrIntegrity` but the messages are precise.

The unrelated `fieldValueByChallenge` helper (state-machine-private) stays in place — that one is a genuine by-challenge iterator over identifier and password, and `dispatchChallenges` needs its multi-kind shape.

## Validation

- `go build ./...` — clean
- `go vet ./internal/service/...` — clean
- `go test ./internal/service/... ./internal/domain/... -count=1` — green
- 8 unit tests in `flow_create_user_for_passkey_handler_test.go` cover: every collected field is written, schema-pinned identifier scope wins over the team fallback, "no identifier field declared" → `ErrIntegrity`, "identifier not in collected" → `ErrIntegrity`, schema-unknown collected keys are skipped, password-challenge fields are defensively skipped, idempotent on `*database.UniqueError`, non-unique writer errors propagate.

## Release notes / changeset

- No changeset required.

## Notes

- The bug pre-dates #415 — the original `domain.FlowCreateUserHandler.HandleProvisional` had the same single-identifier extraction. #415 moved the function from `domain` to `service` and preserved the behavior; this PR fixes it.
- See sibling docs at `docs/design/flowengine/architecture.md` L177–184 for why the passkey verify leg shares one `database.QueryExecutor` with the credential save — the atomicity story is unchanged.

### DIFF
--- a/internal/domain/flow_field_resolver.go
+++ b/internal/domain/flow_field_resolver.go
@@ -42,6 +42,18 @@ type FlowResolvedFields struct {
 	ImplicitOutcomes map[string][]string
 }
 
+// IdentifierField returns the first resolved field whose challenge is
+// [FlowFieldChallengeIdentifier]. Callers that also need the collected
+// value look it up themselves with the returned field's Name.
+func (r FlowResolvedFields) IdentifierField() (FlowField, bool) {
+	for _, f := range r.Fields {
+		if f.Challenge == FlowFieldChallengeIdentifier {
+			return f, true
+		}
+	}
+	return FlowField{}, false
+}
+
 // FlowField is the resolved per-field metadata.
 type FlowField struct {
 	// Name is the user-schema property name this field collects.
--- a/internal/domain/flow_state_machine.go
+++ b/internal/domain/flow_state_machine.go
@@ -785,11 +785,11 @@ func needsPasskeyRegistrationVisitedFields(state *FlowState, in FlowSubmitInput,
 }
 
 func passkeyRegistrationDisplay(resolved FlowResolvedFields, collected map[string]any) (string, string) {
-	_, _, value, ok := FindCollectedFieldByChallenge(resolved.Fields, collected, FlowFieldChallengeIdentifier)
+	idField, ok := resolved.IdentifierField()
 	if !ok {
 		return "", ""
 	}
-	svalue, _ := value.(string)
+	svalue, _ := collected[idField.Name].(string)
 	label := strings.TrimSpace(svalue)
 	if label == "" {
 		return "", ""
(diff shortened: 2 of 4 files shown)


vitorbari:
Closing in favor of a direct refactor: route passkey CreateProvisionalUser through UserService.ApplyActions like the password handler, with an optional caller-supplied ID on domain.NewCreateUser.

PR #434, opened 2026-06-26
refactor: route passkey CreateProvisionalUser through UserService

### LINKED ISSUE #441: bug: Passkey registration drops collected user attributes
**Description:**

During passkey registration, the provisional user created only persists the identifier attribute (e.g. `email`). Any other schema fields collected on earlier steps - `givenName`, `familyName`, etc. - are silently dropped.

Expected: every collected schema field should land on the new user, matching the password-registration path.

vitorbari:
- Route the passkey-register provisional-user write through `UserService.ApplyActions` + `CreateUserAction` (which provides schema-validation)
- Added support to optional pre-assigned ID to `domain.NewCreateUser` and `CreateUserInput`.
- Bugfix: Passkey registration was saving user with ID and email only, other attributes were dropped

Important: passkeys DB operations are not yet part of the same UserService transaction.

### DIFF
--- a/internal/domain/user.go
+++ b/internal/domain/user.go
@@ -57,7 +57,11 @@ type CreateUser struct {
 	Attributes []*CreateAttribute
 }
 
-func NewCreateUser(projectID string, teamID *string, schemabs []byte, muser map[string]any) (*CreateUser, error) {
+// NewCreateUser builds a [CreateUser] from a schema-validated user map.
+// When id is empty a fresh ID is minted; callers with a pre-bound ID
+// (e.g. the passkey-register ceremony, which keys its WebAuthn challenge
+// to a provisional user id at issue time) pass it through here.
+func NewCreateUser(projectID string, teamID *string, id string, schemabs []byte, muser map[string]any) (*CreateUser, error) {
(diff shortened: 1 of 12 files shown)


internal/domain/user.go
```diff
@@ -57,7 +57,9 @@ type CreateUser struct {
 	Attributes []*CreateAttribute
 }
 
-func NewCreateUser(projectID string, teamID *string, schemabs []byte, muser map[string]any) (*CreateUser, error) {
+// NewCreateUser builds a [CreateUser] from a schema-validated user map.
+// id passes through when non-empty; otherwise a fresh one is minted.
+func NewCreateUser(projectID string, teamID *string, id string, schemabs []byte, muser map[string]any) (*CreateUser, error) {
```
wim07101993, line 62: Do we want to have the option of the id not being created by the domain?

vitorbari: Temporarily, until we fix https://github.com/zitadel/nextgen/blob/fix/passkey-via-user-service/internal/domain/flow_state_machine.go#L737

wim07101993 (approved):

Issue #367, opened 2026-06-19
[ARCH-ADR] Error model, mapping, and reporting


muhlemmer:
**Context & Risk:**

The repository already contains partial error handling primitives (a domain.Error type that includes Code, Message, Details, Parent, and Location; API translation logic; and storage error normalization). However, there is no consolidated ADR or architecture decision that defines the expected error taxonomy, contract boundaries, reporting/stacktrace policy, PII redaction rules, and how errors flow from storage → domain → API → logging/export. Without this decision, teams may implement inconsistent error wrapping, leak sensitive details in logs or API responses, and create unstable public error contracts.

This ADR will define:
- Canonical internal error model (confirm/adopt `domain.Error` or propose changes)
- Rules for capturing and reporting error location/stack traces (when and where Location and stack traces are recorded)
- Redaction and PII handling for error Details
- Mapping rules from storage errors to domain errors and from domain errors to API error codes
- Logging and external reporting integration (how `sloggcp.NewErrorReportingHandler` and other handlers are used)
- Stable external error codes vs internal-only codes
- Backwards compatibility rules and migration guidance

**Reference material:**
- Existing domain error type: internal/domain/error.go
- API error handling: internal/api/error_handler.go
- Storage error wrapping: internal/storage/database/dialect/postgres/error.go
- Instrumentation and GCP Error Reporting usage: internal/instrumentation/config.go
- #48 

**Acceptance Criteria:**
- [x] Inventory current usage and gaps (storage, service, api, logging)
- [x] Decide whether to adopt `domain.Error` as canonical or introduce a revised type
- [x] Define policy for Location and StackTrace capture and how it surfaces to logs / GCP error reporting
- [x] Define redaction policy for Details and logging attributes
- [x] Define mapping and translation rules for storage → domain → API
- [x] Draft ADR document and place in docs/adrs/ (with title `ARCH-ADR: Error model, mapping, and reporting`)
- [ ] Review ADR with owners of storage, API, and instrumentation
- [ ] Merge ADR into main branch

**Why now:**
- `nextgen` already implements key pieces; capturing a single canonical ADR ensures the implementation is kept consistent and that operational and security expectations are clear before broader adoption.

livio-a:
ADR drafted and opened as **draft PR #439** (branch `adr-error-handling`): `docs/adrs/027-error-model-mapping-and-reporting.md`.

[...]

Issue #48, opened 2026-04-30
define clear error structure


adlerhurst:
## Original request

Currently the errors returned by the storage do not follow any structure and are therefore not helpful for the domain logic.

The storage package should define a clear error structure which can be used by the domain logic to improve error handling.

---

## ADR requirement

[ADR 030: Error Model, Mapping, and Reporting](https://github.com/zitadel/nextgen/blob/main/docs/adrs/030-error-model-mapping-and-reporting.md) — §3 Storage→domain mapping, §6 storage log leak (worked unique-attribute example), and Follow-up table rows for `database.*Error` capture/`LogValue` plus shared detection helpers.

**Depends on:** PR #521 merged — `internal/errreport` must exist (`Capture`, `Origin`, global toggles).

## Current state

- **v1 storage** ([`internal/storage/database/errors.go`](https://github.com/zitadel/nextgen/blob/main/internal/storage/database/errors.go)): typed `database.*Error` taxonomy exists (`NoRowFoundError`, `UniqueError`, …) but types do not embed `errreport.Origin`; `Error()` renders the wrapped driver with `%v` (for `23505`, `pgconn.PgError.Detail` can expose `value_hash`).
- **Dialect** ([`internal/storage/database/dialect/postgres/error.go`](https://github.com/zitadel/nextgen/blob/main/internal/storage/database/dialect/postgres/error.go), spanner equivalent): `wrapError` normalizes driver errors but does not call `errreport.Capture`.
- **No shared detection helpers** — services duplicate `errors.AsType[*database.UniqueError]` etc.
- **v2 storage** ([`internal/storage/v2/database/error.go`](https://github.com/zitadel/nextgen/blob/main/internal/storage/v2/database/error.go)): already uses `errreport.Capture` (aligned in PR #521).

## Tasks

- [ ] Embed `errreport.Origin` on v1 `database.*Error` types.
- [ ] Call `errreport.Capture` from dialect `wrapError` (postgres + spanner).
- [ ] Add `slog.LogValuer` on integrity / no-row / scan / unknown errors: emit typed fields (`integrity_type`, `table`, `constraint`) only; never driver `Detail` or row values.
- [ ] Add optional `internal/storage/storagemap` (or similar): `NotFound(err)`, `Unique(err)` — **detection only**, not resource-specific mapping.
- [ ] Tests for `LogValue` redaction and deepest-first `Origin` inheritance through the wrap chain.

## Acceptance

[...]

PR #439, opened 2026-06-29
docs: add ADR 030 for error model, mapping, and reporting

livio-a:
## Summary

Adds ADR 030 (`docs/adrs/030-error-model-mapping-and-reporting.md`) plus its index entry in `docs/adrs/README.md`. The ADR is a **Proposed** decision; this PR is documentation only — no code changes.

It records the cross-cutting error decisions for `nextgen`:

- **Canonical model** — adopt `domain.Error` (refinements: `Unwrap`, `Error()` returns only `Message`, drop the `Parent`→`Details` fallback, embed a shared `errreport.Origin`). `database.*Error` stays the storage model and is translated before crossing into the API.
- **Code taxonomy** — `{prefix}.{condition}` codes, stability classes (public stable / opaque / internal-only), sentinel factories.
- **Mapping** — storage→domain in the service layer; domain→API via `api.ErrorDetails` (`code` + `message`, optional `details`).
- **Location / stack / reporting** — centralized in an `internal/errreport` leaf with process-global toggles; deepest-first location inheritance, deepest-first stack inheritance, `slog.LogValuer`, GCP Error Reporting wiring.
- **Redaction & PII** — producer-governed `Details`, log-safe-at-source causes, a worked unique-attribute (EAV / `value_hash`) example, masking limits for schema-driven data, and a list of known leak sites to remediate.

## Validation

- Documentation only; no code paths changed, so no build/lint/test was run.
- Cross-checked every code claim and reference in the ADR against the current source (error types, API leak sites by function name, `instrumentation` config, `sloggcp` interfaces, OpenAPI wire shape, `WithLogging` middleware).

## Release notes / changeset

No changeset required — docs only, no shipped behavior changed (per `.changeset/README.md` decision table).

## Notes

- **Needs owner review** from storage, API, and instrumentation per issue #367 (ACs: review + merge). Kept as a draft until that review.
- **Implementation is intentionally not in this PR.** The doc stays a clean, citable decision; the behavioral changes have real coupling (`domain.Error.Details` consumers, `Error()` test assertions). Follow-up PRs are sequenced in the ADR's "Follow-up work" table: Foundation (`errreport` + `domain.Error`) → Storage capture → Leak remediation → Logging policy → Contract/taxonomy.
- Follow-up implementation issues will be filed once the ADR merges (so they can link stable ADR anchors); #367 closes on merge, and #48 remains the home for the storage detection-helper work.


wim07101993 (commented):
Since storage layer can only return `storage.Error` and domain/service layer can only return `domain.Error`, are we going to change the function signatures accordingliy?

vitorbari:
> Since storage layer can only return `storage.Error` and domain/service layer can only return `domain.Error`, are we going to change the function signatures accordingliy?

From my understanding, everything is just a convention, but the layers still return `error`. So the rules are not really enforced in compile-time.
If the convention is not followed, it would lead to 500 errors.

docs/adrs/027-error-model-mapping-and-reporting.md
```diff
@@ -0,0 +1,363 @@
...
+- `internal` responses use the generic domain message regardless of `Parent` content.
+
+#### Messages
+
+- `Message` is the customer-facing string. It must not contain stack traces, SQL, secret values, or raw driver text.
+- For `req.invalid` and ogen decode errors, prefer normalized messages; avoid leaking parser internals when a stable domain message exists.
+- **`domain.Error.Error()` returns only `Message`; it does not include the parent chain.** `err.Error()` is the Go-default way to render an error, so the default must be safe. The parent is only needed for diagnostics, which are served by `Unwrap()` (programmatic traversal / `errors.Is` / `errors.As`) and `LogValue` (structured logs explicitly emit `parent`, Decision 5). Making `Error()` parent-free means the common `err.Error()` call cannot leak a wrapped cause.
+- This does **not** make every `Error()` safe: non-`domain.Error` values (storage `database.*Error`, ogen framework errors) still include their cause in `Error()`. API code must therefore avoid rendering a **non-domain** error's `Error()` into a client message — translate to a `domain.Error` first, or use the explicit `Message` field via `domainErrorDetails`.
```
adlerhurst: I think they don't have to but _can_ still include their cause.
```suggestion
- This does **not** make every `Error()` safe: non-`domain.Error` values (storage `database.*Error`, ogen framework errors) can still include their cause in `Error()`. API code must therefore avoid rendering a **non-domain** error's `Error()` into a client message — translate to a `domain.Error` first, or use the explicit `Message` field via `domainErrorDetails`.
```

docs/adrs/027-error-model-mapping-and-reporting.md
```diff
@@ -0,0 +1,363 @@
...
+- Trade-off: plain `%v` / non-structured logging of a `domain.Error` no longer shows the cause inline; logs must use the structured path (`slog.Any` → `LogValue`) to capture `parent`. `errors.Is` / `errors.As` / `Unwrap` are unaffected. Confirm at implementation time that no test asserts a chained string via `EqualError` / `ErrorContains` on a `domain.Error`.
```
adlerhurst: What about adding a verbose flag to handle such cases?

wim07101993:
> If the convention is not followed, it would lead to 500 errors.

Should we enforce it to enforce it? I'm pro, why let convention do what the compiler can enforce?

docs/adrs/030-error-model-mapping-and-reporting.md
```diff
@@ -0,0 +1,363 @@
...
+
+1. **Which rule failed** is the error's job — but only partially. The pg dialect captures `table` + `constraint`, here `user_unique_attributes` and a partition PK such as `user_unique_attributes_part_0_pkey`. The storage error's `LogValue` emits those typed fields and drops the driver `Detail`. Two reasons the `Detail` must be dropped: the row it names contains the `value_hash` (a salt-less SHA-256 of, for example, an email — reversible for low-entropy identifiers, so still sensitive), and the PK constraint name is shared by *all* unique attributes, so it does not even reveal which attribute collided:
+
+```
+level=WARN msg="registration failed" code=user.email_taken
+  parent.db_error=integrity_violation parent.integrity_type=unique
+  parent.table=user_unique_attributes parent.constraint=user_unique_attributes_part_0_pkey
+  request_id=... trace_id=... org_id=...
```
adlerhurst, line 293: I would have expected it to be a DEBUG log because it is at database level and it does not imply any application errors. The go example later would log it on debug level.

adlerhurst (commented):
solid work. please don't forget to update the ADR number (27 is already taken)

docs/adrs/030-error-model-mapping-and-reporting.md
```diff
@@ -0,0 +1,363 @@
...
+`domain.Error` stays a **value type**. The refinements above (value-receiver `Unwrap`, embedded `Origin`, `slog.LogValuer`) do not require a pointer type, so the existing sentinel-factory and value-return patterns are unchanged.
+
+**Keep `database.*Error` as the storage-layer model.** Storage errors stop at the service/repository boundary and are translated into `domain.Error` before crossing into API handlers. They are not returned directly to HTTP clients.
+
+Layer contract:
+
+```
+driver/dialect → database.*Error → domain.Error → api.ErrorDetails (+ HTTP status)
+                                      ↓
+                               structured logs / GCP Error Reporting
```
muhlemmer, line 104: Does this diagram mean that the error logline is omitted when `domain.Error` is constructed? This is slightly different from Zitadel, where we do emmit the logline in the middleware, so it is tied to the request details / ID.

I have no problem with emitting the logline in the constructor, but then we have to make sure the request context is passed into it as well.

docs/adrs/027-error-model-mapping-and-reporting.md
muhlemmer: I think we should just set the requirement we only do structured logging. I see no legitimate use-case where we would import any other logger that `slog`.

docs/adrs/027-error-model-mapping-and-reporting.md
```diff
@@ -0,0 +1,363 @@
...
+
+- Every structured error type embeds `errreport.Origin`, which carries a `*sloggcp.ReportLocation` and an optional stack, and implements `sloggcp.ReportLocationError` + `sloggcp.StackTraceError`.
+- `errreport.Capture(parent error, skip int) Origin`:
+  - records a **fresh location** at the creation site when location reporting is enabled;
+  - for the stack, **inherits the deepest existing one**: if the `parent` chain already contains a `sloggcp.StackTraceError`, reuse its stack; otherwise capture a fresh stack when stack reporting is enabled.
+
+#### Location
+
+- Captured at **every** error-construction boundary, storage included — `database.*Error` records its origin in `wrapError`, `domain.Error` records its origin in `newError`.
+- **Location is always local to the creation site; it is never inherited from a parent.** Wrapping is a deliberate "represent this failure as this error now" decision, so the wrapping error's own location is the meaningful primary one. Ancestors keep their own locations in the `Parent` chain.
```
muhlemmer: I would propose to keep the original location of the parent error, and therefore inherited, as we can only send a single location to the logger.

Just an opinion, but I would be more interested in where the error originated in the first place, so the deepest point of the call stack where an error is returned from a 3rd party package. As opposed to the location where the error was wrapped the last time.

docs/adrs/027-error-model-mapping-and-reporting.md
```diff
@@ -0,0 +1,363 @@
...
+- Log the root error with `slog.Any(ErrorAttributeKey, err)`; the handler resolves location/stack through the shared interfaces and `Parent` via `Unwrap()`.
+- `context.Cause` continues to surface through `ExtractErrCause`.
+
+#### Expected client errors (4xx)
+
+- Log at `Info` or `Warn` with `error.code` and request correlation attributes.
+- `middleware.WithLogging` (Inventory) currently logs **all** status ≥ 400 at `Error` with the full response body. Align it: log expected 4xx at `Info`/`Warn`, reserve `Error` for `internal`/5xx, and stop logging the raw response body — log the structured `code` plus safe attributes instead (Decision 6).
+- Expected storage signals (for example `NoRowFoundError` → `*.not_found`) should not trigger stack capture in normal operation; capturing stacks for frequently-expected errors is acceptable only while the debug toggle is on.
+- Do **not** send 4xx domain errors to GCP Error Reporting as service incidents unless rate anomalies trigger alerting separately.
```
muhlemmer: When `sloggcp` is used as log handler, all loglines containing an `error` field end up in error reporting. Eg, we don't explicitly send anything, google parses the log line and if it matches the error reporting format it is collected.

I don't think there's anything wrong with that, we will see error reports are level warn or info. No alerting required, and yet we can gather useful info on problems our client face. For example if we see a sudden spike of 403, it would be easy to spot if we broke something in our own permission code by the hotspot created in error reporting.

docs/adrs/030-error-model-mapping-and-reporting.md
muhlemmer: That contradicts what's written earlier in the ADR:

https://github.com/zitadel/nextgen/blob/a7d7c88f6aed7cbb093f056c1b2710378ad69406/docs/adrs/027-error-model-mapping-and-reporting.md?plain=1#L229-L234

A conflict is a client error and should be logged as WARN in the domain / service layer. It's fine for the DB to produce additional DEBUG logs within its own package scope, but that should be part of the error handling / reporting scope.

docs/adrs/027-error-model-mapping-and-reporting.md
livio-a: Yeah makes sense. Since the location is only logged but never returned in the API we don't leak information and should handle it the same way as a stack trace, which is just a more verbose form and might only be enabled in specific cases, where as the location is probably always enabled.

docs/adrs/027-error-model-mapping-and-reporting.md
livio-a: Yes makes sense. I'll remove the entry. The decision to trigger alerts or rather based on which information we trigger is out of scope of this ADR anyway.

docs/adrs/030-error-model-mapping-and-reporting.md
livio-a: Yes the log would be thrown by the domain / service layer with the storage error being the parent.
I'll add a sentence to make this clearer.

docs/adrs/030-error-model-mapping-and-reporting.md
livio-a: I thought I had already changed this, since i stumbled upon it as well after Cursor created it.
My intention would be to still use the middleware. There are even some gaps / follow ups for that middleware to solve mentioned later. I think Cursor wanted to express that the structure of what's being logged comes out of the domain error and is transformed into error details for the API / client and the structured logs use the domain error info and not the error details (to prevent information leakage)

livio-a:
> > If the convention is not followed, it would lead to 500 errors.
> 
> Should we enforce it to enforce it? I'm pro, why let convention do what the compiler can enforce?

I would not change the function signatures to return `domain.Error` / `database.*Error` as dedicated return parameters.

In Go, the idiomatic contract is still `(T, error)` and `err != nil`. Typed errors are carried **inside** the `error` value and inspected with `errors.Is` / `errors.As`.

A signature like:
```go
func Get(...) (*domain.User, domain.Error)
```
would introduce a second success convention (`domain.Error.Code == ""` / zero value means no error), which is less idiomatic and easier to misuse than the standard `err == nil` check. It also becomes awkward for non-domain failures such as `context.Canceled`, driver errors, or unexpected internal errors — those still naturally implement `error`.

And even with  `(*domain.User, *domain.Error)` and a `nil` check, I would still keep `(T, error)`.
Otherwise, every failure must be converted into that type up front. Anything that is not already a `domain.Error` (storage error, canceled context, framework error) either gets forced into `domain.ErrInternal(...)` immediately or needs a second/error-parallel path anyway.
At the API boundary we would then convert `*domain.Error` back into `error` for handlers/logging, which is extra churn without much gain over the current approach.

Repositories IMO are even allowed to also return a `domain.Error` in specific cases, where the service layer then can decide to return that as is or wrap it if needed (domain.ErrSessionNotFound() is already used like that).

> solid work. please don't forget to update the ADR number (27 is already taken)

I'll do that once all other comments are settled and it's ready to merge.

docs/adrs/030-error-model-mapping-and-reporting.md
```diff
@@ -0,0 +1,378 @@
...
+#### Response body
+
+- Always return `code` and `message` from the `domain.Error`.
+- **`Details` is marshalled to the client whenever a producer sets it** (standard JSON marshalling). Setting `Details` *is* the act of exposing it — there is no separate per-code allow-list.
+- `Details` must contain only **client-actionable, non-sensitive** data (field paths, schema validation issues, conflict keys). Keeping secrets/PII out is the **producer's** responsibility (Decision 6), because the value is client-facing by intent.
+- A details **container type is not a client-redaction mechanism**: a `map`/struct marshals whatever it holds, and no generic type can distinguish intended PII (for example the subject's own email on a self-scoped endpoint) from a leak. Redaction is enforced by producer discipline and review, not by the type.
+
+#### HTTP status
+
+- Status mapping stays in `internal/api/*_errorResponse` functions, keyed by `err.Code`.
```
vitorbari, line 198: Looking at the current `internal/api/*_errorResponse` functions (`userErrorResponse`, `sessionErrorResponse`, tetc), the pattern feels error-prone and repetitive. The same mapping "not-found -> 404" is duplicated once per resource:

```go
// user.go
case domain.ErrUserNotFound().Code:    return errorResponseWithStatusCode(http.StatusNotFound, err)
// team.go
case domain.ErrTeamNotFound().Code:    return errorResponseWithStatusCode(http.StatusNotFound, err)
// session.go
case domain.ErrSessionNotFound().Code: return errorResponseWithStatusCode(http.StatusNotFound, err)
```

What if we add a error class/category to `domain.Error`? For example:

```go
// internal/domain/error.go
type ErrorClass int
const (
    ClassUnknown ErrorClass = iota
    ClassNotFound
    ClassUnauthorized
    ClassForbidden
    // ..
)

// internal/domain/user.go
func ErrUserNotFound() Error {
    return newError(PrefixUser.ErrorCodePrefix("not_found"),
        ClassNotFound, "The user was not found.", nil, nil)
}
```

Then we have to map it only once:

```go
// internal/api/error_handler.go
var classToStatus = map[domain.ErrorClass]int{
    domain.ClassNotFound: http.StatusNotFound,
    // ...
}
```

livio-a: We had categories in the past, but without specific errors — you'd throw a generic `NotFoundError` and that would map to gRPC/HTTP NotFound / 404.

That felt wrong because not every "not found" should be a 404. For example, user-not-found during authentication is often hidden or returned as something else (`precondition failed`, opaque invalid proof, etc.) to avoid enumeration. With the current domain errors we already handle that: specific codes like `ErrAuthAttemptProofRejected` get their own API mapping, independent of a generic "not found" class.

`ErrorClass` on `domain.Error` would partly bring that generic category back — and it also starts to put HTTP semantics in the domain layer. The intention here is to keep that separated: domain exposes `code` + `message` (+ optional `details`); **API** maps code → status (today per-resource `*_errorResponse`, which is repetitive but explicit).

I'm open to deduplicating the API switches (shared helper, suffix conventions for the boring cases, or a central `code → status` table that lives in `internal/api` only) as follow-up. I would not add `ErrorClass` to `domain.Error` in this ADR.

WDYT?

vitorbari: Sounds good to me 👍

muhlemmer (approved):

PR #450, opened 2026-06-30
docs: added adr for token lifecycle

### LINKED ISSUE #335: [ARCH-ADR] Define Architecture for Session Authority, Token Lifecycle, and Revocation
**Context & Risk:**

A production IAM must have a clear server-authoritative model for sessions and tokens. `nextgen` has session/auth-attempt primitives and token metadata, but lacks a consolidated architecture for token classes, refresh token rotation/binding/replay detection, revocation semantics, and how logout/password/factor changes invalidate credentials. This ADR will define the authoritative session model, token families, revocation propagation, and incident/administrative invalidation flows. Without this, token misuse, inconsistent invalidation, and recovery gaps become likely.

**Code & File Links (permalinks):**

- Token domain model (permalink to specific blob): 

https://github.com/zitadel/nextgen/blob/83772d63306892932274292e747fe193263ea432/internal/domain/token.go#L17-L31

- Session/auth model ADR: https://github.com/zitadel/nextgen/blob/main/docs/adrs/010-session-auth-attempt-check-model.md

**Why change the presentation:**
- The issue now links directly to the authoritative code blob so reviewers can inspect exact structs and invariants without me pasting code into the issue body.

**Reference Material:**
- `docs/adrs/010-session-auth-attempt-check-model.md` (nextgen)
- Token domain model (see permalink above)

Suggested `oxidel` ADRs to adapt:
- https://github.com/zitadel/oxidel/blob/main/docs/adr/044-short-lived-edge-access-tokens-and-authoritative-browser-sessions.md

**Acceptance Criteria:**
- [ ] Review relevant `oxidel` ADRs.
- [ ] Draft an ADR specifying token/session classes, refresh rotation and replay handling, revocation storage/propagation, and logout/invalidate-all semantics.
- [ ] Validate the ADR satisfies Epic #249 MVP goals.
- [ ] Review and merge ADR into `nextgen`.

wim07101993:
## Summary

- Added the ADR for token lifecycle

## Release notes / changeset

- No changeset required

## Notes

Issue: #335


docs/adrs/029-token-lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+the token is still active. Additional claims can be added to minimize the amount
+of database round-trips and required compute to validate whether a token is
+active. E.g.: a session token will have the `session_id` claim so that the 
+server can validate whether the session/user is still active.
+
+### Storage
+
+Only tokens which are long-lived (refresh tokens and PATs) should be stored in
+the server because the need to be able to be revoked. Access tokens and session
+tokens should not be stored in the server.
```
livio-a: Doesn't this contradict the statement above where a `jti` and / or `session_id` is stored to verify it server side / handle revocation?

docs/adrs/037-token-lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+mark the token as inactive in the database which will make future validations 
+fail.
+
+
+### Refresh token Rotation & Replay detection
+
+To mitigate token interception and misuse, refresh tokens are single use. When
+a refresh token is used, it is removed/marked as used/inactive in the database.
+In the token response at the end of its exchange, a new refresh token is 
+returned. This token can be used for the next refresh.
```
livio-a, line 83: Kind of edge case, but apparently common problem, as we've had multiple cases in the past, where immediate rotation / revocation of the previous results in issues (SPA / native apps. We might need to either make this rotation optional or just provide some idempotency handling.

docs/adrs/029-token-lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+When a security event occurs (explicit logout, password reset, multifactor 
+credential removal, or administrative user blocking), invalidation propagates 
+via two distinct operational models.
+
+1. First-Party Surfaces (Immediate): Session tokens evaluate directly against 
+   the server on every single call. Administrative lockouts or logouts take 
+   effect instantly on all first-party UI/UX applications.
+2. Third-Party Edges (Eventually Consistent): Active Access Tokens cannot be 
+   revoked once published validated offline. They naturally burn out at the end
+   of their 5-minute lifespan.
```
livio-a: What about refresh tokens? Do they get revoked as well?

livio-a (commented):
I get the overall picture but IMO miss some decision drivers / considerations to go this route. While it seems to be based mostly on oxidel's ADR 044, which is driven by the multi-region / spanner / enterprise grade traffic and to serve those hot paths as fast as possible and to not rely on server side or database calls where not absolutely needed. I'd love to have either these or any other reasons documented in here as well.
Also we should include SAML (assertions) into the ADR.

docs/adrs/029-token-lifecycle.md
```diff
@@ -0,0 +1,80 @@
+# ADR 026: Token lifecycle
+
+> **Status:** Proposed
+> **Date:** 2026-06-30
+> **Context:** Evolution of session security
+
+## Context
+
+We are building an authorization platform which has to be stateless and
```
muhlemmer: The platform isn't stateless, right? We actively store and maintain state. There may be parts of the platform which are in fact stateless, but not as a whole.

docs/adrs/029-token-lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+| Session Tokens              | First-party app | Server | Opaque                               | Session bound                                                                    | First-party applications need to know whether a session is still active or revoked, whether the user was deactivated,... Those are better served by an authoritative session row than by a self-contained browser token.             |
+| Access Tokens               | Third-party app | Client | Signed JWT (but opaque configurable) | Short: minutes (default 5 min)                                                   | For short-lived edge tokens we optimize for performance. If a token is self-contained, no database or even api call is required. This reduces system-load and increases responsiveness of the applications.                          |
```
muhlemmer: Access tokens aren't always short-lived edge tokens. I think this confuses with the AI agentic usecase where we want short-lived tokens to be obtained / exchanged on the edge. However, we still need to integrate with normal applications, where human users login. Their access tokens typically live longer.

Short lived AT helps with stateless validtion. But some customers will still want to call introspection in an old fashioned way. So we should not focus on a single usecase, but talk about how we configure the AT lifetimes instead.

docs/adrs/029-token-lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+
+According to [RFC7519](https://datatracker.ietf.org/doc/html/rfc7519)
+
+By default, provide some standard claims:
+
+```json
+{
+  "iss": "https://auth.example.com",
+  "sub": "123456789",
+  "aud": "the_audience_",
```
muhlemmer: Let's document `aud` as a JSON array.

docs/adrs/029-token-lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+An opaque token always contains a `jti` claim so that the server can verify that
+the token is still active. Additional claims can be added to minimize the amount
+of database round-trips and required compute to validate whether a token is
+active. E.g.: a session token will have the `session_id` claim so that the 
+server can validate whether the session/user is still active.
+
+### Storage
+
+Only tokens which are long-lived (refresh tokens and PATs) should be stored in
+the server because the need to be able to be revoked. Access tokens and session
```
muhlemmer: Access tokens should be stored always. See lifetime comment. As part of Oauth2 compatibilty we will also implement introspection, userinfo. Those endpoints need to verify AT existence on the server. Then there is also the token revokation endpoint. Once that is called, but introspection and userinfo must not return any valid data. Regardless if the AT was long or short lived.

docs/adrs/029-token-lifecycle.md
```diff
@@ -0,0 +1,80 @@
...
+| Refresh Token               | Auth server     | Client | Opaque                               | Long: days/weeks (default 2 weeks)                                               | Because refresh tokens are long-lived, they need to be single-use and revocable. This is to mitigate replay attacks. Refresh tokens need to be exchanged for access-tokens. In the token response, an new refresh token is provided. |
+| Personal Access Token (PAT) | Auth server     | Client | Opaque                               | Very long: months/years/infinite (default 3 months, infinite: not recommendable) | Because PATs are long-lived, they need to be revocable. PATs need to be exchanged for access-tokens.                                                                                                                                 |
```
muhlemmer: What about SAML assertions?

muhlemmer (commented):
I think this ADR focuses too much on a single usecase, agentic AI with short-lived tokens, while we will also need to support many others.

IMO if an edge function wants a short-lived narrow scoped access token which it can self-verify, it can call token exchange with a longer-lived. But this makes it a concern of the edge app, not the IAM platform like zitadel / nextgen.

docs/adrs/029-token-lifecycle.md
wim07101993: What do you mean? Imo access tokens should always be short-lived. Normal applications can handle those as wel by using the refresh token which is included in the token response.

docs/adrs/029-token-lifecycle.md
wim07101993: Done, but it contradicts [the spec](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3) in:
> In the special case when the JWT has one audience, the "aud" value MAY be a single case-sensitive string containing a StringOrURI value.

Agreed, it specifies "MAY" but still it is allowed

docs/adrs/037-token-lifecycle.md
wim07101993: I guess, idempotency on revocation. I will add that.

docs/adrs/029-token-lifecycle.md
wim07101993: In my head, saml assertions fall under access-token. I added clarification for that in the Self-contained Tokens topic.

docs/adrs/029-token-lifecycle.md
muhlemmer: Only if they use the `offline_access` scope. Refresh tokens aren't returned by default

muhlemmer: So what I mean is that we shouldn't make assumptions. There are use cases where the AT should be short lived and minted on the edge. And there are classic use cases where applications only get an access token and MAY get an refresh token for offline access, but this is not a default behavior. There are also use cases where the app does not pass the `offline_access` scope and require user re-login once the AT expires. In those case the AT should be longer living than 5 minutes.

wim07101993: I agree that 5 minutes is a short window if do to stuff from an application/UX point of view. But if an application needs to do more, shouldn't we than document that an application should request that scope?

Second option is to expand the lifetime. But that would mean a less secure default for applications which do request a refresh-token.

Maybe we can create a configuration split for confidential clients and non-confidential clients? For a confidential client (like ssr web-app) we can assume the AT is store more securely than a non confidential one (like SPA). For non confidential apps I would not recommend using an AT-lifetime greater than 15min.

Another option is to split configuration for requests depending on whether they also request a refresh-token. But I got a feeling that the configuration for that will get confusing for users.

@muhlemmer WDYT?

muhlemmer: SPA is an entirely different rabbit hole. There it doesn't matter much if the access token is short lived and a refresh token is also obtained by the SPA. Both live in session storage and can be leaked. In fact, `offline_access` scope was intended for webapps that want to keep a session active without user interaction. The user may not be using the browser, but the backend of the webapp can keep the session active doing timed refresh.

Off course, SPAs can still ask for a refresh token to improve UX when the access token expires, but there is no security benefit of the AT being shorter lived. If tokens are leaked from the browser, the attacker can use the client ID and the refresh token to hijack the session locally.

I think this discussion is moving in the wrong direction. I was just commenting that the ADR should cover all use cases and not make strong assumptions that the AT is a short lived edge tokens. That is a single use case for agents.

The ADR also mentions that database calls are not required for validation. So I was wondering how would an introspection endpoint be implemented?

docs/adrs/035-token-lifecycle.md
```diff
@@ -0,0 +1,113 @@
...
+rotation/ binding/replay detection, revocation semantics, and how
+logout/password/factor changes invalidate credentials. This ADR will define the
+authoritative session model, token families, revocation propagation, and
+incident/administrative invalidation flows. Without this, token misuse,
+inconsistent invalidation, and recovery gaps become likely.
+
+## Decision
+
+|                             | Audience        | Owner  | Token type                               | Lifespan                                                                         | Note                                                                                                                                                                                                                     |
+|:----------------------------|:----------------|:-------|:-----------------------------------------|:---------------------------------------------------------------------------------|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
grvijayan: should we also include id_token as a token type?
https://zitadel.com/docs/guides/integrate/token-exchange#id-token-type

docs/adrs/029-token-lifecycle.md
wim07101993: > There it doesn't matter much if the access token is short lived and a refresh token is also obtained by the SPA. 

I would disagree. Since ATs can be self-contained, revoking them might take some time. Refresh tokens are revoke immediately. (https://datatracker.ietf.org/doc/html/rfc6819#section-3.3)

> That is a single use case for agents.

I would disagree there as well. I see it as I stated. All access tokens are short lived, unless specified otherwise (currently the only exceptions are PAT's). I did not have agents in mind at all when writing this PR.

The idea is, all ATs are short-lived, if a longer lifespan is required, request a refresh token with which you can refresh the access-token. 

> The ADR also mentions that database calls are not required for validation. So I was wondering how would an introspection endpoint be implemented?

The idea is that all information can be retrieved from the token itself. Since all tokens are signed, we know they are not tampered with.

docs/adrs/035-token-lifecycle.md
wim07101993: Good point. I interpreted the issue as only for authentication tokens. Since the ID token is not an authentication token, I did not add it to this list. I can add a separate section on the topic if you want.

grvijayan: ID token is an authentication token, right?

docs/adrs/035-token-lifecycle.md
```diff
@@ -0,0 +1,113 @@
...
+| Refresh Token               | Auth server     | Client | Opaque                                   | Long: days/weeks (default 2 weeks)                                               | Because refresh tokens are long-lived, they need to be single-use. This is to mitigate replay attacks. Refresh tokens need to be exchanged for access-tokens. In the token response, a new refresh token is provided.    |
+| Personal Access Token (PAT) | Auth server     | Client | Opaque                                   | Very long: months/years/infinite (default 3 months, infinite: not recommendable) | PATs need to be exchanged for access-tokens.                                                                                                                                                                             |
```
livio-a: > PATs need to be exchanged for access-tokens. 

What do you mean with that? A Personal Access Token is as the name suggests a ready to use access token.

docs/adrs/032-token-lifecycle.md
```diff
@@ -0,0 +1,113 @@
...
+
+```json
+{
+  "iss": "https://auth.example.com",
+  "sub": "123456789",
+  "aud": [
+    "the_audience_"
+  ],
+  "iat": 1516299022,
+  "exp": 1516239022,
```
livio-a: nitpicking but the token would expired before it's even issued 😄

docs/adrs/035-token-lifecycle.md
livio-a: An id token is to **identify** the user at the application they logged in and should never be used to authenticate.
IMO less of a concern since the format is clear and also expiration, etc is given by the underlying session.

docs/adrs/029-token-lifecycle.md
livio-a: I think we should clarify the storage / revocation part more clearly. From what I read the idea is to use JWTs to allow RPs and resource servers to self validate without additional server calls, but still offer introspection and a revoke endpoint in the future. The latter then requires either introspection or have a security model (in the app) where the expiration of a short lived access token is acceptable and preferred over an additional roundtrip (at least for non high security operations, at which you could anyway do a step-up / re-auth and local validation without introspection).

Storing all (short-lived) access tokens and if it's only their `jti`, adds issuance/replication cost on the hot path server side, even if we use JWTs to improve performance at the RP side.
We should decide whether we really want to store them or go more with oxidel's ADR044 stateless approach (even if the code deviates from that 😄). Synchronized / immediate revocation of access tokens would still be possible with an additional a `jti`-based denylist. Revocation is way less common or called than issuing a token.

One of the treda-offs (and mentioned in the ADR as well) for sure then is the token size, but IMO still ok.

Wrt refresh tokens:
I believe we took the OIDC spec a bit too literal in zitadel and only issue a refresh token if you pass the `offline_access` scope, but OAuth would allow returning one even without, which would allow clients (incl. SPA) to refresh during the current session. These are then typically also revoked on sign out and only the ones with offline_access are kept and need to be revoked explicitly.
Nuance but related to that topic: While we currently immediately rotate/invalidate not only the refresh_token on a refresh grant, we also do the same for the access_token. Dropping the immediate revocation for the access_token, and letting it remain valid until `exp`, might already solve some current issues SPA have with that.

As Flo states: We should be more opinionated and I feel like we currently try again to provide all possible options to clients and won't be able to optimize for one of them.
Based on the past experience (incl. performance optimizations) I'd rather decide for a clearer opinionated way as well rather than having all the options and trying to optimize them one by one again. And to even start with the enterprise-grade design in mind.

livio-a (commented):
see main comment on Tim's thread

docs/adrs/035-token-lifecycle.md
wim07101993: Ow, that must be a misunderstanding on my side. Will adjust

docs/adrs/029-token-lifecycle.md
muhlemmer: I reread token revocation [RFC 7009](https://www.rfc-editor.org/info/rfc7009/) and indeed the refresh token is the only token type that is required for revocation, access tokens are optional. The implementation notes specifically say a short live AT with a revocable refresh token is a valid implementation.

Yes, I may have leaning too much on the current implementation. Let me reread the ADR with this in mind. I'll close this comment and post back with another review.

docs/adrs/037-token-lifecycle.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Refresh token Rotation & Replay detection
+
+To mitigate token interception and misuse, refresh tokens are single use. When
+a refresh token is used, it is removed/marked as used/inactive in the database.
+In the token response at the end of its exchange, a new refresh token is
+returned. This token can be used for the next refresh.
+
+[RFC7009 for token revocation](https://datatracker.ietf.org/doc/html/rfc7009#section-2.1)
+states:
```
muhlemmer, line 86: Please clarify which parts of RFC 7009 we wish to implement.

> Implementations MUST support the revocation of refresh tokens and
   SHOULD support the revocation of access tokens (see Implementation
   Note).

muhlemmer (commented):
With the previous discussion that flared up because of me, I think it would be helpful to include a table of RFCs the implementation should cover and which sections we choose to exclude for the MVP, for the Oauth2 / OIDC related tokens.

wim07101993:
@muhlemmer  Wouldn't specifying which parts of which RFC's should be implemented explode the scope of this ADR? It would result in specifying which auth protocols we are going to implement.

livio-a (commented):
I'm fine with merging this as-is for the MVP / after fixing the version again 😬

muhlemmer (approved):
Looks good to me. Thanks for the rewrite.

PR #454, opened 2026-07-01
docs: credential and factor migration and recovery

### LINKED ISSUE #352: [ARCH-ADR] Define Architecture for Credential & Factor Migration and Recovery

grvijayan:
## Summary

- This ADR documents intended procedures for password/passkey/TOTP lifecycle changes, account recovery, and key-compromise response, and links it from the ADR index.

## Validation

- 

## Release notes / changeset
- No changeset required — no shipped behavior changed.

## Notes

- Closes https://github.com/zitadel/nextgen/issues/352


docs/adrs/035-credential-migration-rotation-and-recovery.md
```diff
@@ -0,0 +1,228 @@
...
+* Keep the old `rpId` and serve it from the new domain via [**Related Origin Requests (ROR)**](https://passkeys.dev/docs/advanced/related-origins/), listing the new origin in the RP's `.well-known/webauthn`. ROR caps the number of related origins and has uneven browser support, so treat it as a transition aid, not a permanent state.
+* Over that transition window, prompt users to register a fresh passkey under the new `rpId`, then retire reliance on the old one.
+
+### 3. TOTP Enrollment and Rotation
+Rotating a TOTP secret mandates invalidating the current shared secret and establishing a new one. Two TOTP secrets cannot coexist the way two passkeys can, so rotation replaces in place and there is no last-factor safeguard; account recovery ([Section 4](#4-account-recovery)) is the fallback if setup is interrupted.
```
wim07101993: Why not? I could see it beeing usefull if a user could add a backup/second one.

docs/adrs/035-credential-migration-rotation-and-recovery.md
```diff
@@ -0,0 +1,228 @@
...
+
+* **Device Change:** When registering a new device to replace an old one, the new TOTP secret replaces the old one. Remove the existing TOTP configuration and register a new one.
+* **Device Loss:** In the event of device loss or compromise, the user authenticates via a recovery method (see [Section 4](#4-account-recovery)). Upon successful recovery, the system immediately prompts the user to re-register their TOTP secret.
+* **Deregistration:** Removing TOTP outright, rather than rotating it, is subject to the same last-factor safeguard as passkeys ([Section 2](#2-passkey-migration)): if TOTP is the account's only remaining usable factor, the user must set up an alternative before it can be removed.
+
+### 4. Account Recovery
+Account recovery acts as the fallback mechanism when a user's primary authentication factor (Passkey, TOTP) is lost or compromised.
+
+#### Recovery Codes
```
wim07101993: Is there a separate set of keys for each factor? (One for passkey, another set for TOTP)

docs/adrs/035-credential-migration-rotation-and-recovery.md
```diff
@@ -0,0 +1,228 @@
...
+
+Three configured durations govern the signing key lifecycle:
+- **Propagation window:** how long a newly created key waits before it signs anything. It is the maximum cache lifetime Nextgen advertises for the JWKS response (the `Cache-Control` max-age, a project-level config with a default of 5 min), so clients holding a cached copy have had a chance to pick up the new key before it signs.
+    - `active_from` defaults to `now() + propagation_window` while the project already has an active key.
+    - `active_from = now()` for the first key of the project, where no cached JWKS can exist yet, and for a key replacing a compromised one, where the risk of an unresolvable `kid` is preferable to the risk of continuing to sign with a key an attacker holds.
+- **Grace period:** how long a key stays in the JWKS after its successor takes over.
+    - This should be at least as long as the longest-lived self-contained token the key signed, so outstanding tokens expire before their `kid` disappears.
+    - Stored per project with the token configuration, alongside the maximum active-key age, and defaulting to the longest of that project's self-contained token lifetimes.
```
wim07101993: I think it is a proper solution to keep keys around for as long as tokens live. I have questions about feasibility though.

1. According to [this draft ADR](https://github.com/zitadel/nextgen/pull/450) we do not store any info about the tokens, so we don't know what the latest expiration time is.
2. If we would have this info in the database, won't this be an expensive operation? We would have to get the max exp for all tokens of a project
3. What about tokens which live indefinitely (possible for PATs)

docs/adrs/039-signing-key-rotation-and-incident-response.md
```diff
@@ -0,0 +1,228 @@
...
+
+- **Automatic:** a sweep finds projects whose current signing key is older than the project's maximum active-key age and which have no successor queued, and inserts one. Projects that set the age to zero are skipped. The same sweep also covers a project left with no eligible signing key (a new project, or one made keyless by a deletion): it generates a key with `active_from = now()` so signing can resume. A project with automatic rotation disabled (age zero) relies on the admin to create that key.
+- **Manual:** the admin calls the create endpoint. `active_from` defaults to `now() + propagation_window` and may be set later, but not earlier: a key that signs before the propagation window elapses risks an unresolvable `kid`. Moving it also moves the predecessor's `retired_at` to match. Only the emergency path sets `active_from = now()`.
+
+#### Emergency Compromise Handling
+If an active signing key is compromised:
+1. Generate a new signing key pair and insert it with `active_from = now()`, which retires the compromised key at the same instant, so the replacement signs immediately.
+2. Delete the compromised key. It leaves the JWKS and isn't chosen for signing.
```
wim07101993, line 126: > isn't chosen for signing.

It should also not be available for verification either. I guess this is already is so because "delete" would be a hard delete in the database? But we do not want to validate against a key which an attacker might hold.

docs/adrs/035-credential-migration-rotation-and-recovery.md
```diff
@@ -0,0 +1,228 @@
...
+2. The previous KEK stays in configuration as a decrypt-only key, so its private
+   key can still unwrap the DEKs stored under it.
+3. For each project, unwrap the DEK with the old KEK private key, re-wrap it with the
+   new KEK's public key, and update the wrapped DEK in the database.
+4. Remove the old KEK from configuration once every project's DEK has been re-wrapped,
+   since nothing is encrypted under it any longer.
+
+#### Emergency KEK Compromise Handling
+A compromised KEK exposes every DEK it wrapped, and therefore every secret those DEKs
+protected is exposed in plaintext to anyone who also held a copy of the database.
```
wim07101993: Indeed, everyone who holds both a kek and the database can encrypt everything. However, this is a highly unlikely event I think. A more "common" event is that the kek might get compromised, but the database stays untouched. In that case we should only rotate the kek and wrap all deks again. The deks themselves are not compromised since nobody got access to the database.

docs/adrs/035-credential-migration-rotation-and-recovery.md
```diff
@@ -0,0 +1,228 @@
...
+* Recovery is not a guarantee against lockout. A user who loses their recovery codes and their email access has no path back in. We accept that outcome, because the only way to open one is to let support staff restore access to someone who can prove nothing, and an attacker can do that too.
+* Magic-link recovery floors an account's security at the security of its email inbox, because it bypasses every enrolled factor. For an account otherwise protected by a passkey this is a downgrade. Projects that cannot accept that downgrade can disable magic-link recovery ([Section 4](#4-account-recovery)).
+* Emergency KEK compromise handling ([Section 6](#6-encryption-at-rest-key-rotation)) is not a larger routine rotation. It is a full platform credential compromise: every signing key must be rotated, every opaque token revoked, and every third-party secret re-issued by the third party that owns it. Recovery time is therefore partly outside the platform's control.
```
wim07101993: Not necessarily, only if the db and kek are compromised at the same time.

docs/adrs/035-credential-migration-rotation-and-recovery.md
```diff
@@ -0,0 +1,228 @@
...
+1. During an account recovery event, what is the exact policy for invalidating a user's pre-existing active sessions and revoking outstanding refresh tokens? Does recovery revoke every session, or only the ones bound to the lost factor?
+
+2. Should a retired signing key pair be kept as audit evidence past the point it is purgeable ([Section 5](#5-signing-key-lifecycle))? Section 5 already settles when a key becomes deletable: once it leaves the JWKS after its grace period, every self-contained token it signed has expired, so the row can be purged. The open question is only whether to retain the key pair beyond that as rotation/audit evidence, and for how long.
+
+3. Should the KEK be per project as well as the DEK ([Section 6](#6-encryption-at-rest-key-rotation))? A per-project KEK would contain even a KEK compromise to one project, closing the gap a per-project DEK under a global KEK leaves open. But the KEK is provisioned externally (cert-manager, per [ADR 029](029-cryptography-secrets-and-key-lifecycle.md#master-key)), so one KEK per project means provisioning and rotating a key pair per tenant through that external tooling, and coupling tenant onboarding to it.
```
wim07101993: Unfortunately not possible. The KEK would need to be stored somewhere. 
- We cannot store it in the db since it would need to be stored encrypted, which we would need an encryption key for which has the KEK as the root. 
- We cannot store it in config if it needs to be per project.

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: I wrote this based on the existing implementation in Zitadel. I'll revisit this one. Thanks for pointing it out!

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: Nope, just one set. It's not attached to a factor, but to a user.

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: yes, that's the assumption for this point

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: yes, this is the absolute worse scenario when both the kek and the database are comproised.
for just the kek, the usual rotation procedure defined in the section above this one should apply. I could explicitly mention it.

docs/adrs/039-signing-key-rotation-and-incident-response.md
grvijayan: > I guess this is already is so because "delete" would be a hard delete in the database?

yes, it is a hart delete when the key is compromised

> It should also not be available for verification either.

I thought this was implied with "leaves the JWKS". 
I didn't specify that verification would also stop immediately because clients that have a JWKS cache might continue to verify/validate until they refresh

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: let me revisit this after the recent changes to the tokens ADR that you mentioned.

wim07101993: After the latest changes the PATs were removed from the scope. However, the point still stands imo. We will probably add PATs in the future since they are a widely used concept in auth.

grvijayan: Regarding the persistence/removal of keys from the DB: this depends on the lifetime of a **self-contained token**, which is configured per project ([defaulting to 5 mins in the tokens ADR](https://github.com/zitadel/nextgen/pull/450/changes#diff-70442640588acf8e15282d8785ab471f745383ec72d618c47296a5dd1a2a8ec4R23)), so, no DB storage/lookup of expiration would be needed.

I don't see how PATs would influence this as they're supposed to be opaque tokens and their validation would involve decrypting and DB lookup. Isn't this the case?

> We will probably add PATs in the future

yeah, I am planning to frame it similarly in this ADR as well

wim07101993: If a token has a lifetime of 3 years, we would keep a key set around for those entire 3 years. Even if the keys are rotated after 1 year. If a PAT does not have a expiration date, the keys will be stored forever.

I think we need to figure out how we can handle that case. Do we want to keep it around forever?

docs/adrs/035-credential-migration-rotation-and-recovery.md
```diff
@@ -0,0 +1,228 @@
...
+* Each code is strictly single-use.
+* Because each code is single-use, the set depletes over time. When a user runs low on or exhausts their codes, they are prompted to regenerate a fresh set; regenerating invalidates any codes remaining from the old set.
+* A user only has recovery codes if they enrolled an MFA factor. A password-only user relies on ordinary password reset, which is out of scope here.
+
+#### Magic Links
+* A single-use magic link is generated and sent to the user's verified email address.
+* Links enforce a strict, configurable expiration window (e.g., 15 minutes) and are only valid for a single use.
+* Magic-link recovery is a per-project setting and can be disabled. Because it bypasses every enrolled factor and reduces account security to email possession, high-assurance projects (for example under a NIST profile per [ADR 029](029-cryptography-secrets-and-key-lifecycle.md#nist)) can turn it off.
+
+Post-recovery, the user should be prompted to re-register their lost authentication factors (Passkey, TOTP, etc.) to ensure they have at least one valid factor for future logins.
```
livio-a: Unsure about this. We've since a couple of cases in the past, which just want to use magic links as the standard / only factor as alternative to passwords, so basically a username + OTP email flow

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: But will the PATs be self-contained tokens? I thought the plan is to make them opaque (JWE) as well.
As far as I understand, JWEs don't require signing keys. 
Please let me know if I am missing some context here.

grvijayan: > Do we want to keep it around forever?

Also, in case of manual rotation, the keys would be around forever if they don't rotate. Not sure how we should handle it (maybe a product decision?)

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: fixed [here](https://github.com/zitadel/nextgen/pull/454/changes#diff-1b5619bbffa85a13946741be78a4decf49bf0bb43cdebe0786a2d96f2be9b8ecR282)

docs/adrs/035-credential-migration-rotation-and-recovery.md
wim07101993: In the original ADR I stated that PATs would be self contained. Since the application which holds the token needs to be able to see what the user can do offline. 

An opaque token would be indeed a JWE and indeed don't need to be signed, JWTs don't need to be signed either. But both would be signed. For JWTs it is obvious why I think. For JWEs we would do this to keep tokens integer even when one of the keys is compromised. The DEK will ensure nobody can read the token and the signature will ensure nobody can modify the decrypted token.

> Not sure how we should handle it 

Me neither. Product decision is fine by me, but I think we need to write it down in that case. Just to document it.

grvijayan: > For JWEs we would do this to keep tokens integer even when one of the keys is compromised. 

Could you please add this clarification to the tokens ADR? I inferred that it need not be signed and used that as a basis for parts of my ADR. Adding a clarification would avoid the same confusion for others.

docs/adrs/035-credential-migration-rotation-and-recovery.md
```diff
@@ -0,0 +1,405 @@
...
+  uneven browser support, so treat it as a transition aid, not a permanent state.
+- Over that transition window, prompt users to register a fresh passkey under the new `rpId`, then
+  retire reliance on the old one.
+
+### 3. TOTP Enrollment
+
+A user can enroll more than one TOTP secret, the same way they can enroll more than one passkey. The
+secrets are independent, and at verification a submitted code is accepted if it matches any enrolled
+secret for the user. Adding an authenticator is enrollment, and retiring one is a separate, optional
+removal.
```
livio-a: Looks like you were too fast to go with multi device. IMO this is not required and i'm unaware of any other provider or service supporting this. And for a reason, if you need a backup, use another method like recovery code, for mutli device: password managers and other apps support TOTP or use passkeys, which are already designed for this.
Also having two or even more secrets means either iterating over them on a check or requiring the user to pre-select.

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: also, per the diagram in this [ADR](https://github.com/zitadel/nextgen/blob/main/docs/adrs/029-cryptography-secrets-and-key-lifecycle.md), opaque tokens are only encrypted 😕

docs/adrs/035-credential-migration-rotation-and-recovery.md
```diff
@@ -0,0 +1,405 @@
...
+- A single-use magic link is generated and sent to the user's verified email address.
+- Links enforce a strict, configurable expiration window (e.g., 15 minutes) and are only valid for a
+  single use.
+- Magic-link recovery is a per-project setting and can be disabled. Because it bypasses every
+  enrolled factor and reduces account security to email possession, high-assurance projects (for
+  example under a NIST profile per [ADR 029](029-cryptography-secrets-and-key-lifecycle.md#nist))
+  can turn it off.
+
+Post-recovery, the user should be prompted to re-register their lost authentication factors
+(Passkey, TOTP, etc.) to ensure they have at least one valid factor for future logins.
```
livio-a: Almost every (feature) request wrt magic links until now was not about having magic links as recovery but rather as first class auth method / alternative to password-based approach.

docs/adrs/035-credential-migration-rotation-and-recovery.md
```diff
@@ -0,0 +1,405 @@
...
+#### Emergency Compromise Handling
+
+If an active signing key is compromised:
+
+1. Generate a new signing key pair and insert it with `active_from = now()`, which retires the
+   compromised key at the same instant, so the replacement signs immediately.
+2. Delete the compromised key. It leaves the JWKS and isn't chosen for signing.
+3. Revoke the project's sessions, refresh tokens, and PATs (out of scope in the current MVP
+   ([ADR 032](032-token-lifecycle.md#personal-access-tokens))). These opaque tokens are encrypted
+   under the DEK and are independent of the signing key, so the compromise does not touch them.
```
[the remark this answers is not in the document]
grvijayan: will be fixed after the tokens ADR is merged

docs/adrs/035-credential-migration-rotation-and-recovery.md
wim07101993: The diagram does not specify anything about signing. It illustrates how the encryption/decryption hierarchy works. Maybe it is not well enough documented. But it only states things about encryption, not signing. (Except how the signing keys are stored because they need to be encrypted)

grvijayan: it was confusing nevertheless 😅 

let me just ask this: 
so, all tokens (access tokens, session, refresh) are signed (self-contained) / signed + encrypted (opaque) - is this inference correct?
and PAT would also be signed + encrypted, when added in the future?

docs/adrs/035-credential-migration-rotation-and-recovery.md
```diff
@@ -0,0 +1,405 @@
...
+  `retired_at` to match. Only the emergency path sets `active_from = now()`.
+
+#### Emergency Compromise Handling
+
+If an active signing key is compromised:
+
+1. Generate a new signing key pair and insert it with `active_from = now()`, which retires the
+   compromised key at the same instant, so the replacement signs immediately.
+2. Delete the compromised key. It leaves the JWKS and isn't chosen for signing.
+3. Revoke the project's sessions, refresh tokens, and PATs (out of scope in the current MVP
```
livio-a: i'd explicitly make that optional. currently it reads like a must do.

docs/adrs/039-signing-key-rotation-and-incident-response.md
```diff
@@ -0,0 +1,405 @@
...
+3. For each project, unwrap the DEK with the old KEK private key, re-wrap it with the new KEK's
+   public key, and update the wrapped DEK in the database.
+4. Remove the old KEK from configuration once every project's DEK has been re-wrapped, since nothing
+   is encrypted under it any longer.
+
+#### Emergency KEK Compromise Handling
+
+Per the key hierarchy in
+[ADR 029](029-cryptography-secrets-and-key-lifecycle.md#signingencryption-keys), the DEK protects
+four classes of material: signing keys, opaque tokens, third-party secrets, and authenticator
```
livio-a, line 184: Maybe it's too much, but: do we encrypt opaque/JWE tokens with the DEK, or should that be a separate token-encryption key?

The DEK feels like an at-rest key to me — stuff that stays in our DB, not something we put on issued tokens. Reusing it for JWE widens the blast radius, and it couples lifecycles: any DEK rotation then means re-encrypting everything else it protects (signing keys, IdP secrets, TOTP, …), not just dealing with outstanding tokens. I’d rather have a dedicated key for that, stored under the DEK the same way signing keys are, unless 029 meant them to be the same on purpose.

livio-a (changes requested):

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: Thanks for sharing the context.
Would you rather we retained the current design from Zitadel?

docs/adrs/039-signing-key-rotation-and-incident-response.md
grvijayan: > do we encrypt opaque/JWE tokens with the DEK, or should that be a separate token-encryption key?

per 029, it seems to be the same DEK
not sure if there was a discussion about a separate one for token-encryption
cc: @wim07101993

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: sorry, do you mean `Revoke the project's sessions, refresh tokens, and PATs` should be optional?

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: I didn't mean that we use magic-link feature only for recovery, but yeah, we would need a way to discern that this was a `recovery` flow rather than the usual `auth` flow for this to be applicable: `Post-recovery, the user should be prompted to re-register their lost authentication factors`

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: Does this feel more like a product decision to you? 🤔 
If we don't want to strictly enforce factors (other than email OTP), we can do that. We just need to decide what we want to do.

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: related discussion: https://github.com/zitadel/nextgen/pull/454/#discussion_r3579134610

will resolve this to keep the conversation in a single thread

docs/adrs/035-credential-migration-rotation-and-recovery.md
wim07101993: all auth tokens are signed, opaque tokens are also encrypted. 

A pat will not be encrypted I think. However, it could be if the customer wants to.

The ADR you mentioned does not specify anything about token by the way. It only states how we handle encryption. And shows how it would be handled for encrypted tokens (opaque). Another ADR exists for tokens and their lifetimes which also specifies what kinds of tokens we have.

docs/adrs/035-credential-migration-rotation-and-recovery.md
wim07101993: Maybe it can be a policy?

docs/adrs/039-signing-key-rotation-and-incident-response.md
wim07101993: in 029, the conclusion was to use a single DEK for both token-encryption and third party encryption.

Since they don't leave the system, it would be very difficult to leak. If one gets leaked anyway the chances of other keys also being leaked is so big, we should rotate them all anyway imo.

docs/adrs/035-credential-migration-rotation-and-recovery.md
wim07101993: Imho recovery keys are a last resort for when you lost all other methods. Which might be combined with a policy that when one is used, other factors must be renewed as well.

That being said, I get that this might be more complex that what we would like to handle in the mvp.

docs/adrs/039-signing-key-rotation-and-incident-response.md
livio-a: By third party encryption I assume you refer to OIDC secrets, etc, which stay internally yes, but encrypted token like JWT are exposed. That's one of the reasons why we use rotation on the signing keys. IMO encryption is not different in that case and the risk is about the same and would require us to provide rotation and revocation of a encryption key and in that case i would want to prevent rotating the DEK and re-encrypting all it's underlying data.

docs/adrs/035-credential-migration-rotation-and-recovery.md
livio-a: Exactly. The sentence to felt like it's required or very much implied to do. And then in the second part mentions, they're not affected and it's just a precaution. IMO in a real scenario I would even abstain from just revoking them as well (at least if there's not yet some evidence or possibility), since it will just add additional load and make it harder to evaluate what's really compromised. Maybe you even fallback to some PAT as break class if needed.

docs/adrs/039-signing-key-rotation-and-incident-response.md
grvijayan: > provide rotation and revocation of a encryption key and in that case i would want to prevent rotating the DEK and re-encrypting all it's underlying data.

we have the option to rotate the KEK without having to re-encrypt all the underlying data

docs/adrs/035-credential-migration-rotation-and-recovery.md
livio-a: Yes I was also wondering if we should treat recovery as a separate ADR, since this can get very complex rather soon. And as mentioned somewhere factors like recovery codes weaken other, more secure one like passkeys.
I don't think we need to solve that right now and we can also add the base here and add a follow up on defining the user recovery process. Might also then include what should be checked before you can actually reset / request a password reset, add or remove a MFA, etc.

Wrt TOTP I'd definitely prefer a single device / registration.

docs/adrs/035-credential-migration-rotation-and-recovery.md
livio-a: Hmm after commenting https://github.com/zitadel/nextgen/pull/454#discussion_r3587728491 I wonder if it really would make sense to split that process our or defer it to later and just keep the base management and migration path here. What's your opinion?

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: thanks for the clarification.

I'll revert the totp changes

regarding user recovery process: I can create a separate ticket to define an adr for this so we can cover all the scenarios and keep it brief in this one.

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: makes sense
this adr is doing too many things

docs/adrs/035-credential-migration-rotation-and-recovery.md
grvijayan: I split them as two ADRs now.

docs/adrs/039-signing-key-rotation-and-incident-response.md
wim07101993: The rotating the kek would not do anything if dek is compromised. What Livio is saying is that when a dek get's compromised, we would need to re-encrypt all keys in the project. Which is not wrong.

I don't see how that key could get compromised if it never leaves the system and is stored encrypted in the database. But we could add another layer in between. A separate encryption key for opaque tokens. However, that is out of scope for this PR I think. An amendment should be made to ADR029 in that case.

grvijayan: >  when a dek get's compromised, we would need to re-encrypt all keys in the project

yes and this is already covered in this [section](https://github.com/zitadel/nextgen/pull/454/changes#diff-b959bc0ed299867cb7a0654796bce06a69dad3dce3a3c268eed837f6104a9565R194)

> A separate encryption key for opaque tokens. However, that is out of scope for this PR I think. An amendment should be made to ADR029 in that case.

👍

grvijayan: > A separate encryption key for opaque tokens.

now i am wondering that this would also be stored in the database right? has the same risk of being compromised as the global DEK, isn't it? or am i missing something? 🤔

wim07101993: From a db perspective yes. But I think what @livio-a is hinting at is that if someone could break the encryption by figuring out the key from an opaque token, we would have to re-encrypt everything.

livio-a: Yeah — if someone has the DB (and can unwrap), TEK and DEK fall the same way. That’s not what I’m optimizing for.

The case I care about is: the encryption key is also used on material that leaves the system (opaque/JWE). If we ever need to rotate/retire that key because of exposure or crypto concerns, I don’t want that to mean re-encrypting signing keys, IdP secrets, TOTP, etc. With proper GCM use it’s unlikely an attacker recovers the key from tokens alone, but I’d still want to keep the blast radius small.

Fine to leave single-DEK as 029 says for now and open an amendment there. Happy if this ADR just doesn’t harden “opaque tokens = DEK” as the long-term model.

grvijayan: I've added a note to indicated that a dedicated token encryption key could be used in the future (and 029 will be accordingly amended)

grvijayan: created an issue to track this: https://github.com/zitadel/nextgen/issues/559

docs/adrs/039-signing-key-rotation-and-incident-response.md
```diff
@@ -0,0 +1,310 @@
...
+If an active signing key is compromised:
+
+1. Generate a new signing key pair and insert it with `active_from = now()`, which retires the
+   compromised key at the same instant, so the replacement signs immediately.
+2. Delete the compromised key. It leaves the JWKS and isn't chosen for signing.
+3. The project's sessions and refresh tokens may also be revoked. These opaque tokens are signed by
+   the same key but also encrypted under the DEK, so a signing-key compromise alone cannot forge them: an
+   attacker would also need the DEK. Revoking them is a precaution against a broader breach, forcing
+   re-authentication. PATs are out of scope in the
+   current MVP ([ADR 032](032-token-lifecycle.md#personal-access-tokens)), and their exposure depends
```
livio-a: the version / link will need to be updated

livio-a (commented):
We need to align version number once again, but LGTM

docs/adrs/039-signing-key-rotation-and-incident-response.md
grvijayan: Yes, I'll update this after Wim's ADR is merged.

livio-a (approved):

PR #456, opened 2026-07-02
feat: implemented correlation between schemas

wim07101993:
## Summary

- Added a way to correlate multiple schemas using a `userType`.

## Validation

- go test ./...

## Release notes / changeset

- Changeset: .changeset/add-schema-correlation.md - Added a way to correlate schemas over revisions

## Notes

-


api/openapi/endpoints/schemas/list-schemas-response.yaml
```diff
@@ -0,0 +1,13 @@
...
+  required:
+    - id
+    - createdAt
+  properties:
+    id:
+      type: string
+      description: The unique identifier for this user schema.
+    createdAt:
+      type: string
+      format: date
```
vitorbari:
```suggestion
      format: date-time
```
so that we can properly display when it was created

vitorbari (commented):
Should we update the default user schema as part of this PR?
https://github.com/zitadel/nextgen/blob/main/api/openapi/endpoints/schemas/examples/default-human-user-schema.json

internal/storage/database/condition.go
```diff
@@ -250,6 +250,59 @@ func NewTextIgnoreCaseCondition[T Text](col Column, op TextOperation, value T) C
...
+// String implements [Column].
+func (c jsonBFieldColumn) String() string {
+	return "database.jsonBFieldColumn"
+}
+
+// WriteQualified implements [Column].
+func (c jsonBFieldColumn) WriteQualified(builder *StatementBuilder) {
+	builder.WriteString("payload->>")
+	builder.WriteArg(c.field)
+}
```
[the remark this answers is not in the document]
IAM-marco: Valid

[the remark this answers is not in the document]
IAM-marco: What if the json object is not called `payload`?

IAM-marco (changes requested):

wim07101993:
> What if the json object is not called payload?

Yes, copilot also remarked that. I need to make that more flexible. It also said this won't work for spanner. So a solutions is needed there as well.

IAM-marco:
Disregard spanner for the moment. We'll take care of it in due time

api/openapi/endpoints/schemas/examples/default-human-user-schema.json
```diff
@@ -3,6 +3,7 @@
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "metaSchema": "${SERVER_URL}/user-schema.json",
   "$id": "${USER_SCHEMA_URL}",
```
vitorbari, line 5: I think we should remove this as well, since we want to encourage developers to not pre-define an `$id` and let the server define it.

internal/api/integration_test/schema_test.go
```diff
@@ -176,3 +194,115 @@ func TestGetSchema(t *testing.T) {
...
+	project, err := harness.EnsureProjectService(t).Create(t.Context(), nil)
+	require.NoError(t, err)
+
+	client, err := helpers.NewApiClient(harness.EnsureTestServer(t).URL)
+	require.NoError(t, err)
+	client.SetToken(project.ProjectSecret)
+
+	testCases := []struct {
+		name            string
+		userType        string
```
vitorbari: `objectType`

api/openapi/endpoints/schemas/examples/default-human-user-schema.json
wim07101993: I changed it back because a lot of tests use the provided id for fetching the default schema

vitorbari (approved):

IAM-marco (approved):

PR #457, opened 2026-07-02
feat(database): implement storage v2 Spanner dialect

adlerhurst:
## Summary

Implements the storage v2 Spanner dialect with native `cloud.google.com/go/spanner` APIs, plus review/CI follow-ups:

- `queryExecutor.Query` consume-callback API; pool path uses `ReadWriteTransaction` for `INSERT … THEN RETURN`
- Spanner integration tests enabled in CI via `moon run server:test-spanner`
- Spanner `json_schemas.object_type` made nullable (matches Postgres)
- Spanner JSON column binds use `spanner.NullJSON` so ProjectService seed defaults work on the v2 native-client txn bridge
- Schema create passes the payload into Resolve (avoids Spanner commit-time duplicate from HTTP re-fetch/re-insert)
- Review follow-ups: close `Ping`’s `Single()` txn, preserve `domain.Error` / `UnimplementedError` in `wrapError`, rename cursor typo, Spanner stubs return `database.UnimplementedError` mapped by service to `domain.ErrNotImplemented`
- `TestListUsers` is Postgres-only until the user repository supports GoogleSQL

## Validation

```sh
go test ./internal/storage/database/... ./internal/storage/v2/dialect/spanner/... ./internal/service/...
```

CI `full-pr` (including `moon run server:test-spanner`) was green on the prior revision; this follow-up is covered by the unit tests above.

## Release notes / changeset

No changeset required — internal storage dialect / migration / CI wiring.

## Notes

- Flow-definition v2 CRUD beyond the create-path bridge remains deferred (stubs return `UnimplementedError`).
- User EAV repository remains Postgres-oriented; Spanner user list e2e is deferred with the repository work.

<div><a href="https://cursor.com/agents/bc-3156eb1e-18a7-4058-92e2-60033c05bda2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3156eb1e-18a7-4058-92e2-60033c05bda2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


internal/storage/v2/dialect/spanner/executor.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func (e txnExecutor) QueryRow(ctx context.Context, sql string, args ...any) rowScanner {
+	return &txnQueryRow{
+		ctx:  ctx,
+		txn:  e.txn,
+		sql:  sql,
+		args: args,
+	}
+}
```
adlerhurst: This code look ways too complex. IMO the queryExecutor interface can define the native spanner methods like `Query`, `Update`, `Read`, `ReadRow` so that the statements can use these methods instead of doing all this wrapping.

adlerhurst (commented):
The code currently heavily tries to mimic postgresql. The goal is to implement the dialect using native spanner tooling and not having a 1:1 clone of the postgres dialect implementation

internal/storage/v2/dialect/spanner/placeholders.go
```diff
@@ -0,0 +1,36 @@
...
+			}
+			continue
+		}
+		b.WriteByte(ch)
+		i++
+	}
+	return b.String()
+}
+
+func isDigit(b byte) bool { return b >= '0' && b <= '9' }
```
adlerhurst: What is this function used for? the spanner dialect should never care about postgresql style placeholders?

IAM-marco:
**Claude:** 

Context for enabling the `spanner_integration` test suite, found while wiring the postgres integration tests into CI (https://github.com/zitadel/nextgen/pull/461).

**What currently blocks `spanner_integration`:**
1. **Compile failure in the API harness** — `internal/api/integration_test/helpers/project.go` calls `h.ServiceDB(t)`, but `ServiceDB` is defined only in `helpers/pg_database.go` (`//go:build !spanner_integration`); `helpers/spanner_database.go` has no equivalent. So `go test -tags spanner_integration ./internal/api/integration_test/...` fails to build with `h.ServiceDB undefined`.
2. **Unimplemented v2 spanner statements** (all `panic("unimplemented")`): `internal/storage/v2/dialect/spanner/statement_project.go` (CreateProject/GetProjectByID/DeleteProjectByID/ListProjects), `statement_flow_definition.go` (same shape), and `tx.go` (`Transaction.Exec/Query/QueryRow`).

**To turn the suite green:**
- Implement the v2 spanner project + flow-definition statements. Note `queryExecutor` is `any` here (`statement.go`) and `newStatements` is built from both `*spanner.Client` (pool) and `*spanner.ReadWriteTransaction` (tx), so the impls must handle both. Use the schema at `internal/storage/database/dialect/spanner/migration/sql/000001_projects.sql` and the postgres impls (`internal/storage/v2/dialect/postgres/project.go`) as the behavioral contract — e.g. `CreateProject` must populate `CreatedAt`/`UpdatedAt`, and `GetProjectByID` should map not-found to `internal/storage/database`.`NoRowFoundError` to match the postgres dialect.
- Add a spanner `ServiceDB` helper in `helpers/spanner_database.go` (mirroring `pg_database.go`) that builds a v2 spanner pool, so the API harness compiles and runs under the tag. `spannerv2.Dialect{Database: dsn}.Connect(ctx)` builds a `*Client` from a DSN; an exported constructor from an existing `*spanner.Client` may be cleaner if you want to reuse the harness client.

**CI wiring (once green):** re-add the deferred step in `.github/workflows/ci.yml` — there's a placeholder comment pointing at this PR:
```yaml
- name: Run Go integration tests (spanner)
  if: ${{ steps.ci-mode.outputs.mode == 'full' }}
  run: moon run server:test-spanner
```

**Storage-level v2 test pattern to mirror:** the same branch adds `internal/storage/v2/dialect/postgres/project_test.go` — a `postgres_integration` test that stands up a **pure v2 pool**: start the DB via `dbtest.Postgres`, recover the DSN, migrate with the pool-independent `migration.Migrate(ctx, *sql.DB)`, then `DecodeConfig(dsn)` + `Connect` a v2 pool and exercise the statements. The spanner analog would use `dbtest.Spanner` (emulator + `SPANNER_EMULATOR_HOST`), DSN `projects/test-project/instances/test-instance/databases/test-database`, the spanner `migration.Migrate` on a go-sql-spanner `*sql.DB`, and `spannerv2.Dialect{...}.Connect`.

internal/storage/v2/dialect/spanner/tx.go
```diff
@@ -2,42 +2,103 @@ package spanner
...
+type transaction struct {
+	txn *spanner.ReadWriteTransaction
 	statements
 }
 
 // Exec implements [database.QueryExecutor].
-func (t *Transaction) Exec(ctx context.Context, stmt string, args ...any) (int64, error) {
-	panic("unimplemented")
+func (tx transaction) Exec(ctx context.Context, stmt string, args ...any) (int64, error) {
+	s := buildStatementFromPostgresSQL(stmt, args).statement()
```
IAM-marco: Claude:

`args` is forwarded to the variadic `buildStatementFromPostgresSQL(sql string, args ...any)` without spreading, so the whole `[]any` is packed into a single variadic element. `buildStatement` then binds the entire slice to `@p1` and never creates `@p2…@pN`, so any bridged query with placeholders binds the wrong value (a slice) and leaves the remaining params unbound — a runtime query failure. Use `args...`.

Same issue in `Query` (L27) and `QueryRow` (L34).

https://github.com/zitadel/nextgen/blob/8325e3ad5bc1bcd989ac94d4bd4e099b19083965/internal/storage/v2/dialect/spanner/tx.go#L19-L21

internal/storage/v2/dialect/spanner/tx.go
```diff
@@ -2,42 +2,103 @@ package spanner
...
+	n, err := tx.txn.Update(ctx, s)
+	return n, wrapError(err)
 }
 
 // Query implements [database.QueryExecutor].
-func (t *Transaction) Query(ctx context.Context, stmt string, args ...any) (dbold.Rows, error) {
-	panic("unimplemented")
+func (tx transaction) Query(ctx context.Context, stmt string, args ...any) (dbold.Rows, error) {
+	s := buildStatementFromPostgresSQL(stmt, args).statement()
```
IAM-marco: Same missing spread as L20: pass `args...` so each value maps to its own `@pN` parameter.

internal/storage/v2/dialect/spanner/tx.go
```diff
@@ -2,42 +2,103 @@ package spanner
...
+	iter := tx.txn.Query(ctx, s)
+	return &bridgeRows{iter: iter}, nil
 }
 
 // QueryRow implements [database.QueryExecutor].
-func (t *Transaction) QueryRow(ctx context.Context, stmt string, args ...any) dbold.Row {
-	panic("unimplemented")
+func (tx transaction) QueryRow(ctx context.Context, stmt string, args ...any) dbold.Row {
+	s := buildStatementFromPostgresSQL(stmt, args).statement()
```
IAM-marco: Same missing spread as L20: pass `args...` so each value maps to its own `@pN` parameter.

internal/storage/v2/dialect/spanner/client.go
```diff
@@ -28,27 +28,25 @@ func (c *Client) Close(ctx context.Context) error {
...
-		return nil
-	})
+	return wrapError(iter.Do(func(*spanner.Row) error { return nil }))
 }
 
-// Transaction implements [database.Pool].
-func (c Client) Transaction(ctx context.Context, fn func(ctx context.Context, tx service.AllStatements) error) error {
+// Transaction implements [service.Transactioner].
+func (c Client) Transaction(ctx context.Context, fn func(ctx context.Context, tx service.Statementer[service.AllStatements]) error) error {
 	_, err := c.client.ReadWriteTransaction(ctx, func(ctx context.Context, rwt *spanner.ReadWriteTransaction) error {
```
IAM-marco, line 44: I read the docs of ReadWriteTransaction:

> To limit the number of retries, set a deadline on the Context rather than using a fixed limit on the number of attempts.

This piece of information is buried in here and not immediately visible to people calling `Client.Transaction()`. If they pass a context without deadline, this could potentially stall the system, don't you think?

If so, I propose to:

- Either surface these bits of info into the docs of `Transaction`, i.e. we should explicitly state that a context with deadline needs to be passed
- Or put checks in here on the context to make sure the input context has a deadline set.

internal/storage/v2/dialect/spanner/project.go
IAM-marco, line 1: Please see https://github.com/zitadel/nextgen/pull/457#issuecomment-4891952229

internal/storage/v2/dialect/spanner/pattern.go
IAM-marco, line 1: This file is identical to the one under the `postgres` package. Should go under `all`

IAM-marco (changes requested):

internal/storage/v2/dialect/spanner/project.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// CreateProject implements [service.ProjectStatements].
+func (ps projectStatements) CreateProject(ctx context.Context, project *domain.Project) error {
+	previewOrigins, err := encodePreviewOrigins(project.PreviewOrigins)
+	if err != nil {
+		return wrapError(err)
+	}
+	stmt := buildStatement(createProjectStmt, project.ID, project.ProjectSecret, project.PreviewSecret, previewOrigins).statement()
+	_, err = collectOneRow(ps.db.Query(ctx, stmt), func(row *spanner.Row) (struct{}, error) {
```
IAM-marco: `createProjectStmt` is `INSERT … THEN RETURN` (DML), but it is executed via `ps.db.Query`, which for the pool executor runs on `c.client.Single()` — a read-only single-use transaction (`client.Query`, [db.go#L23-L24](https://github.com/zitadel/nextgen/blob/7726d0105b25e70e2ed6aaf43a5227c1a153d78a/internal/storage/v2/dialect/spanner/db.go#L22-L25)). Spanner rejects DML in read-only transactions, so `CreateProject` fails at runtime on the pool path. `DeleteProjectByID` already uses the write path (`ps.db.Update`, which wraps a `ReadWriteTransaction`); the INSERT-with-`THEN RETURN` needs the same — run it inside a `ReadWriteTransaction` and read the returned row from that transaction.

This isn't caught because the covering test [`TestProjectStatements_CRUD`](https://github.com/zitadel/nextgen/blob/7726d0105b25e70e2ed6aaf43a5227c1a153d78a/internal/storage/v2/dialect/spanner/project_test.go#L40-L48) drives exactly this pool path (`pool.(*Client).Statements()` → `CreateProject`), but the `spanner_integration` suite is not enabled in CI (`.github/workflows/ci.yml` L65 placeholder; deferred per cursor[bot] status in r3595564156).

Can we enable the spanner integration tests on the CI?

https://github.com/zitadel/nextgen/blob/7726d0105b25e70e2ed6aaf43a5227c1a153d78a/internal/storage/v2/dialect/spanner/project.go#L41-L45

internal/storage/v2/dialect/spanner/compiler.go
```diff
@@ -0,0 +1,252 @@
...
+
+type statementCompiler struct {
+	strings.Builder
+	args []any
+}
+
+func (c *statementCompiler) Reset() {
+	c.Builder.Reset()
+	c.args = nil
+}
```
IAM-marco, line 20: Could test

IAM-marco (changes requested):
My main gripe is about the spanner tests being disabled

internal/service/schema_service.go
```diff
@@ -81,13 +81,19 @@ func (s *SchemaService) CreateSchema(ctx context.Context, input CreateSchemaInpu
...
+	_, err = s.schemaResolver.Resolve(ctx, tx, input.ProjectID, model.URL, model.Schema)
 	if err != nil {
 		return nil, domain.ErrInternal(err).WithMessage("failed to resolve schema when creating")
 	}
 
 	err = tx.Commit(ctx)
 	if err != nil {
+		if _, ok := errors.AsType[*database.IntegrityViolationError](err); ok {
+			return nil, domain.ErrJSONSchemaAlreadyExists().WithParent(err)
+		}
```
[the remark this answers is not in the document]
IAM-marco: drunk

internal/storage/v2/dialect/spanner/db.go
```diff
@@ -0,0 +1,154 @@
...
+type client struct {
+	client *spanner.Client
+}
+
+func newClientDB(c *spanner.Client) queryExecutor {
+	return client{client: c}
+}
+
+func (c client) Query(ctx context.Context, stmt spanner.Statement, consume func(*spanner.RowIterator) error) error {
+	_, err := c.client.ReadWriteTransaction(ctx, func(ctx context.Context, txn *spanner.ReadWriteTransaction) error {
```
IAM-marco: **Claude:**

This correctly fixes `CreateProject`'s `INSERT … THEN RETURN`, but `client.Query` now wraps *every* query in a read-write transaction, and the other caller is a plain `SELECT`: [`ListProjects`](https://github.com/zitadel/nextgen/blob/9a9ebae415a0968c07d631f45c48f933d96dd356/internal/storage/v2/dialect/spanner/project.go#L79-L84). A Spanner read-write transaction takes shared read locks for the duration of the scan, needs a Commit RPC, and is subject to abort/retry — costs a pure read shouldn't pay. Single-use read-only transactions exist for this, and it's what this path used before.

It's also inconsistent within the file: `GetProjectByID` is an equally pure read and correctly stays read-only via `ReadRow` → `c.client.Single()`.

Consider splitting the executor's read path from the DML-with-returning path: a `Single()`-backed method for `ListProjects`, and the read-write wrapper only for `CreateProject`. `tx.Query` is already correct and unaffected, so this is only about the pool-level executor.

https://github.com/zitadel/nextgen/blob/9a9ebae415a0968c07d631f45c48f933d96dd356/internal/storage/v2/dialect/spanner/db.go#L26-L31

IAM-marco (changes requested):
Thanks for the changes 👍 . I think there's still a valid change request from Claude. Let me know what you think

internal/storage/v2/dialect/spanner/db.go
adlerhurst: I have now split the `Query`-method into `Query` for reads and `Write` for writing and reading data. I think this is the cleanest approach.

IAM-marco (approved):

PR #459, opened 2026-07-03
docs: add wide events ADRs for audit stream and events API

### LINKED ISSUE #350: [ARCH-ADR] Define Architecture for Audit Logging, Observability, and Compliance
**Context & Risk:**

Epic #249 assumes observability and compliance requirements that the `nextgen` POC does not fully satisfy. We need an ADR to specify audit logging, PII sanitization, retention policies, immutable append-only storage options, and streaming to external compliance systems (SIEM/event pipeline). This ADR will define what to log (auth events, admin actions, token operations), how to protect logs (PII redaction, integrity), and how to expose logs for incident response and regulatory review.

**Reference Material:**

- Epic #249: https://github.com/zitadel/nextgen/issues/249
- Current ADRs (search for audit/log): https://github.com/zitadel/nextgen/tree/main/docs/adrs

- https://github.com/zitadel/oxidel/blob/main/docs/adr/050-audit-and-audit-log-streaming.md

**Acceptance Criteria:**
- [ ] Inventory current events logged by nextgen and gaps vs. compliance needs.
- [ ] Define an append-only audit event model and storage recommendations (e.g., write-once S3, dedicated audit DB).
- [ ] Define PII sanitization rules, retention policies, and export/streaming interfaces.
- [ ] Draft ADR and review with Security and Compliance.
- [ ] Merge ADR into `nextgen` before implementation.

adlerhurst:
## Summary

Docs-only ADRs for nextgen audit logging (relational SoT + semantic wide events):

- **[ADR 048](docs/adrs/048-wide-events-internal-audit-primitive.md)** — internal `events` table, categories, Path A (batched request events) + Path B (`AllStatements` / `InsertEvent`), deny-by-default PII + `x-audit`
- **[ADR 049](docs/adrs/049-events-api-retention-export.md)** — unified `GET /events`, time-only retention, first-class managed sinks, project-scoped get

Also updates ADR index, ADR 028 Related, resource-map, api README, and `user-schema.md`.

### Review decisions (latest round)

- **Pre-claim (a):** emit/store always; gate list/get/shipper until claim (ADR 046)
- **No `resource_scope_index` for events;** `GET /events/{id}` is project-scoped
- **Retention:** time-only (not sink-gated); `events_purged_undelivered` metric when aging out undelivered rows
- **Sinks:** first-class CRUD, managed `sink_…` ids; ≤1 deployment stdout + ≤1 deployment webhook + ≤1 project webhook (additive)
- **Path A:** bounded buffer, drop new enqueue + metric, retry/backoff, recompute `wait`
- **`x-sensitive`:** kept complementary to `x-audit` (still used by OpenAPI/console/config)

**Note:** [#710](https://github.com/zitadel/nextgen/pull/710) also claims ADR 048 (Dev Inbox) — number collision if both merge; coordinate renumber.

## Validation

- Docs-only; no `moon ci` / Go tests run
- Grep: no sink-gated purge; pre-claim store+visibility; events not indexed in RSI

## Release notes / changeset

Docs-only — **no changeset** per [`.changeset/README.md` decision table](.changeset/README.md#decision-table).

## Notes

PR title remains conventional `docs:` (allowed by `.github/semantic.yml`).

closes https://github.com/zitadel/nextgen/issues/350

<div><a href="https://cursor.com/agents/bc-ba77f572-bfcf-4606-86d6-51ee221d365e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ba77f572-bfcf-4606-86d6-51ee221d365e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


docs/adrs/029-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,382 @@
...
+remain until ported per [ADR 028](028-storage-v2-statements-and-dialects.md).
+There is **no dual emission** during migration — entities emit events only after
+their v2 statement port lands.
+
+**Pre-claim policy:** Per [claim-flow](../design/platform/claim-flow.md), a
+project begins emitting auditable events only after claim. Pre-claim activity is
+not retroactively audited. This is product policy, not a technical limitation of
+the storage model.
+
+This ADR adapts the wide-event model from oxidel ADR-023 for nextgen tenancy
```
adlerhurst: link the adr using markdown links

docs/adrs/029-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,382 @@
...
+common queries.
+
+```sql
+CREATE TABLE zitadel_nextgen.events (
+    -- Identity
+    project_id          TEXT        NOT NULL
+    , id                BIGINT      GENERATED ALWAYS AS IDENTITY
+    , event_type        TEXT        NOT NULL       -- 'user.deactivated', 'request.api', etc.
+    , category          TEXT        NOT NULL       -- 'request', 'auth', 'session', 'admin', 'entity', 'signal'
+    , sequence          BIGINT      NOT NULL       -- per-project monotonic ordering
```
adlerhurst: This will kill performance. I don't think that we need a sequence.

docs/adrs/029-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,382 @@
...
+    , created_at        TIMESTAMPTZ NOT NULL DEFAULT now()
+
+    -- WHO
+    , actor_id          TEXT
+    , actor_type        TEXT                    -- 'human', 'service', 'system'
+
+    -- WHAT
+    , aggregate_id      TEXT
+    , aggregate_type    TEXT                    -- 'user', 'session', 'project', etc.
+    , resource_type     TEXT
```
adlerhurst: I am not sure if we need it in that detail. nextgen does not know about aggregates. What if we instead use a entity_type and entity_id column instead?

docs/adrs/029-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,382 @@
...
+    , client_id         TEXT        NOT NULL DEFAULT ''
+    , token_id          TEXT        NOT NULL DEFAULT ''
+    , delegation_type   TEXT        NOT NULL DEFAULT ''   -- 'direct', 'delegated', 'pat_shared', 'exchanged'
+
+    -- WHERE (device context)
+    , fingerprint       TEXT        NOT NULL DEFAULT ''
+
+    -- SDK (informational only)
+    , sdk_name          TEXT        NOT NULL DEFAULT ''
+    , sdk_version       TEXT        NOT NULL DEFAULT ''
```
adlerhurst: Not sure if we need to store these information because it feels more like telemetry (logging or tracing) than worth storing on each event.

docs/adrs/029-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,382 @@
...
+    , request_id        TEXT
+    , session_id        TEXT
+    , flow_id           TEXT
+
+    -- Arbitrary data
+    , payload           JSONB       NOT NULL DEFAULT '{}'
+    , metadata          JSONB       NOT NULL DEFAULT '{}'
+
+    -- Export lifecycle (see ADR 030)
+    , shipped_at        TIMESTAMPTZ
```
adlerhurst: What if there are multiple subscribers?

docs/adrs/029-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,382 @@
...
+**Indexes** (define when query paths are concrete):
+
+- `(project_id, created_at, id)` — keyset list pagination ([ADR 027](027-cursor-based-pagination.md))
+- Partial indexes on high-cardinality filter columns: `category`, `actor_id`,
+  `client_id`, `session_id`, `request_id`, `event_type`
+
+**Per-project sequence:** Assign `sequence` inside the same transaction as the
+event INSERT (e.g. `SELECT COALESCE(MAX(sequence), 0) + 1 FROM events WHERE
+project_id = $1 FOR UPDATE` or a dedicated counter row). Sequence gives stable
+ordering within a project independent of clock skew on `created_at`.
```
adlerhurst: IMO having a sequence will kill performance because all calls which insert events get serial instead of parallel per project.

docs/adrs/030-events-api-retention-export.md
```diff
@@ -0,0 +1,289 @@
...
+### 3. External export (`shipped_at`)
+
+A background **shipper** process delivers events to configured external sinks
+before retention deletes them.
+
+**Shipper loop:**
+
+1. Poll: `SELECT ... FROM events WHERE shipped_at IS NULL ORDER BY created_at, id LIMIT $batch_size`
+2. Deliver batch to configured sink.
+3. On successful delivery: `UPDATE events SET shipped_at = now() WHERE (project_id, id) IN (...)`
```
adlerhurst: As asked before. How does the sink handle multiple subscribers?

docs/adrs/030-events-api-retention-export.md
```diff
@@ -0,0 +1,289 @@
...
+        url: https://siem.example.com/zitadel/events
+        headers:
+          Authorization: "Bearer ${SIEM_TOKEN}"
+      - type: stdout
+```
+
+### 4. Threat detection integration pattern
+
+The server **emits facts**; policy execution stays **external** in v1. There is
+no built-in rules engine.
```
adlerhurst: It's not part of this ADR to define where the rules engine will live.

docs/adrs/049-events-api-retention-export.md
```diff
@@ -0,0 +1,296 @@
...
+
+```sql
+CREATE TABLE zitadel_nextgen.event_deliveries (
+    project_id      TEXT        NOT NULL
+    , event_id      BIGINT      NOT NULL
+    , sink_id       TEXT        NOT NULL   -- config name, e.g. "siem", "local-collector"
+    , delivered_at  TIMESTAMPTZ NOT NULL DEFAULT now()
+    , PRIMARY KEY (project_id, event_id, sink_id)
+    , FOREIGN KEY (project_id, event_id)
+        REFERENCES zitadel_nextgen.events (project_id, id)
```
fforootd, line 260: This retention design needs one more detail before it is implementable: the purge deletes from `events`, but every purge-eligible event has at least one `event_deliveries` child row, and this FK does not cascade. As written, the retention job will fail with a foreign-key violation. Please either add `ON DELETE CASCADE` here or make the retention algorithm delete matching `event_deliveries` rows before deleting `events`.

docs/adrs/030-events-api-retention-export.md
```diff
@@ -0,0 +1,296 @@
...
+  [ADR 027](027-cursor-based-pagination.md). No total count. List omits
+  `deliveries` to keep payloads small.
+
+**Pagination:** Keyset cursor on `(created_at, id)` within `project_id`,
+consistent with [ADR 027](027-cursor-based-pagination.md) and v2 storage list
+options.
+
+**Permissions:** Callers require an `events.read` permission (exact RBAC shape
+TBD — reference platform authz when specified). Team-scoped credentials may
+restrict list results to events where `actor_id` or affected resources belong
```
fforootd: This access rule is not safe enough for team-scoped credentials. `actor_id` and affected resources are not immutable team-scope facts: users/resources can move teams, and some events can involve cross-team actors or resources. To avoid leaking historical events across teams, the event row should carry the immutable authorization scope captured at emit time, for example `team_id`/`scope_id`/resource owner context, and filtering should use that stored scope instead of recomputing membership later.

docs/adrs/029-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,376 @@
...
+- Optional `X-SDK-Name` and `X-SDK-Version` headers may be read for operational
+  logging ([`zlog`](../../internal/instrumentation/zlog/)) and OTEL Tier 3
+  export — they are **not persisted** on audit event rows.
+- After the handler completes, if the request was **authenticated**, insert one
+  `category=request` wide event containing: `operation_id`, HTTP method, route
+  template, status code, duration, and actor/delegation dimensions from context.
+- Response header `X-Request-Id` echoes the assigned `request_id`.
+
+Request events use a **separate INSERT** outside the handler's entity
+transaction (fire-and-forget after response). Domain mutations co-locate their
```
fforootd: For request-wide audit events, `fire-and-forget after response` can silently lose the only audit record for read/API access if the process exits, the DB insert fails, or shutdown races the background work. Since these events are part of the audit trail, please specify a durable path: either persist synchronously before completing the request, or enqueue into a durable outbox/queue with retry and shutdown draining. In-memory best-effort batching is fine for operational telemetry, but risky as the audit primitive.

docs/adrs/030-events-api-retention-export.md
fforootd: Related to the previous point: `GET /events/{id}` also needs a resolvable authorization scope. With only the path ID, the server cannot know which project/team scope to authorize before looking up the event. Please either include `project_id` in the get route/query, require the event ID to be globally scoped and lookup-safe, or define a prefetch pattern that cannot leak existence/timing across scopes.

docs/adrs/029-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,376 @@
...
+```
+
+#### Path A — Request-wide events (authenticated API)
+
+Extend the existing request-ID middleware
+([`internal/api/middleware/request_id.go`](../../internal/api/middleware/request_id.go))
+into a **RequestContext** middleware:
+
+- Extract or generate `request_id` (W3C traceparent-compatible 128-bit hex).
+- Stash optional `session_id`, `flow_id`, and `fingerprint` from request headers
```
fforootd: Please avoid trusting caller-provided headers for forensic correlation fields. A client can forge `session_id`, `flow_id`, or `fingerprint` and stitch unrelated activity into another timeline. These should come from server-side session/flow/device state after authentication where possible; if any client hint is retained, it should be explicitly marked as untrusted metadata rather than a primary audit dimension.

docs/adrs/048-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,376 @@
...
+
+User mutations through the header/data/registry model
+([ADR 008](008-users-eav-store.md)) require special emit rules when ported to v2
+statements:
+
+| Operation | Event | Payload |
+|-----------|-------|---------|
+| Create user | `user.created` | `user_id`, schema ref — no attribute values |
+| Patch attributes | `user.updated` | `changed_keys[]` only |
+| Unique violation | `user.create.failed` | key name — not value |
```
fforootd, line 398: This failure event needs a separate emission path. In Postgres, a uniqueness violation aborts the current transaction until rollback, so the same transaction cannot simply continue and insert `user.create.failed`. Please specify savepoints, preflight uniqueness checks, or a separate failure-event transaction/outbox so the audit behavior is actually achievable.

docs/design/api/README.md
```diff
@@ -55,6 +55,8 @@ for this design PR live in:
 - [`../flowengine/api/flow-api.yaml`](../flowengine/api/flow-api.yaml)
 - [`../flowengine/api/session-api.yaml`](../flowengine/api/session-api.yaml)
 
-Specs for auth_attempts, flat api_keys, events, audit_events, imports, and capabilities are not yet written.
+Specs for auth_attempts, flat api_keys, imports, and capabilities are not yet written.
+Events are partially specified in [ADR 030](../adrs/030-events-api-retention-export.md);
```
fforootd: This relative link is one directory short. From `docs/design/api/README.md`, `../adrs/...` resolves to `docs/design/adrs/...`, which does not exist. Please change it to `../../adrs/030-events-api-retention-export.md`.

docs/design/flowengine/user-schema.md
```diff
@@ -13,12 +13,17 @@ What matters here is the contract: which schema annotations exist, how the flow
...
+| `x-audit: true` | Field | Audit emitter | Field value may appear in audit event payloads (allowlist; deny-by-default) |
 | `x-editable: true` | Field | Flow Engine | Field appears in profiling / self-service flows |
 | `x-unique: "project"` | Field | Flow Engine | Server validates uniqueness on form submit (per-project scope) |
 | `x-claim: "claims.email"` | Field | Flow Engine | Maps to SSO/OIDC claim for auto-population |
 | `x-auth-methods` | Schema | Policy Engine | Which auth methods this user type supports (narrows what policy can require) |
 
+Audit event payloads use a **deny-by-default** PII policy: user attribute values
+are omitted unless a field is explicitly marked with `x-audit`. See
+[ADR 029](../../adrs/029-wide-events-internal-audit-primitive.md) §8. The older
+`x-sensitive` annotation is not used for audit redaction.
```
fforootd: This update correctly separates audit allowlisting from `x-sensitive`, but it now makes the annotation table look like `x-sensitive` no longer exists. It is still part of the OpenAPI user-property schema and the CLI presets still emit it for phone/password fields. Please keep `x-sensitive` documented for its non-audit meaning, then add the clarification that it is deprecated/ignored only for audit redaction and `x-audit` is the audit allowlist.

docs/adrs/030-events-api-retention-export.md
```diff
@@ -0,0 +1,296 @@
...
+- Configurable per deployment via server config (suggested default: **30 days**).
+- Background job deletes event rows only when **every enabled sink** has a
+  delivery record and the retention window has elapsed:
+
+```sql
+DELETE FROM zitadel_nextgen.events e
+WHERE e.project_id = $1
+  AND e.created_at < now() - $retention_interval
+  AND NOT EXISTS (
+    SELECT 1 FROM configured_sinks s
```
fforootd: This example reads like executable SQL, but `configured_sinks` is described below as an in-memory config set, not a relation. Please either label this as pseudo-SQL or rewrite it with an executable shape such as a generated `VALUES`/CTE list of enabled sink IDs.

docs/adrs/029-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,376 @@
...
+);
+```
+
+Per-sink export delivery is tracked in a separate `event_deliveries` table; see
+[ADR 030 §3](030-events-api-retention-export.md).
+
+**Indexes** (define when query paths are concrete):
+
+- `(project_id, created_at, id)` — keyset list pagination ([ADR 027](027-cursor-based-pagination.md))
+- Partial indexes on high-cardinality filter columns: `category`, `actor_id`,
```
fforootd: The index guidance should be tightened before this becomes migration guidance. `category` is a small enum, and partial indexes only help for sparse predicates; several listed columns are `NOT NULL DEFAULT ''`, so a partial index would not filter much unless the predicate excludes the empty value. Please describe the concrete intended predicates/indexes, or keep this at the level of `add indexes when query paths are proven`.

docs/design/flowengine/user-schema.md
```diff
@@ -13,12 +13,20 @@ What matters here is the contract: which schema annotations exist, how the flow
...
 | `x-editable: true` | Field | Flow Engine | Field appears in profiling / self-service flows |
 | `x-unique: "project"` | Field | Flow Engine | Server validates uniqueness on form submit (per-project scope) |
 | `x-claim: "claims.email"` | Field | Flow Engine | Maps to SSO/OIDC claim for auto-population |
 | `x-auth-methods` | Schema | Policy Engine | Which auth methods this user type supports (narrows what policy can require) |
 
+Audit event payloads use a **deny-by-default** PII policy: user attribute values
+are omitted unless a field is explicitly marked with `x-audit`. See
+[ADR 029](../../adrs/029-wide-events-internal-audit-primitive.md) §8.
+`x-sensitive` remains in the OpenAPI user-property schema and CLI presets
+(e.g. phone/password); it is **deprecated only for audit redaction**. Audit
```
adlerhurst: The project is still in alpha therefore we don't need to deprecate the audit redaction.

docs/adrs/029-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,443 @@
...
+  export — they are **not persisted** on audit event rows.
+- Client headers named `session_id`, `flow_id`, or `fingerprint`, if accepted at
+  all, go into `metadata.client_hints` as untrusted hints only.
+- After the handler completes, if the request was **authenticated**, insert one
+  `category=request` wide event containing: `operation_id`, HTTP method, route
+  template, status code, duration, and actor/delegation dimensions from context.
+- Response header `X-Request-Id` echoes the assigned `request_id`.
+
+**Durability:** Request-wide events are inserted **synchronously** before the
+HTTP response is flushed to the client. The insert uses a **separate DB
```
adlerhurst: this could cause high load on the database if we flush on each api call. I think the tradeoff of batching events and either flush after duration or x amount of events would be fine. 
This might introduce a new field in the events table:
- `created_at` always `now()` the time the event got written to the database
- `occurred_at`: now()- duration since the request hit the api. (meaning the go server uses `time.Now()` and `time.Since` to calculate the duration the event was waiting until batch flush)

docs/adrs/029-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,443 @@
...
+
+OpenTelemetry distributed tracing solves **latency debugging across services**,
+not business-dimension queries. Operational logs via [`zlog`](../../internal/instrumentation/zlog/)
+are ephemeral and not tenant-queryable. We need a first-class, append-only,
+queryable audit stream stored in the database.
+
+**Scope boundary:** Event emission is defined for **v2 statements only**
+([`internal/storage/v2/`](../../internal/storage/v2/),
+[`internal/service/statement.go`](../../internal/service/statement.go)). v1
+repositories under [`internal/storage/database/repository/`](../../internal/storage/database/repository/)
```
adlerhurst: recent refactorings makes this comment outdated.

docs/adrs/049-events-api-retention-export.md
```diff
@@ -0,0 +1,336 @@
...
+
+**Immutability scope:** Event content is immutable for the duration rows remain
+in the database. Retention purge is an explicit, scheduled operation — not an
+application UPDATE that alters event content.
+
+Link to [ADR 024](024-user-team-lifecycle-ownership.md): user/team lifecycle
+purge policies are separate from event retention; this ADR does not define
+user-data purge windows.
+
+### 3. External export (per-sink delivery)
```
adlerhurst, line 214: I think we should re-discuss the delivery options. It might be enough to have a small set of sinks in the beginning (stdout and one webhook per project)

docs/adrs/048-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,503 @@
...
+([ADR 028](028-storage-v2-statements-and-dialects.md)). Product persistence is
+fully on `AllStatements`. Mutating statement methods that need an audit record
+call `InsertEvent` in the same transaction once `EventStatements` exists.
+Operational request middleware logs ([ADR 030](030-error-model-mapping-and-reporting.md)
+MaskConfig / `zlog`) are **not** the audit source of truth.
+
+**Pre-claim policy:** Per [ADR 046](046-claim-lifecycle-v2.md) (and
+[claim-flow](../design/platform/claim-flow.md)), a project begins emitting
+auditable events only after claim. Pre-claim activity is not retroactively
+audited. This is product policy, not a technical limitation of the storage model.
```
adlerhurst: We should discuss this part. IMO we could gain important details if we still store these events.

adlerhurst: @fforootd whats your opinion?

docs/design/flowengine/user-schema.md
```diff
@@ -13,12 +13,20 @@ What matters here is the contract: which schema annotations exist, how the flow
...
+| `x-audit: true` | Field | Audit emitter | Field value may appear in audit event payloads (allowlist; deny-by-default) |
 | `x-editable: true` | Field | Flow Engine | Field appears in profiling / self-service flows |
 | `x-unique: "project"` | Field | Flow Engine | Server validates uniqueness on form submit (per-project scope) |
 | `x-claim: "claims.email"` | Field | Flow Engine | Maps to SSO/OIDC claim for auto-population |
 | `x-auth-methods` | Schema | Policy Engine | Which auth methods this user type supports (narrows what policy can require) |
 
+Audit event payloads use a **deny-by-default** PII policy: user attribute values
+are omitted unless a field is explicitly marked with `x-audit`. See
+[ADR 048](../../adrs/048-wide-events-internal-audit-primitive.md) §8.
+`x-sensitive` remains for API/flow payload redaction (OpenAPI user-property
```
vitorbari, line 26: I don't think there is any usage of `x-sensitive`. If we want to adopt deny-by-default behaviour, we can drop `x-sensitive`.

docs/adrs/048-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,503 @@
...
+  satisfy the tuple predicate as an efficient index range scan.
+- **Deferred:** Additional indexes when concrete query paths and predicates are
+  measured. Avoid speculative partial indexes on `NOT NULL DEFAULT ''` columns
+  unless predicates exclude empty strings.
+
+**Ordering:** List and export use keyset pagination on `(created_at, id)` within
+`project_id` per [ADR 027](027-cursor-based-pagination.md). Forensic "when did
+it happen" uses `occurred_at`.
+
+### 2. Six orthogonal correlation scopes
```
fforootd: Nit: the heading says six scopes, but the table below has eight rows — six correlation dimensions plus the two tenancy columns (`project_id`, `team_id`).

```suggestion
### 2. Correlation scopes (six dimensions plus tenancy)
```

---

docs/adrs/048-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,503 @@
...
+### 3. Event categories
+
+| `category` | When emitted | Sampling |
+|------------|--------------|----------|
+| `request` | Every **authenticated** API call (wide event at request end) | None |
+| `auth` | Token issue/revoke/exchange, login outcomes, MFA decisions | None |
+| `session` | Session create/end, step-up, handoff | None |
+| `admin` | Team, API key, config push, RBAC mutations | None (mutations only) |
+| `entity` | User/project lifecycle semantic events | None |
+| `signal` | Platform bot/abuse signals ([ADR 019](019-captcha-gate-and-bot-signals.md)) | None |
```
fforootd: Suggestion: make the durability contract explicit per category instead of leaving it implicit in §4's one-liner ("only `category=request` is batched"). We effectively have three data types with distinct write guarantees — stateful (`entity`/`admin`) and transient (`auth`/`session`) co-commit in-TX, fire-and-forget (`request`) batches — and this table is the natural single place to state that. It also forces any future category to declare its guarantee instead of inheriting one by omission.

Bonus: it pins down `signal`, which today has no declared path. It happens to work out (captcha/bot signals verify through the `auth_attempts` challenge/verify mutation per ADR 019, so Path B applies), but right now that's an inference across two ADRs rather than a statement.

```suggestion
| `category` | When emitted | Emission path → durability | Sampling |
|------------|--------------|----------------------------|----------|
| `request` | Every **authenticated** API call (wide event at request end) | Path A → batched flush (N/T); visibility ≤ T; bounded crash-loss window (§4) | None |
| `auth` | Token issue/revoke/exchange, login outcomes, MFA decisions | Path B → in-TX with the mutating statement; no loss | None |
| `session` | Session create/end, step-up, handoff | Path B → in-TX with the mutating statement; no loss | None |
| `admin` | Team, API key, config push, RBAC mutations | Path B → in-TX with the mutating statement; no loss | None (mutations only) |
| `entity` | User/project lifecycle semantic events | Path B → in-TX with the mutating statement; no loss | None |
| `signal` | Platform bot/abuse signals ([ADR 019](019-captcha-gate-and-bot-signals.md)) | Path B → in-TX with the auth-attempt/check mutation | None |
```

---

docs/adrs/048-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,503 @@
...
+VALUES (..., now() - $wait_interval, now(), ...);
+```
+
+   - `occurred_at` — when the action happened (`now() - wait`)
+   - `created_at` — when the row was written (`now()` / `DEFAULT now()`)
+4. List/export keyset pagination stays on `(created_at, id)`; forensic timing
+   uses `occurred_at`.
+5. On process shutdown: **drain** the buffer (flush remaining events) before
+   exit. Insert failure is logged and surfaced (metric / alert); it is not
+   silently dropped.
```
fforootd: Suggestion: item 5 covers shutdown, but the buffer's failure semantics are still open — and they're the one place where the `request`-category guarantee stays fuzzy:

- **Bounded buffer + overflow policy.** Is the buffer bounded, and what happens when it fills during a DB outage — block the request path, or drop with a metric? Either is defensible; unspecified is not (an unbounded buffer is a memory leak during outages).
- **Retry.** Is a failed flush retried with backoff, or only "logged and surfaced"? As written, one failed INSERT could still lose up to N events — loudly, but lost. Retry-with-backoff plus the bounded buffer is what makes the batching guarantee crisp.
- **`wait` on retry.** If retried, note that `wait = time.Since(enqueuedAt)` is recomputed at the actual insert, so `occurred_at` stays honest across retries.

One adjacent clock note: "DB `now()`" isn't one thing across dialects — Postgres `now()` is transaction-start time (which also means all rows of one batch flush share an identical `created_at`, making the `(created_at, id)` tie-breaker load-bearing — good that §1 already has it), while Spanner `CURRENT_TIMESTAMP()` is statement-eval time and commit timestamps carry read-back restrictions. Dialect divergence is already a listed risk; one sentence here would cover it.

---

docs/adrs/048-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,503 @@
...
+  (wait ≈ 0); no separate wait parameter is required.
+- Event `id` is minted via dialect `NewManagedID` before insert
+  ([ADR 047](047-dialect-id-generation.md)).
+
+Example event types (non-exhaustive; catalog grows with statement coverage):
+
+| `event_type` | `category` | Trigger |
+|--------------|------------|---------|
+| `request.api` | `request` | Authenticated HTTP handler completes |
+| `project.created` | `entity` | `CreateProject` statement |
```
fforootd: This row contradicts the pre-claim policy in Context: in the primary create-first-claim-later path (ADR 003 → ADR 046), `CreateProject` runs *before* claim, so under the current policy `project.created` would never be emitted for CLI-created projects. This feeds directly into the open thread on the pre-claim paragraph — two ways to resolve it:

- **(a) Store pre-claim events, gate *visibility* at claim.** Best forensic completeness (abuse investigations want exactly this history), and visibility is already scope-gated. But ADR 049 §4 then needs rework: "shipper/API return empty for unclaimed projects" combined with "never purge until every enabled sink has delivered" means unclaimed-project events accumulate forever. If we go this way, unclaimed-project retention must be decoupled from sink delivery (purge after a fixed window regardless).
- **(b) Keep the policy, add a `project.claimed` anchor event** so every audited timeline has a defined start, and drop/qualify this row (e.g. "post-claim creation paths only").

Suggestion: I'd lean (a) for the forensic value — but only together with the ADR 049 retention fix.

---

docs/adrs/048-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,503 @@
...
+| `user.created` | `entity` | User create statement |
+| `user.updated` | `entity` | User patch statement |
+| `user.deactivated` | `entity` | User lifecycle statement |
+| `auth.token.issued` | `auth` | Token create statement |
+| `auth.token.revoked` | `auth` | Token revoke statement |
+| `auth.check.failed` | `auth` | Check verification failure |
+| `session.established` | `session` | Auth attempt handoff |
```
fforootd, line 333: Suggestion: add a `session.ended` row here, and say who emits it on **passive expiry**. Timeout has no user-triggered mutating statement — so if expired sessions are reaped by a cleanup job, the reaper's delete/expire statement is the Path B emission point. Without that sentence, "session create/end" in §3 is only half covered for the transient-data timeline.

---

docs/adrs/048-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,503 @@
...
+SAVEPOINT is preferred on Postgres because it keeps a single connection and
+commit for the request, and `ROLLBACK TO SAVEPOINT` is cheaper than aborting
+the whole TX and starting a new one.
+
+Optional **preflight** uniqueness checks reduce how often the conflict path
+runs (happy path stays one TX), but races still need the dialect-specific path
+above.
+
+**Not supported:** emitting a failure event inside an aborted transaction
+without a savepoint (Postgres), or after Spanner abort without a new TX.
```
fforootd, line 425: Suggestion: generalize this dialect rule beyond `user.create.failed` — it applies to *any* failure event emitted after an aborted TX (`auth.check.failed` on constraint violations, future failure types). And make explicit that on Spanner the separate-short-TX failure event is inherently **weaker than co-commit**: a crash between rollback and the new TX loses it. That's unavoidable, but stating it keeps the per-category guarantee honest (failure events are the one Path B case that can degrade to at-most-once on Spanner).

---

docs/adrs/048-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,503 @@
...
+    -- Identity (dialect-minted managed ID per ADR 047; e.g. evt_<opaque>)
+    project_id          TEXT        NOT NULL
+    , id                TEXT        NOT NULL
+    , event_type        TEXT        NOT NULL       -- 'user.deactivated', 'request.api', etc.
+    , category          TEXT        NOT NULL       -- 'request', 'auth', 'session', 'admin', 'entity', 'signal'
+    , occurred_at       TIMESTAMPTZ NOT NULL       -- when the action happened (DB clock; see §4)
+    , created_at        TIMESTAMPTZ NOT NULL DEFAULT now()  -- when the row was inserted (DB clock)
+
+    -- Authorization scope captured at emit time (immutable)
+    , team_id           TEXT                    -- NULL = project-scoped; set for team-scoped credential context
```
fforootd, line 72: Nit/suggestion: the schema mixes two "absent" encodings — `team_id` and the WHO/WHAT/correlation columns are nullable, while the HOW/WHERE block below is `NOT NULL DEFAULT ''`. Both are workable; a one-line stated convention (e.g. "correlation scopes and identity dims are nullable = not applicable; delegation/device dims are `''` = always resolved, possibly empty") would prevent churn when this becomes a real migration.

---

docs/adrs/049-events-api-retention-export.md
```diff
@@ -0,0 +1,353 @@
...
+| `client_id` | string | Application or agent |
+| `session_id` | string | Session correlation |
+| `flow_id` | string | Login flow correlation |
+| `request_id` | string | HTTP request correlation |
+| `fingerprint` | string | Device correlation |
+| `entity_type` | string | Resource type affected |
+| `entity_id` | string | Resource ID affected |
+| `team_id` | string | Filter by emit-time team scope |
+| `created_after` | RFC 3339 timestamp | Inclusive lower bound on `created_at` |
+| `created_before` | RFC 3339 timestamp | Exclusive upper bound on `created_at` |
```
fforootd: Suggestion: ADR 048 tells forensic users to prefer `occurred_at` when timing matters, but the API only filters on `created_at` — so the API can't query on the very column the model recommends for timing. With the ≤ T (~1s) skew bound that's mostly fine in practice, but then say so here explicitly, or list `occurred_after`/`occurred_before` as a future extension (they can't be the keyset column, but they can be predicates).

---

docs/adrs/049-events-api-retention-export.md
```diff
@@ -0,0 +1,353 @@
...
+  "actor_type": "human",
+  "entity_type": "token",
+  "entity_id": "tok_987",
+  "client_id": "app_dashboard",
+  "token_id": "tok_abc",
+  "delegation_type": "direct",
+  "delegation_id": "",
+  "grantor": "",
+  "fingerprint": "fp_device123",
+  "request_id": "req_128bit_hex",
```
fforootd: Nit: ADR 048 Path A defines `request_id` as bare W3C-traceparent-compatible 128-bit hex — no `req_` prefix. The example should match:

```suggestion
  "request_id": "0af7651916cd43dd8448eb211c80319c",
```

---

docs/adrs/049-events-api-retention-export.md
```diff
@@ -0,0 +1,353 @@
...
+- Background job deletes eligible `events` rows; matching `event_deliveries`
+  rows are removed by `ON DELETE CASCADE` (see §3).
+- An event is eligible only when **every enabled sink** has a delivery record
+  and the retention window has elapsed.
+
+Executable retention pattern (Postgres; `$2` = enabled sink ID array from
+server config at job start — not a database table):
+
+```sql
+-- $1 = project_id, $2 = TEXT[] of currently enabled sink IDs
```
fforootd: The sink table (§3) says the webhook is **per project** ("project config"), but this comment says `$2` comes from **server config**. If `$2` were global and included `webhook`, events of projects *without* a configured webhook would never get a delivery row — and never purge, which is exactly the stranded-events risk listed in Consequences. The SQL is already per-project (`$1`), so suggest pinning the sink set per project too:

```suggestion
-- $1 = project_id, $2 = TEXT[] of sink IDs enabled for THIS project at job start
--      (stdout when enabled for the deployment; webhook only if this project configured one)
```

Same for the config sketch below: worth one sentence on where the webhook actually lives (deployment YAML vs per-project resource config), since the YAML example currently shows it deployment-wide. Micro-nit: the statement mixes positional `$1`/`$2` with a named `$retention_interval` placeholder.

---

docs/adrs/049-events-api-retention-export.md
```diff
@@ -0,0 +1,353 @@
...
+    Deliver[POST to sink URL or write stdout]
+    Record["INSERT event_deliveries"]
+  end
+  Poll --> Deliver --> Record
+```
+
+**Shipper behavior:**
+
+- One shipper loop **per enabled sink** (or one loop iterating sinks).
+- Poll: events with no `event_deliveries` row for that `sink_id`.
```
fforootd: Suggestion (future-optimization note, not v1 scope): this poll is an anti-join that rescans a growing table per sink. A per-sink watermark cursor on `(created_at, id)` keeps it proportional to new events. One sentence here makes it a known follow-up instead of a rediscovery under load.

---

fforootd (commented):
Reviewed with a specific lens: the three data types we distinguish and the write guarantee each should get — (1) **stateful** (users, domain objects), (2) **transient** (sessions, tokens), (3) **fire-and-forget** (HTTP requests and the like). The mapping onto this design works out:

| Data type | Categories | Path | Guarantee in this PR |
|---|---|---|---|
| 1. Stateful | `entity`, `admin` | B | Same-TX co-commit with the mutating statement |
| 2. Transient | `auth`, `session` | B | Same-TX co-commit (sessions/tokens are authoritative rows per ADR 010/037, so their lifecycle statements carry the event) |
| 3. Fire-and-forget | `request` | A | Batched flush (N=100 / T=1s), bounded loss window, drain on shutdown |

Types 1 and 2 get something strictly **stronger** than a dual write: since events live in the same database as the entities, Path B co-commits both rows atomically — no reconciliation problem, no "state changed but no log" race. That's the right call. Type 3 as batching matches the earlier discussion, and the DB-`now()` / `occurred_at` split is a clean way to keep one time authority.

The inline comments are all suggestions, aimed at making that contract explicit and closing the few spots where a guarantee is fuzzy or two sections disagree:

- State the per-category path/durability in the §3 table (also pins down `signal`, which currently has no declared path).
- Specify Path A failure semantics — bounded buffer, overflow policy, flush retry — the one place the type-3 guarantee is still soft.
- Resolve `project.created` vs the pre-claim policy (ties into the open thread on the pre-claim paragraph).
- ADR 049: per-project vs server-config sink set for retention (has a real never-purge consequence), `occurred_at` filterability, plus small nits.

Process note: the branch is behind `main`; 048/049 are still free today (main tops out at 047), so worth rebasing/updating and landing before the numbers get taken.

---

docs/adrs/048-wide-events-internal-audit-primitive.md
livio-a: +1 leaning **(a)** — store pre-claim events, gate visibility/export at claim.

The forensic gap in the current paragraph is real: create-first-claim-later means `CreateProject`, config push, claim init/complete, and unclaimed-mode agent work all happen before any auditable timeline exists. That's exactly the window you'd want for abuse investigations, and it's why `project.created` contradicts the policy as written.

Agree **(a) only works with the ADR 049 retention fix**: unclaimed projects need a fixed retention window decoupled from sink delivery, otherwise "never purge until every enabled sink has delivered" + "API/shipper empty for unclaimed" = events accumulate forever.

One addition while we're here: if we go (a), please also resolve the **`resource_scope_index` lifecycle** gap (separate thread, but related). Registering every event — especially high-volume `request` — in a table designed for durable flat-by-ID resources, with no purge when `events` rows are deleted, gets worse if we also store the full pre-claim window. I'd pair (a) with **not indexing events in `resource_scope_index`** (project-scoped `GET /events/{id}` instead).

**(b)** (`project.claimed` anchor only) is cleaner doc-wise but leaves the pre-claim blind spot Livio flagged — fine if we explicitly accept that tradeoff for v1 product policy, not if we pretend it's a technical limitation.

Suggest: update Context to (a), add ADR 049 §4 retention exception for unclaimed projects, qualify/remove the standalone pre-claim "no emit" rule, and fix the `project.created` row accordingly.

docs/adrs/048-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,503 @@
...
+([ADR 047](047-dialect-id-generation.md)): Go `NewManagedID` / statement helper
+supplies the PK on INSERT (no SQL `IDENTITY` / client-supplied resource PK).
+
+At emit time, capture `project_id` and optional `team_id` from the resolved
+credential scope (`ScopeContext` per
+[url-architecture.md](../design/api/url-architecture.md)). Team-scoped list/get
+filters use the **stored** `team_id`, not recomputed membership. See
+[ADR 049](049-events-api-retention-export.md) for API authorization rules.
+
+Also register each event in the global `resource_scope_index`
```
livio-a: **Blocker: `resource_scope_index` lifecycle**

ADR 048 registers every event (including high-volume `category=request`) in `resource_scope_index`. ADR 049 retention deletes `events` and cascades `event_deliveries`, but ADR 033 defines no cleanup for scope-index rows. After purge, index entries for deleted events remain → unbounded growth.

`url-architecture.md` describes this as a small table for globally-addressable **resources**; time-retained audit rows do not fit that model.

Suggestion: **do not use `resource_scope_index` for events**. Use project-scoped get (`GET /events/{id}` with `project_id` from query/credential → load row → authorize on emit-time `team_id`). Alternative: index only non-`request` categories + explicit purge in the retention job.

Either way, please resolve before merge — the current text will silently leak storage.

docs/adrs/049-events-api-retention-export.md
```diff
@@ -0,0 +1,353 @@
...
+
+**Application DB role:**
+
+- `INSERT` and `SELECT` on `events` — no `UPDATE` or `DELETE` from application
+  code during normal operation.
+- `INSERT` on `event_deliveries` — shipper only.
+- The retention background job uses a dedicated role with `DELETE` permission on
+  `events` (and relies on FK cascade for `event_deliveries`).
+
+**Retention policy:**
```
livio-a, line 181: Nit for a later pass: append-only + 30d retention + one row per authenticated call is a natural fit for time-range partitioning on `created_at` so retention can drop partitions instead of big correlated DELETEs. Does not need to be v1-mandatory for a Proposed ADR, but worth a short implementation note if we expect high-traffic projects.

docs/adrs/048-wide-events-internal-audit-primitive.md
```diff
@@ -0,0 +1,503 @@
...
+- Client headers named `session_id`, `flow_id`, or `fingerprint`, if accepted at
+  all, go into `metadata.client_hints` as untrusted hints only.
+- After the handler completes, if the request was **authenticated**, enqueue one
+  `category=request` wide event containing: `operation_id`, HTTP method, route
+  template, status code, duration, and actor/delegation dimensions from context
+  (including `delegation_id` / `grantor` when acting under an agent delegation
+  per [ADR 033 §5](033-internal-permission-management.md)).
+- Response header `X-Request-Id` echoes the assigned `request_id`.
+
+**Durability (batched insert, DB clock):** Request-wide events are **not**
```
livio-a, line 252: Optional clarity, not blocking for Proposed: Path B events are TX-co-committed; Path A `request` events are batched and can be lost on crash before drain. Risks section is honest about that — might be worth one sentence distinguishing transactional audit (auth/admin/entity/…) from best-effort access telemetry (`request`) so compliance readers do not treat them as the same durability bar. Fine to defer to implementation/operator docs if preferred.

docs/adrs/049-events-api-retention-export.md
```diff
@@ -0,0 +1,353 @@
...
+- List response wraps items in `{ "data": [...], "next_page_token": "..." }` per
+  [ADR 027](027-cursor-based-pagination.md). No total count. List omits
+  `deliveries` to keep payloads small.
+
+**Pagination:** Keyset cursor on `(created_at, id)` within `project_id`,
+consistent with [ADR 027](027-cursor-based-pagination.md) and v2 storage list
+options. Index: `(project_id, created_at, id)`.
+
+**Permissions and scope:**
+
```
livio-a, line 126: Small doc drift: ADR 049 retires `/audit_events`, but `system-permission-catalog.md` still has separate `event.read` / `audit_event.read`. Either update the catalog in a follow-up PR or add an explicit note here that `audit_event.read` is retired with the unified `/events` surface.

docs/adrs/048-wide-events-internal-audit-primitive.md
adlerhurst: This relates to [my comment](https://github.com/zitadel/nextgen/pull/459#discussion_r3718968897) and we came to the same conclusion. I therefore plan for a.

fforootd (approved):

livio-a (approved):

PR #468, opened 2026-07-07
docs: added ADR for adding query functionality for list endpoints

### LINKED ISSUE #406: Project API A: API contract (OpenAPI)
Part of #405.

## Goal

Define the OpenAPI contract for the Project API: create, get, update, and list. This is the source of truth. The Go server types are generated from it.

See the epic (#405) for the full list. The ones that affect the contract:

- The Project model is `id`, `name`, `createdAt`, `updatedAt`, `projectSecret`, `previewSecret`, `previewOrigins`. This is today's model plus `name`.
- Keep the preview fields, following the platform docs. The secrets (`projectSecret`, `previewSecret`) are returned on create only.
- No `status` field.
- Update is `PATCH` with partial updates. The only mutable field is `name`.
- List sorts by creation date (asc or desc) and filters on creation date with equals, greater than, lower than. Cursor pagination.
- No delete endpoint.

## Dependencies

- None. This is a foundation task. It can start right away.
- It unblocks sub-issue E (API handlers), which needs the generated types.

## Test first (TDD)

Add or extend the OpenAPI conformance fixtures and a spec validation check for the new paths and schemas before, or while, editing the YAML. Follow the same pattern used for `flow_definition` fixtures.

## Tasks

- Add `name` to the create request. Keep `previewOrigins`. So the create request becomes `{ name, previewOrigins }`.
- Add `name` to the create response. Keep `projectSecret`, `previewSecret`, `previewOrigins`, `createdAt`. So the create response becomes `{ id, name, projectSecret, previewSecret, previewOrigins, createdAt }`.
- Change the get response to return `id`, `name`, `previewOrigins`, `createdAt`, `updatedAt`. No secrets.
- Add an update request schema for `PATCH /projects/{id}` with `{ name }`.
- Add a list response schema (`projects[]` plus `next_page_token`). Each item uses the same fields as the get response.
- Add `PATCH /projects/{id}` and `GET /projects` to the methods files.
- For the list endpoint, reuse `components/parameters/page-token.yaml`, add a sort direction (asc or desc on creation date), and add creation-date filter parameters (equals, greater than, lower than).
- Limit: reuse `components/parameters/limit.yaml`, which sets a default (20) and a maximum (100). The list must have a default limit that applies when the caller sends none, and an upper bound. This matters because the storage layer treats a limit of 0 as "no limit", so the default and the maximum have to be enforced above storage (see sub-issue D). Keep the values in the one shared parameter so they are easy to change later.
- Register the new paths in `openapi/openapi-spec.yaml`.
- Run `go generate ./...` inside `api/` so `api/generated/` has the new types.

## Reference

Mirror the conventions used by `api/openapi/endpoints/flow_definitions/` and `api/openapi/endpoints/sessions/`. These already have list, filters, and cursor pagination, so they are good examples for the list shape.

## Boundary with the configuration surface

Do not collide with the config sub-resource path namespace. The paths `/projects/{id}/config`, `/projects/{id}/config/versions`, and `/projects/{id}/config/drift` are reserved by a separate draft spec (`docs/design/platform/api/config-api.yaml`) and a separate ticket. `GET /projects/{id}` in this work returns the project resource only, not the config blob. Coordinate path and operationId names so the two specs merge cleanly. See the epic section "Relationship to the configuration surface".

## Notes

The secrets (`projectSecret`, `previewSecret`) are returned only on create. They are not part of the get or list response. `previewOrigins` is not a secret, so it can be returned on get and list.

wim07101993:
## Summary

- Added an ADR for adding query functionality for list endpoints

## Release notes / changeset

- No changeset required

## Notes

- Linked PR: https://github.com/zitadel/nextgen/pull/462


docs/adrs/029-openapi-querying.md
```diff
@@ -0,0 +1,201 @@
...
+language would be a lot of complexity added to the product. Therefore, it would
+be better to use a JSON-like query language in the body.
+
+#### JSON-like query language
+
+To make the query language readable and definable in OpenAPI, we use a JSON 
+structure.
+
+The sorting is an array of items so that a sort-by, sort-then-by can be done.
+Each element describes a field and direction.
```
vitorbari: https://github.com/zitadel/nextgen/pull/462#issuecomment-4902727020

wim07101993: Good point, will remove

docs/adrs/031-openapi-querying.md
```diff
@@ -0,0 +1,200 @@
...
+        $ref: filter-field.yaml
+      direction:
+        $ref: /components/schemas/sort-direction.yaml
+  filter:
+    type: array
+    items:
+      type: object
+      required:
+        - field
+        - operation
```
vitorbari, line 137: are there cases where `value` isn't required?

docs/adrs/029-openapi-querying.md
```diff
@@ -0,0 +1,200 @@
...
+  - less_than_or_equal
+  - greater_than
+  - greater_than_or_equal
+```
+
+This results in a filter request which looks like: 
+
+```json
+{
+  "sorting": [
```
vitorbari: This has to be updated to a single object

vitorbari:
We can also provide both `GET /{resource}` and `POST /{resource}/search` endpoints, because the GET would allow for simple full retrieval of the complete data set when we want to perform search+sort in the UI (for resources we expect a small number of entries). It would be the same as `POST /{resource}/search` without any filtering applied, just more idiomatic.

docs/adrs/031-openapi-querying.md
wim07101993: If the customer wants to filter by an empty value, I think.

vitorbari: shall we add a `is_empty` filter?

wim07101993: we could, but then the value still needs to be nullable

docs/adrs/031-openapi-querying.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Decision
+
+### HTTP-method
+
+#### TL;DR
+
+We use HTTP-`POST` for querying but should also implement `QUERY` as soon as it
+is available in Ogen. Next to the `POST` method we still allow for a `GET` but
+it won't have filter/sort functionality.
```
IAM-marco, line 24: > but should also implement `QUERY` as soon as it is available in Ogen.
Is there a timeline for this? Or are we talking about days? Months? Years?

docs/adrs/031-openapi-querying.md
```diff
@@ -0,0 +1,208 @@
...
+
+Once Ogen supports OpenAPI v3.2, we should implement it so that clients can use
+it if they want. But since `QUERY` is not yet widely adopted, we should provide
+an alternative anyway.
+
+#### `POST`
+
+When being pedantic, `POST` is not meant for querying. However, it does fit our
+needs: it allows for a request body and is widely supported. Where POST lacks, 
+is in caching. Browsers, reverse proxies, api gateways all don't cache `POST`
```
IAM-marco, line 73: Could we use etags for caching?

docs/adrs/031-openapi-querying.md
```diff
@@ -0,0 +1,208 @@
...
+
+To make the query language readable and definable in OpenAPI, we use a JSON 
+structure.
+
+The sorting is an object with a field and direction. We do not use an array,
+which would enable a sort-by, sort-then-by feature because the storage backend
+currently does not support it.
+
+The filter is an array which combines all statements into an AND-clause. Each
+element describes a field, operation and value. OR-operations are out of scope,
```
IAM-marco, line 101: Why are ORs out of scope? I think they are supported by the backend

docs/adrs/031-openapi-querying.md
vitorbari: years or decades until it is widely adopted

docs/adrs/031-openapi-querying.md
vitorbari: Only possible IMO if we use a query language, or complex json structure to differentiate OR vs AND

docs/adrs/031-openapi-querying.md
wim07101993: Depends on how fast the community moves I guess. It is already adopted in the openapi 3.2.x spec, but since 3.1.x is still not that widely adopted, it can indeed take some time. 

But whether it is widely adopted does not really matter I think. We should adopt it as soon ogen does imo. People can choose not to use it.

Where it does matter is on when we remove the POST and simple GET endpoints. But I wouldn't count soon for that.

docs/adrs/031-openapi-querying.md
wim07101993: As far as I understand it, no. Most proxies, browsers, cdns,... don't support it and will ignore it.

docs/adrs/031-openapi-querying.md
wim07101993: Indeed, if we want to add or clauses, we need to expand the json query language. It might be something for the future. But as for now, I do not think it is a must.

I think most queries can be formed using only the operations provided in an and clause. There is a hidden "OR" like feature already built in by using the `in` and `not_in` operations.

IAM-marco (approved):

PR #472, opened 2026-07-08
docs: added ADR for abuse prevention and ratelimiting

### LINKED ISSUE #351: [ARCH-ADR] Define Architecture for Abuse Prevention and Operational Resiliency
**Context & Risk:**

Epic #249 expects operational resilience and abuse protections that are not yet specified in the `nextgen` POC. This ADR will define rate limiting strategies, brute-force protection, credential stuffing mitigation, global and project-scoped throttles, CAPTCHA/factor escalation triggers, and how to handle DB replication lag during auth flows (e.g., read-after-write guarantees for critical revocation paths). Without this, the platform is vulnerable to DoS, account compromise, and inconsistent behavior during failover/replication scenarios.

**Reference Material:**

- Epic #249: https://github.com/zitadel/nextgen/issues/249
- Current session/auth ADRs: https://github.com/zitadel/nextgen/blob/main/docs/adrs/010-session-auth-attempt-check-model.md

- https://github.com/zitadel/oxidel/blob/main/docs/adr/045-rate-limiting-and-abuse-protection.md

**Acceptance Criteria:**
- [ ] Inventory current throttles and auth attempt counters in nextgen.
- [ ] Define rate-limiting architecture (edge vs service-level), sliding window vs token bucket, and distributed enforcement options.
- [ ] Define brute-force mitigation (backoff, CAPTCHA, progressive delays) and detection thresholds.
- [ ] Define behavior and compensations for DB replication lag in auth/revocation flows.
- [ ] Draft ADR and review with SRE and Security teams.
- [ ] Merge ADR into `nextgen` before implementation.

conblem:
## Summary

- added ADR for abuse prevention and ratelimiting

## Release notes / changeset

- No changeset required

## Related

- Closes #351


docs/adrs/041-abuse-prevention-operational-resiliency.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Brute-Force Mitigation
+
+Nextgen stores durable failed-attempt state per user credential/factor
+and uses that state consistently across auth attempts.
+
+Proof verification updates the credential/factor counter in the same logical
+operation as the verification result. A failed proof increments the counter and
+records the failure time. A successful proof resets the counter for that proof
+type and records the successful verification time.
```
muhlemmer, line 71: Refine "operation" a bit further. Do you mean Request or DB Transaction? I'm asking because we need to be careful here. Given:

1. A secret like a password is obtained from the database before it can be verified.
2. Verification is a hash operation which is typically bound to CPU time
3. The counter needs to be updated on failure.

If this all happens in the same transaction, we risk DB pool exhaustion as the a connection stays occupied mid-transaction for the duration of the crypto operation. Hence we should specify a behavior that reading a hashed secret should happen in a first transaction that gets terminated. Post-check updates should happen in a second transaction.

Developing further on this I would propose to update the counter when the secret is read for verification, within the same round-trip if possible. This way we can protect against parallel brute-forcing. Updating the counter post-failure allows a short time window of sending multiple password attempts in parallel requests, which may then result in exceeding the policy.

My proposed request flow would be:

1. Initial transaction:
    1. Secret is obtained and counter is increased. The query returns the old value of the counter.
    2. Policy check on counter ->  lock user / block request if policy is exceeded.
    3. Commit
2. Perform verification / crypto operation
3. Second transaction:
   1. Reset counter on success
   2. Push related success / failure events (audit log ADR is not merged yet, but here is where that should go, IMO)
   3. Commit

CC @adlerhurst?

docs/adrs/041-abuse-prevention-operational-resiliency.md
```diff
@@ -0,0 +1,115 @@
...
+
+- after each failed proof, calculate an exponential backoff delay from the
+  current failure count, bounded by the fixed initial maximum delay
+
+The initial backoff defaults are fixed service behavior, not customer-facing
+configuration:
+
+- failures 1-4: no artificial delay
+- failure 5: 10 second delay
+- each later failure doubles the delay, capped at 5 minutes
```
muhlemmer, line 83: How should the back-off be enforced? 

1. Do we flat-out deny the request with a 423 or 429? (cheapest, easy to implement)
2. Current Zitadel delays the response using a tarpit. I'm not a big fan of this because it keeps the HTTP connection open for the duration of tarpit. Platforms like GCP have a max of concurrent requests on a container, so it is very easy to exceed the budget by sending invalid passwords and sending all available connections into the tarpit.
3. Use tech like ALTCHA.[^1] They provide both a [Go](https://github.com/altcha-org/altcha-lib-go) and [NPM](https://www.npmjs.com/package/altcha) package which allows setting a PoW target which we can increase on failure, moving the cost of brute-forcing entirely to the client. More failed attempts will result in higher CPU consumption for the client. (little more expensive to implement)

[^1]: Originally proposed on slack: https://zitadel.slack.com/archives/C098X56G06R/p1777366864101389

docs/adrs/041-abuse-prevention-operational-resiliency.md
muhlemmer, line 1: Can we add a list of positives / negatives (risks) for the decisions taken in this ADR?

PR #478, opened 2026-07-08
docs: ADR 035 — Environment Releases for Configuration Resources

### LINKED ISSUE #449: ADR: define architecture for environments and releases
## Context:

User-schema versioning (https://github.com/zitadel/nextgen/issues/445) is the first concrete step toward a broader model for managing source-controlled configuration across environments. Before we commit to building that broader model, the architecture needs to be developed and reviewed as a proper ADR.

High-level idea (needs further development):

- **Many resource kinds become versioned**, not just user-schemas — also flows, IdPs, branding, apps, policies. Each save creates an immutable revision.
- **Releases bundle revisions** for a specific environment. A release is an immutable atomic snapshot of which revisions of which resources are bound to an environment at a given moment.
- **Environments** (development, preview, production) each have their own release log. Non-production environments auto-apply on clean validation; production goes through an approval step before a release becomes current.
- **Promotion** between environments is "apply the same revision IDs to the next environment" — identical content flows unchanged from dev to production.
- **Rollback** is forward-creating: produces a new release whose contents copy a historical one, preserving the audit log.

This generalizes the schema-versioning work into a unified pattern for all resources defined as code, with explicit safety for production changes.

## References: 

https://www.figma.com/board/1XbYrhyGvGMKn6joEsLvmY/Zitadel-NextGen?node-id=1240-3786&t=JEZQ9D6jd43yDGjS-4

vitorbari:
Adds ADR introducing configuration environments and the release model.


docs/adrs/031-configuration-environments.md
```diff
@@ -0,0 +1,214 @@
...
+
+An environment either runs the previous release or the new one, never a mixture. A partial failure (release constructed, activation refused) leaves the environment unchanged and the release available for a later attempt.
+
+## CLI and drift
+
+Direct writes to the per-resource CRUD APIs remain available. Editing a resource through the dashboard, MCP, or a direct API call produces a new immutable revision but leaves every environment's activated release unchanged. The change is saved, not live. To make it live, the user constructs a new release that includes the drafted revision and activates it, the same pattern as Vercel's "you edited env vars, redeploy to apply."
+
+The CLI is source of truth for release construction. `zitadel apply` packages `.zitadel/` as-is; drafts made outside the CLI are not incorporated and become superseded by the next apply. `zitadel plan` compares the local bundle against server-side drafts and surfaces any drafts not represented locally, so the user can pull them into `.zitadel/` before applying — or deliberately overwrite them.
```
grvijayan:
```
`zitadel plan` compares the local bundle against server-side drafts
```
how would we distinguish if we wanted to compare the activated release in an env vs. the local bundle before deciding to apply? I was of the understanding that plan provides a preview of what's going to change before `apply`. I don't completely get how server-side drafts picture here between `plan` and `apply`. Could you please clarify?

vitorbari:
> Looked at the ADR — right direction, the model is sound. Immutable releases + envs as pointers is what makes instant promote/rollback possible, and the single-transaction bundle endpoint fixes today's non-atomic sync loop. Feedback in two parts:

> *1. Terminology (the TBD) — survey of how deploy platforms name our two concepts:*
• Vercel: artifact = _deployment_ (fused with the act — every deployment is born live at its own URL; when the act needs a name even they say "rolling _release_")
• Cloudflare Workers: artifact = _version_, assignment = _deployment_
• Heroku / Fly / Sentry: artifact = _release_, verb = _deploy_ (Sentry is literally our model: a deploy is a release sent to an environment)

> Everyone agrees on the verbs: deploy / promote / rollback. Only the artifact noun differs — and our releases exist unattached to any env, so "deployment" would be a misnomer (plus k8s owns that word for our operator audience). Proposal:
you *deploy* a *release* to an *environment*; the act is recorded as a *deployment*; the env *runs* the release of its latest deployment

> Concretely: `zitadel deploy` replaces `apply` (once deploy is the verb, apply has no distinct job); inert construction = `zitadel releases create`; activation = `POST /environments/{env}/deployments`; the env audit log = `GET /environments/{env}/deployments` instead of overloading `GET /releases?env=`. And drop `plan`: Terraform needs it because apply mutates live infra — our deploy is inert, git owns the content diff (if releases record their source commit), drift detection goes to `status`, removal warnings into the deploy confirmation. That's also the answer to the "rebuilding terraform" concern: no state reconciliation, just snapshots and pointers. (Resolves @grvijayan's plan thread too.)

> Team question: (a) adopt deploy-platform vocabulary at all? (b) artifact noun: release or deployment? My vote: yes + release.

> *2. Smaller items:*
• 031 is taken on main (openapi querying) → renumber 032, add README index row; typo "commited"
• flow examples disagree: handle ref inside `content` vs. sibling — pick one
• `rel_2026-07-08_14:22` → opaque id per ADR 011; add provenance (message, author, git SHA) — releases list is useless without it, and it's what makes git-as-diff work
• closed-boundary footgun: missing `.zitadel/idps/` dir ⇒ release drops all idps. Local-authoritative is right, but deploy must print removals vs. the target env's active release before activating
• bare `zitadel deploy` needs a default env and zero prompts; #449 said "auto-apply on non-prod" — the ADR drops that silently, record it either way
• `promote --from staging --to prod` (env→env) and `rollback --env prod` defaulting to previous deployment — no copying release ids on the happy path
• idempotency: unchanged tree → no-op returning existing release id, not release spam
• drift section says "pull drafts into .zitadel/" but no command does that → `zitadel pull` or explicit deferral
• inner loop: is every local save a release? Decides whether local dev feels like Vercel — name it as a follow-up
• one sentence: re-deploying the same release re-resolves `${env.X}`/`${secrets.Y}` — that's the "edited env vars → redeploy" affordance
• your meeting note "UI promotes/rollbacks, doesn't assemble" resolves the open TODO — fold it in
• +1 env lifecycle/data isolation as own ADR (incl. the users-pin-schema-revision exception)
• ADR 007 (plan/apply) needs a superseded-in-part note when this lands

docs/adrs/031-configuration-environments.md
vitorbari: Dropped `plan` completly
`zitadel status` now compares local vs. each environment's currently deployed release

vitorbari:
> Core concepts + API are solid, agree. Feedback is now basically all CLI dev-ex:

> Bigger three:
• CLI can't construct without deploying, or deploy by id — API separates construction from activation but the CLI fuses them. CI shape "build release on merge, walk the same id through envs" is impossible (promote needs a source env). Suggest zitadel releases create + zitadel deploy --release <id> --env <env>
• Same-content re-run: no-op the deployment too — if the env already runs that release, "nothing to do" instead of a new deployment (history spam). --force as escape hatch for the future template-re-resolution case. Symmetric nit: rollback should target the previous distinct release, not previous deployment
• TOCTOU on removal preview — someone else deploys between preview and confirm → you confirmed against a stale baseline. One field fixes it: optional expected_current_deployment_id on POST .../deployments, 409 → recompute + re-prompt. Cheap now, painful to retrofit
Quality pass:
• status: local can be ahead and behind at once (you edited schema, colleague deployed idp swap) — needs a diverged state; ahead/behind implies an ordering that doesn't exist
• pull assumes you know what was dashboard-edited, but drafts are enumerable via per-kind lists — status --include-drafts or pull --list, or at least an explicit deferral line
• dirty tree: deploying with uncommitted changes records a git_sha that doesn't match content → git diff mechanism silently lies. Record a dirty flag or warn on deploy
• promote confirm shows handle-set diff only (no modifications) — exactly where you want fidelity for prod. Both releases carry git_sha, so print/run the ready-made git diff <from>..<to> -- .zitadel/. That's the moment the provenance metadata pays off
• deployment record: store source_environment / rolled_back_from, not just the reason enum

grvijayan (approved):

fforootd (approved):

PR #489, opened 2026-07-09
docs: add ADR 031 — permission catalogs (portable relational FGA core)

muhlemmer:
## Summary

This is 1 of 3 PRs splitting the original permission-management ADR draft from #374 into focused documents, per feedback on that PR. This one covers the shared, foundational decision; the other two (#490 internal, #491 external) build on it.

- Add ADR 031: one permission engine, two catalogs (system and app-group), over a portable relational storage/resolver core shared by internal and external authorization.
- Add a glossary for FGA-specific vocabulary (relation, tuple, userset, TTU, closure, catalog, IR, Leopard index, etc.), since the original ADR's dense FGA language was hard to follow.
- Promote the cross-cutting terms — `principal`, `scope`, `resource_scope_index`, `delegation` — into the canonical `docs/design/glossary.md` as a new **§ 5 Authorization (FGA)** section, instead of only defining them locally in the ADR. Other design docs (`api/hierarchy.md`, `api/authz.md`, `api/url-architecture.md`, `platform/project-team-modeling.md`) already use these terms without ever defining them. Also broadens the existing **grant** definition in `docs/design/glossary.md` §4 to cover permission/relation bindings, reconciling it with ADR 031's framing. ADR 031's own glossary keeps only the vocabulary that's genuinely FGA-engine-internal (tuple, userset, TTU, closure, IR, Leopard index) and has no audience outside the permission-management ADRs; ADR 032/033 point to the new global section for the promoted terms.
- Reword the original "ordinary access changes write ordinary indexed rows" line, which reviewers on #374 found unclear and asked to be clarified (`docs/adrs/027-permission-management.md#discussion_r3498909418` and follow-ups). The new text explicitly answers the "100k-member group → 100k inserts?" question (no — one row per real assignment, expansion happens at check time via indexed semi-joins) and names the technique as Leopard-style index flattening, linking to the referenced [Leopard indexing article](https://thediligentengineer.com/leopard-the-c-of-o1-indexing-system).
- Add a plain-language "Introduction" explaining the product use case before the technical decision, per PM feedback that the original ADR jumped straight into low-level implementation.
- Index the ADR in `docs/adrs/README.md`.

[...]

PR #374, opened 2026-06-22
docs: add permission management ADR

muhlemmer:
## Summary

- Add ADR 027 proposing a portable relational FGA core for internal and external permissions.
- Clarify that Zitadel reuses `github.com/openfga/language/pkg/go` for OpenFGA parsing and owns profile validation, IR compilation, storage, and query planning.
- Treat roles as optional catalog bundles/aliases rather than required FGA primitives.
- Add constraints from tenant isolation, cross-project identity, and agent delegation requirements.
- Index the ADR in `docs/adrs/README.md`.

## Release notes / changeset

- No changeset required — no public npm package files changed.

## Closes

- Closes #349

Issue #333, opened 2026-06-17
[ARCH-ADR] Define Architecture for Cross-Project Identity and Collaboration


muhlemmer:
**Context & Risk:**

Cross-project identity is explicitly unresolved in Epic #249: the current user model scopes users to a single project, which complicates consultant/agency and platform-staff workflows. A public MVP needs a defined model for safe and understandable cross-project identity and collaboration.

This ADR will define:
- how one human is represented across multiple projects (linking vs global identity)
- when project-scoped duplication is acceptable and when it is not
- staff/operator/agency flows for granting cross-project access
- account lifecycle and deprovisioning semantics across projects
- UI/CLI/SDK implications for cross-project identity flows

Without this, everyday workflows for agencies, consultants, and platform staff will be error-prone or insecure.

**Reference Material:**
- Epic: https://github.com/zitadel/nextgen/issues/249
- Hierarchy doc: https://github.com/zitadel/nextgen/blob/main/docs/design/api/hierarchy.md
- User/Team lifecycle ADR: https://github.com/zitadel/nextgen/blob/main/docs/adrs/024-user-team-lifecycle-ownership.md

Suggested oxidel ADRs to adapt:
- https://github.com/zitadel/oxidel/blob/main/docs/adr/036-staff-access-support-grants.md
- https://github.com/zitadel/oxidel/blob/main/docs/adr/041-cloud-customer-portal-collaboration.md
- https://github.com/zitadel/oxidel/blob/main/docs/adr/042-projects-apps-owner-org-authz.md

**Acceptance Criteria for Engineer:**
- [ ] Research current state/limitations in `nextgen`.
- [ ] Review any linked `oxidel` ADRs for foundational UX/design choices.
- [ ] Draft an ADR defining the technical approach (data models, security boundaries, edge cases, algorithms).
- [ ] Ensure the proposed architecture satisfies the MVP UX goals in Epic #249.
- [ ] Review ADR with Engineering Lead.
- [ ] Merge approved ADR into `nextgen` before implementation begins.

muhlemmer:
[...]

PR #490, opened 2026-07-09
docs: add ADR 032 — internal permission management (system catalog)

muhlemmer:
## Summary

This is 1 of 3 PRs splitting the original permission-management ADR draft from #374 into focused documents. This one covers internal, Zitadel-owned resource authorization; it builds on the shared catalog model in the companion "decision of catalogs" PR (ADR 031, #489).

- Add ADR 032: `resource_scope_index`, the resolver/list-filtering mechanics for Zitadel's resource-oriented API (including the 403/404 denial contract), and audited agent delegation.
- Add a plain-language "Introduction" explaining the use case (resource-scoped checks, agent delegation) before the implementation detail, per PM feedback on the original ADR.
- Add a short glossary of internal-specific terms (`resource_scope_index` now lives in the shared `docs/design/glossary.md`; this ADR keeps credential class, system catalog, agent delegation), pointing back to ADR 031 and the shared glossary rather than duplicating them.
- Index the ADR in `docs/adrs/README.md`.

**Scope change:** staff/support access grants (the four-tier `support.read`…`support.admin` table) and the break-glass operator escape hatch were removed from this ADR's scope. Review discussion surfaced that this is squarely the subject of [issue #333, "Define Architecture for Cross-Project Identity and Collaboration"](https://github.com/zitadel/nextgen/issues/333), which is blocked by this PR series and references the same oxidel ADR-036/041 prior art this document was partially reimplementing. That content now belongs in #333's future ADR instead of here. This ADR keeps agent delegation (a different, still-in-scope concern — machine principals acting on Zitadel's own resources) and adds explicit "out of scope, see #333" pointers so the removal reads as deliberate rather than an oversight.

**System catalog schema (new §1 content):** describes that the system catalog ships a Zitadel-defined default schema built from ADR 031's primitives — RBAC-style bundles today, with room to grow into ReBAC-style relations (e.g. user groups) later — compiled through the same OpenFGA parser/profile/compiler pipeline now defined in [ADR 031 §2](031-permission-catalogs.md#2-openfga-parser-and-profile-compiler). Explicitly notes this isn't a permanent restriction to Zitadel-only authorship: customizing the system catalog's schema is anticipated future work, motivated by cases like a customer's AI agent managing that customer's own Zitadel resources and eventually needing a tailored internal permission schema. Tracked as a new Follow-up rather than decided here.

**Naming:** standardized on "Zitadel" as the product name in prose (matching ADR 033), instead of "nextgen". "nextgen" remains only where it's the literal repo name in URLs, since the codebase hasn't migrated into the zitadel repo yet.

Note: the original draft (#374) numbered itself ADR 027, but `main` has since gained ADRs up to 030 (027 is now "Cursor-Based Pagination"). This series renumbers to 031–033.

## Release notes / changeset

- No changeset required — docs only, no public npm package files changed.


docs/adrs/032-internal-permission-management.md
```diff
@@ -0,0 +1,273 @@
...
+drives both access decisions and audit export — there is no separate access
+path that bypasses the grant check. Cross-project privileged access is
+modeled through four tiers in the system catalog:
+
+| Tier | Permitted scope |
+|---|---|
+| `support.read` | Read resources, sessions, event streams, and settings |
+| `support.write` | Above plus reset credentials, revoke sessions, suspend/unsuspend principals |
+| `support.config` | Above plus modify resource configuration and policy |
+| `support.admin` | Full delegated access including impersonation grants |
```
muhlemmer: This sounds more like roles than permissions. Eg `support.read` would imply `user.read`, `session.read`. Do we really need these separate permissions, or can we think about a way to compound FGA schemas to achieve the same use case. For example have a system -> project -> team inheritance. Or would that overly complicate the implementation?

muhlemmer: Moved out of scope, should be defined as part or #333.

muhlemmer:
Closing this in favor of #493, which consolidates ADR 031, 032, and 033 into a single PR. Splitting them across three PRs turned out to add more review overhead than it saved — they're too tightly coupled (shared glossary, constant cross-referencing) to review independently. All the content and review feedback from this PR (including the staff/support-tier discussion, now resolved by descoping to #333) is carried forward into #493.

---

PR #515, opened 2026-07-09
fix(webauthn): enforce x-auth-methods for passkey actions

fforootd:
## Bug

Elina (alpha.15 testing): setting `x-auth-methods.passkey.enabled: false` in the user schema and applying succeeds, but /login and /register still offer and accept passkeys.

**Root cause:** `x-auth-methods` had exactly one production consumer — the field resolver's password check. Passkeys are action-shaped, not field-shaped: the state machine issued WebAuthn challenges purely on `stepHasActionKind(step, FlowActionKindPasskey)` and never consulted the schema. Note the asymmetry this fixes: disabling *password* was already enforced (flows collecting `x-auth-methods#password` fail validation); passkey actions now get the same treatment.

### 1. Definition-time rule (mirrors the password-field precedent)

- **Go** — `validatePasskeyActionsEnabled` in `ValidateFlowDefinition` (schema-dependent phase): rejects any step declaring a `passkey` or `passkey_register` action when the pinned schema's `x-auth-methods.passkey.enabled` is not true. An absent keyword counts as disabled, matching `xAuthMethodsReader.IsEnabled` and the password precedent.
  > `step "identifier": action "passkey" offers passkey but "passkey" is not an enabled authentication method`
- **TS port** — same rule in `packages/config/src/validate.ts` (`schema/passkey-actions`, goRef `validatePasskeyActionsEnabled`), message verbatim-identical; drift-audit test extended with the shared literal. CLI `plan` fails pre-mutation naming every offending step; the api-mock picks the rule up automatically (shared validator).
- Preset hygiene: both presets keep passkey enabled — `config:test` (incl. `presets.test.ts`) green.

### 2. No runtime double-check — by design

Adopted from @vitorbari's point on #519: a flow **pins its schema revision by URL** — schema edits mint a new revision, and repinning a flow re-validates it — so a validated flow's passkey verdict cannot change at runtime. The state machine trusts validated definitions here exactly as it does for every other rule in the validator (nothing else is runtime-re-checked either). The rationale is recorded on `validatePasskeyActionsEnabled` so the "why no runtime check" question doesn't recur.

An earlier revision of this branch carried a runtime gate + render-side hiding (review rounds 1–3); both are removed as redundant. Consequence, stated plainly: flows applied **before** this rule keep offering and accepting passkeys exactly as applied — their pinned revision is their truth — and the violation is named by their next `plan`/`apply`.

## Tests

- Go: validator table tests (disabled / absent / `passkey_register` / enabled-passes). `go test ./internal/...` green.
- TS: vectors both directions incl. `passkey_register`, drift audit extended; journey test — scaffold defaults, flip `passkey.enabled` to `false`, `plan` fails pre-mutation naming `identifier`/`password`/`register`. `moon run config:test cli:test api-mock:test` + typecheck/lint green.

## Notes

- #499 and #500 are on main and this branch is synced — the diff is the passkey enforcement only. The new Go rule (`validatePasskeyActionsEnabled`) is registered in the TS port's rule registry, which the bidirectional drift audit from #499 now demands for any validator addition.
- Changeset: `@zitadel/config`, `@zitadel/cli`, `@zitadel/server` (patch).


fforootd:
might overlap with https://github.com/zitadel/nextgen/pull/519

internal/domain/flow_state_machine.go
```diff
@@ -700,14 +708,36 @@ type passkeyPhaseResult struct {
...
+	// make the schema lie. Should the product ever want enrollment-only
+	// enforcement, gate the issue legs below instead of this
+	// whole-ceremony refusal.
+	if !pc.passkeyOn {
+		state.ClearPendingChallenge()
+		msg := "error.passkey_disabled"
+		rendered := r.buildStep(state, step, resolved, &msg, nil, nil, false)
+		state.IssuedAt = r.now()
+		return passkeyPhaseResult{handled: true, halt: &FlowStepResult{State: state, Step: rendered}}, nil
+	}
```
vitorbari: I don't think this is needed during runtime. The flow definition constraints are already applied during creation, and the user-schema is immutable after that.

There is nothing wrong with this, but if we port all validation rules to runtime, it will increase the complexity of the state_machine, which is already quite complex.

fforootd: Hm fair, I removed this in https://github.com/zitadel/nextgen/pull/515/commits/d5c6634b5da9e7a6c77705fbf39e3dd3cc04652f

vitorbari (approved):
Closes #519

PR #519, opened 2026-07-09
feat: require passkey auth method for passkey actions

vitorbari:
## What

A flow that declares an action with `kind: passkey` or `kind: passkey_register` now fails validation unless the user schema enables the `passkey` authentication method under `x-auth-methods` (`x-auth-methods.passkey.enabled == true`).


fforootd:
superseeded by #515

PR #520, opened 2026-07-10
docs: add status column to ADR index table

muhlemmer:
## Summary

Added a `Status` column to the Architecture Decision Records (ADR) index table in `docs/adrs/README.md`. Each ADR now explicitly declares its status (e.g., Proposed, Accepted, Draft, Withdrawn, Superseded, Implemented) alongside its title and summary.

Updated the parsing logic in `scripts/check-adrs-index.mjs` and corresponding test fixtures to handle the new column, ensuring the validation script correctly extracts and validates all four columns (ID, Title, Status, Summary).

## Validation

- Updated regex pattern in `check-adrs-index.mjs` to parse the new 4-column table format
- Updated type definitions and test fixtures in `check-adrs-index.test.ts` to include the `status` field
- All test cases updated to reflect the new table structure with sample status values

## Release notes / changeset

No changeset required — no shipped behavior changed. This is a documentation and internal tooling update.

## Notes

The status values assigned to existing ADRs reflect their current state based on the ADR content (e.g., ADR-003 marked as "Withdrawn" per its content, ADR-008 marked as "Implemented", ADR-023 marked as "Superseded"). The validation script now enforces that all ADRs have an explicit status in the index.

https://claude.ai/code/session_01J2CoMDoRKzcVrphEQisSmT


grvijayan (approved):

wim07101993:
Should we also specify somewhere when to use which status? I see a lot of proposed and accepted. What is the difference for us? I would guess proposed means it is open in a PR and accepted that the PR was merged. But that does not correlate with reality.

muhlemmer:
> Should we also specify somewhere when to use which status? I see a lot of proposed and accepted. What is the difference for us? I would guess proposed means it is open in a PR and accepted that the PR was merged. But that does not correlate with reality.

I would say:

1. Proposed is still subject to change (even after merge)
2. Accepted is ready to implement (a separate PR can move an ADR from proposed to accepted
3. Implemented all code is there

Feel free to send a PR with a proposal ;)

PR #522, opened 2026-07-13
fix(api): align WithLogging with ADR 030

livio-a:
## Summary

Aligns request logging middleware with [ADR 030](docs/adrs/030-error-model-mapping-and-reporting.md) expected-client-error policy:

- **4xx** responses log at `Warn` (`request rejected`) instead of `Error`
- **5xx** responses log at `Error` (`request failed`)
- Stops logging the raw serialized `response` body (removes the main API-leak amplification path in middleware)
- Extracts only the wire `code` field into structured `error_code` for diagnostics

## Validation

- `go test ./internal/api/middleware/... -count=1`
- `go test ./internal/api/... -count=1`

## Release notes / changeset

Changeset: `.changeset/align-with-logging.md` — `@zitadel/server` patch: ADR 030 request logging alignment (4xx/5xx severity, `error_code` only, no raw bodies).

## Notes

- Largely independent of the errreport foundation (PR #521); can merge in either order.
- Does not plumb domain errors via context yet — `error_code` is parsed from the JSON envelope when present. Non-JSON error bodies omit `error_code`.
- API leak remediation at handler call sites remains a separate follow-up.

### DIFF
--- a/internal/api/middleware/logging.go
+++ b/internal/api/middleware/logging.go
@@ -3,6 +3,7 @@ package middleware
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"log/slog"
 	"net/http"
 	"time"
@@ -44,7 +45,7 @@ func WithLogging(next http.Handler) http.Handler {
 
 		lrw := &loggingResponseWriter{
 			ResponseWriter: w,
-			body:           bytes.NewBuffer(nil),
+			errorBody:      bytes.NewBuffer(nil),
 		}
 		next.ServeHTTP(lrw, r.WithContext(ctx))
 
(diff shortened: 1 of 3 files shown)


internal/api/middleware/logging.go
```diff
@@ -53,15 +54,21 @@ func WithLogging(next http.Handler) http.Handler {
 		}
 
 		operationID, _ := GetOperationIDContext(ctx)
+		attrs := []any{
+			slog.Int("status_code", lrw.statusCode),
+			slog.Duration("duration", time.Since(start)),
+			slog.String("operation_id", operationID),
+		}
+		if code := extractWireErrorCode(lrw.errorBody.Bytes()); code != "" {
```
wim07101993, line 75: Won't this add a rather big performance over head by deserializing each response? Cant't we only do that on error responses?

livio-a: Good catch — agreed on the intent.

Success bodies were already **not buffered** (`Write` only copies into `errorBody` for ≥400), so `extractWireErrorCode` on the 2xx path was effectively a no-op (`len == 0`). Still looked like we might decode every response.

Pushed a fix that returns early on `< 400` so JSON parse only runs on the error path.

wim07101993 (approved):

PR #524, opened 2026-07-13
feat: harden the passkey-first golden journey end to end

fforootd:
## Summary

Fixes from a live agent-driven run of the passkey-first golden journey (empty dir → `zitadel setup --preset passkey-first` → register/sign-out/sign-in in the browser), hardened across all three layers:

- **CLI guidance**: setup failure hints and `next_commands` reconstruct the full invocation (`--preset`, `--renderer`, `--dev-port`, `--non-interactive`) — following the printed retry verbatim no longer silently drops the requested preset. HTTP 404s map to a new `E_NOT_FOUND` code (exit 4) instead of `E_VALIDATION`/exit 3; a 404 without the platform error envelope names the URL and asks whether the target is a Zitadel platform API. Passkey-first `AGENTS.md` tells agents to verify the loop via the password fallback or a CDP WebAuthn virtual authenticator.
- **Ceremony UX**: `<zl-passkey>` renders a cancellable "Waiting for your passkey…" pending state (light DOM, so the `unsafeHTML` step string stays byte-identical and the ceremony never restarts mid-flight); timeouts get their own copy instead of reading as cancellations; new `zl-passkey-started` event and `pending-label`/`cancel-label`/`silent` attributes.
- **Error dialect**: flow field validation errors travel as `error.<field>_<rule>` text keys (`StepError()`, format aliased to `_invalid`) and localise client-side — catalog-known keys render inline on their field, unknown fields resolve through generic `error.field_<rule>` fallbacks (en/de/it), and an inline-routed key whose field is not on the step downgrades to a visible banner instead of vanishing. Users see "Please enter a valid email" instead of "flow field email: format".
- **Stale errors**: step-error banners are dismissible and clear on field edit (edited field's inline error too) via imperative DOM surgery behind a non-reactive flag — typed input survives, and a dismissed banner cannot flicker back on the loading re-render.
- **Session identity**: `GET /sessions/me` hydrates `name` (conventional `name`, else `given_name` + `family_name`) and `email`; `<zitadel-session>` greets users by identity instead of the raw `user_…` id.

## Validation

- `apps/cli`: `pnpm exec vitest run` — 104 files / 759 tests green; `moon run cli:typecheck cli:lint`.
- `packages/components`: `pnpm exec vitest run --project unit` (27 files / 255 tests) and `--project browser` (37 tests); `moon run components:typecheck components:lint`.
- `go test ./internal/domain/ ./internal/service/ ./internal/api/` green (postgres integration tests are build-tagged and run in CI).
- `moon run cli-journey-e2e:e2e-local` — full journey matrix green, including passkey-only register/logout/login via CDP virtual authenticator against the rebuilt packages.
- Live golden journey against a fresh scaffold: empty-email submit → localized inline error → clears on edit; real passkey registration → sign-out → one-tap passkey sign-in in an embedded browser.
- `corepack pnpm exec changeset status` parses all four changesets.

## Release notes / changeset

- Changeset: `.changeset/setup-retry-flags-not-found.md` — @zitadel/cli: retry-flag preservation, `E_NOT_FOUND` (exit 3→4 for 404s — scripts branching on it should update), AGENTS.md agent note.
- Changeset: `.changeset/passkey-pending-and-error-clearing.md` — @zitadel/components: passkey pending/cancel/timeout UX, dismissible self-clearing error banners.
- Changeset: `.changeset/flow-error-text-keys.md` — @zitadel/server + @zitadel/components: the validation-error key dialect end to end.
- Changeset: `.changeset/session-identity-fields.md` — @zitadel/server: sessions/me `name`/`email` hydration.

## Notes

- The render-string stability constraint (`unsafeHTML` rebuilds the step subtree on any output change, reconnecting `<zl-passkey>` and wiping typed input) is now load-bearing in three places; comments at each seam document it. Latent trap flagged in review: `identity` in the Liquid context varies per keystroke — safe today because no step copy interpolates `{0}`, but a future "Welcome back, {0}" title would break the invariant.
- Tenant templates that implement only the documented banner outlet still can't render inline-keyed errors for fields that ARE on the step — that exposure predates this PR and belongs to the deferred structural template validator; the new field-presence downgrade closes the case where the error would otherwise render nowhere at all.
- Known cosmetic follow-up from review: a wrong-server 404 whose body has no `message` prints the URL twice in the `E_NOT_FOUND` message.
- Mixed-version note: new components against an older server (raw `flow field …` strings) fall back to today's verbatim banner; old components against the new server render the raw keys.


internal/domain/flow_state_machine.go
```diff
@@ -378,7 +378,7 @@ func (r *FlowStateMachineRuntime) resolveInputs(pc *processCtx) (FlowResolvedFie
 func (r *FlowStateMachineRuntime) validateAndMerge(pc *processCtx, resolved FlowResolvedFields) (*FlowStepResult, error) {
 	if validationErr := r.fields.Validate(resolved, pc.in.Fields); validationErr != nil {
 		if errs, ok := errors.AsType[FlowFieldValidationErrors](validationErr); ok {
-			step := r.buildStep(pc.state, pc.currentStep, resolved, new(errs.Error()), nil, nil)
+			step := r.buildStep(pc.state, pc.currentStep, resolved, new(errs.StepError()), nil, nil)
 			pc.state.IssuedAt = r.now()
 			return &FlowStepResult{State: pc.state, Step: step}, nil
```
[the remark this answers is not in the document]
fforootd: False positive: the repo builds with Go 1.26, where `new(expression)` is valid (spec change landed in 1.26); `go build ./...` and the full `internal/domain` test suite compile and pass with this form.

packages/components/src/orchestrator/zitadel-login.ts
```diff
@@ -614,15 +638,68 @@ export class ZitadelLogin extends LitElement {
...
+    if (!this.stepErrorDismissed) {
+      this.stepErrorDismissed = true;
+      for (const alert of root.querySelectorAll("zl-alert[data-zl-step-error]")) {
+        alert.remove();
+      }
+    }
+    const escaped = fieldName.replace(/["\\]/g, "\\$&");
+    const field = root.querySelector<HTMLElement & { invalid?: boolean; error?: string }>(
+      `zl-field[name="${escaped}"]`,
+    );
```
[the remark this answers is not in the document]
fforootd: Addressed in 8ef945fd: replaced the interpolated selector with querySelectorAll("zl-field") + getAttribute("name") equality, so free-form schema field names never enter CSS selector syntax.

.changeset/flow-error-text-keys.md
```diff
@@ -0,0 +1,14 @@
...
+
+Flow field validation errors now travel as localisation keys instead of
+developer strings: `step.error` carries `error.<field>_<rule>` per violation
+("; "-joined, format spelled `_invalid` to match the catalog), and the login
+components localise them — catalog-known keys render inline on their field,
+unknown fields resolve through new generic `error.field_<rule>` fallbacks
+interpolated with the step's field label (en/de/it). A key routed inline whose
+field is not on the step downgrades to a visible banner message instead of
+disappearing. Users see "Please enter a valid email" instead of
+"flow field email: format".
```
vitorbari, line 14: `step.error` was originally a single string for simplicity, but we can expand it to a proper data model that can carry errors in a more structured way.

We probably just have to split between field validation errors vs other generic errors:
- field validation errors: are usually rendered right next to the input field
- generic errors: rendered separately


for example:

 ```json
{
  "error": {
    "fields": {
      "email": [
        { "rule": "format", "text_key": "error.email_invalid" }
      ],
      "x-auth-methods#password": [
        { "rule": "required", "text_key": "error.x-auth-methods#password_required" },
        { "rule": "min_length", "text_key": "error.x-auth-methods#password_min_length" }
      ]
    },
    "general": [
      { "text_key": "auth_attempt.password_invalid" }
    ]
  }
}
```

Comment (vitorbari) [reply .changeset/flow-error-text-keys.md:14]: Or a single `errors` array...

**Review by bastionstack — APPROVED** [review]

File: .changeset/flow-error-text-keys.md (hunk 3 above)
Comment (fforootd) [reply .changeset/flow-error-text-keys.md:14]: Thats a good idea, lets merge this PR and then create a follow up for that! How does that sound?

File: apps/cli/src/commands/setup/index.ts [hunk 4]
```diff
@@ -220,7 +220,13 @@ export default class Setup extends BaseCommand {
           this.meta.cliVersion,
           projectName,
           issuer,
-          framework.id,
+          {
+            // Resolved values, not raw flags: the wizard may have picked the
+            // preset interactively, and the retry must reproduce that choice.
+            ...retryOptionsFromFlags(flags),
+            framework: framework.id,
+            preset: answers.preset,
```
IAM-marco, line 230: The "resolved values, not raw flags" comment is only honored for `preset`: `retryOptionsFromFlags(flags)` still takes `devPort` from the raw `--dev-port` flag. When the user picks a port interactively at the DevPortPrompt (folded into `framework.devPort` and the issuer at https://github.com/zitadel/nextgen/blob/b6710fef700a751f7e5bb90183f768e19c5612c2/apps/cli/src/commands/setup/index.ts#L200-L205), a failure in `createProjectWithLocalHint` prints a retry command without `--dev-port`, so following it verbatim can scaffold on a different port/issuer than the one just confirmed. That is the failure mode `.changeset/setup-retry-flags-not-found.md` says this PR fixes. Suggest adding `devPort: answers.devPort` alongside `preset: answers.preset` here.

IAM-marco (dismissed):

apps/cli/src/commands/setup/index.ts
fforootd: Confirmed and fixed in 669635d9 — `devPort: answers.devPort` now sits alongside `answers.preset` in the override, so an interactively picked port (and therefore the issuer derived from it) survives into the printed retry. The two cloud-failure tests pin the full string including `--dev-port`.

vitorbari (approved):

PR #526, opened 2026-07-14
chore: set argon2id as default password hashing algorithm

### LINKED ISSUE #503: Crypto: Change default password hashing from bcrypt to argon2id
ADR requirement (§Hashing → Passwords and secrets): "By default, we should use argon2id for password/secret hashing."

Current state: `cmd/server/server.go:324` sets `password_hasher.hasher.algorithm = crypto.HashNameBcrypt` (bcrypt cost 10, bounds 10–16). The passwap wrapper already fully supports argon2id (`internal/crypto/passwap.go:82,336`), so the primitive exists, only the default is wrong.

Tasks:
- [ ] Flip the default algorithm to `HashNameArgon2id` with RFC 9106 recommended params.
- [ ] Keep bcrypt (and legacy verifiers) registered as verifiers so existing bcrypt hashes still validate and transparently rehash on next successful login (passwap already does this: add a test).
- [ ] Update default `HashLimitsConfig` bounds for argon2.
- [ ] Docs/config sample update.

Acceptance: New passwords hash with $argon2id$; existing bcrypt hashes verify and get rehashed to argon2id on verify.

wim07101993:
## Summary

- set argon2id as default password hashing algorithm

## Validation

- `go test ./...`

## Release notes / changeset

- Changeset: .changeset/default-argon2id-password-hashing.md

## Notes

- Issue: #503


cmd/server/server_test.go
```diff
@@ -26,3 +29,30 @@ func TestLoadConfigReadsPostgresDatabaseEnv(t *testing.T) {
 	assert.Equal(t, "postgresql://postgres@localhost:5432/nextgen?sslmode=disable", postgresConfig)
 	assert.Equal(t, "4D61737465726B65794E65656473546F48617665333243686172616374657273", cfg.Server.EncryptionKey)
 }
+
+// TestLoadConfigDefaultsToArgon2id locks in the ADR 029 default: with no
+// password_hasher overrides, the built hasher produces argon2id hashes while
+// still verifying (and rehashing) pre-existing bcrypt hashes.
+func TestLoadConfigDefaultsToArgon2id(t *testing.T) {
+	configPath := filepath.Join(t.TempDir(), "nextgen.yaml")
+	_, err := os.Create(configPath)
```
IAM-marco: `os.Create` returns an open file handle that is never closed, leaking a file descriptor. The same pattern was flagged in PR #228 and fixed in [runtime_test.go](https://github.com/zitadel/nextgen/blob/85eed4a086493041bf6095b8cce9d7bd13697638/cmd/server/runtime_test.go#L18-L20) by using `os.WriteFile`. Consider `require.NoError(t, os.WriteFile(configPath, nil, 0o600))` here as well.

IAM-marco (changes requested):

IAM-marco (approved):

PR #533, opened 2026-07-14
feat: added data encryption keys per project

### LINKED ISSUE #505: Crypto: Introduce KEK→DEK envelope encryption (replace the single flat key)
ADR requirement (Encryption → Signing/encryption keys, Storage): Master Key (KEK) wraps a Data Encryption Key (DEK); the DEK encrypts signing keys, third‑party secrets and opaque tokens. Keys stored in the DB fall under envelope encryption.

Current state (PARTIAL/gap): `cmd/server/server.go:485` builds one AES‑256‑GCM crypter directly from the master key and uses it to encrypt everything (currently just opaque tokens). There is no DEK layer, no `crypto.Encrypter` implementation beyond the interface (`internal/crypto/encryption.go`), and no key table.

Tasks:
- [x] Design a keys storage table (id, purpose, project scope, key (encrypted), algorithm, state, createdAt/expiresAt).
- [x] Generate a DEK, encrypt (wrap) it with the master key (KEK), persist it encrypted.
- [x] Add an envelope Crypter that loads+unwraps the DEK and performs AES‑GCM for data at rest.
- [ ] Route opaque‑token encryption (and future secret encryption) through the DEK rather than the raw master key.

Don't cache keys for now. It opens the door for stale cache which we do not want to tackle right now.

Acceptance: Encrypted data at rest is protected by a DB‑stored DEK that is itself wrapped by the master key; rotating the master key (Issue 4) only requires re‑wrapping the DEK, not re‑encrypting data.

wim07101993:
## Summary

- Added data encryption keys (DEKs) (storage, service and domain layers)
- Refactored some tests because they broke after implementation
- Moved `tokengen` (`TokenGenerator` and `TokenVerifier`) to a service (`TokenService`)
- Because the project secrets now use the DEK, all integration tests needed to be updated
- Moved `projectSecret` and `previewSecret` out of project because it would create circular dependency: DEK is needs project ID, Project needs DEK to create secrets

## Validation

- `go test ./...`

## Release notes / changeset

- no changeset because no public api's changed


internal/storage/v2/dialect/spanner/statement_crypto_keys.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/zitadel/nextgen/internal/domain"
+	"github.com/zitadel/nextgen/internal/service"
+	"github.com/zitadel/nextgen/internal/storage/v2/database"
+)
+
+type cryptoKeyStatements struct{ statement }
+
+// CreateEncryptionKey implements [service.CryptoKeyStatements].
+func (s cryptoKeyStatements) CreateEncryptionKey(ctx context.Context, key *domain.EncryptionKey) error {
+	panic("not implemented")
```
fforootd: Since `cryptoKeyStatements` is now part of the Spanner `statements` set and `projectService.Create` always calls `CreateEncryptionKey`, a Spanner-backed run would hit this `panic` and take the process down on project creation.

Recommendation: return an error instead of panicking so the dialect degrades gracefully until it's implemented — e.g. `return domain.ErrNotImplemented()` here and `return nil, domain.ErrNotImplemented()` in `GetEncryptionKey`. Same idea as the other not-yet-implemented v2 spots.

internal/api/flow.go
```diff
@@ -207,11 +207,22 @@ func (h *Handler) GetFlowStep(ctx context.Context, params api.GetFlowStepParams)
...
-	payload, err := h.crypter.Decrypt(raw)
+
+	header, err := domain.DecodeJWEHeader(raw)
+	if err != nil {
+		return nil, errFlowCookieInvalid
+	}
+
+	decrypter, err := h.keyService.GetCrypter(ctx, header.KeyID, header.EncryptionAlgorithm)
+	if err != nil {
+		return nil, err
```
fforootd, line 220: Small suggestion: a `GetCrypter` failure here is returned raw, so a cookie with a valid-looking JWE header but an unknown `kid` surfaces as a `dek.not_found` error. That code isn't recognized by `mapFlowErrorStatus`/`mapFlowGetError`, so it falls through to the internal-error path and returns a 500 (plus an error-report) for what's really just an untrusted/expired cookie.

Every other failure in `openState` maps to `errFlowCookieInvalid` — it'd be nice to do the same on key-resolution failure so these come back as a clean 401/404:

```go
decrypter, err := h.keyService.GetCrypter(ctx, header.KeyID, header.EncryptionAlgorithm)
if err != nil {
    return nil, errFlowCookieInvalid
}
```

(The `SecurityHandler` paths already collapse verify errors to a 401, so this is the one spot that leaks.)

internal/service/keys.go
```diff
@@ -0,0 +1,113 @@
...
+	kek op.Crypto,
+) KeyService {
+	return &keyService{
+		db:  db,
+		kek: kek,
+	}
+}
+
+func (s *keyService) GetEncryptionKey(ctx context.Context, keyID string, algorithm jose.ContentEncryption) (*domain.EncryptionKey, error) {
+	key, err := s.db.Statements().GetEncryptionKey(ctx, database2.And(
```
fforootd, line 44: Not for this PR — just flagging for whenever key rotation/removal lands: this lookup filters on id + algorithm but not `state`, so an `expired` or `removed` key still resolves and can decrypt. That's exactly what you want for rotation continuity, but once there's a revocation path we'll probably want `KeyStateRemoved` to stop decrypting here (otherwise `IsRemoved()` ends up unused). Might be worth a short comment noting the intentional "any non-removed state still decrypts" rule.

fforootd (commented):
Really nice change — the envelope-encryption model and the per-project DEK resolution read cleanly, and the test coverage (especially the recursive KEK→DEK resolution and the partial-unique-index invariant) is great. A few recommendations inline, mostly small. I left out the "existing projects have no DEK / 000001 edited in place" point on purpose since we're still in alpha and DBs get reset freely — worth revisiting only if we ever commit to a no-reset upgrade path before GA.

internal/storage/v2/dialect/spanner/statement_crypto_keys.go
wim07101993: i took inspiration from the `statement_project.go` which does exactly the same: 

https://github.com/zitadel/nextgen/blob/2cf426e0bbe9d27059d748f16272bd1674408dc0/internal/storage/v2/dialect/spanner/statement_project.go#L14-L16

I thought those api's were still only placeholders for an actual implementation

fforootd: Hm yeah that could be... I guess @adlerhurst or @muhlemmer know more about this

internal/api/flow.go
wim07101993: I added a case which checks for domain errors. I don't think we should hide internal errors. Otherwise these won't get logged.

fforootd (approved):

internal/storage/v2/dialect/spanner/statement_crypto_keys.go
wim07101993: In the mean time, some implementations have been made. So I implemented the integration with spanner as well.

fforootd (approved):

PR #551, opened 2026-07-16
feat: add use-case selection to zitadel setup

### LINKED ISSUE #448: feature: app-type selection on zitadel setup
## Context:

Today, zitadel setup provisions every project with the same default user-schema and the same default login flow. As we expand support for different application patterns, "the right default" diverges.

## Proposal:

During `zitadel setup`, the CLI prompts for the type of app being built. Based on the answer, it copies a different set of pre-defined schemas + flow definitions into `.zitadel/`.

App-type categories (working names; final names TBD during implementation):

Template|Required|Optional
--|--|--
Basic web app|email|displayName
Consumer|email|displayName, firstName, lastName, country, preferredLanguage
Business|email|firstName, lastName, ~companyName~, jobTitle, department

## Mechanics:

- Prompt fires before any `.zitadel/` file is written.
- Each app type maps to a pre-defined bundle of source files
- The CLI copies the selected bundle on first setup

## References:

https://www.figma.com/board/1XbYrhyGvGMKn6joEsLvmY/Zitadel-NextGen?node-id=1247-1331&t=JEZQ9D6jd43yDGjS-4

vitorbari:
## What

Adds a second, **composable** axis to `zitadel setup` (closes #448): the **use case** owns the schema field set; the **sign-in preset** owns the flow. `zitadel setup` now asks **"Who will sign in to your app?"** *before* the sign-in preset:

| Use case | Fields collected |
|---|---|
| `minimal` (default) | `email` |
| `consumer` | `email`, `givenName`, `familyName` |
| `business` | `email`, `givenName`, `familyName`, `companyName` |

The login flow's `register` step is **derived** from the chosen field set instead of a hard-coded list.

Screenshot 2026-07-17 at 06 17 19 1

## Notes

- Go-embedded templates untouched.
- The choice is stored in `zitadel.json` as `useCase`.
- **Agents/non-interactive** — `--use-case` enum flag, defaults to `minimal`

## Manual tests / Screenshots

Use-Case|Prompt|Result
--|--|--
minimal|Screenshot 2026-07-17 at 06 14 49|Screenshot 2026-07-17 at 06 15 00
consumer|Screenshot 2026-07-17 at 06 19 01|Screenshot 2026-07-17 at 06 18 52
business|Screenshot 2026-07-17 at 06 24 46|Screenshot 2026-07-17 at 06 26 46


.changeset/use-case-selection.md
```diff
@@ -0,0 +1,20 @@
+---
+"@zitadel/config": patch
+"@zitadel/cli": patch
+---
+
+`zitadel setup` now asks "Who will sign in to your app?" and scaffolds the
+matching schema fields: `minimal` (email only, today's default), `consumer`
```
fforootd: Small but important wording thing: today's no-flag scaffold collects `email`, `givenName`, `familyName`, and `dateOfBirth` on register (both presets), so `minimal` isn't quite "today's default" — this PR actually slims the default scaffold down to email-only, and `dateOfBirth` drops out of every use case entirely.

I'm genuinely fine with leaner-by-default (it fits the quickstart golden path), but I'd love for it to read as a stated decision rather than a side effect. Suggestion along the lines of: "the default scaffold now collects email only; given/family name move to `consumer`/`business`, and `dateOfBirth` is no longer scaffolded." If we'd rather keep continuity with today's output, defaulting to `consumer` would be the alternative — your call, both work for me.

One more piece of the same decision: `api-mock`'s canned schema shrinks with it (`getDefaultHumanUserSchema()` bare call in `platform-handlers.ts`). CI is green, so nothing in-repo depends on the extra fields — just noting it belongs to the same choice.

packages/config/src/defaults.ts
```diff
@@ -59,11 +59,118 @@ function presetTemplates(preset: string): { schema: unknown; flow: unknown } {
...
+};
+
+/**
+ * JSON-Schema bodies for use-case fields the shipped templates don't already
+ * define (they own `email`/`givenName`/`familyName`). Authored in the same
+ * style as those templates — no `x-claim`, since the backend maps identity by
+ * attribute name and the user-property meta-schema defines no claim keyword.
+ */
+const EXTRA_FIELD_BODIES: Record<string, Record<string, unknown>> = {
+  companyName: {
```
fforootd, line 110: `register.field.companyName` doesn't exist in any builtin locale yet, so the `business` register step renders the humanised fallback — English "Company name" in all three locales, next to otherwise-translated siblings in de/it, and without `.placeholder`/`.help` support. Would you mind adding the key to `en`/`de`/`it` as part of this PR?

Related: `preset-coverage.spec.ts` ("every text key a shipped preset flow can produce must resolve in every builtin locale") still renders flows at the default use case, so `consumer`/`business` field keys now sit outside the guard that exists to catch exactly this. Iterating `SETUP_USE_CASES × SETUP_PRESETS` there would keep that contract honest for the new axis.

packages/config/src/defaults.ts
```diff
@@ -59,11 +59,118 @@ function presetTemplates(preset: string): { schema: unknown; flow: unknown } {
...
+  }
+  return { ...schema, required: ["email"], properties };
+}
+
+/**
+ * Derive the flow's register-step fields from the use case rather than
+ * copying a hard-coded list into every bundle (#448): the register step
+ * collects exactly what the schema defines. Other steps are untouched.
+ */
+function applyUseCaseToFlow(
```
fforootd, line 169: Now that the register step's `fields` are always overwritten here, the four-field lists authored in both flow templates become dead content, and `dateOfBirth` (still authored in both schema templates) can't appear in any composed output. Nothing is broken — but the next person editing a template's register fields will wonder why nothing changes. I'd suggest either trimming the templates to match reality, or a short note in them that `properties` serves as the field-body catalog and the register `fields` list is placeholder, overwritten per use case. Either works for me.

apps/cli/src/lib/setup-resources.ts
```diff
@@ -50,16 +52,19 @@ export async function materializeSetupResources(opts: {
...
 }): Promise<MaterializeSetupResourcesResult> {
   await mkdir(join(opts.cwd, FLOWS_DIR), { recursive: true });
   await mkdir(join(opts.cwd, SCHEMAS_DIR), { recursive: true });
 
   const filesWritten: string[] = [];
   const preset = opts.preset ?? DEFAULT_SETUP_PRESET;
+  const useCase = opts.useCase ?? DEFAULT_SETUP_USE_CASE;
 
-  const { $id: _templateId, ...schemaBody } = getDefaultHumanUserSchema({ preset }) as {
+  const { $id: _templateId, ...schemaBody } = getDefaultHumanUserSchema({ preset, useCase }) as {
```
fforootd, line 67: The config-package matrix covers the composition itself really well; this plumbing is the one seam without automated coverage — nothing asserts that `--use-case business` actually lands `companyName` in the written `.zitadel/` files (today that's proven by the screenshots). One case in `setup-resources.test.ts` with a non-default use case would close the loop cheaply. Just a suggestion, not a blocker.

apps/cli/src/commands/setup/prompts/use-case.ts
```diff
@@ -0,0 +1,33 @@
...
+ * `--use-case` is authoritative; non-interactive runs keep the minimal
+ * default seeded by the command.
+ */
+export class UseCasePrompt implements SetupPrompt {
+  async ask(answers: SetupAnswers, ctx: PromptContext): Promise<SetupAnswers> {
+    if (ctx.useCaseFromFlag) {
+      return answers;
+    }
+    const labels: Record<SetupUseCase, string> = {
+      minimal: "Just me or a small group — email only",
```
fforootd: Non-blocking copy thought: the labels lead with audience size/type, but the axis is really the field set — e.g. a consumer app that wants email-only ends up picking the option labelled "Just me or a small group". Maybe lead with the fields and keep the audience as the flavor?

Related follow-up we don't need to solve here (pre-existing): the `en` locale labels `register.field.email` as "Work email", which the `consumer` use case now shows to consumers.

fforootd (commented):
Really nice — this is exactly the shape #448 was hoping for: the use case owns the field set, the preset owns the flow, and the two compose instead of multiplying into a bundle per pair. The 6-pair hygiene matrix against the real flow validator gives me a lot of confidence in the composition, and I'm glad to see `buildUserSchema`/`fieldPreset` go — that catalog had quietly drifted from the shipped templates (snake_case names, `x-claim`, password-in-properties), so consolidating on one source of field defaults removes a real trap. The wiring parity with `preset` is spot on too: flag seeding, prompt skip, telemetry re-record after prompts, retry-flag reproduction (with a test), additive `zitadel.json` — and the prototype-safety guard with its own tests is a nice touch.

Two things I'd recommend we settle before merging (details inline):

1. **The no-flag default changes more than the changeset suggests.** Today's scaffold collects `email` + `givenName` + `familyName` + `dateOfBirth`; with this PR the default (`minimal`) collects `email` only, and `dateOfBirth` disappears from every use case. Leaner-by-default is fine by me — I'd just like it to be a stated decision rather than a side effect.
2. **`register.field.companyName` has no locale entry yet**, so the `business` register step falls back to humanised English in de/it — and the coverage spec that exists to catch exactly this doesn't iterate use cases yet.

A few smaller, non-blocking suggestions inline as well (dead template content, one test at the CLI seam, copy thoughts). Happy to help with any of these if useful.

packages/config/src/defaults.ts
vitorbari: `companyName` key added to en/de/it

packages/config/src/defaults.ts
vitorbari: I have trimmed the default template. `dateOfBirth` was removed.

apps/cli/src/commands/setup/prompts/use-case.ts
vitorbari: The ida was to show the audience first ("Just me or a small group", "Consumers", "Business") and keep it more generic so that the answer is used beyond only user-fields.

.changeset/use-case-selection.md
vitorbari: I have reduced the default schemas to the minimal (email only).

vitorbari:
Updated prompt:

Screenshot 2026-07-17 at 13 07 04

apps/cli/src/commands/setup/prompts/use-case.ts
fforootd: Fair — agreed, audience-first is the right call if the answer is meant to outgrow fields, and it justifies the useCase name. Let's just keep the '— fields' suffix honest as more consequences attach to the choice, and keep in mind a consumer app may pick 'just me' purely to get fewer fields; fine while the value stays guidance-only. Resolving

fforootd (approved):

PR #563, opened 2026-07-20
feat: tenant login templates as editable config (ADR 037)

fforootd:
## Summary

Tenant login templates become an uploadable, editable-config resource ([ADR 037](docs/adrs/037-tenant-login-templates-editable-config.md)), closing the loop the widget was already half of: `zitadel branding eject --design <name>` → edit `login.liquid` → `zitadel plan` → `zitadel apply` → the login renders it.

- **Docs**: ADR 037 (immutable per-project branding revisions, latest-revision live resolution, validation split, sibling-`.liquid` dialect, design catalog); `template-security.md` Layer 4 rewritten into 4a (CLI-authoritative LiquidJS validation) / 4b (server lexical gate — the server is Go, the dialect is LiquidJS, so a Go AST check would validate the wrong language); `templates.md` records the grouping decision (one global template per project; storage and wire shape are decoupled) plus the design catalog and eject workflow.
- **API**: new `branding` endpoint group (`POST /branding`, `GET /branding`, `GET /branding/{id}`, `branding.read/write` scopes), mirroring the schemas convention (immutable ⇒ create/read only, query-param `project_id`). The shared `Branding` component doubles as the write shape, so what you publish is exactly what flow responses emit.
- **Server**: domain model + lexical validator (size/UTF-8/`<script>`/`<style>`/inline `on*=` handlers/`javascript:`/`| raw`; the handler pattern only matches inside open tags so Liquid like `{% assign online = true %}` doesn't false-positive), migration `000012` in both dialects, repository with latest-per-project resolution (`created_at` DESC, ULID-id tiebreak — Postgres freezes `NOW()` per transaction), service, handlers, and the emission swap: all three flow handlers resolve the latest revision per project and degrade to the built-in default so a broken branding lookup can never take the login down.
- **`@zitadel/config`**: `./template` — the authoritative validator (real liquidjs parse, `{% mandatory_gates %}` required, banned patterns mirrored to the Go gate via `goRef`s); the `branding.json` config dialect meta-schema (sourced in `api/openapi/endpoints/schemas/`, drift-copied like the others); the design catalog `centered` / `split` / `split-right` / `minimal` (`centered` is a drift-tested byte-identical copy of the bundled default template).
- **`@zitadel/components`**: split/minimal layout chrome in `layout-chrome.css`, the `mandatory_gates` tag name single-sourced from `@zitadel/config/template`, and a spec rendering every catalog design through the full pipeline (engine → DOMPurify → gates patcher) against real atoms.
- **CLI**: `BrandingSyncer` (schema-style `revisioned` semantics; the descriptor references the template via `liquid_template_file`, inlined for hashing/upload so a `.liquid` edit plans a `revise`, split back out on write-back; path-traversal guard; singleton — see review response), `zitadel branding eject [--design]` with an interactive picker, and `zitadel setup --design <name>` which scaffolds and publishes revision 1 during setup.

## Review response (codex, 2026-07-20)

All five findings addressed, plus the branch is merged with main:

1. **Frozen installs (P1)** — `@zitadel/config` was in both `dependencies` and `devDependencies` of `packages/components` while the lockfile only recorded the dev entry; dropped the duplicate and regenerated `pnpm-lock.yaml`. `pnpm install --frozen-lockfile` passes again.
2. **Migration collision (P1)** — main took `000011` (`user_team_lifecycle`); branding is now `000012` in both dialects. (Main also took ADR numbers 035 *and* 036, so the ADR here is renumbered to **037** throughout; it now cross-references ADR 035 — branding revision ids are exactly what a configuration release can pin.)
3. **`font_url` document-level CSS (P1, security)** — the field is now **read-only in v1**: the component must load the font stylesheet into `document.head` (shadow-scoped `@font-face` never registers faces), so a writable URL would hand `branding.write` page-wide CSS over the embedding app, bypassing the template sandbox. `POST /branding` rejects a non-empty value, the config dialect omits it, the CLI refuses it at plan time, and ADR 037 records the safe-delivery follow-up options (FontFace API against font binaries, or a pure-font-CSS allowlist).
4. **Stray descriptors published as revisions (P2)** — `ResourceSyncer` gains `singletonFile`; branding declares `branding.json` and the scan hard-fails (`E_VALIDATION` + hint) on any other `*.json` in `.zitadel/branding/` before anything is validated or uploaded.
5. **Advertised-but-ignored pagination (P2)** — `listBranding` drops `offset`/`page_token` instead of shipping dead parameters (schemas has the same gap behind a `TODO` awaiting ADR 031's list-query design; all list endpoints should gain real querying together).

## Review response (codex re-review + self-review, head `bf98fa2d`)

The second codex pass found four blockers; all fixed, with the settled model: **managing templates and rendering them during login are different planes.**

1. **Project binding (P0)** — every branding management operation now requires the bearer's token project to match the requested `project_id`; foreign projects answer exactly like nonexistent ones (anti-oracle: create → 400 `brnd.invalid`, read → 404). Two-project integration test proves it against real tokens.
2. **Scope enforcement (P0)** — writes require `project.write`|`branding.write`, reads `project.write`|`branding.read`. The browser-plane **preview secret** (`project.read` only, shipped to visitors' browsers by design) loses all management access — before this gate a leaked preview token could publish login templates. Blanket per-operation scope enforcement stays with ADR 036 (scopes are minted into tokens, so retroactive enforcement would strand every existing secret); until then `project.write` implies the `branding.*` scopes, documented in the OpenAPI scope descriptions and ADR 037's new "Access model" section. Negative tests: guard unit table + preview-secret 403 integration tests. The same gaps exist for schemas/flows/users/teams (pre-existing) — spun off as a follow-up task.
3. **Broken standalone bundle (P1)** — `@zitadel/config` is now inlined into `dist/standalone.mjs` (library build keeps it as a declared npm dependency). Root cause went deeper than the missing `noExternal` entry: `components:build` had **no moon dep edge on `config:build`**, so the two raced in the CI graph and rolldown silently left the unresolvable import external — and moon cached the broken artifact. Edge added, plus a `standalone-artifact.spec.ts` guard (no bare package imports, no `node:` builtins) with `components:test` now depending on the build. **Round-3 follow-up (head `30629b61`)**: the release twin `components:build-release` had the same missing `config:build-release` edge (the release aggregate only passed by incidental ordering); edge added, and the checks moved to `scripts/check-standalone-artifact.mjs`, which `build-release` now runs against its own artifact (verified from clean config output, and the checker exits 1 on a bad file) while the spec imports the same function for the dev build.
4. **plan/apply asymmetry on asset URLs (P1)** — the zod dialect now mirrors the server's absolute-https rule for `logo_url`/`hero_url`, the meta-schema carries `^https://` patterns for editor feedback, and negative tests cover the CLI validate path (apply mutates schemas/flows before branding, so plan must reject exactly what the server would).

Earlier self-review fixes on `d4451c5a`: unknown project on `POST /branding` → 400 (FK `IntegrityViolationError` mapped, schema-service convention), `resolveBranding` logs before degrading to default branding, the inline-handler ban treats `/` as an attribute separator (``) in both gates, and the latest-revision index gained the ULID tiebreak column.

## Validation (post-merge, head `214a7376`)

- `go test ./...` and `go test -tags postgres_integration ./internal/...` (embedded Postgres) — green, including new domain/service/repository/API-integration branding suites (publish → flow response carries it → second revision supersedes → bare project falls back to default; hostile template → 400 `brnd.invalid`; `font_url` → 400; unknown id → 404). One unrelated flake: parallel packages can race migrations in `internal/storage/v2/dialect/postgres` on a shared DB (passes serially; pre-existing, tracked separately).
- `moon run server:check-generate` — codegen drift clean.
- `moon run config:test` (121), `moon run components:test` (354), `moon run cli:test` (786) — green; `tsc --noEmit` on cli; oxlint clean on touched files; `pnpm install --frozen-lockfile` green.
- The centered-design drift test caught main's #404 template changes during the merge — the whole catalog is refreshed to the new field loop (testid filter, checkbox/select error passthrough).
- **Live golden-path smoke** (pre-merge): built server from this branch against embedded Postgres → created a project → `branding eject --design split` → `plan` (1 create) → `apply` (revision published, write-back parity: empty re-plan) → edited `login.liquid` → `plan` shows `revise` → `apply` → `GET /branding` lists two revisions newest-first and the latest carries the edit → `<script>` template fails `plan` with `E_VALIDATION` / `no-script-tag` before any network call.

## Release notes / changeset

- Changeset: `.changeset/tenant-login-templates-branding.md` — minor for `@zitadel/server`, `@zitadel/api`, `@zitadel/config`, `@zitadel/cli`; patch for `@zitadel/components`.

## Notes

- **Decisions** (settled during planning): revisioned-immutable rather than mutable PUT (consistent with schema immutability; audit + rollback for free, no repin machinery since nothing references branding revisions); live latest-revision resolution rather than pinning at flow start (templates are chrome — a mid-flow change re-skins the next step but can't reshape in-flight data; `branding.yaml`'s old "at flow creation time" wording updated); wire `layout` enum stays `centered|split` — richer designs travel *as templates*, the escape hatch templates.md always reserved.
- `setup --design` is flag-only (no wizard prompt) — `branding eject` owns the interactive design picker, keeping the setup wizard short. It composes with main's new `--use-case` flag.
- Vocabulary rule (canonical in glossary §6): **branding** = the resource, **template** = the Liquid artifact you edit, **design** = a named catalog starting point, **layout** = the wire degrade enum.
- In-CLI prose must say "the `branding eject` command", never backtick-`zitadel branding eject` — the contract test extracts suggested command ids from source prose and bare `branding` is not a visible command id.
- Third-party-consumer caveat recorded in `template-security.md`: the API returns raw template strings; the safety guarantees hold for rendering through `@zitadel/components`.
- Not covered: `setup --design` against a real framework scaffold (unit-tested only), and no real-browser screenshot of the split design yet — worth one manual look before merge.


bastionstack (approved):

docs/adrs/040-tenant-login-templates-editable-config.md
```diff
@@ -0,0 +1,190 @@
...
+
+### 2. Resolution: latest revision, resolved per response
+
+Every flow response (`create`, `submit`, `get step`) resolves the
+**latest branding revision for the project** at response-build time and
+falls back to the built-in default (`layout: centered`, no template) when
+none exists. Templates are chrome, not data: a mid-flow template change
+re-skins the next step but cannot reshape in-flight flow state, so live
+pickup is harmless and gets fixes out immediately. This supersedes the
+earlier "inherited at flow creation time" wording in the wire doc.
```
peintnermax, line 82: This conflicts with ADR 035

ADR 035 says branding is one of the resources a release pins, and that an environment only sees what's in its currently deployed release.

But this PR resolves branding as "latest revision per project": every flow response just picks the newest row for the project (`GetLatest` = `ORDER BY created_at DESC LIMIT 1`). Environments and releases are never checked.

That means:
- A `POST /branding` goes live immediately in **all** environments, including prod — no deployment step.
- Revisions that were never part of any release still show up at runtime, which ADR 035 explicitly forbids.
- Staging and prod can never have different branding.

I think branding should resolve through the environment's active release, like schemas and flows do. 
If instant-latest is really what we want, then ADR 035 should remove branding from the pinned resources and ADR 037 should probably say so explicitly.

peintnermax (commented):
I found a contradiction to ADR 035 (releases and revisions)

internal/service/branding.go
```diff
@@ -0,0 +1,97 @@
...
+			return nil, nil
+		}
+		return nil, domain.ErrInternal(err).WithMessage("failed to resolve latest branding revision")
+	}
+	return branding, nil
+}
+
+// List returns revisions for the project, newest first.
+func (s *BrandingService) List(ctx context.Context, projectID string) ([]*domain.Branding, error) {
+	brandings, err := s.brandingRepo.List(ctx, s.pool,
```
grvijayan, line 96: a limit would be nice to avoid overfetching. 

also, other resources are moving to use v2Pool ([example](https://github.com/zitadel/nextgen/blob/main/internal/service/project.go)) - should we use that here instead of migrating later?

grvijayan (approved):

Issue #408, opened 2026-06-24
Project API C: storage implementation (Postgres and Spanner)


IAM-marco:
Part of #405.

Blocked by: https://github.com/zitadel/nextgen/pull/457

## Goal

Implement the storage layer for the Project on storage v2 (#315, merged). Add the `name` column, add `name` to the Postgres statements, add `UpdateProject`, make delete cascade to child resources, and implement the Spanner statements (which are all stubs today).

## Current state (after #315, merged to main)

- Postgres: `internal/storage/v2/dialect/postgres/project.go` implements `CreateProject`, `GetProjectByID`, `ListProjects`, and `DeleteProjectByID`. `DeleteProjectByID` is a hard `DELETE FROM ... WHERE id = $1`. The `projectSchema` binds each `ProjectField` to a column.
- Spanner: `internal/storage/v2/dialect/spanner/statement_project.go` exists but every method is `panic("unimplemented")`.
- Table: `zitadel_nextgen.projects` (see `internal/storage/database/dialect/postgres/migration/sql/000001_projects.sql` and the Spanner equivalent). Columns today: `id, created_at, updated_at, project_secret, preview_secret, preview_origins`. There is no `name` and no `deleted_at`.

## Decisions that shape this work

See the epic (#405). The ones that affect storage:

- Add `name` (required) and keep the preview columns.
- Delete is a hard delete with cascade. Deleting a project removes the project row and its child resources. Use `ON DELETE CASCADE` foreign keys or delete the child rows explicitly in the delete path. `#315` already ships `DeleteProjectByID` as a hard `DELETE`; keep it and add the cascade. No `deleted_at`, no soft delete.
- `name` has no unique index for now (workspace deferred).
- Filter on creation date with equals, greater than, lower than. Sort by creation date (asc or desc). These are already supported by the #315 filter AST and keyset pagination.

## Dependencies

- #315 (merged). This issue extends the project statements it added.
- Sub-issue B (#407), which adds `Name` to the model, adds `ProjectFieldName`, and adds `UpdateProject` to the interface.

## Test first (TDD)

Write repository integration tests first (Postgres and Spanner) for create (with `name`), get, update, list (creation-date filter and sort and cursor pagination), and delete (the project row and its child resources are removed). Then implement.

## Tasks

[...]

PR #579, opened 2026-07-21
feat: add project Update, List, and Delete to ProjectService

### LINKED ISSUE #409: Project API D: service layer
Part of #405.

## Goal

Add the service layer methods for update, list, and delete, and make create and get use the new model (with `name`, without the preview fields).

See the epic (#405) for the full list. The ones that affect the service:

- Create takes a `name` in addition to the preview origins it takes today. It still mints the preview secret. Keep the current preview handling and add `name` (pass it through `NewProject`).
- Get returns `id`, `name`, `previewOrigins`, timestamps.
- Update is partial and changes `name` (calls the new `UpdateProject` statement).
- List sorts by creation date (asc or desc) and filters on creation date with equals, greater than, lower than. Cursor pagination.
- List limit: the service must apply a default limit when the caller sends none, and enforce a maximum. The storage layer treats a limit of 0 as "no limit", so this cannot be left to storage. Use the shared limit values (default 20, maximum 100).
- Delete is a hard delete with cascade (through the storage delete). Deleting a project removes the project and its child resources. There is no delete API, but the service method should exist so a future cronjob can call it.

## Dependencies

- Sub-issue B (#407). The service uses the statement interface through the generated mocks.
- Runs in parallel with sub-issue C. It does not wait for real storage, because the tests use the mocks.

## Test first (TDD)

Write service unit tests first, using the statement mocks:

- `Create` with a `name` (keeping the existing preview handling).
- `Get` returning `name` and `previewOrigins`.
- `Update` (partial, changes `name`).
- `List` (creation-date filter and sort and cursor pagination, including the default limit and the maximum limit).
- `Delete` (hard delete with cascade, logic only).

Then implement.

## Tasks

- Update `Create` to take a `name`. Keep the existing preview handling (accept preview origins, mint the preview secret).
- Add `Update` (partial, `name`).
- Add `List` (filter, sort, cursor pagination). Apply the default limit when none is given and enforce the maximum limit.
- Add `Delete` (hard delete with cascade, logic only, no API).
- Make `Get` return `name` and `previewOrigins`.
- Keep the transaction handling consistent with the existing `projectService`.

## Reference

`internal/service/project.go` is the file to change. `internal/service/flow_definition.go` shows a service with create, update, get, list, and delete for comparison.

grvijayan:
## Summary

Builds out the project use-case surface in the service layer for #409:
`Update`, `List`, and `Delete` on `projectService`. (`Create`'s `name`
parameter landed separately in #554.)

- **Update** — partial update of a project's name. Empty id/name return the
  corresponding domain validation errors; a missing row maps to
  `ErrProjectNotFound`, other failures to `ErrInternal`.
- **List** — filter, sort, and cursor-paginate projects through the v2 pool.
  Sort defaults to `createdAt` asc with `id` as a tiebreaker for stable keyset
  pages; the limit is normalized to `[1, 100]` (default 20), since the storage
  layer treats limit 0 as "no limit". Validation errors carry client-facing
  details, and storage cursor errors are mapped to domain request errors via
  `mapListError` so storage codes never leak past the service. Shared
  list-query helpers and the `Sorting`/`Filter` DTOs live in `list.go` for
  reuse by other resource `List` endpoints.
- **Delete** — hard-delete a project through the v2 pool, wrapping storage
  failures in a domain internal error. There is no delete API; the method
  exists so a future cronjob can call it. The cascade to child resources and
  the idempotent no-op for a missing project are handled by the storage delete
  (tracked in https://github.com/zitadel/nextgen/issues/408).

## Additional Changes:
- Removed name as a sortable/filterable project field.

## Testing

TDD against the mocked statement interface (`testAllStatements` +
`servicemocks.MockPool`). Covers Update (validation, ok, not-found, wrapped
error), List (defaults, limit clamping, sorting, filters, cursor passthrough,
error mapping, and validation short-circuits), Delete (missing id, ok, wrapped
error), and a `previewOrigins` round-trip assertion on Get.

- `go test ./internal/service/` — green
- `go vet ./internal/service/` — clean

## Notes

Changeset: `.changeset/project-service-update-list-delete.md`

Closes #409


internal/service/project.go
```diff
@@ -144,6 +156,166 @@ func (s *projectService) createDefaultLoginFlowDefinitions(ctx context.Context,
...
+	project := &domain.Project{
+		ID:   id,
+		Name: name,
+	}
+	if err := s.v2Pool.Statements().UpdateProject(ctx, project); err != nil {
+		if _, ok := errors.AsType[*database.NoRowFoundError](err); ok {
+			return nil, domain.ErrProjectNotFound()
+		}
+		return nil, domain.ErrInternal(err).WithMessage("failed to update project")
+	}
```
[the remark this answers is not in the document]
grvijayan: storage implementation will be fixed in follow-up PRs

internal/service/flow_test.go
```diff
@@ -44,16 +47,25 @@ func (s testAllStatements) GetProjectByID(ctx context.Context, id string) (*doma
...
 }
 
-func (testAllStatements) UpdateProject(context.Context, *domain.Project) error {
-	panic("unexpected call to UpdateProject")
+func (s testAllStatements) UpdateProject(ctx context.Context, project *domain.Project) error {
+	if s.updateProject != nil {
+		return s.updateProject(ctx, project)
+	}
+	return nil
 }
```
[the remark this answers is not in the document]
grvijayan: this is the existing convention for the project statements in this helper

internal/service/project.go
```diff
@@ -147,3 +169,164 @@ func (s *projectService) Get(ctx context.Context, id string) (*domain.Project, e
...
+	default:
+		return nil, domain.ErrRequestInvalid()
+	}
+}
+
+// projectField maps an API field name to its [domain.ProjectField].
+func projectField(field string) (domain.ProjectField, error) {
+	switch field {
+	case projectFieldName:
+		return domain.ProjectFieldName, nil
```
IAM-marco: `projectField` maps `"name"` to `domain.ProjectFieldName` here, so `List` now accepts `name` as a sortable/filterable field via `projectFilter`/`projectOrderBy`. But neither dialect's `projectSchema` defines a `FieldBinding` for `domain.ProjectFieldName` — see [postgres/project.go#L109-L140](https://github.com/zitadel/nextgen/blob/f41eaf743a532dca8b8db44d0fa074a42c9e2d0f/internal/storage/v2/dialect/postgres/project.go#L109-L140) (ID/CreatedAt/UpdatedAt/ProjectSecret/PreviewSecret/PreviewOrigins all have one; Name does not), and the same gap exists in the Spanner schema.

`Schema.binding` [panics on an unmapped field](https://github.com/zitadel/nextgen/blob/f41eaf743a532dca8b8db44d0fa074a42c9e2d0f/internal/storage/v2/database/schema.go#L23-L29), and that path is reached from `List` via `compileFilter`/`compileOrderBy` -> `schema.SQLName` -> `binding`. A real request sorting or filtering projects by `name` will crash the server, not return an error. All current tests mock `v2Pool.Statements()`, so this path is untested.

Either add a `FieldBinding` for `domain.ProjectFieldName` to both dialect schemas (the `projects` table also has no `name` column yet per `000001_projects.sql`), or reject `name` here until that lands.

internal/service/project_test.go
```diff
@@ -308,3 +315,438 @@ func TestProjectService_Get(t *testing.T) {
...
+			},
+		},
+		{
+			name: "filter equals name",
+			req: service.ListProjectsRequest{
+				Filters: []service.Filter{{Field: "name", Operation: "equals", Value: "acme"}},
+			},
+			result: &v2database.ListResult[*domain.Project]{},
+			checkOpts: func(t *testing.T, opts *v2database.ListOptions[domain.ProjectField]) {
+				assert.Equal(t, v2database.And[domain.ProjectField](
```
IAM-marco:
```suggestion
				assert.Equal(t, v2database.And(
```

internal/service/project_test.go
```diff
@@ -308,3 +315,438 @@ func TestProjectService_Get(t *testing.T) {
...
+			},
+		},
+		{
+			name: "filter greater_than createdAt parses RFC3339",
+			req: service.ListProjectsRequest{
+				Filters: []service.Filter{{Field: "createdAt", Operation: "greater_than", Value: createdAt.Format(time.RFC3339)}},
+			},
+			result: &v2database.ListResult[*domain.Project]{},
+			checkOpts: func(t *testing.T, opts *v2database.ListOptions[domain.ProjectField]) {
+				assert.Equal(t, v2database.And[domain.ProjectField](
```
IAM-marco:
```suggestion
				assert.Equal(t, v2database.And(
```

IAM-marco (changes requested):

internal/service/project.go
grvijayan: I didn't touch anything related to api/storage layer as we have separate tickets for them.

As the `List` method isn't called anywhere yet, should we leave this as is as we will anyway add this binding as part of the storage implementation?

IAM-marco: Why not return an error? The specs require filtering and sorting only by creation date

IAM-marco: Or ignore `name` if it's passed in the params

grvijayan: sorry, i missed that
will fix it.

grvijayan: the openapi contract mentions the `name` field though: https://github.com/zitadel/nextgen/blob/main/api/openapi/endpoints/projects/query/filter-field.yaml#L3 🤔

IAM-marco: I missed that during review it seems. It's not requested in the tickets either. Can you remove it from the input params for the list request?

grvijayan: but why not though? we have similar [queries](https://github.com/zitadel/zitadel/blob/main/proto/zitadel/org/v2/query.proto#L20) in Zitadel (e.g., one could list resources where the name [contains/startswith](https://github.com/zitadel/zitadel/blob/main/proto/zitadel/object.proto#L84) some string value)

edit: i can remove it if we don't need this; i am just curious

IAM-marco: For the MVP I asked Elina what she wanted, and she told me creation date is enough.

For sorting, if you add another parameter you then also need to be able to specify which direction you want. Currently we support only sorting in one single direction per request. That's because it complicates things on the storage side. I don't recall the exact issue (I think performance) but I can ask Silvan if you're interested.

For filtering, a date and a string require different comparison operators at SQL level:
- Date greater/lower than are fine. What is the meaning though of a string greater/lower than another?
- `string contains` is a different construct from `date contains` in sql

Yes it can be done, but it gets more complicated than we need. Also, all of the implementation above needs to be done 2 times: postgres and spanner.

grvijayan: removed `name` from the project filter fields

internal/service/project_test.go
```diff
@@ -73,139 +73,508 @@ func createMockedProjectService(t *testing.T) (svc service.ProjectService,
...
+	}
 
-			got, err := svc.Create(context.Background(), "", nil, false)
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
 
-			assert.Nil(t, got)
-			assert.Error(t, err)
+			svc, _, _, _, _, _, _, _, _, _, statements := createMockedProjectService(t)
```
IAM-marco, line 294: 😆

grvijayan: I know 😅 
this test file went through 4-5 different merge conflict resolutions and this is the current state😛

internal/service/project.go
```diff
@@ -158,3 +175,150 @@ func (s *projectService) Get(ctx context.Context, id string) (*domain.Project, e
 	project, err := s.v2Pool.Statements().GetProjectByID(ctx, id)
 	return project, mapStorageError(err)
 }
+
+func (s *projectService) Update(ctx context.Context, id, name string) (*domain.Project, error) {
```
IAM-marco, line 179: I'm wondering if `name` needs to be a pointer. Because API side, `Update` is a PATCH operation, so `name` is optional over there: api/openapi/endpoints/projects/by_id/patch-project-request.yaml . WDYT?

IAM-marco: Ohh maybe, on API side, if `name == nil` then no call to `service.Update()`

IAM-marco (approved):

internal/service/project.go
grvijayan: will handle this in the API task

PR #582, opened 2026-07-21
fix(database): serialize concurrent schema migrations across processes

fforootd:
## Summary

- Serialize concurrent `migration.Migrate` calls sharing one database, on both dialects. Parallel `go test -tags postgres_integration ./internal/...` packages against a single `ZITADEL_TEST_POSTGRES_URL` database intermittently died with `partial migration error (type:sql,version:4): duplicate key value violates unique constraint "pg_type_typname_nsp_index"` — two goose runs both saw a migration as pending and raced its DDL (`CREATE TYPE` has no `IF NOT EXISTS`; concurrent `CREATE SCHEMA IF NOT EXISTS` also collides on `pg_namespace_nspname_index`). The same race exists for several server nodes starting at once, so the fix lives in `migration.Migrate`, the single choke point all pool/test paths funnel through.
  - **Postgres**: goose's own `WithSessionLocker` holds `pg_advisory_lock` on the same connection goose migrates on (schema bootstrap runs under the same lock first), so single-connection pools work and the session dying releases the lock with it. A `postgres_integration` unit test pins the single-connection guarantee.
  - **Spanner** (no advisory locks): a claimed lease row in `zitadel_migration_lock` with expiry-based takeover for crashed holders — and, from review, the holder renews `acquired_at` every 60s while goose runs under a fencing context that cancels the moment holdership can't be proven, so long schema changes (index backfills) keep their lease and a taken-over holder stops instead of racing its successor. Waiters poll **read-only** at 1s and bootstrap is check-first with bounded `FailedPrecondition` retries plus a post-claim grace, because the emulator rejects DDL overlapping any read-write transaction or other schema change (real Spanner queues instead).

After the repository-to-statements refactor wave landed on `main` (#566–#573), this PR is exactly the migration serialization: the two `migrations.go` files, the pg locker unit test, and the changeset. The auth-attempt determinism work this PR carried for a while (client-generated ids via a dialect-owned `spanner.NewEphemeralID`, plus an ADR 011 amendment) was superseded by #573: the v2 statements run their writes inside the native spanner client's `ReadWriteTransaction`, whose retry re-runs the whole closure — database-generated `THEN RETURN` ids are replay-safe there — so that code and the ADR amendment were dropped and the id-strategy threads are closed as moot. (Historically this PR also fixed `json_schemas.object_type` nullability, the spanner-tag `ServiceDB` build break, and v2 CRUD test isolation; #457, #533, and #601 shipped their own versions.)

## Validation

- Race repro/proof harness (throwaway, not committed): 4 concurrent `Migrate` calls per fresh database. Pre-fix: Postgres failed round 0 with the exact reported error; Spanner failed round 1 with `FailedPrecondition` at version 3. Post-fix: 20/20 rounds clean on each dialect.
- Pre-merge, `go test -tags postgres_integration ./internal/... -count=1` against one embedded Postgres: three sweeps with `DROP SCHEMA zitadel_nextgen CASCADE` between plus one warm re-run — all green. Same command `-tags spanner_integration` against one shared emulator: two sweeps with the database recreated between — all green.
- Post-merge (with #457 in): postgres sweep 20 packages ok, 0 failed. Spanner sweep vs one shared emulator: 19 ok, zero migration-race errors, and `api/integration_test` now runs (and passes) its real tests — the first genuine two-package concurrent migration through the lease, alongside live test traffic. The new `internal/storage/v2/dialect/spanner` `TestProjectStatements_CRUD` initially failed in this mode — it expected its list to hold only its own project, and its fixed ID collided across runs; fixed here (`244feb75`) by filtering the list to the test's own project and deriving the ID per run, mirroring the postgres v2 pattern. Three further sweeps (fresh, warm re-run, fresh): that package green in all. A residual flake in the repository auth-attempt tests under shared-emulator concurrent traffic was then root-caused and fixed here too (`794bef64`): go-sql-spanner replays aborted read-write transactions comparing result checksums, and `THEN RETURN id` on `GENERATED … AS IDENTITY` columns diverges on replay (`ErrAbortedDueToConcurrentModification`), which the repository's `!found`-before-`rows.Err()` ordering masqueraded as "auth attempt not found". The spanner auth-attempt writes now use client-generated random INT64 ids (timestamps were already client-side), and both dialects' scans check `rows.Err()` before concluding not-found. Verified: four further full shared-emulator sweeps (fresh/warm alternating) with zero auth-attempt failures — the baseline was one failing subtest per sweep — plus a clean postgres sweep covering the reordered pg scan. The same latent `THEN RETURN` pattern exists in the session/token spanner paths but their returns include DB-generated timestamps (a bigger semantic call) and have not flaked; left untouched. CI is unaffected throughout (per-package emulator containers).
- Review-round re-verification: the goose-SessionLocker postgres design passed 20/20 rounds of 4 concurrent fresh-database migrations plus a fresh-schema sweep including the new single-connection unit test; the renewed/fenced spanner lease passed two further 10-round races and two full sweeps; after merging #533 both final sweeps are green (pg 22 ok / spanner 21 ok, zero migration-race errors) with the crypto-keys migrations running under the locks.
- Post-#573 verification: shared-database sweeps on the final tree — postgres 22 packages ok, spanner 21 ok, zero migration-race errors, and zero auth-attempt failures under the native-client transactions (empirically confirming #573 dissolved the replay-divergence flake this PR previously patched in v1). The single failing package per dialect is #601's `TestProjectStatements_List`, whose whole-table assertions fail on any shared database; pre-existing upstream, tracked separately.
- `go build ./internal/...`, `go vet` under both tags, `gofmt` clean.

## Release notes / changeset

- Changeset: .changeset/serialize-concurrent-migrations.md — serialize concurrent schema migrations (pg advisory lock / spanner lease row) (`@zitadel/server`).

## Notes

- The Postgres lock is held on its own connection while goose runs on the pool, so `Migrate` needs a pool allowing ≥2 connections (noted in the doc comment; all defaults qualify).
- Spanner deliberately has no `000010_passkey_registrations` counterpart: `PasskeyRegistrationRepository` is Postgres-only (pgx import, schema-qualified table). Whoever ports it should take the next free migration number — goose's provider rejects out-of-order versions by default.
- The spanner lease's emulator hardening (read-only waiter polling, claim grace, bootstrap retries) exists because the emulator flatly rejects overlapping DDL/read-write transactions; on real Spanner, DDL queues and the lease alone suffices.

Co-Authored-By: Claude <noreply@anthropic.com>


internal/storage/v2/dialect/spanner/project_test.go
```diff
@@ -39,13 +42,16 @@ func TestProjectStatements_CRUD(t *testing.T) {
 
 	stmts := pool.(*Client).Statements()
 
+	// Unique per run: the database may be shared with other test packages and
+	// persist across runs, so a fixed ID would collide.
 	project := &domain.Project{
-		ID:             "proj_v2_crud",
+		ID:             "proj_v2_crud_" + strconv.FormatInt(time.Now().UnixNano(), 10),
 		ProjectSecret:  "project-secret",
```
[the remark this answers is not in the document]
fforootd: Addressed in f717552f. The ID now includes the sanitized test name plus a ULID, so it no longer relies on wall-clock resolution for uniqueness across packages or overlapping runs.

internal/storage/database/dialect/postgres/migration/migrations.go
```diff
@@ -14,11 +16,41 @@ var sqlFiles embed.FS
...
+// across both the schema creation and goose's run: two concurrent goose runs
+// can each see a migration as pending and race its DDL (CREATE TYPE has no IF
+// NOT EXISTS, so the loser fails on pg_type's unique index). The lock is held
+// on a dedicated connection while goose runs on the pool, so db must allow at
+// least two open connections.
+func Migrate(ctx context.Context, db *sql.DB) (err error) {
+	conn, err := db.Conn(ctx)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
fforootd: Addressed in f717552f by removing the two-connection requirement rather than only rejecting it. Schema bootstrap releases its connection, then goose reacquires the same advisory lock as a SessionLocker and runs version checks plus SQL migrations on that locked connection. I also added a postgres_integration regression test with MaxOpenConns(1).

internal/storage/database/dialect/spanner/migration/migrations.go
```diff
@@ -2,17 +2,63 @@ package migration
...
-func Migrate(ctx context.Context, db *sql.DB) error {
+const (
+	// lockTable holds the single lease row that serializes Migrate across all
+	// connections and processes sharing one database. Spanner has no advisory
+	// locks (the postgres dialect uses pg_advisory_lock), so mutual exclusion
+	// is a claimed row instead.
+	lockTable = "zitadel_migration_lock"
+	// leaseSeconds bounds how long a holder that died without releasing can
+	// block other migrators; a live holder finishes and releases well within it.
+	leaseSeconds = 600
```
fforootd, line 32: Could we avoid allowing this lease to expire while its holder can still be running? acquired_at is never renewed, so after 600 seconds another node can claim the row and enter goose while the original DDL is still active, recreating the exact race this PR fixes. Spanner schema changes are long-running operations, and index backfills or data validation can exceed a fixed ten-minute assumption: https://cloud.google.com/spanner/docs/schema-updates#schema_update_performance. Because the original holder is not fenced, the grace period does not make takeover safe. This needs either renewal plus fencing, a hard deadline that is proven to terminate the original DDL before takeover, or a non-expiring lock with explicit recovery.

internal/storage/database/repository/auth_attempt.go
```diff
@@ -428,10 +435,33 @@ func (a *spannerAuthAttempt) scan(rows database.Rows, attempt *domain.AuthAttemp
...
+	return nil
+}
+
+// spannerRandomID returns a random positive INT64 for columns whose identity
+// would otherwise be database-generated. Supplying the id keeps the INSERT
+// deterministic: go-sql-spanner retries aborted transactions by replaying
+// their statements and comparing results, and a THEN RETURN of a fresh
+// IDENTITY draw diverges on replay, failing the whole transaction with
+// ErrAbortedDueToConcurrentModification.
+func spannerRandomID() (int64, error) {
```
fforootd: Could we keep this workaround aligned with the accepted identifier contract before merging it? ADR 011 says auth-attempt and check IDs are database-generated bit-reversed identities and are not supplied on insert, while ADR 028 places ephemeral ID generation in the dialect layer. Generating random INT64s in the repository bypasses the Spanner sequence, loses its uniqueness guarantee, and creates an undocumented exception. The transaction-replay failure is real, but this should either use a dialect-owned deterministic allocation strategy that preserves uniqueness, or update the ADR and add an explicit collision/retry plan.

internal/storage/database/dialect/postgres/migration/migrations.go
```diff
@@ -14,23 +16,83 @@ var sqlFiles embed.FS
...
+	return nil
+}
+
+// migrationSessionLocker lets goose hold the advisory lock on the same
+// *sql.Conn it uses for version checks and SQL migrations.
+type migrationSessionLocker struct{}
+
+func (migrationSessionLocker) SessionLock(ctx context.Context, conn *sql.Conn) error {
+	if _, err := conn.ExecContext(ctx, "SELECT pg_advisory_lock($1)", migrationLockID); err != nil {
```
adlerhurst, line 83: It might be worth doing it in a transaction and use `pg_advisory_xact_lock`. This would ensure a lock gets cleaned up if the process dies midway.

adlerhurst: The unlock would then get a simple commit or rollback.

internal/storage/database/dialect/spanner/migration/migrations.go
```diff
@@ -2,17 +2,63 @@ package migration
...
+	pollInterval = 250 * time.Millisecond
```
adlerhurst: If we need to wait for 5 minutes i would reduce the poll intervall.
```suggestion
	pollInterval = 1 * time.Second
```

internal/storage/database/dialect/spanner/migration/migrations.go
```diff
@@ -25,3 +71,139 @@ func Migrate(ctx context.Context, db *sql.DB) error {
...
+
+// holderID returns a random identity for this Migrate call, so releases and
+// lease reclaims only ever touch our own claim.
+func holderID() (string, error) {
+	var b [16]byte
+	if _, err := rand.Read(b[:]); err != nil {
+		return "", fmt.Errorf("generate migration lease holder id: %w", err)
+	}
+	return hex.EncodeToString(b[:]), nil
+}
```
adlerhurst, line 147: i would use the `NewULID` function of the `idgen` package instead of having some magic here.

internal/storage/database/repository/auth_attempt.go
```diff
@@ -428,10 +435,33 @@ func (a *spannerAuthAttempt) scan(rows database.Rows, attempt *domain.AuthAttemp
...
+	var b [8]byte
+	if _, err := rand.Read(b[:]); err != nil {
+		return 0, fmt.Errorf("failed to generate id: %w", err)
+	}
+	id := int64(binary.BigEndian.Uint64(b[:]) &^ (1 << 63))
+	if id == 0 {
+		id = 1
+	}
+	return id, nil
 }
```
adlerhurst: looks like claude had fun spreading random id generators across the project 😄 
Use `NewULID` of the `genid` package.

adlerhurst: or add a function in the genid package to generate uuid v4 if that improves performance in spanner

internal/storage/database/dialect/spanner/migration/migrations.go
```diff
@@ -2,17 +2,63 @@ package migration
...
 //go:embed sql/*.sql
 var sqlFiles embed.FS
 
-func Migrate(ctx context.Context, db *sql.DB) error {
+const (
+	// lockTable holds the single lease row that serializes Migrate across all
+	// connections and processes sharing one database. Spanner has no advisory
+	// locks (the postgres dialect uses pg_advisory_lock), so mutual exclusion
+	// is a claimed row instead.
+	lockTable = "zitadel_migration_lock"
```
adlerhurst, line 29: when and where do we create this table?

adlerhurst (commented):
nice catch. I wasn't aware that goose doesn't provide a mechanic to catch these errors internally.

internal/storage/database/dialect/spanner/migration/migrations.go
fforootd: Implemented renewal + fencing: the holder re-stamps acquired_at every 60s, and goose runs under a cancel-cause context that fences it off the moment the lease row names another holder or renewal fails past a full lease window — takeover now only ever follows a provably non-renewing holder.

internal/storage/database/repository/auth_attempt.go
fforootd: Took the update-the-ADR path: generation moved to the dialect-owned spanner.NewEphemeralID per ADR 028, and ADR 011 now documents the Spanner client-generated exception with the replay rationale and an explicit collision plan (fail the operation, retry at flow level — in-transaction retry would be replay-non-deterministic again); sequence-backed pre-allocation is noted as the follow-up that would also cover sessions/tokens.

internal/storage/database/dialect/spanner/migration/migrations.go
fforootd: Applied the suggestion: pollInterval is now 1s.

internal/storage/database/dialect/spanner/migration/migrations.go
fforootd: Adopted: holderID now uses idgen.NewULID().New("mig").

internal/storage/database/repository/auth_attempt.go
fforootd: Centralized as suggested, but in the spanner dialect (spanner.NewEphemeralID) rather than idgen: these are INT64 columns so a ULID/UUID string cannot fit, and ADR 028 places ephemeral id generation in the dialect layer — ADR 011 now documents the exception.

internal/storage/database/dialect/spanner/migration/migrations.go
fforootd: The lockTable doc comment now answers this: ensureLeaseRow creates the table and its row lazily on a migrator's first run against a database, deliberately outside the goose chain because the serialization must exist before goose itself can run safely.

internal/storage/database/dialect/postgres/migration/migrations.go
fforootd: Resolved differently in f717552f: the lock stays session-scoped but moves into goose's own `WithSessionLocker`, which holds `pg_advisory_lock` on the same connection goose runs migrations on — an enclosing transaction cannot wrap goose's run (each migration commits its own transactions on that connection, and Postgres cannot nest), so a xact-scoped lock does not fit the locker contract. The cleanup property you were after holds regardless: a dying session releases its advisory locks on disconnect, and `SessionUnlock` verifies the release and errors if the lock was not held. Single-connection pools are pinned by `TestMigrateSupportsSingleConnectionPool`.

fforootd:
All review feedback is addressed and every thread is resolved; the branch is green on both integration sweeps.

- **Postgres lock** (@adlerhurst): the final design is f717552f's goose `WithSessionLocker` — `pg_advisory_lock` held on the very connection goose migrates on. A transaction-scoped lock can't wrap goose's run (each migration commits its own transactions on that connection, no nesting), and the crash-cleanup property holds for session locks too: disconnect releases them, and `SessionUnlock` verifies the release. Single-connection pools are pinned by `TestMigrateSupportsSingleConnectionPool`.
- **Spanner lease expiry** (@fforootd): renewal + fencing in b0e534f7 — the holder re-stamps `acquired_at` every 60s, and goose runs under a cancel-cause context that fences it off the moment the lease row names another holder or renewal fails past a full lease window. Waiter polling is 1s; the holder id is a `mig_<ulid>`.
- **ADR 011 identifier contract** (@fforootd): 2ab19e84 moves generation to the dialect-owned `spanner.NewEphemeralID` (per ADR 028) and amends ADR 011 with the Spanner client-generated exception — replay-determinism rationale plus an explicit collision plan (fail the operation, retry at flow level). Sequence-backed pre-allocation is noted as the follow-up that would also cover sessions/tokens.
- **Centralized id generation** (@adlerhurst): the lease holder uses `idgen.NewULID`; the INT64 ephemeral ids stay dialect-owned per ADR 028, since a ULID/UUID string cannot fit those columns.
- **Lock table lifecycle** (@adlerhurst): documented on `lockTable` — created lazily by `ensureLeaseRow`, deliberately outside the goose chain.

Verification on the merged tree (including #533's crypto-keys migrations running under the locks): postgres 20/20 concurrent-migration rounds plus a fresh-schema sweep (22 packages ok, single-connection test included); spanner two 10-round lease races plus full sweeps (21 packages ok); zero migration-race errors throughout.

internal/storage/database/dialect/spanner/ephemeral_id.go
```diff
@@ -0,0 +1,36 @@
...
+// client-side; per ADR 028 the generation strategy is dialect-owned, which is
+// why this lives here and not in idgen. Uniform randomness spreads keys like
+// bit-reversed identities do, avoiding range hotspots.
+//
+// Uniqueness is probabilistic (collision odds ~2^-63 per insert against the
+// existing keyspace). A collision surfaces as a key-conflict error on the
+// insert; retrying inside the same transaction would be non-deterministic on
+// replay, so callers let the operation fail and leave the retry to the flow
+// level. See ADR 011 § Spanner client-generated exception.
+func NewEphemeralID() (int64, error) {
```
adlerhurst: use an id generated by the database instead of adding this slop code. spanner already has id generation functions built in.

internal/storage/database/repository/auth_attempt.go
```diff
@@ -446,15 +456,19 @@ func (a *spannerAuthAttempt) Create(ctx context.Context, client database.QueryEx
 		ttlNanos = new(attempt.TimeToLive.Nanoseconds())
 	}
 
-	var attemptID database.Identity
-	err := client.QueryRow(ctx,
-		`INSERT INTO auth_attempts (project_id, required_checks, time_to_live, session_id, created_at) VALUES ($1, $2, $3, $4, $5) THEN RETURN id`,
-		attempt.ProjectID, req, ttlNanos, sessionIDArg(attempt.SessionID), now).
-		Scan(&attemptID)
+	// Dialect-generated id: a database identity read back via THEN RETURN
+	// diverges on transaction replay (see spanner.NewEphemeralID).
```
adlerhurst: why should it be an issue if the id is regenerated on a transaction retry? IMO it doesn't make a difference. The tests can just check if the id is not empty and all good.

internal/storage/v2/dialect/spanner/project_test.go
```diff
@@ -39,11 +42,14 @@ func TestProjectStatements_CRUD(t *testing.T) {
 
 	stmts := pool.(*Client).Statements()
 
+	// Unique per run: the database may be shared with other test packages and
+	// persist across runs, so a fixed ID would collide.
 	project := &domain.Project{
-		ID:             "proj_v2_crud",
+		ID:             "proj_v2_crud_" + strings.ReplaceAll(t.Name(), "/", "_") + "_" + ulid.Make().String(),
```
adlerhurst: adding a new library just for this doesn't make sense. If it really has to be unique just use time.Now()

adlerhurst (changes requested):
The code touched is too much. I expected the changes to be in the storage package and the adr

internal/storage/v2/dialect/spanner/project_test.go
fforootd: Simplified in 8993c037: the id is now t.Name() plus time.Now().UnixNano(), stdlib only. (ulid came in via an earlier review pass and is already a transitive dependency through idgen, but it is indeed unnecessary here.)

internal/storage/database/repository/auth_attempt.go
fforootd: It unfortunately does make a difference, independent of any test assertion: go-sql-spanner retries an aborted read-write transaction by replaying its statements and comparing results (checksum_row_iterator.go in the driver), and a replayed `THEN RETURN id` that yields a different identity fails the comparison, so the driver aborts the whole transaction with `ErrAbortedDueToConcurrentModification` — the write is lost, not just renamed. That was the root cause of the repository auth-attempt flakes this PR chased down: one subtest failed per shared-emulator sweep before the change, zero across six sweeps after, with no test asserting id values anywhere. Happy to walk through the driver trace if useful.

internal/storage/database/dialect/spanner/ephemeral_id.go
fforootd: Database-generated was the starting point, and it is what breaks under the driver's transaction replay (see the sibling thread): any in-transaction draw — `IDENTITY`, `GENERATE_UUID()`, or `GET_NEXT_SEQUENCE_VALUE` — produces a different value on replay and fails the retry, and `GENERATE_UUID()` is additionally a STRING, which ADR 011 rejects for ephemeral ids to stay INT64-aligned with Postgres. The one genuinely database-generated deterministic option is pre-allocating from a bit-reversed sequence *outside* the transaction, which I proposed as follow-up (it would cover sessions/tokens too); the current client-side value implements the ADR-amendment path fforootd requested in the earlier thread. Whichever direction you two settle on, I can implement it.

fforootd: Moot as of #573: the v1 repository this generator served is deleted, so the helper and the ADR amendment are dropped from this PR (merge 3ddb01d5). Your v2 statements get the database-generated ids you wanted safely, because the native client's ReadWriteTransaction re-runs the whole closure on abort instead of replaying recorded statements — verified with shared-emulator sweeps on the merged tree: zero auth-attempt failures. The PR is now only the migration serialization.

internal/storage/database/repository/auth_attempt.go
fforootd: Settled by #573 rather than by either of us: under go-sql-spanner's statement-replay retry the regenerated id failed the whole transaction (that was the measured flake), but your v2 statements run inside the native client's ReadWriteTransaction whose retry re-runs the entire closure, so a fresh id is used consistently and database-generated ids are safe there. Confirmed empirically on the merged tree — shared-emulator sweep with zero auth-attempt failures. The v1 code this thread discussed is gone from the PR.

adlerhurst (approved):

PR #584, opened 2026-07-22
fix(api): normalize ogen decode errors to req.invalid

livio-a:
## Summary

- Normalize `OgenErrorHandler` decode/validation failures to the stable `req.invalid` domain message instead of echoing raw ogen `err.Error()` text (ADR 030).
- Include the Copilot follow-up for [#583](https://github.com/zitadel/nextgen/pull/583): name only the offending field on malformed flow `Fields` JSON (`invalid JSON for field "…"`), without wrapping `json.Unmarshal` parser text.

## Note

#583 auto-merged before Copilot’s review comment was addressed. The DX fix had been pushed to `flow-error` and replied on the review thread, but that commit was not in the merge. It is included here so the review is closed out on `main`.

## Validation

- `go test ./internal/api/ -count=1 -run 'TestOgenErrorHandler|TestSubmitFlow'`

## Release notes / changeset

Changeset: `.changeset/normalize-ogen-decode-errors.md` — `@zitadel/server` patch: normalize ogen decode messages and keep safe field names on flow `Fields` decode errors.

## Notes

- Remaining ADR 030 follow-ups: #48 (storage), contract hardening (status map dedup, selective `details`, schema sensitivity).

### DIFF
--- a/internal/api/flow.go
+++ b/internal/api/flow.go
@@ -97,7 +97,9 @@ func (h *Handler) SubmitFlowStep(ctx context.Context, req *api.FlowSubmitRequest
 	if fields, ok := req.Fields.Get(); ok {
 		decoded, err := decodeFlowFields(fields)
 		if err != nil {
-			return errorResponseWithStatusCode(http.StatusBadRequest, domain.ErrRequestInvalid()), nil
+			// Name the offending field only — do not surface json.Unmarshal text (ADR 030).
+			return errorResponseWithStatusCode(http.StatusBadRequest,
+				domain.ErrRequestInvalid().WithMessage(err.Error())), nil
 		}
 		submitReq.Fields = decoded
 	}
@@ -448,7 +450,7 @@ func decodeFlowFields(raw map[string]jx.Raw) (map[string]any, error) {
 	for k, v := range raw {
 		var decoded any
 		if err := json.Unmarshal(v, &decoded); err != nil {
-			return nil, fmt.Errorf("decode field %q: %w", k, err)
+			return nil, fmt.Errorf("invalid JSON for field %q", k)
 		}
 		out[k] = decoded
 	}
--- a/internal/api/flow.go
+++ b/internal/api/flow.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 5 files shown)


internal/api/flow.go
```diff
@@ -97,7 +97,9 @@ func (h *Handler) SubmitFlowStep(ctx context.Context, req *api.FlowSubmitRequest
 	if fields, ok := req.Fields.Get(); ok {
 		decoded, err := decodeFlowFields(fields)
 		if err != nil {
-			return errorResponseWithStatusCode(http.StatusBadRequest, domain.ErrRequestInvalid()), nil
+			// Name the offending field only — do not surface json.Unmarshal text (ADR 030).
+			return errorResponseWithStatusCode(http.StatusBadRequest,
+				domain.ErrRequestInvalid().WithMessage(err.Error())), nil
 		}
```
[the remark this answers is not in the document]
livio-a: Done — took the stricter path: `WithParent` is a log-safe `flowFieldDecodeError` (safe `Error()` + `LogValue` with only `kind`/`field`). It still `Unwrap`s to the `json.Unmarshal` cause for `errors.As`, but we do **not** attach/log the raw unmarshal string (can embed payload fragments / PII).

internal/api/flow.go
```diff
@@ -448,7 +450,7 @@ func decodeFlowFields(raw map[string]jx.Raw) (map[string]any, error) {
 	for k, v := range raw {
 		var decoded any
 		if err := json.Unmarshal(v, &decoded); err != nil {
-			return nil, fmt.Errorf("decode field %q: %w", k, err)
+			return nil, fmt.Errorf("invalid JSON for field %q", k)
 		}
```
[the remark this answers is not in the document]
livio-a: Agreed. `decodeFlowFields` now returns `*flowFieldDecodeError`: sanitized `Error()`, `Unwrap()` to the original cause, and `LogValue` that omits the unmarshal text so logs stay free of payload/PII fragments.

internal/api/flow.go
```diff
@@ -443,12 +447,36 @@ func toFlowStepComplete(c domain.FlowStepComplete) api.FlowStepComplete {
...
+// Error() names only the field for clients; Unwrap preserves the json.Unmarshal
+// cause for errors.Is/As. LogValue omits the cause string (it can embed payload
+// fragments / PII).
+type flowFieldDecodeError struct {
+	field string
+	err   error
+}
+
+func (e *flowFieldDecodeError) Error() string {
+	return fmt.Sprintf("invalid JSON for field %q", e.field)
```
fforootd: Could we call this `invalid value for field %q` instead? By the time `decodeFlowFields` runs, ogen `RawAppend` has already syntax-validated the field value. For example, the valid body `{"action":"submit","fields":{"score":1e1000}}` reaches this branch because `encoding/json` cannot represent the number as a `float64`, but the response currently says the JSON itself is invalid. Truly malformed JSON is rejected earlier by ogen and gets the generic `req.invalid` message. A small regression test with `1e1000` would keep those two paths distinct.

livio-a: Makes sense... let me / cursor change it

livio-a: Updated — message is now `invalid value for field %q`, with a `1e1000` regression test so it stays distinct from ogen’s structural `req.invalid` path.

fforootd (approved):

Issue #410, opened 2026-06-24
Project API E: API handlers


IAM-marco:
Part of #405.

## Goal

Wire the remaining endpoints in the API handler layer: `patchProject` and `queryProjects`, and make `getProject` return the full response. This connects the generated server interface to the service layer. The two new handlers must apply the authorization guard from PR #580.

## Decisions that shape this work

See the epic (#405) for the full list. The ones that affect the handlers:

- The contract already shipped (PR #462, adjusted by PR #580). The operations are `patchProject` (`PATCH /projects/{id}`, body `{ name }`, partial) and `queryProjects` (`POST /projects/query`). There is no `GET /projects`.
- `queryProjects` takes a request body with `limit`, `page_token`, `sorting { field, direction }`, and `filter[] { field, value, operation }`. Filter fields are `name` and `createdAt`.
- Get returns `id`, `name`, `previewOrigins`, `createdAt`, `updatedAt` (per the shipped `get-project-response`).
- No delete endpoint. Do not add a delete handler.

## Authorization (PR #580)

PR #580 bound the management API to the token's project and locked out the preview secret. It states that `patchProject` and `queryProjects` "pick the guard up when implemented". Concretely:

- `PatchProject`: call `requireProjectAccess(ctx, projectID, projectAccess, opWrite)` first. A foreign project and a nonexistent project must answer identically with `proj.not_found` (the anti-oracle rule; for projects both miss answers are `domain.ErrProjectNotFound()`). Insufficient scope answers 403 `proj.permission_denied`. The mappings already exist in `projectErrorResponse`.
- `QueryProjects`: the operation has no project parameter, so it is bound to the token's project by construction (like `ListUsers`). Only the scope check applies, and the handler must restrict results to the token's project. A project secret lists at most its own project.
- The preview secret must be rejected on both operations. The `projectAccess` row accepts only `project.write`; `project.read` is the browser-plane preview secret's scope.
- The spec-side security fix for `queryProjects` (`project.read` to `project.write`) is tracked in #624, which blocks this issue.

## Dependencies

- #624 (amendments from the PR #580 authorization model). Blocks this issue: the `queryProjects` security fix should land before or together with the handlers.
- #408 (storage, in review as PR #601). The service methods exist (PR #579), but the `UpdateProject` statement is not implemented until #601 lands.
- Closed and already available: #406 (contract and generated types), #407 (domain model), #409 (service methods).

## Test first (TDD)

[...]

PR #625, opened 2026-07-24
fix(api): require `project.write` for queryProjects

### LINKED ISSUE #624: Project API F: amendments from the PR #580 authorization model
Part of #405.

## Goal

Amend already-shipped Project API pieces so they match the authorization model from PR #580. That PR bound the management API to the token's project and locked the preview secret out of management operations. Two shipped pieces need small amendments. This ticket exists so we do not edit closed tickets (#406, #409) or the in-review storage work (#408).

## Current state (after PR #580)

- `internal/api/authz.go` has `requireProjectAccess(ctx, projectID, resourceAccess, op)`. The `projectAccess` row lists no finer scopes: only `project.write` reaches project management operations. `project.read` is the browser-plane preview secret's scope and must not gate a management read.
- `getProject` (`api/openapi/endpoints/projects/by_id/methods.yaml`) was changed by #580 to require `project.write`, with a comment explaining why.
- `queryProjects` (`api/openapi/endpoints/projects/query/methods.yaml`, shipped by PR #462 under #406) still declares `security: oauth2: [project.read]`. This is drift: with that scope the preview secret could query projects.
- `ProjectService` gained `Update`, `List`, and `Delete` through PR #579 (under #409). `queryProjects` has no project parameter, so the handler must restrict results to the token's project (like `ListUsers`). The service needs to support that restriction.

## Dependencies

- PR #580 (merged). This ticket applies its model to shipped pieces.
- Blocks #410 (API handlers). The spec security fix should land before or together with the handlers.
- Do not change #408 or its PR #601. No storage change is required by #580. The reviewer of #601 should only be aware that the handlers will call the statements behind the guard.

## Test first (TDD)

Extend the tests PR #580 introduced, before implementing:

- `internal/api/authz_internal_test.go`: scope table entries showing that only `project.write` reaches `queryProjects` and that `project.read` (the preview secret) is denied.
- `internal/api/integration_test/authz_test.go`: a preview-secret probe against `queryProjects`.
- Service unit tests showing `List` restricted to a given project id, and the default and maximum limits applied.

Then implement.

## Tasks

- Amendment 1 (contract, shipped under #406 by PR #462): change the `queryProjects` security from `project.read` to `project.write` in `api/openapi/endpoints/projects/query/methods.yaml`. Copy the explanatory comment style from `getProject` in `by_id/methods.yaml`. Regenerate `api/generated` with `go generate ./...` in `api/`.
- Amendment 2 (service, shipped under #409 by PR #579): verify `ProjectService.List` can restrict results to the token's project. If it cannot express that restriction, add it. Also verify the service applies the default limit when the caller sends none and enforces the maximum limit (the storage layer treats a limit of 0 as "no limit", so this must happen in the service). The shared values are in `api/openapi/components/schemas/limit.yaml` (default 20, maximum 100).

## Reference

`internal/api/authz.go` (the guard and the `projectAccess` row), `api/openapi/endpoints/projects/by_id/methods.yaml` (the `project.write` comment to copy), `internal/service/project.go` (the merged service), PR #580 (the access model and its tests).

grvijayan:
## Summary

Amendments from PR #580's authorization model (#624):

- queryProjects declared `project.read`, the browser-plane preview secret's scope, which must not gate an operator read. Flip the contract to `project.write` and regenerate.
- `ProjectService.List` gains an opt-in `ProjectID` restriction so a project-bound caller sees only its own project; left empty the list spans all projects for a broader-scoped caller.
- Adds a scope-table test case in `authz_internal_test.go` pinning that queryProjects maps to `project.write`, plus a skipped integration probe in `integration_test/authz_test.go` for #410 to un-skip once the queryProjects handler exists. No docs changes.

## Validation
- `moon run server:generate` (OpenAPI regen; only the queryProjects scope changed)
- `go test ./...`

## Release notes / changeset
- Changeset: `.changeset/project-query-authz-amendments.md`

## Notes

Closes https://github.com/zitadel/nextgen/issues/624

### DIFF
--- a/.changeset/project-query-authz-amendments.md
+++ b/.changeset/project-query-authz-amendments.md
@@ -1,0 +1,8 @@
+---
+"@zitadel/server": patch
+---
+
+`queryProjects` now requires the `project.write` scope. Its contract still
+declared `project.read`, the browser-plane preview secret's scope, which must
+not gate an operator read — aligning it with the project management access
+model from ADR 037.
--- a/.changeset/project-query-authz-amendments.md
+++ b/.changeset/project-query-authz-amendments.md
@@ -5,4 +5,4 @@
 `queryProjects` now requires the `project.write` scope. Its contract still
 declared `project.read`, the browser-plane preview secret's scope, which must
 not gate an operator read — aligning it with the project management access
-model from ADR 037.
+model from ADR 036.
--- a/api/generated/oas_security_gen.go
+++ b/api/generated/oas_security_gen.go
@@ -154,7 +154,7 @@ var oauth2ScopesOAuth2 = map[string][]string{
 		"project.write",
 	},
 	QueryProjectsOperation: []string{
-		"project.read",
+		"project.write",
 	},
 	RevokeSessionOperation: []string{
 		"session.delete",
--- a/api/openapi/endpoints/projects/query/methods.yaml
+++ b/api/openapi/endpoints/projects/query/methods.yaml
@@ -3,9 +3,12 @@ post:
   summary: Query projects
   tags:
     - Projects
-  security: 
+  security:
     - oauth2:
-        - project.read
+        # project.write, not project.read: project.read is carried by the
+        # browser-plane preview secret, so it cannot gate an operator read
+        # until ADR 036 splits the credential planes into distinct schemes.
+        - project.write
   requestBody:
     description: Request to query projects.
     required: true
--- a/internal/api/authz_internal_test.go
+++ b/internal/api/authz_internal_test.go
@@ -64,6 +64,9 @@ func TestResourceAccessScopes(t *testing.T) {
 		{"team.read cannot write", teamAccess, opWrite, "team.read", false},
 
 		{"project.read does not read project management state", projectAccess, opRead, "project.read", false},
+		// queryProjects has no project parameter of its own and binds to the
+		// token's project like getProject: only project.write reaches it.
+		{"project.write reaches queryProjects (opRead)", projectAccess, opRead, "project.write", true},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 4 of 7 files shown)


.changeset/project-query-authz-amendments.md
```diff
@@ -0,0 +1,8 @@
+---
+"@zitadel/server": patch
+---
+
+`queryProjects` now requires the `project.write` scope. Its contract still
+declared `project.read`, the browser-plane preview secret's scope, which must
+not gate an operator read — aligning it with the project management access
+model from ADR 037.
```
IAM-marco: Claude:

This cites ADR 037, which is [Token Lifecycle](https://github.com/zitadel/nextgen/blob/8c2e7f26b1af78641a7b97dcb5b30ac40650e8a4/docs/adrs/037-token-lifecycle.md) and unrelated. The project management / credential-plane access model is [ADR 036 — API Credential Planes](https://github.com/zitadel/nextgen/blob/8c2e7f26b1af78641a7b97dcb5b30ac40650e8a4/docs/adrs/036-api-credential-planes.md), which explicitly governs the project query/patch surfaces and the preview secret. This PR's own spec comment already cites 036 ([methods.yaml:5-8](https://github.com/zitadel/nextgen/blob/8c2e7f26b1af78641a7b97dcb5b30ac40650e8a4/api/openapi/endpoints/projects/query/methods.yaml#L5-L8)), so the changeset should read ADR 036 to match.

IAM-marco (changes requested):

.changeset/project-query-authz-amendments.md
grvijayan: oops
fixed in https://github.com/zitadel/nextgen/pull/625/commits/09305513550d2914473faaee9f828156c2c04bcb

IAM-marco (approved):

PR #629, opened 2026-07-24
chore: run spanner integration tests against a real test instance (#604) 

### LINKED ISSUE #604: Replace spanner emulator with a test instance
As we activated the spanner test the pipeline got flaky and we randomly get errors like `Transaction: 1784714109495306 aborted due to another transaction getting priority. The emulator only supports one transaction at a time.`

This is due to the fact that the emulator only supports one transaction at a time. Therefore I propose the following changes:

- change database configuration to allow a dynamic schema or database name
- create a spanner instance which is only used for tests in a google test project (once not for each test run)
- use the credentials of the spanner test instance in the `spanner_integration` tests
- extend the tests to create the database first before running the tests and delete it after the tests.

IAM-marco:
## Summary

The Cloud Spanner emulator only supports one transaction at a time, so concurrent `spanner_integration` tests are flaky (`Transaction ... aborted due to another transaction getting priority. The emulator only supports one transaction at a time`). This lets the integration suites run against a real, long-lived Spanner test instance, giving each run its own isolated database. Fixes #604.

- New `internal/storage/database/dialect/spanner/testdb` helper: `Provision` creates a uniquely named database on an existing instance (via `ZITADEL_TEST_SPANNER_INSTANCE=projects/<p>/instances/<i>`, ADC auth) and drops it on teardown. Database ids are `itest_<run-id>_<random>`, always including a random suffix so the two test binaries in one CI run never collide.
- Emulator path refactored to reuse the same `CreateDatabase` helper (emulator kept as the local default).
- Both suites (repository + API integration) wired for the new mode. Precedence: `ZITADEL_TEST_SPANNER_INSTANCE` > `ZITADEL_TEST_SPANNER_URL` > emulator container > Postgres.
- CI authenticates to Google Cloud with Workload Identity Federation on trusted runs and targets the shared instance; fork PRs fall back to the emulator. The auth + spanner steps run last in the job so their (currently expected) failure does not skip the unrelated build/journey/e2e steps.
- `CONTRIBUTING.md` documents the new env var and precedence.

> [!IMPORTANT]
> **The `Authenticate to Google Cloud (Spanner test instance)` step is expected to FAIL on this PR** and will keep failing on trusted runs until the GCP infrastructure and `GCP_*` secrets are provisioned (see prerequisite below). This is intentional: a red check is a visible signal that the real test instance is not yet configured, rather than silently masking it. It runs last so every unrelated step still executes and validates.

## Validation

- `gofmt -l` on all touched files: clean.
- `go vet -tags spanner_integration ./internal/storage/database/dialect/spanner/... ./internal/storage/database/repository/ ./internal/storage/database/dbtest/ ./internal/api/integration_test/`: clean.
- `go vet -tags postgres_integration ./internal/storage/database/repository/` (stub path): clean.
- ci.yml validated as well-formed YAML.
- Not run: the integration suite itself was not executed locally (no Docker for the emulator path / no GCP credentials for the instance path). Relying on CI's emulator run for now.

## Release notes / changeset

- No changeset required — no shipped behavior changed (test-only Go code under the `spanner_integration` build tag, CI workflow, and docs).

## Notes

**CI auth choice:** Workload Identity Federation (keyless) was chosen. The considered alternative was a service-account JSON key stored as a GitHub secret: simpler to set up, but a long-lived credential to manage and rotate. WIF avoids the standing secret.

**One-time infra prerequisite (not in this PR).** Before trusted CI actually targets the real instance, ops must provision:
1. A Google Cloud test project with the Spanner API enabled.
2. One small Spanner instance (minimal processing units / regional config).
3. A service account with **Spanner Database Admin** (create/drop databases) + **Spanner Database User** (read/write).
4. A Workload Identity Federation pool + provider bound to `zitadel/nextgen`; grant the SA `roles/iam.workloadIdentityUser`.
5. GitHub Actions config: secrets `GCP_WORKLOAD_IDENTITY_PROVIDER` and `GCP_SERVICE_ACCOUNT`, and variable `SPANNER_TEST_INSTANCE = projects/<project>/instances/<instance>`.

Until these exist, the auth step fails by design on trusted runs (see the note above). Because it runs last and the trailing artifact-upload steps use `always()`/`failure()`, no unrelated step is skipped or fails on its account.


.github/workflows/ci.yml
```diff
@@ -128,6 +129,27 @@ jobs:
...
         run: env -u CI -u GITHUB_ACTIONS moon run console-e2e:e2e-real
 
+      # Spanner integration runs last on purpose: until the GCP infrastructure
+      # and the GCP_* secrets are provisioned (see PR notes), the auth step
+      # below fails on trusted runs by design, as a visible signal that the real
+      # test instance is not configured yet. Keeping it at the end prevents that
+      # expected failure from skipping the unrelated steps above. Fork PRs skip
+      # auth and fall back to the emulator.
+      - name: Authenticate to Google Cloud (Spanner test instance)
+        if: ${{ steps.ci-mode.outputs.mode == 'full' && (github.event_name != 'pull_request' || !github.event.pull_request.head.repo.fork) }}
```
adlerhurst: wouldn't it make sense to test if the secrets are set?

adlerhurst: if they are not set we could still run the emulator tests (with `-parallel 1 -p 1`)

PR #601, opened 2026-07-22
feat: project storage

grvijayan:
## Summary

This PR introduces a required `projects.name` column, and updates `CreateProject` statement to persist the name in DB, implements `UpdateProject` in the v2 storage dialects (Postgres + Spanner), and adds cascade delete migrations so project-scoped resources are removed when a project is deleted.

## Validation

- `go test ./...`
- `go test -v -tags postgres_integration -timeout=10m ./...`
- `go test -v -tags spanner_integration -timeout=10m ./...`

## Release notes / changeset

- Changeset: `.changeset/project-name-column-delete-cascade-migrations.md`

## Notes
- Closes https://github.com/zitadel/nextgen/issues/408

PR #634, opened 2026-07-26
fix: mirror the real login flow, and add a real-data console dev loop

bastionstack:
# Which Problems Are Solved

- `@zitadel/api-mock` served a **combined email+password card**, but the real default flow (`packages/config/defaults/default-login.json`) is a **split** `identifier` → `password`. That mock backs the orchestrator specs, the Storybook orchestrator story, and both demo e2e suites — so all of them were proving a screen the server never emits.
- The mock keyed credential fields `password`; the real flow declares `x-auth-methods#password` and the server answers `req.invalid` to the short name. The orchestrator could submit a key the backend refuses with no mock-backed test able to catch it.
- The console users list and detail read a `username` property no shipped user schema defines, so every row fell back to the email and sort-by-name was really sort-by-email. The specs passed because their fixtures invented `username` too.
- No defined way to develop the console against real data, even though it manages an instance — the pieces existed but were buried in `apps/console-e2e` as Playwright `webServer` entries.

# How the Problems Are Solved

**api-mock mirrors the real flow**

- `identifierStep` collects the email only; machine wires `identifier --submit--> password --submit--> done`.
- Added the `back` action the engine injects on any step with a predecessor (ADR 022).
- Credential fields on `password` and `register-password` use the schema pointer, exported as `PASSWORD_FIELD`.
- Invalid-credential injection moved from `identifier` to `password` — the first request that can fail authentication — reading the address from flow context.
- Consumers updated to the two-step walk: orchestrator specs, both demo e2e suites, Storybook story docs. One liquid test renamed to claim *template capability*, not default-flow shape.
- Two exceptions kept and documented: the identifier's `register`/`recover` links (so those screens stay reachable), and the `passkey-upsell` pair (direct actor injection).
- `packages/api-mock/AGENTS.md` names the flow definition as the authority and requires fixtures be diffed against it, so this cannot drift again silently.

**Console display name**

- New `userDisplayName` mirrors the platform's `User.DisplayName` (`internal/domain/user.go`, behind `GET /sessions/me`'s `name`): explicit `name`, else given + family joined, camelCase first, snake_case accepted; falls back to email then id.
- Shared by the list and detail screens. Specs rewritten onto real schema shapes, plus a regression test that an unrelated attribute is never read as a name.

**Real-data console dev loop**

- `moon run console:dev-real` boots an ephemeral real instance (binary + embedded Postgres, no Docker) via `@zitadel/testing`, bootstraps a project, seeds a fixed sign-in user plus eight list users, and starts Vite bound to it — printing the credentials. `--seed-only` skips Vite. The project secret goes to the child via env, never to disk.
- `apps/console/README.md` + `AGENTS.md` document all three backends and why the real one is needed: `listUsers` requires `user.read`, carried only by the project secret — the publishable key is deliberately refused (`internal/api/user.go`).

# Additional Changes

- `CLAUDE.md` is now tracked, mirroring the already-tracked `.cursorrules` (the repo ignores `.cursor/`, `.claude/`, `.agents/` but tracks the root bootstrap pointing at scoped `AGENTS.md` files). It was untracked-but-not-gitignored, so a root `git add -A` would have committed personal content. Per-developer rules moved out; skill references reworded because those skills live in the gitignored `.claude/skills/`.

# Additional Context

- Rebased onto `main` after #617 was squash-merged. The earlier diff here included that PR's commits, because these console fixes build on its `_authed` route layout; it is now 23 files. A Go test fix that was needed while #617 was open (#601 had removed the `flowDefinitionRepo` return from `createMockedProjectService`) was dropped — #617 fixed it on its own rebase.
- Subsequently updated with `main` to pick up #603, which also touches `orchestrator/liquid.spec.ts` and `orchestrator/zitadel-login.ts`. The two sets of changes to that spec file are disjoint (the renamed template-capability test survives intact, not duplicated) and CI is green on the merged state.
- No changeset: `@zitadel/api-mock` and `@zitadel/console` are changeset-ignored, and the only publishable package touched (`@zitadel/components`) is spec-only.

## Validation

Re-run in full after the rebase onto post-#617 `main`; CI is additionally green on the merged state that includes #603:

- `moon ci :lint :typecheck :build :test :test-browser` — 28 tasks, zero failures (api-mock 40, components 357 + 46 browser, storybook 20, console 25)
- `go build ./...`, `go test ./cmd/server ./internal/service`, `gofmt -l cmd internal` — clean
- `demo-next-e2e:e2e`, `demo-nuxt-e2e:e2e` — two-step walk in a real browser
- `console-e2e:e2e-real` (3/3, incl. seeded user in list + detail)
- Manual via `console:dev-real`: signed in through the split flow (identifier → password, `Back` rendering), nine seeded users with real names and initials in light + dark, no console errors


peintnermax (approved):
lgtm. dev-real is handy 👍

PR #637, opened 2026-07-27
feat: delete user

### LINKED ISSUE #631: Frontend: User detail — overview, authentication, and project permissions
## Context

## Delivered

## Out of scope

## Blocked

## Reference

wim07101993:
## Summary

- Added endpoint to delete a user

## Validation

- go test ./... -tags postgres_integration

## Release notes / changeset

- Changeset: .changeset/delete-user-endpoint.md

## Notes


elinashoko:
oh awesome @wim07101993 tnx for this, it was on my to-do to create a ticket :D

api/openapi/endpoints/users/by_id/methods.yaml
```diff
@@ -40,3 +40,41 @@ get:
...
+    parameters:
+      - $ref: ../../../components/parameters/project-id.yaml
+      - name: user_id
+        in: path
+        required: true
+        schema:
+          $ref: ../../../components/schemas/user-id.yaml
+    security:
+      - oauth2:
+          - user.write
```
grvijayan: we could have a separate permission for user deletion, similar to `session.delete` and `flow_definition.delete`, what do you think?

internal/api/user.go
```diff
@@ -35,6 +35,22 @@ func (h *Handler) CreateUser(ctx context.Context, req *api.User, params api.Crea
 	return convertUsingJson[api.CreateUserResponse](u)
 }
 
+func (h *Handler) DeleteUserByID(ctx context.Context, params api.DeleteUserByIDParams) (api.DeleteUserByIDRes, error) {
```
grvijayan, line 39: Claude:

Deleting a user who has ever logged in returns 500

`sessions.user_id` references `users(project_id, id)` with no `ON DELETE` clause — so NO ACTION:

- internal/storage/database/dialect/postgres/migration/sql/000007_user_agents_and_sessions.sql:27
- internal/storage/database/dialect/spanner/migration/sql/000007_user_agents_and_sessions.sql:25

Every other referencing table is handled: tokens, credentials, and the attribute tables cascade; team_memberships is ON DELETE RESTRICT, which is exactly why `DeleteUserByID` deletes memberships explicitly first. Sessions were missed. Any user with a session row fails the DELETE FROM users with an FK violation, and the action funnels that into domain.ErrInternal(err).WithMessage("failed to delete user") → 500.

grvijayan: we could add an integration test as well if we're fixing this

internal/service/user.go
```diff
@@ -316,6 +326,33 @@ func (o *SetPasswordUserAction) Apply(ctx context.Context, tx StatementerWithQue
...
+}
+
+func (o *DeleteUserAction) Prepare(_ context.Context, _ database.QueryExecutor) error {
+	return nil
+}
+
+func (o *DeleteUserAction) Apply(ctx context.Context, tx StatementerWithQueryExecutor[AllStatements]) error {
+	err := tx.Statements().DeleteUserByID(ctx, o.ProjectID, o.UserID)
+	if err != nil {
+		if _, ok := errors.AsType[*database.NoRowFoundError](err); ok {
```
grvijayan, line 384: we could make the deletes idempotent (similar to other resources), not return an error if it's `NoRowFoundError`

internal/api/user.go
```diff
@@ -35,6 +35,22 @@ func (h *Handler) CreateUser(ctx context.Context, req *api.User, params api.Crea
...
+	if err := requireProjectAccess(ctx, string(params.ProjectID), userAccess, opWrite); err != nil {
```
grvijayan: the handler layer checks for `opWrite` while the api definition has `user.delete` scope defined. 

i see [userAccess](https://github.com/zitadel/nextgen/blob/main/internal/api/authz.go#L79-L80) has only read/write defined at the moment, so, we'll need to add delete as well (please also add tests similar to the other resources when you update authz)

internal/storage/v2/dialect/postgres/migration/sql/000007_user_agents_and_sessions.sql
```diff
@@ -24,7 +24,10 @@ CREATE TABLE zitadel_nextgen.sessions (
     , PRIMARY KEY (project_id, id)
     , UNIQUE (project_id, token_id)
     , FOREIGN KEY (project_id) REFERENCES zitadel_nextgen.projects(id) ON DELETE CASCADE
-    , FOREIGN KEY (project_id, user_id) REFERENCES zitadel_nextgen.users(project_id, id)
+    , CONSTRAINT fk_sessions_user
```
grvijayan, line 27: fyi, Flo expressed concern in one of my PRs about editing existing migrations instead of adding new ones for changes ([comment](https://github.com/zitadel/nextgen/pull/646#discussion_r3681979810)). If this is a deliberate decision, please just note it down in the PR description, or add a new migration.

internal/api/user.go
wim07101993: 🤦

internal/storage/v2/dialect/postgres/migration/sql/000007_user_agents_and_sessions.sql
wim07101993: Yes, I know my answer is the same as yours in that pr :)

internal/api/user.go
```diff
@@ -35,6 +35,22 @@ func (h *Handler) CreateUser(ctx context.Context, req *api.User, params api.Crea
 	return convertUsingJson[api.CreateUserResponse](u)
 }
 
+func (h *Handler) DeleteUserByID(ctx context.Context, params api.DeleteUserByIDParams) (api.DeleteUserByIDRes, error) {
+	if err := requireProjectAccess(ctx, string(params.ProjectID), userAccess, opDelete); err != nil {
```
grvijayan, line 40: `userAccess` currently has only `user.read` and `user.write` scopes - this also needs to be updated for the new delete scope to be usable when we have tokens with finer scopes (although at the moment everything is project-scoped and would be covered anyway). 

https://github.com/zitadel/nextgen/blob/658ce78926b96240bcae583fee2e042283991b30/internal/api/authz.go#L77-L85

grvijayan (approved):

PR #640, opened 2026-07-27
feat: list user passkeys

### LINKED ISSUE #631: Frontend: User detail — overview, authentication, and project permissions

wim07101993:
## Summary

- Added list passkeys endpoint

## Validation

- `go run ./... -tags postgres_integration`

## Release notes / changeset

- Changeset: .changeset/list-user-passkeys.md


internal/service/user.go
```diff
@@ -169,6 +175,40 @@ func (s *UserService) ListUsers(ctx context.Context, input ListUsersInput) ([]ma
...
 
+func (s *UserService) ListPasskeys(ctx context.Context, input ListPasskeysInput) ([]*domain.UserPasskey, error) {
+	passkeys, err := s.v2Pool.Statements().ListUserPasskeys(
+		ctx, &v2database.ListOptions[domain.UserPasskeyField]{
+			Filter: v2database.And(
+				v2database.Equal(v2database.Col(domain.UserPasskeyFieldProjectID), input.ProjectID),
+				v2database.Equal(v2database.Col(domain.UserPasskeyFieldUserID), input.UserID),
+			),
+			Pagination: v2database.Page[domain.UserPasskeyField]{
+				Limit: 100, // arbitrary limit of 100, assuming nobody has 100 passkeys attached to their user
```
grvijayan: we could add an `OrderBy` similar to `ListUsers` query to keep the ordering deterministic when the results go over one page. 

also, we don't have to hardcode the limit if we define it in the api definition and apply defaults

internal/api/integration_test/user_test.go
```diff
@@ -423,3 +423,96 @@ func TestGetMyUser(t *testing.T) {
...
+
+			if assert.IsType(t, &api.ListUserPasskeysResponse{}, resp, helpers.MustMarshal(t, resp)) {
+				passkeys := resp.(*api.ListUserPasskeysResponse)
+				require.Len(t, passkeys.Passkeys, 0)
+			}
+		})
+	})
+
+	t.Run("error", func(t *testing.T) {
+		t.Run("unknown user returns 404", func(t *testing.T) {
```
grvijayan: as we have just one test case running here, one of the two`t.Run()`s is redundant

api/openapi/endpoints/users/by_id/passkeys/methods.yaml
```diff
@@ -0,0 +1,58 @@
+get:
```
grvijayan, line 1: we can define limit, page_token, etc., in the api definition (per adr 031) and then apply default limits if unset in the service layer - we did this recently in [ProjectService](https://github.com/zitadel/nextgen/blob/main/internal/service/project.go#L286) (there are generic helpers for querying in `list.go`)

internal/service/user.go
```diff
@@ -169,6 +175,40 @@ func (s *UserService) ListUsers(ctx context.Context, input ListUsersInput) ([]ma
 	return users, nil
 }
 
+func (s *UserService) ListPasskeys(ctx context.Context, input ListPasskeysInput) ([]*domain.UserPasskey, error) {
```
grvijayan: please add unit tests

internal/domain/user_passkey.go
```diff
@@ -343,6 +343,7 @@ func VerifyPasskeyRegistration(challenge *PasskeyRegistrationChallenge, attestat
 		SignCount:       int64(credential.Authenticator.SignCount),
 		BackupEligible:  credential.Flags.BackupEligible,
 		BackupState:     credential.Flags.BackupState,
+		Name:            challenge.DisplayName,
```
grvijayan: Claude: 

`Name: challenge.DisplayName` uses the WebAuthn user display name, not a per-credential label. In the real flow that is the user's email (`passkey_upsell_flow_test.go:142`), and `passkeyRegistrationLabels` falls back to username then `"Passkey account"`. So every passkey a user registers gets the same name. That defeats the point of `name` on a list endpoint. The test only looks right because the helper passes distinct display names, which no production caller does.

Either plumb a real per-credential name through registration, or drop `name` from this PR.

I am unsure because my domain knowledge related to Passkeys is limited. Please check this.

internal/api/integration_test/user_test.go
```diff
@@ -423,3 +423,96 @@ func TestGetMyUser(t *testing.T) {
 		)
 	})
 }
+
+func TestListPasskeys(t *testing.T) {
+	t.Parallel()
+
+	dependencies := func() (project *domain.Project, user map[string]any, client *helpers.ApiClient) {
```
grvijayan: Claude: 

`dependencies()` closes over the outer `t` and runs inside `t.Parallel()` subtests. `require.NoError` calls `FailNow`, which `testing` only allows from the goroutine running that test. A setup failure would be attributed to the parent and would not abort the subtest. Make it `dependencies(t *testing.T)`

api/openapi/endpoints/users/by_id/passkeys/methods.yaml
```diff
@@ -0,0 +1,58 @@
...
+  operationId: listUserPassKeys
```
grvijayan:
```suggestion
  operationId: listUserPasskeys
```

internal/storage/v2/dialect/postgres/user.go
```diff
@@ -94,6 +94,13 @@ WHERE project_id = $1 AND user_id = $2
 	deleteUserStmt = `
 DELETE FROM zitadel_nextgen.users
 WHERE project_id = $1 AND id = $2
+`
+
+	userExistsQuery = `
```
grvijayan: nit: `userExistsStmt` to follow the convention of other queries in this file (also in spanner)

api/openapi/endpoints/users/by_id/passkeys/list-user-passkeys-response.yaml
```diff
@@ -0,0 +1,24 @@
...
+properties:
+  passkeys:
+    type: array
+    items:
+      type: object
+      required:
+        - id
+        - name
+        - created_at
+      properties:
```
grvijayan, line 14: list endpoints of other resources also return a `next_page_token` as part of the contract for the next fetch in case there are more results.  

session: https://github.com/zitadel/nextgen/blob/main/api/openapi/endpoints/sessions/session-list-response.yaml#L10
flow_definition: https://github.com/zitadel/nextgen/blob/main/api/openapi/components/flows/flow-definition-list-response.yaml#L9

wim07101993: I didn't add these initially because the list of passkeys will never be a large collection. And adding cursors would add bloat to the api. But for consistency sake it might make sense to add them anyways.

internal/api/integration_test/user_test.go
wim07101993: good catch

internal/service/user.go
wim07101993: IMO there is not much to unit tests here. All the logic represented here is already tested in the integration test

grvijayan (approved):

PR #644, opened 2026-07-28
docs: leave room for multiple projects in the self-hosted Console

elinashoko:
## Summary

- Clarify that the initial self-hosted Console manages one default project.
- Keep project switching and multi-project management outside the initial scope.
- Leave the deployment model open to supporting multiple projects through the same Console later.
- Align the Console agent guidance with the updated ADR wording.

## Validation

- Reviewed the branch diff against `main`.
- Confirmed the ADR no longer describes self-hosting as an “exactly one project” or “one-project world” model.
- Documentation-only change; no runtime tests required.

## Release notes / changeset

## Notes

This does not change the initial implementation: the Console still opens one default project and has no project switcher. It only avoids making that initial scope a permanent constraint.


peintnermax (approved):
lgtm 👍

PR #645, opened 2026-07-28
fix(api): align system permission catalog and OpenAPI scopes

### LINKED ISSUE #420: Define System Permission Catalog & Default Bundles
## Context

[ADR 032](https://github.com/zitadel/nextgen/blob/main/docs/adrs/032-permission-catalogs.md) defines two catalogs: a **system catalog** of Zitadel-owned permissions for internal API resources, and an **app-group catalog** for customer-owned permissions. This issue covers the system catalog side.

The system catalog replaces legacy level-specific role semantics (instance/org/project/project-grant). Since the nextgen API is resource-oriented, permissions must use stable, resource-scoped action names. [oxidel ADR-042](https://github.com/zitadel/oxidel/blob/main/docs/adr/042-projects-apps-owner-org-authz.md) established the naming pattern to follow: `<resource>.<verb>`, e.g. `project.create`, `project.read`, `project.write`, `project.delete`, `project.app.read`, `project.app.write`, `project.app.delete`. Extend this pattern to all nextgen resource types.

## Tasks

- Enumerate all internal Zitadel resource types and their required operations using the `<resource>.<verb>` naming convention from [oxidel ADR-042](https://github.com/zitadel/oxidel/blob/main/docs/adr/042-projects-apps-owner-org-authz.md)
- Define optional default bundles (e.g. `admin`, `viewer`) as RBAC-style convenience aliases over system permissions — roles are optional, not required
- Document which permissions are required for each planned API endpoint
- Update [OpenAPI security declarations](https://github.com/zitadel/nextgen/blob/main/docs/design/api/authz.md) to reference final permission names
- Staff/support privilege tiers and break-glass operator access are out of scope for this issue — tracked by [issue #333](https://github.com/zitadel/nextgen/issues/333)

## References

- [ADR 032 §1 — One model, two catalogs](https://github.com/zitadel/nextgen/blob/main/docs/adrs/032-permission-catalogs.md#1-one-model-two-catalogs)
- [ADR 033 §1 — System catalog](https://github.com/zitadel/nextgen/blob/main/docs/adrs/033-internal-permission-management.md#1-system-catalog)
- [Authorization Design](https://github.com/zitadel/nextgen/blob/main/docs/design/api/authz.md)
- [issue #333 — Cross-Project Identity and Collaboration](https://github.com/zitadel/nextgen/issues/333)
- [oxidel ADR-042: Projects and Apps Use Owner-Org AuthZ](https://github.com/zitadel/oxidel/blob/main/docs/adr/042-projects-apps-owner-org-authz.md)

livio-a:
## Summary

- Define the canonical system permission catalog with singular `{resource}.{verb}` names, default bundles, credential constraints, and endpoint mappings.
- Separate branding, domain, feature, allowed-origin, signing-key, and webhook permissions from `project.*`, while retaining project grants as their scope boundary.
- Align OpenAPI endpoint security, OAuth declarations, generated code, and authorization checks with singular `flow_definition.*`, `session.*`, and `auth_attempt.*` scopes.
- Align supporting API design documentation, including the shipped flat `/branding` surface and `/schemas` inventory.
- Address review feedback: bump the scope-alignment changeset to `minor`, document that `project.write` umbrella behaviour and `HandleOAuth2` scope checks are not yet enforced, and disambiguate the session credential (`nextgenSession` / `__nextgen_session`) from operator `session.*` scopes on by-id endpoints.

## Validation

- `go generate ./...`
- `npx --yes @redocly/cli@2.31.5 lint api/openapi/openapi-spec.yaml --format=github-actions`
- `go build ./...`
- `go test ./internal/api/... -count=1`
- `git diff --check`

## Release notes / changeset

- Changeset: `.changeset/openapi-permission-scope-alignment.md` — `@zitadel/server` **minor**; renaming plural OAuth scopes to singular is a breaking contract change for clients minting the old names.

## Notes

- The create/delete verb-model split and `user.set_password` endpoint wiring remain a follow-up; `PUT /users/{id}/password` still requires `user.write` today.
- `domain.*`, `feature.*`, `allowed_origin.*`, `signing_key.*`, and `webhook.*` are declared final scope names; their resource-specific endpoints and enforcement remain planned API work.
- The current authorization compatibility layer still lets legacy `project.write` reach finer resource scopes (`scopeAllowed` in `internal/api/authz.go`). The catalog intentionally does not define that as permission inheritance; `authz.md` and the catalog drift notes now call this out explicitly.
- Declaring a scope on an operation does not yet gate the request: `HandleOAuth2` verifies the bearer but does not consult the per-operation requirement list. Enforcement only happens where handlers call `requireProjectAccess`. Session and auth-attempt handlers do not today. Wiring that check belongs with the verb-model follow-up.
- Related bug found while clarifying session security schemes: `DELETE /sessions/{session_id}` cleared the caller's `__nextgen_session` cookie and signed console admins out when revoking any session. Fixed separately in #658.
- Remaining product questions (SCIM management, billing endpoints, project creation gating, and possible user directory/PII separation) are recorded in the catalog and do not block review.
- Staff/support access remains out of scope (#333).

Closes #420 partially.


api/openapi/endpoints/sessions/methods.yaml
```diff
@@ -19,7 +19,7 @@ post:
     - Sessions
   security:
     - oauth2:
-        - sessions.write
+        - session.write
```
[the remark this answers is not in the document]
livio-a: Addressed in f23afe75.

`createSession` now states that the returned `session_token` is an end-user session credential delivered as the `__nextgen_session` cookie and authorises `GET`/`DELETE /sessions/me` under the `nextgenSession` scheme. The by-id operations remain operator endpoints behind `session.read` / `session.delete`. `revokeSession`'s description and 401 wording were updated to match.

A related behaviour bug (by-id revoke clearing the caller's cookie) is fixed separately in #658.

docs/design/api/authz.md
```diff
@@ -38,7 +38,7 @@ A user can be in the platform project (a developer/admin) or a customer project
 
 ## Permission resolution
 
-Permissions are dotted strings: `users.read`, `projects.settings.write`, `team.memberships.write`. The resolver answers: "for this principal, in this resolved scope, is this permission granted?"
+Permissions are flat `{resource}.{verb}` strings such as `user.read`, `project.write`, and `team_membership.write`. Multi-word resource types use `_` (for example `team_membership`, `flow_definition`), not dot nesting. Scope (which project or team) comes from the resolved grant, not from nesting in the permission name. Parent-resource permissions do not inherit into separately cataloged resources: for example, `project.write` does not imply `branding.write`, `allowed_origin.write`, or `webhook.write`. Bundles may grant those permissions together. The resolver answers: "for this principal, in this resolved scope, is this permission granted?" See [`system-permission-catalog.md`](system-permission-catalog.md) for the full list.
```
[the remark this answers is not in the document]
livio-a: Addressed in f23afe75.

`authz.md` now has an explicit **Not enforced yet** callout under permission resolution: the compatibility layer in `scopeAllowed` still treats legacy `project.write` as an umbrella that satisfies finer scopes including `branding.write`. The catalog drift notes also document that declared scopes are not yet enforced by `HandleOAuth2`.

.changeset/openapi-permission-scope-alignment.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zitadel/server": patch
```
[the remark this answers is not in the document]
livio-a: Addressed in f23afe75.

Bumped `.changeset/openapi-permission-scope-alignment.md` to `minor` and added an explicit breaking-change note listing the seven old plural scope names that clients must stop minting.

docs/design/api/system-permission-catalog.md
```diff
@@ -0,0 +1,644 @@
...
+| `team` | provisioning-boundary | **`team_admin` has `team.write` but not `team.create`** — only `project_admin` creates teams (new tenant boundary). |
+| `user` | provisioning-boundary | Onboarding/SCIM (`create`) vs profile edits (`write`). Plus a **credential-tier** split: `user.set_password` (account-takeover-grade) is separate from `user.write`. Stock admin bundles grant all three; custom grants can hold `user.write` alone. |
+| `team_membership` | provisioning-boundary | Invites/direct add (`create`) vs role change (`write`). **`sk_team_` grants `.write` but not `.create`/`.delete`.** |
+| `grant` | soft split | New access binding (`create`) vs change existing binding (`write`) — delegation control. |
+| `api_key` | action-verbs | `create`, `read`, `delete`, `rotate`, `revoke` — no `api_key.write`. |
+| `signing_key` | action-verbs | `create`, `read`, `delete` — manual rotation creates a successor; emergency replacement creates the successor before deleting the compromised key. |
```
wim07101993: shouldn't there be a rotate verb like with api_key, and no create?

docs/design/api/system-permission-catalog.md
```diff
@@ -0,0 +1,644 @@
...
+
+**OpenAPI reality:** several resources still declare a single `.write` scope
+meaning “create and manage.” The catalog names below are the **target** model;
+see [§ Drift notes](#drift-notes).
+
+| Class | Meaning |
+|---|---|
+| **provisioning-boundary** | `create` ≠ `write` is **normative now** — bundles or credential classes already depend on it. |
+| **soft split** | `create` vs `write` is useful for custom grants; default bundles may grant both. |
+| **manage** | No `create`/`write` split — `write` means create *and* manage (`read`, `write`, and `delete` where a delete endpoint exists). Matches OpenAPI `*.write` = create-and-manage. Split only if a real RBAC boundary emerges. |
```
wim07101993: can't we create a manage verb?

docs/design/api/system-permission-catalog.md
```diff
@@ -0,0 +1,644 @@
...
+
+| Permission | Intended operations | Notes |
+|---|---|---|
+| `billing.read` | View tier, subscription status, seat usage, invoices/receipts | From platform docs: Free / Pro / Enterprise tiers, seat counts. |
+| `billing.write` | Change plan/tier, payment method, billing contact | Gated to team owner (see [`platform/secret.md`](../platform/secret.md) capability matrix). Claim attaches the billing relationship. |
+
+Until endpoints exist, treat `billing.*` as reserved catalog names — do not
+grant them via `team.write`.
+
+### API keys
```
wim07101993: Do we want api keys? Not the place for the discussion but if we add them here in the documentation, it assumes we will. I would assume we only allow for client credentials grants from oauth2. In that case there is no api keys to manage and thus no permission for it

docs/design/api/system-permission-catalog.md
```diff
@@ -0,0 +1,644 @@
...
+api_key.create, api_key.read, api_key.delete, api_key.rotate, api_key.revoke,
+event.read
+```
+
+### `team_member`
+
+Basic team-scoped access.
+
+```
+team.read,
```
wim07101993, line 466: do we want team members to have permissions outside their own account? E.g.: with `user.read` they will be able to list all the members of a team. Which in case of an online shop might not be what we want.

Or I am misunderstanding `team_member`

docs/design/api/system-permission-catalog.md
```diff
@@ -0,0 +1,644 @@
...
+```
+
+### Origin-bound browser nonce
+
+Only the operations the nonce was minted for. No system catalog permissions
+are grantable — browser nonces gate auth flow operations, not management.
+
+---
+
+## Drift notes
```
wim07101993, line 536: Don't know whether this is the place but the `sk_proj...` and `sk_team...` don't exist in the codebase outside of documentation.

docs/design/api/system-permission-catalog.md
```diff
@@ -0,0 +1,644 @@
...
+## Permission catalog
+
+### Projects
+
+> **Verb model:** provisioning-boundary — `create` ≠ `write` is normative.
+
+| Permission | Endpoints | Notes |
+|---|---|---|
+| `project.create` | `POST /projects` (authenticated / on-prem path) | Two creation modes share this URL. **Cloud self-service:** anonymous `POST /projects` (`security: []`) + claim flow — no catalog permission (see [`../platform/claim-flow.md`](../platform/claim-flow.md)). **On-prem / admin-provisioned:** authenticated `POST /projects` gated by `project.create`, no claim — the platform project + default team are seeded at install ([`hierarchy.md`](hierarchy.md#self-hosted-exposes-the-same-api-shape-as-cloud)). How one endpoint reconciles both gatings is an [open question](#open-questions). |
+| `project.read` | `GET /projects`, `GET /projects/{id}` | Project resource only. Does not imply project-scoped configuration permissions. |
```
fforootd: This is the one row I'd like to talk through, because the repo currently says the opposite on purpose in three places:

- `api/openapi/endpoints/projects/by_id/methods.yaml:16` and `api/openapi/endpoints/projects/query/methods.yaml:8` both carry an inline comment: *"project.write, not project.read: project.read is carried by the browser-plane preview secret, so it cannot gate an operator read until ADR 036 splits the credential planes."*
- `internal/api/authz.go:113` — `projectAccess` deliberately lists no finer read scope, same reasoning.
- `internal/domain/project.go:72` — the preview secret is minted with exactly `["project.read"]`.

My guess is you mean the post-ADR-036 world, where `project.read` is no longer the browser scope and this becomes safe. That's completely defensible! But as written a reader can't tell, and this is the one row in the catalog where implementing it literally against today's credential planes would hand a browser-delivered secret operator read access to project resources. Every smaller delta got a drift note, so its absence here reads as "this one's settled".

Would you mind either adding the constraint to the Notes column, or stating on the row that it assumes ADR 036 has landed? Also, small thing while you're here: `GET /projects` doesn't exist — project listing is `POST /projects/query`.

docs/design/api/system-permission-catalog.md
```diff
@@ -0,0 +1,644 @@
...
+the corresponding resource permission.
+
+### Teams
+
+> **Verb model:** provisioning-boundary — `team_admin` has `team.write` but not `team.create`.
+
+| Permission | Endpoints | Notes |
+|---|---|---|
+| `team.create` | `POST /teams` | |
+| `team.read` | `GET /teams`, `GET /teams/{id}` | Team resource only. Does **not** imply `team_membership.*` or `user.*`. |
```
fforootd: The Endpoints columns mix shipped and aspirational surface without distinguishing them, which I think undersells how careful the rest of the doc is. Today: `/teams` is POST-only (no `GET /teams`), and `PATCH`/`DELETE` don't exist on `/users/{id}` or `/teams/{id}`, nor does `DELETE /projects/{id}`.

The Schemas section already solves this nicely — "`PATCH`/`DELETE /schemas/{id}` are **not yet exposed**" is unambiguous and costs one clause. Applying that same marker to the Projects / Users / Teams / Flow definitions rows would make the whole table something you can check a client against, rather than something you have to cross-reference `openapi-spec.yaml` for. Totally fine as a follow-up if you'd rather land this and sweep after.

docs/design/api/system-permission-catalog.md
```diff
@@ -0,0 +1,644 @@
...
+| `schema.write` | `POST /schemas`, `PATCH /schemas/{id}` | Create + manage. `POST /schemas` exists (OpenAPI `schema.write`); `PATCH` **not yet exposed**. |
+| `schema.delete` | `DELETE /schemas/{id}` | **Not yet exposed** — target shape. |
+
+### Flow definitions
+
+> **Verb model:** manage — `write` = create + manage; `delete` separate (already in OpenAPI).
+
+| Permission | Endpoints | Notes |
+|---|---|---|
+| `flow_definition.read` | `GET /flow-definitions`, `GET /flow-definitions/{id}` | Includes validate, simulate (read-only probes). |
```
fforootd: Two small path corrections in this block:

- the real paths use underscores — `/flow_definitions`, not `/flow-definitions` (see `openapi-spec.yaml:114-121`)
- there's no `archive` lifecycle endpoint; it's `POST /flow_definitions/{id}/deactivate`

Also `flow_definition.read` mentions "validate, simulate (read-only probes)" — worth marking those as planned, since neither endpoint exists yet.

docs/design/api/credentials.md
```diff
@@ -37,30 +37,50 @@ All three variants carry the same prefix and differ only by metadata bound at cr
...
-  platform.*                  (cross-project — note: the platform itself is a project)
+  user.set_password                  (account takeover — human/owner op only)
+  project.*
+  branding.*
+  domain.*
+  feature.*
+  allowed_origin.*
+  signing_key.*
+  webhook.*
+  team.*
```
fforootd: Question rather than an objection: `team.*` under MUST NEVER means an `sk_team_` can't `GET /teams/{its own id}` — while `team_member`, the weakest bundle in the catalog, does get `team.read`. The old list never addressed the team resource itself, so this is a new restriction landing quietly in a doc marked LOCKED.

If it's deliberate (a machine token has no business reading team metadata) it'd be great to say so in a parenthetical, since the deny list is meant to drive an enumerating test suite and "why is this here" will come up. If it isn't, `team.read` scoped to its own team probably belongs under MAY.

docs/design/api/credentials.md
```diff
@@ -37,30 +37,50 @@ All three variants carry the same prefix and differ only by metadata bound at cr
 
 ## `sk_team_…` narrow permission model — LOCKED
 
-`sk_team_…` is not just "a project token with team scope." It is a constrained credential class with an explicit allowlist. Lateral movement from a team token to project administration must be mechanically impossible.
+`sk_team_…` is not just "a project token with team scope." It is a constrained credential class with an explicit allowlist. Anything not listed under **MAY** is denied; **MUST NEVER** emphasizes high-risk names. Lateral movement from a team token to project administration must be mechanically impossible.
 
 ```text
 sk_team_… MAY:
```
fforootd, line 43: The flattening itself is right, and I like that the replacement table justifies each rename. One thing worth making explicit while this list is being rewritten: the old `team.users.*` names encoded the tenant boundary *in the permission string*, so a resolver bug degraded to "denied". Flat `user.write` puts the entire containment on grant-scope resolution, so the same bug degrades to project-wide user write.

Given the note two paragraphs down calls a single escape here a cross-tenant admin escalation, it might be worth stating the compensating requirement directly — `sk_team_` grants must carry a resolver-enforced team scope — and adding "team token reads/writes a user outside its team" to the enumerated deny-list suite you're already asking for. Doc-level only; no change needed to the names.

api/openapi/endpoints/sessions/by_id/methods.yaml
```diff
@@ -24,7 +24,7 @@ get:
           schema:
             $ref: ../session-response.yaml
     '401':
-      description: Missing or invalid session token.
+      description: Missing or invalid credential.
```
fforootd: Tiny wording thing with a practical consequence: singular "credential" is the odd one out — projects (×3), teams (×2), users and the sessions list operation all say "Missing or invalid credentials."

More usefully: #658 changes this same 401 line in the `delete` block to the plural form, so the two PRs will conflict here, and whichever way that resolves the `get` one above stays singular. Switching both to "Missing or invalid credentials." here removes the conflict and the inconsistency in one go.

(The disambiguation of the operator revoke path vs `/sessions/me` in the description below is a real improvement, by the way — that was genuinely confusing before.)

api/openapi/security/oauth2.yaml
```diff
@@ -13,13 +13,32 @@ flows:
...
-      flow_definitions.write: Permission to create and manage flow definitions
-      flow_definitions.delete: Permission to delete flow definitions
+      flow_definition.read: Permission to read flow definitions
+      flow_definition.write: Permission to create and manage flow definitions
+      flow_definition.delete: Permission to delete flow definitions
       project.read: Permission to read projects
       project.write: Permission to create and manage projects
       branding.read: Permission to read branding revisions
       branding.write: Permission to publish branding revisions
+      domain.read: Permission to read project domains
```
fforootd: Non-blocking, and I understand the goal is name stability so clients aren't rewritten later. But these fourteen (`domain.*`, `feature.*`, `allowed_origin.*`, `signing_key.*`, `webhook.*`) are referenced by zero operations, back no endpoints, and nothing mints or validates scopes yet — so "clients can request the final names" doesn't buy much today, while the published contract advertises permissions that grant nothing.

It's also a bit asymmetric: `api_key.*`, `grant.*` and `team_membership.*` are equally final and equally unimplemented, but stay out. Either way is fine by me — appending something like "(reserved; endpoints not yet available)" to these descriptions would keep the contract honest without giving up the naming commitment.

fforootd (commented):
Really nice piece of work — having one canonical place for permission names has been missing for a while, and the verb-model classes (provisioning-boundary vs manage vs runtime) are a much better framing than "CRUD everywhere". The drift notes and the explicit "declared ≠ enforced" section are exactly the kind of honesty that makes a design doc usable a month from now. Thanks also for the singular rename — that was a paper cut every time I read `oauth2.yaml`.

I checked out the branch and verified the mechanical side:

- `go generate ./api/...` leaves the tree clean, so the generated output is reproducible and `server:check-generate` will pass
- `go build ./...` and `go test ./internal/api/` pass
- no leftover plural names anywhere in Go, TS, api-mock, console or CLI, and every scope referenced on an operation is declared in `oauth2.yaml`
- nothing in the repo mints these scopes today (`internal/domain/project.go:61,72` only ever mints `project.write`/`project.read`, and the token endpoint doesn't validate requested scopes), so the rename has no runtime effect. `minor` plus the explicit breaking note reads exactly right to me.
- all the new doc links and anchors resolve (checked by hand, since there's no link checker in CI)

Comments are all on the catalog and docs — nothing on the rename itself. Two that I couldn't anchor inline because the lines aren't in the diff:

**The ADRs still carry the retired names.** The catalog opens by implementing ADR 032 §1 / ADR 033 §1 and retires `projects.settings.write` → `project.write`, but `docs/adrs/033-internal-permission-management.md:107` still uses `project.settings.write` as its illustrative name, and `docs/adrs/032-permission-catalogs.md:59,148` use `users.read`. Since root `AGENTS.md` treats ADRs as the decision record, that's the same drift this PR is closing everywhere else — a one-line example fix in each would finish the job, or a pointer to the catalog if you'd rather not touch accepted ADRs.

**`resource-map.md:198` still labels `DELETE /sessions/{id}` as `# logout`.** That's precisely the confusion you disambiguate in the OpenAPI description, and `/sessions/me` is missing from that block entirely. Since the PR already edits this file it's a cheap one to fold in.

Nothing here is a blocker except the `project.read` row — that one I'd genuinely like to understand before it becomes the thing people implement against. Happy to be wrong about it.

docs/design/api/system-permission-catalog.md
livio-a: Done — Notes + Drift: target **operator** read after ADR 036; today preview carries `project.read`, operator get/query stay on `project.write`. Also fixed listing → `POST /projects/query`. See latest push.

docs/design/api/system-permission-catalog.md
livio-a: Kept `create` / `read` / `delete` on purpose. Per ADR 039, rotation is create-a-successor (then retire), not in-place secret rotate like a typical opaque key. We explicitly do **not** add `signing_key.rotate` — that would imply the wrong lifecycle. Notes in the catalog spell this out now.

docs/design/api/system-permission-catalog.md
livio-a: Agreed; we marked several rows **Not yet exposed** (projects delete, teams write/delete, users delete, schema delete, flow validate/simulate). Happy to do a fuller sweep as follow-up if anything still reads as shipped when it isn’t — or call out remaining gaps you still see.

docs/design/api/system-permission-catalog.md
livio-a: “Manage” stays a **verb-model class** only (full lifecycle: `write` = create+update, plus `delete` where an endpoint exists). Grantable strings stay `write` / `delete` — no `*.manage`. Holding `write` alone still doesn’t authorize delete. Documented in the class table + drift notes.

docs/design/api/system-permission-catalog.md
livio-a: Done earlier — underscores, deactivate (not archive), validate/simulate marked planned.

docs/design/api/system-permission-catalog.md
livio-a: Agreed this wasn’t settled — we **parked** first-class `api_key.*` (removed from catalog, bundles, and `sk_team_` MAY). Opaque `sk_proj_` / `sk_team_` service tokens remain for bootstrap create/claim; management CRUD vs OAuth2 client-credentials vs secret-rotate-only is open question #3. Thanks for the push.

docs/design/api/credentials.md
livio-a: Done — MAY includes `team.read` (own team only); MUST NEVER is create/write/delete, not blanket `team.*`.

docs/design/api/system-permission-catalog.md
livio-a: `team_member` is meant as **org staff** (legacy org-member), not shop end-customers. `user.read` is roster visibility inside the team grant for operators. Clarified that audience on the bundle. End-user product roles would need a thinner bundle (or no membership bundle) if we ever model customers that way.

docs/design/api/credentials.md
livio-a: Done — compensating note: `sk_team_` grants must carry resolver-enforced team scope; deny-list suite should include “team token reads/writes a user outside its team.”

docs/design/api/system-permission-catalog.md
livio-a: Intentional for this PR: the catalog locks **credential classes and permission names**; wire minting of the `sk_*` prefix format is follow-up. Today’s project/preview secrets are the stand-ins (JWE + scopes). Same pattern as reserved scopes — docs precede implementation. Happy to add a one-line “not minted under this prefix yet” if you want it on the credential section.

api/openapi/endpoints/sessions/by_id/methods.yaml
livio-a: Done in rebase with #658 — plural “credentials”; no Set-Cookie on operator revoke.

api/openapi/security/oauth2.yaml
livio-a: Done — descriptions marked reserved / endpoints not yet available.

fforootd (approved):

PR #646, opened 2026-07-28
feat: add team names with per-project uniqueness

### LINKED ISSUE #618: Team: add name field
## Context

Part of #598. Team's `name` field is required but doesn't exist anywhere yet: not in the migration, the domain model, or the OpenAPI contract.

## Scope / Tasks

- Add a `name` column to the teams table, on both postgres and spanner.
- Require `name` at creation and reject empty values, matching how Project already treats its `name` field.
- Add a validation error for an invalid team name.
- Update the Create and Get request/response schemas to include `name`.
- Return `name` from the Create and Get endpoints.

## Acceptance Criteria

- [x] Creating a team requires a non-empty `name`.
- [x] Empty `name` returns 400.
- [x] `name` is persisted and returned by Get.
- [x] Migration adds `name` on both postgres and spanner.

## Out of scope

- Renaming a team (Update) is tracked separately in #619.

grvijayan:
## Summary

- Teams now carry a name. It is required on create and returned by the create and get endpoints.
- Team names are unique per project, ignoring case. The same name can be reused in a different project.
- Team and project names are trimmed, and whitespace-only names are rejected.

## Validation

- `go vet ./...` and `go test ./internal/domain/... ./internal/service/...`
- `go test -tags postgres_integration ./internal/storage/v2/dialect/postgres/... ./internal/api/integration_test/...`
- `go test -tags spanner_integration ./internal/storage/v2/dialect/spanner/... ./internal/api/integration_test/...`
- `moon run server:generate` (generated client regenerated from the OpenAPI source)
- Full `moon ci` was not run locally.

## Release notes / changeset

- Changeset: `.changeset/team-name-field.md` — team name field and per-project uniqueness (`@zitadel/server`)

## Notes

- Postgres enforces uniqueness with a `lower(name)` index. Spanner cannot index an expression, so it uses a `STORED` generated column instead.
- Two test commits are included: fixture IDs no longer come from the clock (that caused a flake), and the team service tests are table driven.
- Also fixes a pre-existing leak: the team membership test ran its cleanup with `t.Context()`, which is cancelled before cleanups run, so it left a project row behind on every run and broke `TestProjectStatements_List` under `-count>1`.
- `TestProjectStatements_List` also asserted on an unfiltered list, which is what turned that leak into a failure there. Scoping those assertions is in #638, since it touches the project tests.


livio-a (approved):

livio-a (approved):
Looks good. Approving with the understanding that the `ensureTeam` UniqueError swallow is limited to the optional user-bootstrap path and is acceptable for now.

grvijayan:
> Approving with the understanding that the ensureTeam UniqueError swallow is limited to the optional user-bootstrap path and is acceptable for now.

Thank you for the approval. 
Yes, this is during user bootstrapping. This change was already there though and I missed it 🤔 
The service layer returns an `domain.ErrTeamAlreadyExists()` [error](https://github.com/zitadel/nextgen/blob/main/internal/service/team.go#L38). But I think it'd be better to update the bootstrapping path to also return the error. I'll fix this in the next PR.

internal/storage/v2/dialect/postgres/migration/sql/000002_teams.sql
```diff
@@ -4,6 +4,7 @@ CREATE TABLE zitadel_nextgen.teams(
     REFERENCES zitadel_nextgen.projects (id)
     ON DELETE CASCADE
     , id TEXT COLLATE "C" NOT NULL CHECK (id <> '')
+    , name TEXT NOT NULL CHECK (name <> '' AND length(name) <= 200)
```
fforootd, line 7: I think this one needs to become a new migration rather than an edit of `000002`.

`Migrate` is documented as "already-applied migrations are skipped", and the provider is built with only `WithTableName`/`WithSessionLocker` — no checksum verification. So on any database where `goose_db_version` already records version 2 (existing alpha deploys, dev DBs, and the persistent Postgres people use for re-runnable `-tags postgres_integration` runs), the `name` column, its CHECK, and `uq_teams_project_name` never get created — silently, with no error. `createTeamStmt` and `teamQuery` then fail at runtime with `column "name" does not exist`. CI stays green because it migrates a fresh database, which is exactly what hides it.

There's a good precedent to follow in `000011_user_team_lifecycle.sql`, which added `teams.status` as a new numbered migration with `ALTER TABLE … ADD COLUMN` plus an explicit backfill.

Rough shape: restore `000002`, then add `000014_team_names.sql` doing add nullable `name` → backfill → `SET NOT NULL` + CHECK → create the unique index. For the backfill value, `'team-' || id` matches what `ensureTeam` already derives and keeps the unique index satisfiable. On the Spanner side `name_lower` would become `ALTER TABLE teams ADD COLUMN name_lower STRING(MAX) AS (LOWER(name)) STORED` before the index.

If the deliberate call is "we're in alpha, every database gets recreated" then that's completely fair — could we just note it in the PR description? It's worth being explicit since it diverges from 000011 and would bite anyone running against a persistent DB.

internal/bootstrap/users/ensure.go
```diff
@@ -43,9 +43,12 @@ func ensureProject(ctx context.Context, stmts service.AllStatements, projectID s
 }
 
 func ensureTeam(ctx context.Context, stmts service.AllStatements, projectID, teamID string) error {
+	// The bootstrap header carries no team name, so derive a placeholder
+	// name from the team ID to satisfy the NOT NULL name column.
 	err := stmts.CreateTeam(ctx, &domain.Team{
 		ProjectID: projectID,
 		ID:        teamID,
+		Name:      "team-" + teamID,
```
fforootd, line 51: Small knock-on effect worth a look: before this PR `teams` had exactly one unique constraint (the PK), so the `UniqueError ⇒ the team already exists ⇒ return nil` below was exact. Now there are two.

If this derived placeholder happens to collide case-insensitively with a *different* team's name in the same project, the insert trips `uq_teams_project_name`, `ensureTeam` reports success, the row doesn't actually exist, and the following user insert fails on the team FK with an error that points somewhere else entirely.

Probably rare in practice, and `ensureProject` is fine since `projects` has no unique index on name. Cheapest fix is probably a `GetTeamByID` confirmation after swallowing. Checking the constraint name instead would only work on Postgres today — `spanner/error.go` builds `NewUniqueError("", "", err)` with an empty constraint name.

internal/service/team.go
```diff
@@ -27,12 +28,15 @@ func NewTeamService(v2Pool *DB) *TeamService {
...
 
 func (s *TeamService) CreateTeam(ctx context.Context, input CreateTeamInput) (*domain.Team, error) {
-	model, err := domain.NewTeam(input.ProjectID)
+	model, err := domain.NewTeam(input.ProjectID, input.Name)
 	if err != nil {
 		return nil, err
 	}
 
 	if err := s.v2Pool.Statements().CreateTeam(ctx, model); err != nil {
+		if _, ok := errors.AsType[*database.UniqueError](err); ok {
```
fforootd, line 37: Same conflation one layer up: a `(project_id, id)` PK collision would surface to the caller as a 409 "a team with this name already exists". IDs are generated, so this is close to unreachable in practice — mentioning it mostly because the information to tell them apart is already in hand. The test just below constructs `database.NewUniqueError("teams", "uq_teams_project_name", nil)` *with* the constraint name, and the production path doesn't look at it.

If you do go for a constraint-name check, one thing that'd help: the two dialects currently name the same invariant differently (`uq_teams_project_name` on PG, `idx_teams_project_name` on Spanner). Aligning them now would make that check possible later.

internal/storage/v2/dialect/spanner/migration/sql/000002_teams.sql
```diff
@@ -4,17 +4,30 @@
...
+    CONSTRAINT chk_teams_name CHECK (name <> ''),
     CONSTRAINT fk_teams_project
         FOREIGN KEY (project_id)
         REFERENCES projects (id)
         ON DELETE CASCADE,
 ) PRIMARY KEY (project_id, id)
 -- +goose StatementEnd
+-- +goose StatementBegin
+-- Team names are unique per project, case-insensitive.
+CREATE UNIQUE INDEX idx_teams_project_name
```
fforootd, line 22: Nit, and only worth it while you're in here: this is `idx_teams_project_name` while Postgres calls the same invariant `uq_teams_project_name`. Same name in both would make it easier to key error handling off the constraint later (see the service comment). No strong feelings on which name wins.

The `name_lower` STORED-generated-column workaround for Spanner's lack of expression indexes is a nice solution, and the comment explaining *why* it's there is exactly the kind I'm glad to find later.

internal/api/integration_test/team_test.go
```diff
@@ -25,30 +28,123 @@ func TestCreateTeam(t *testing.T) {
...
 			resp, err := client.CreateTeam(t.Context(), req, params)
 			require.NoError(t, err)
 
-			assert.IsType(t, &api.CreateTeamBadRequest{}, resp, helpers.MustMarshal(t, resp))
+			badRequest, ok := resp.(*api.CreateTeamBadRequest)
+			require.True(t, ok, helpers.MustMarshal(t, resp))
+			assert.Equal(t, api.ErrorCode("req.invalid"), badRequest.Code)
+		})
+
+		// The contract carries minLength/maxLength, so the generated client
```
fforootd, line 116: Good call flagging this explicitly. One gap it leaves: because the generated client rejects these before anything goes over the wire, nothing pins what the *server* does with a raw `{"name": ""}` or a 201-character name. The ogen request decoder should reject both, but that's the path a non-generated client (curl, another SDK) takes, and it's the one that isn't covered.

Not asking for it in this PR if you'd rather keep the scope tight — just noting it so it doesn't read as covered.

internal/domain/team_test.go
```diff
@@ -0,0 +1,121 @@
...
+		{
+			name: "team name over the length limit",
+			args: args{
+				projectID: "proj_1",
+				name:      strings.Repeat("a", domain.TeamNameMaxLength+1),
+			},
+			wantErr: domain.ErrTeamNameInvalid(),
+		},
+		{
+			name: "team name is trimmed before the length check",
```
fforootd, line 96: Tiny inconsistency this case surfaces: ogen validates `minLength`/`maxLength` against the raw string (it counts runes, so the multi-byte cases above are genuinely consistent end to end — nice), but the domain trims *before* measuring. So a 200-char name wrapped in whitespace is 204 runes and gets rejected at the contract with a generic `req.invalid`, meaning the behaviour this case guarantees can't actually be reached through the API.

The domain behaviour itself seems right to me. Might just be worth a note that trim-then-measure is domain-only, so nobody later reads this as an API guarantee.

api/openapi/endpoints/teams/methods.yaml
```diff
@@ -32,6 +33,13 @@ post:
         application/json:
           schema:
             $ref: ../../components/error-details.yaml
+    '409':
```
fforootd, line 36: Optional, and arguably a separate PR: the two schemas `gen_error_schemas` produced (`team-already_exists.yaml`, `team-name_invalid.yaml`) aren't referenced from here — 400 and 409 both still point at the generic `error-details.yaml`. Practical effect is that generated clients can't discriminate by type, which is why the integration test has to string-compare `conflict.Code`.

`auth_attempts` and `sessions` already use a `*-error-response.yaml` oneOf if you want a pattern to copy. Since you're adding the error codes anyway, this might be the natural moment — but no objection to leaving it.

internal/storage/v2/dialect/spanner/team_test.go
```diff
@@ -43,3 +45,54 @@ func TestTeamStatements_CRUD(t *testing.T) {
...
+	stmts := testClient.Statements()
+
+	project := newTestProject(uniqueProjectID(t))
+	require.NoError(t, stmts.CreateProject(ctx, project))
+	t.Cleanup(func() { _ = stmts.DeleteProjectByID(context.Background(), project.ID) })
+
+	t.Run("empty", func(t *testing.T) {
+		team := newTestTeam(project.ID, "team_v2_name_empty")
+		team.Name = ""
+		// Spanner reports a violated CHECK as OutOfRange, which wrapError does
```
fforootd, line 88: Thanks for writing down *why* the assertion is weaker here — that saves the next person the same investigation.

If you feel like it: adding `codes.OutOfRange` alongside `FailedPrecondition, InvalidArgument` in `spanner/error.go`'s switch would let this assert `CheckError` like the over-length case does. It'd also mean a Spanner CHECK violation surfaces as a 400 rather than a 500 — unreachable through the API today since the domain trim check fires first, so this is defence-in-depth only, and `wrapError` is pre-existing. Entirely fine to leave as-is.

fforootd (commented):
Really nice piece of work — the layering is clean and the test coverage is thorough. A few things I especially liked:

- Rune semantics line up at 200 across all four layers (PG `length()`, Spanner `STRING(200)`, ogen's validator, and `utf8.RuneCountInString`). That's the easy thing to get subtly wrong and it's right here.
- `uniqueSuffix` → `rand.Text()` is a genuine flake fix, and applying it consistently across both dialects was the right instinct.
- The `t.Context()`-in-`t.Cleanup` catch in `team_membership_test.go` is a good find, and I appreciate that the PR description is upfront about the `TestProjectStatements_List` interaction instead of quietly working around it.
- The table-driven rewrite plus `newMockedTeamService` cuts real duplication, and the `if setupStmt != nil` guard neatly avoids an unmet `Statements()` expectation on the validation-only cases.

One thing I'd like to sort out before merge — the in-place edit of `000002_teams.sql` (comment inline). Everything else is small and can be taken or left.

Verified locally: `go build ./...` clean, `go test ./internal/domain/... ./internal/service/...` pass, `go vet` clean under both `postgres_integration` and `spanner_integration` tags. I didn't run the integration suites against a live DB.

internal/storage/v2/dialect/postgres/migration/sql/000002_teams.sql
grvijayan: > If the deliberate call is "we're in alpha, every database gets recreated" 

yes, this was the intention as we're allowed to break stuff at the moment.
i'll keep this in mind for the future PRs.

fforootd: Ok works for me

api/openapi/endpoints/teams/methods.yaml
grvijayan: Would be nice as a separate PR because this would require refactoring the service-layer implementation of Teams. `gen_openapi_errors` only picks up services declared as interfaces with `// errors:` comments, and TeamService is a concrete struct right now. 

I tried it spec-only to see what we'd get. The `createTeam` docs page currently renders five identical `{ "code": "string", "message": "string", "details": {} }` bodies (without the error code, but with HTTP status). With the `oneOf` they become one named example per error code, but no longer list the HTTP statuses. 

This pattern is present in most other endpoints/resources too. We could cover all of them in a separate refactor ticket, if this is what we want to do.

before:
image


after:
image

internal/service/team.go
grvijayan: Postgres returns the constraint name and Spanner doesn't. To branch on it we'd need to expose it on [IntegrityViolationError](https://github.com/zitadel/nextgen/blob/main/internal/storage/v2/database/integrity_errors.go#L75), which seems like overkill for a case that is close to unreachable in practice. Will add a comment explaining this.

internal/bootstrap/users/ensure.go
grvijayan: this is being addressed in my next [PR](https://github.com/zitadel/nextgen/pull/672/changes#diff-01880e3bfcade08c733e76f565b0d23488d6f28060b6455167bf961d447c78fbL45) for the teams API

PR #649, opened 2026-07-29
feat: project scoped encryption and signing keys

### LINKED ISSUE #507: Crypto: Per-project signing keys stored envelope-encrypted
ADR requirement (Scope, §Signing/encryption keys): "a separate set of keys per project"; signing keys are asymmetric, private keys never leave the system, stored per the data-at-rest (envelope) decision.

Current state (NOT FOUND): No signing keys are generated, stored, or scoped anywhere. `tokengen/jwt.go` can sign given an injected jose.Signer but nothing constructs keys (`rsa.GenerateKey`/`jose.NewSigner` absent from non-test code).

Tasks:
- [ ] Generate DEK per project.
- [ ] Store the private key DEK‑wrapped in the keys table (Issue 3); store/expose the public key.
- [ ] Provide a signer that loads the active project signing key on demand.
- [ ] Created repositories/database tables to store keys

No caching, it opens the door for stale caches, which we don't want to handle right now.

Acceptance: Each project has its own active RSA signing key persisted envelope‑encrypted; a token can be signed with the correct project key.

### LINKED ISSUE #510: Crypto: Encrypted storage for third-party secrets (IdP secrets, API keys)
ADR requirement (Third party secrets): IdP secrets, API keys, etc. must be stored encrypted (AES-GCM) in the DB — they can't be hashed because they must be sent to other systems.

Current state (NOT FOUND): No IdP/API-key secret storage exists. `internal/secrets/generator.go` mints random `sk_…` strings but is only used in test helpers; there's no encrypted-secret persistence type.

Tasks:
- [ ] Add a reusable encrypted-secret value type + column convention, encrypting via the DEK (Issue https://github.com/zitadel/nextgen/issues/505).
- [ ] Storage/repository helpers to write/read/decrypt third-party secrets, scoped per project.
- [ ] Use it as the storage primitive when IdP/API-key features land.

Acceptance: A third-party secret can be stored AES-GCM-encrypted under the project DEK and decrypted for outbound use; plaintext never hits the DB.

wim07101993:
## Summary

- added project scoped encryption and signing keys

## Validation

- `go test ./...`

## Release notes / changeset

- Changeset: .changeset/project-scoped-encrytption-and-signing-keys.md

## Notes

**Note for contributors:** if you already ran this from `main`, wipe `nextgen-data` and the schema — the rename from `encryption_keys`/`keks/` to `master_keys`/`master-keys/` orphans existing wrapped keys, and the in-place `000012` edit won't apply to DBs that already have that migration.


internal/domain/signing_key.go
```diff
@@ -0,0 +1,122 @@
...
+	"crypto/rand"
+	"fmt"
+	"time"
+
+	"github.com/go-jose/go-jose/v4"
+	"github.com/zitadel/nextgen/internal/crypto"
+)
+
+const (
+	PrefixSigningKey ResourcePrefix = "sig_key"
```
livio-a, line 14: Prefixes disagree: `sig_key` here, `encryption_key` in `internal/domain/encryption_key.go:15`. Worth settling on one while the codes are still cheap to change.

This one is tangled up with the `enc_key-*.yaml` comment on `api/openapi/components/schemas/errors/enc_key-not_found.yaml` — whichever way we go decides which schema files survive. If we pick `enc_key`, `PrefixEncryptionKey` changes to `"enc_key"` and the `encryption_key-*.yaml` set gets deleted. If we keep `encryption_key`, `sig_key` here probably becomes `signing_key`.

api/openapi/components/schemas/errors/enc_key-not_found.yaml
```diff
@@ -2,13 +2,13 @@
...
-  Error code: dek.not_found
+  Error code: enc_key.not_found
 allOf:
   - $ref: ../../error-details.yaml
   - type: object
     properties:
       code:
         type: string
-        const: dek.not_found
+        const: enc_key.not_found
```
livio-a, line 12: These four were renamed from `dek-*.yaml` and rewritten to `enc_key.*`, but nothing emits that code — the domain prefix is `encryption_key` (`internal/domain/encryption_key.go:15`), and the correctly generated `encryption_key-*.yaml` files are already in the tree, untouched here.

I ran `go run ./cmd/gen_error_schemas` on this branch: zero diff, and it never emits an `enc_key-*` file. So these are hand-edited files carrying a `DO NOT EDIT` header for a code that can't be returned. The generator only writes, never prunes, which is why CI stays green.

The `dek-*.yaml` files were stale leftovers from the earlier `encryption_key` → `dek` rename and should have been plain deletions. Which set survives depends on the prefix question I raised on `internal/domain/signing_key.go:14`: keep `encryption_key` and delete these four, or move the domain to `enc_key` and delete the `encryption_key-*.yaml` set instead.

internal/domain/project.go
```diff
@@ -83,6 +85,48 @@ func (p *Project) PreviewSecret(encrypter op.Encrypter) (string, error) {
 	return previewSecret, nil
 }
 
+// GenerateNewKeySet creates the project's key encryption key, wrapped by the
+// deployment's master key, and the purpose-scoped keys wrapped by that KEK.
+func (p *Project) GenerateNewKeySet(masterKey crypto.Crypter) (*ProjectKeySet, error) {
```
livio-a, line 90: The property this PR exists to establish — purpose keys wrapped by the project KEK, not directly by the master key — has no test asserting it. `GenerateNewKeySet` and `ProjectKeySet.Activate` have no domain coverage, and `internal/service/project_test.go:91` only counts four `CreateEncryptionKey` calls, which passes just as well if all four were wrapped by the master key.

Decoding each purpose key's JWE header and asserting `kid` equals the KEK's ID (and the KEK's `kid` is the master key's) would pin it down cheaply. `GetProjectSigningKey` / `GetProjectSigner` are also untested at the service layer.

internal/service/keys.go
```diff
@@ -67,27 +69,27 @@ func (s *keyService) GetCrypter(ctx context.Context, keyID string, algorithm jos
...
-		return nil, domain.ErrInternal(err).WithMessage("failed to get DEK from the database")
+		return nil, domain.ErrInternal(err).WithMessage("failed to get encryption key from the database")
 	}
-	return dek, nil
+	return key, nil
 }
 
-func (s *keyService) GetProjectDEKCrypter(ctx context.Context, projectID string) (op.Crypto, error) {
-	dek, err := s.GetProjectDEK(ctx, projectID)
+func (s *keyService) GetProjectCrypter(ctx context.Context, projectID string, purpose domain.EncryptionKeyPurpose) (op.Crypto, error) {
```
livio-a, line 87: Small note, not a blocker: this is now two queries instead of one. The purpose key is wrapped by the project KEK rather than a configured master key, so `getCrypterOfKey` misses the in-memory hit and falls through to `GetCrypter`, which reloads the KEK from the database. Inherent to the extra hierarchy level, and it's on the flow-cookie and token-mint paths, so maybe worth an issue for caching later.

internal/storage/v2/dialect/spanner/crypto_keys.go
```diff
@@ -132,6 +141,60 @@ func (s cryptoKeyStatements) scanEncryptionKey(row *spanner.Row) (*domain.Encryp
...
+	).statement()
+	return s.db.Write(ctx, stmt, func(iter *spanner.RowIterator) error {
+		_, err := collectOneRow(iter, func(row *spanner.Row) (struct{}, error) {
+			return struct{}{}, row.Columns(&key.ID, &key.CreatedAt)
+		})
+		return err
+	})
+}
+
+// GetEncryptionKey implements [service.CryptoKeyStatements].
```
livio-a: Copy-pasted doc comment: says `GetEncryptionKey`, function is `GetSigningKey`.

internal/service/user.go
```diff
@@ -32,6 +32,12 @@ type SetPasswordInput struct {
 	IsPasswordChangeRequired bool
 }
 
+type UpdateUserInput struct {
```
livio-a: `UpdateUserInput` has no references anywhere and isn't related to project key sets. Leftover from another branch?

livio-a (commented):
Nice refactor — the updated docs make the new key hierarchy easy to follow. Rotation safety looks good, and local build/tests pass.

I found one error-schema inconsistency, one test gap, and a few small cleanups below.

Worth a short note for contributors: renaming `encryption_keys` / `keks/` to `master_keys` / `master-keys/` orphans any existing wrapped keys (old config is ignored, a new master key is generated). Combined with the in-place `000012` edit, wipe `nextgen-data` and the schema if you already ran this from `main`.

api/openapi/components/schemas/errors/enc_key-not_found.yaml
wim07101993: The rename was done after code-gen, that is why they are modified. Will clean up

internal/service/keys.go
wim07101993: Yes caching would be required in the future. However in the ADR we explicitly said not to implement that yet.

It is theoretically can be even more than 2 calls, depending on which key you are fetching. The `getCrypterOfKey` does a recursive search because the kek to decrypt the key could be encrypted with another key in the database. E.g.: to decrypt a token

1. get kid from token-header (points to dek)
2. get dek from database
3. get kid from dek-header (points to project-kek)
4. get kek from database
5. get kid from kek-header (points to master key)
6. get correct master key from memory

Depending on how many levels of keys, there could be many roundtrips to the database. Maybe this can be optimized in the future.

internal/service/user.go
wim07101993: must have slipped in from another work tree

livio-a (approved):
Approving. Follow-ups look good — especially the `enc_key` rename and the new `GenerateNewKeySet` wrapping tests.

I pushed a small docs fix so `docs/operations/encryption-keys.md` uses the `enc_key.*` error codes after the prefix rename.

PR #658, opened 2026-07-29
fix(api): stop revoking a session by id from clearing the caller's cookie

livio-a:
## Summary

- `DELETE /sessions/{session_id}` returned the same `Set-Cookie: __nextgen_session=; Max-Age=0` deletion as the self-service `DELETE /sessions/me`. Revoking a session by id is an operator action on a session that is usually **not** the caller's, so clearing the caller's cookie signs the operator out.
- The console's session list is the only by-id caller (`apps/console/src/routes/_authed/sessions/index.tsx`). It revokes over the same-origin `/api` base with fetch's default `credentials: "same-origin"`, so the browser sends the admin's cookie and honours the deletion in the response. The console's own logged-in state *is* that cookie (Console ADR 0003: the `_authed` guard calls `GET /sessions/me` with credentials), so revoking anyone's session destroyed the admin's own session and the next guard evaluation redirected to `/login`. The 15s session cache in `apps/console/src/auth/session.ts` only delayed the symptom.
- Cookie clearing now belongs solely to `RevokeMySession`, the operation that acts on the cookie's own session. The header is dropped from the by-id operation in OpenAPI and the handler; regenerating removes `SetCookie` from `RevokeSessionNoContent`, so it cannot be reintroduced without a compile error.

## Validation

- `go generate ./...` (committed regenerated output)
- `npx @redocly/cli lint api/openapi/openapi-spec.yaml` — valid
- `go build ./...`
- `go test ./internal/api/... -count=1` — pass

## Release notes / changeset

- Changeset: `.changeset/revoke-session-cookie-scope.md` — `@zitadel/server` patch: revoking a session by id no longer clears the caller's session cookie.

## Notes

- Found while reviewing session security schemes for #645 (system permission catalog). That PR corrects the stale prose on this operation — it still reads "This is the logout operation. The session_token issued at creation ... is required," which has been untrue since the operation moved to the `session.delete` OAuth scope. The two PRs are kept apart deliberately: this one is the behavior fix, #645 owns the documentation. Whichever merges second may need a trivial conflict resolution in `api/openapi/endpoints/sessions/by_id/methods.yaml`, since the edits are adjacent.
- No test pinned the old header, and none is added: the field no longer exists on the generated response type, which is a stronger guarantee than an assertion.
- Not addressed here: `HandleOAuth2` does not enforce that a token carries the scopes an operation declares, so `session.delete` on this operation is still a declaration rather than a gate. Tracked in the permission-catalog work.


internal/api/session.go
```diff
@@ -127,9 +127,11 @@ func (h Handler) RevokeSession(ctx context.Context, params api.RevokeSessionPara
 	if err != nil {
 		return nil, err
 	}
-	return &api.RevokeSessionNoContent{
-		SetCookie: deleteSessionCookie(),
-	}, nil
+	// No Set-Cookie: this operation revokes a session by id on behalf of an
```
fforootd, line 130: Good comment — it puts the reasoning exactly where the next person will look for it.

One thing I'd like to make sure lands in the tracking issue alongside the scope-enforcement gap you mention in the description: the sharper half here is arguably project binding rather than scopes. `params.ProjectID` goes straight into `DeleteSessionInput` with no check that the bearer is actually bound to that project, so a valid token for project A can revoke a session in project B given its id. `GetSession` and `ListSessions` have the same shape.

Not for this PR — the diff is better for staying narrow — but `requireProjectAccess` already lives one file over (`internal/api/authz.go:127`) and would be a two-line addition when the permission-catalog work reaches these operations. Worth naming the cross-tenant angle explicitly there, since it reads a bit differently from "scopes are declared but not gated".

.changeset/revoke-session-cookie-scope.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zitadel/server": patch
+---
+
+Stop `DELETE /sessions/{session_id}` from clearing the caller's `__nextgen_session` cookie. Revoking a session by id is an operator action on someone else's session, so clearing the caller's cookie signed the operator out — in the console, revoking any session from the session list destroyed the admin's own session. Cookie clearing now happens only on `DELETE /sessions/me`, which acts on the cookie's own session.
```
fforootd: Tiny wording suggestion, since this becomes public release notes: the sentence states the console symptom as observed, but I couldn't reproduce it end-to-end — `sessionService.List` returns `ErrNotImplemented` (`internal/service/session.go:116`, stubbed since #319), so the session-list loader 501s against a real backend and never renders a Revoke button, and the orval msw mock doesn't emit response headers either.

The API-level defect is real by construction, so nothing changes about the fix. I'd just phrase the console part conditionally — "…would have signed the operator out; in the console it would destroy the admin's own session" — so the changelog stays accurate. Unless you did hit it somewhere I haven't looked, in which case I'm curious where, because that would mean the list is wired up in a way I've missed.

api/openapi/endpoints/sessions/by_id/methods.yaml
```diff
@@ -57,11 +57,8 @@ delete:
   responses:
     '204':
       description: Session revoked successfully.
-      headers:
-        Set-Cookie:
-          $ref: ../../../components/headers/set-cookie-session-token.yaml
     '401':
-      description: Missing or invalid session token.
+      description: Missing or invalid credentials.
```
fforootd, line 64: This matches the house wording now — every other oauth2-secured operation in the spec says "Missing or invalid credentials.", and only the genuinely cookie-secured `/sessions/me` pair keeps the session-token phrasing. 👍

Two optional freebies while you're in this file, both fine to leave to #645 if you'd rather keep the split clean:

- `getSession`'s 401 on line 27 still reads "Missing or invalid session token." though it's oauth2-secured too — same one-line fix you just made for `delete`.
- The `delete` description on line 45 still says "This is the logout operation. The session_token issued at creation … is required," which this PR now directly contradicts (and the new handler comment says the opposite). If #645 slips, the published spec ships that contradiction right next to the 401 you corrected — might be worth the three lines here to avoid the window.

fforootd (commented):
Nice catch, and a nice fix — the semantics argument in the description is exactly right: the by-id operation acts on someone else's session, so it has no business touching the caller's cookie. Dropping the field from the generated type rather than just not setting it is the detail I liked most; that's a stronger guarantee than a test, and I agree no new test is warranted.

I re-ran your validation plus a few checks of my own, everything holds:

- `go generate ./api/...` produces no diff — the committed generated output matches regeneration exactly.
- `go build ./...` and `go test ./internal/api/...` pass.
- The console really is the only by-id caller. The login orchestrator signs out through `revokeMySession` (`packages/components/src/orchestrator/api-client.ts:103`), so `<zitadel-logout>` / `<zitadel-session>` are untouched.
- Nothing in `@zitadel/api-mock` needs to follow: the hand-written routes only implement `/sessions/me` and `/auth/end-session`, and the conformance assertion at `packages/api-mock/src/spec-conformance.spec.ts:147` pins the cookie on `/sessions/me`, which you kept. Your "no test pinned the old header" claim checks out.
- `Set-Cookie` now remains on create, exchange and `/sessions/me` — precisely the operations that mint or own the cookie.
- The leftover cookie in the self-revoke case is harmless: `DeleteSessionByID` hard-deletes the row (`internal/storage/v2/dialect/postgres/session.go:90`) and the console guard fails closed on any error (`apps/console/src/auth/session.ts:52`), so it still lands on `/login`.

Three small notes inline, none blocking — happy for this to go in as is.

One optional follow-up for later, definitely not for this PR: once `listSessions` is implemented, revoking *your own* row from the console list will now be silent for up to the 15s session cache before the guard bounces you. Routing the row whose `session_id` matches the current session through `signOut()` (or disabling it) would make that immediate.

internal/api/session.go
livio-a: 👍 Agreed — filing the project-binding/`requireProjectAccess` gap as a follow-up PR now, named as cross-tenant rather than just undeclared scopes.

.changeset/revoke-session-cookie-scope.md
livio-a: 👍 Softened the changeset to conditional wording — listSessions is still stubbed, so the console path was hypothetical.

api/openapi/endpoints/sessions/by_id/methods.yaml
livio-a: 👍 Took the freebies here so the published spec stays consistent if #645 slips.

api/openapi/endpoints/sessions/by_id/methods.yaml
```diff
@@ -57,9 +57,6 @@ delete:
   responses:
     '204':
       description: Session revoked successfully.
-      headers:
-        Set-Cookie:
-          $ref: ../../../components/headers/set-cookie-session-token.yaml
     '401':
       description: Missing or invalid session token.
```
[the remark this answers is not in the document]
livio-a: 👍 Already fixed in cba7cb86; getSession 401 aligned in the follow-up commit.

internal/api/session.go
livio-a: Follow-up opened: https://github.com/zitadel/nextgen/pull/669

PR #664, opened 2026-07-30
test: share statement contract suites across dialects

adlerhurst:
## Summary

Shared statement-contract integration suite under `internal/storage/v2/stmttest` against `service.AllStatements` / `service.Pool`, so Postgres and Spanner prove the same domain behavior without duplicating suites per dialect.

Build-tag registration opens every compiled dialect; `forEachDialect` runs each suite once per dialect (`postgres`, `spanner`), with normal `t.Run` for case names. CI still uses one tag per job; locally both tags may run in one process. Dialect-local leftovers are compilers, migrations, `withTransaction` unit tests, and raw DML for `SeedProjectsTiedAt`. Spanner FK failures that arrive as `FailedPrecondition` now map to `ForeignKeyError` so typed integrity assertions match Postgres.

Documented in [ADR 041](docs/adrs/041-storage-statement-contract-tests.md).

## Validation

- `go test -tags postgres_integration -timeout=15m ./internal/storage/v2/stmttest/...`
- `go test -tags spanner_integration -timeout=20m ./internal/storage/v2/stmttest/...`
- `go test -tags 'postgres_integration,spanner_integration' -timeout=25m ./internal/storage/v2/stmttest/...`

## Release notes / changeset

- Spanner FK `FailedPrecondition` → `ForeignKeyError`: patch changeset included (`.changeset/spanner-fk-error-mapping.md`).
- Shared stmttest harness / ADR: no extra changeset (test + docs only).

## Notes

Addresses review feedback: multi-dialect harness (both tags allowed), `t.Context()` used directly, ADR 041 for the shared suite approach. Harness helper is `forEachDialect` (dialect loop only; case names via `t.Run`).

<div><a href="https://cursor.com/agents/bc-4a384e09-dd22-4ed8-9eb9-19fc9a88c620"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-4a384e09-dd22-4ed8-9eb9-19fc9a88c620"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


internal/storage/v2/stmttest/main_test.go
```diff
@@ -0,0 +1,43 @@
...
+var seedProjectsTiedAt func(ctx context.Context, ids []string, createdAt time.Time) error
+
+func TestMain(m *testing.M) {
+	os.Exit(run(m))
+}
+
+func run(m *testing.M) int {
+	ctx := context.Background()
+
+	pool, stop, err := openPool(ctx)
```
adlerhurst: It feels like this will only work if the tests provider either the postgres or spanner tag but not both. Would it make sense to allow both as well? If we provide a runTest function which is called by the tests instead of `t.Run` we could loop over the given pools.

internal/storage/v2/stmttest/user_passkey_test.go
```diff
@@ -89,14 +94,11 @@ func TestUserPasskeyStatements_CreateGetListDelete(t *testing.T) {
 
 func TestUserPasskeyStatements_Update(t *testing.T) {
 	ctx := t.Context()
```
adlerhurst: this unnecessary assignment feels weird to read. Use `t.Context()` directly instead of creating a variable in all tests.

adlerhurst (commented):
I think it might be worth adding an ADR to make it clear for everybody.

PR #665, opened 2026-07-30
feat(database): add SQLite dialect and use it as local default

adlerhurst:
## TL;DR

SQLite becomes the zero-config **local** storage dialect (`&lt;data_dir&gt;/zitadel.db`). Postgres/Spanner stay production/testcontainer paths. Portable `stmttest` covers the statement contract; Spanner dual CI path is unchanged.

## How to review (suggested order)

1. **Product default / wiring** — skim first  
   - [`cmd/server/server.go`](cmd/server/server.go), [`apps/cli/src/commands/start.ts`](apps/cli/src/commands/start.ts)  
   - [`docs/operations/nextgen.example.yaml`](docs/operations/nextgen.example.yaml), [`docs/quick-start/configuration.md`](docs/quick-start/configuration.md)  
   - [`.changeset/sqlite-local-default.md`](.changeset/sqlite-local-default.md)

2. **SQLite dialect core** — main review surface  
   - Connect/DSN: [`internal/storage/v2/dialect/sqlite/dialect.go`](internal/storage/v2/dialect/sqlite/dialect.go) (pragmas, `:memory:` isolation, `0600`, URI-escaped paths)  
   - Schema: [`internal/storage/v2/dialect/sqlite/migration/sql/000001_init.sql`](internal/storage/v2/dialect/sqlite/migration/sql/000001_init.sql) (incl. ADR 024 `team_memberships` RESTRICT + user FK)  
   - Compiler: [`internal/storage/v2/dialect/sqlite/compiler.go`](internal/storage/v2/dialect/sqlite/compiler.go) + shared [`internal/storage/v2/dialect/compare/null_aware.go`](internal/storage/v2/dialect/compare/null_aware.go)  
   - Pool / statements: `pool.go`, `statement.go`, entity files under `dialect/sqlite/`

3. **Cross-dialect contract** — `stmttest` + small postgres/spanner touchups  
   - [`internal/storage/v2/stmttest/`](internal/storage/v2/stmttest/) (string fold `Übung`, nil-`ExpiresAt` cursor, membership FK)  
   - Auth-attempt / token accessor parity in postgres+spanner (no Spanner CI dual-path change)

4. **Mechanical / docs / cleanup** — lower risk  
   - Removed unused embedded Postgres (`dialect/postgres/embedded/`, CLI reap)  
   - Devcontainer / CONTRIBUTING / AGENTS / `@zitadel/testing` comments  
   - `go.mod` / `go.sum`

## Risk notes

- **Local-only default**: production Postgres/Spanner configs unchanged; file DB holds crypto key material → Path mode chmods `0600`.  
- **SQLite ASCII `LOWER`**: ignore-case LIKE folds in SQL; docs already caveat Unicode divergence from Postgres.  
- **Greenfield SQLite migration**: single `000001_init.sql` (no upgrade path from earlier SQLite drafts).  
- **History**: branch includes several merge-from-main commits; prefer path-based review over commit-by-commit.

## Validation

- `go test ./internal/storage/v2/database/ ./internal/storage/v2/dialect/sqlite/ ./internal/storage/v2/dialect/compare/`  
- `go test -tags=sqlite_integration ./internal/storage/v2/stmttest/ -run 'StringFilter|Token|TeamMembership'`  
- CI `full-pr` on latest push

## Release notes / changeset

Changeset present: [`.changeset/sqlite-local-default.md`](.changeset/sqlite-local-default.md) (`@zitadel/server` minor). Follow-up refactors are internal-only (no extra changeset).

## Notes

Review threads on the PR are resolved. Latest cleanup (`1b152920`) dedupes NULL-aware keyset compare into `dialect/compare` (net −168 lines across the three compilers).

<div><a href="https://cursor.com/agents/bc-65c2ca65-3a7a-4301-94fb-febeac23d53d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-65c2ca65-3a7a-4301-94fb-febeac23d53d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


internal/storage/v2/dialect/sqlite/compiler.go
```diff
@@ -0,0 +1,256 @@
...
+		if filter.IgnoreCase {
+			c.WriteString("LOWER(")
+			c.WriteString(col)
+			c.WriteString(")")
+			value = strings.ToLower(filter.Value)
+		} else {
+			c.WriteString(col)
+		}
+		c.WriteString(" LIKE ")
+		pattern.CompileLikePattern(c, filter.Match, value)
```
fforootd: **Bug: the escaped `LIKE` pattern needs an `ESCAPE` clause on SQLite.** `pattern.EscapeLikePattern` backslash-escapes `%`/`_`/`\`, which works on Postgres (backslash is the default escape) and Spanner (GoogleSQL escapes with backslash natively) — but SQLite has **no default escape character**, so the backslash becomes a literal:

```
sqlite> SELECT 'a%b' LIKE 'a\%b';             -- 0: the literal value no longer matches itself
sqlite> SELECT 'a\Zb' LIKE 'a\%b';            -- 1: matches backslash + anything instead
sqlite> SELECT 'a%b' LIKE 'a\%b' ESCAPE '\';  -- 1: correct
```

Since `_` appears in most emails/loginnames, any `Contains`/`StartsWith`/`EndsWith` needle containing `_`, `%`, or `\` silently over- or under-matches.

```suggestion
		pattern.CompileLikePattern(c, filter.Match, value)
		c.WriteString(` ESCAPE '\'`)
```

A compiler test with a needle like `100%_a\b` would lock this in together with the case-sensitivity fix on the DSN.

internal/storage/v2/dialect/sqlite/dialect.go
```diff
@@ -0,0 +1,113 @@
...
+		return "", fmt.Errorf("sqlite: resolve path: %w", err)
+	}
+	return "file:" + filepath.ToSlash(abs) + "?" + sqlitePragmaQuery, nil
+}
+
+// _txlock=immediate acquires the write lock at BEGIN so deferred read
+// transactions cannot upgrade into SQLITE_BUSY_SNAPSHOT after another writer
+// commits. busy_timeout covers remaining lock waits if MaxOpenConns is raised.
+// Pragmas are applied via DSN so every opened connection gets them.
+const sqlitePragmaQuery = "_txlock=immediate&_pragma=busy_timeout(30000)&_pragma=foreign_keys(1)&_pragma=journal_mode(WAL)&_pragma=synchronous(NORMAL)"
```
fforootd: **Parity: SQLite `LIKE` is case-insensitive for ASCII by default** (`'ABC' LIKE 'abc'` → 1), so the non-`IgnoreCase` `StartsWith`/`Contains`/`EndsWith` filters behave case-insensitively here but case-sensitively on Postgres and Spanner — which cuts against the "local behaves like prod" premise of the default flip.

Every connection already gets its pragmas from this DSN, so it's one more entry (the `IgnoreCase` path is unaffected — it LOWERs both sides explicitly):

```suggestion
const sqlitePragmaQuery = "_txlock=immediate&_pragma=busy_timeout(30000)&_pragma=foreign_keys(1)&_pragma=journal_mode(WAL)&_pragma=synchronous(NORMAL)&_pragma=case_sensitive_like(1)"
```

Worth pairing with a small integration assertion that a wrong-case `Contains` misses, since no stmttest suite covers string filters yet.

Related, probably just a documented limitation for a local dialect: modernc's `LOWER()` only folds ASCII (no ICU), so `IgnoreCase` matching on non-ASCII (`Ü`/`ü`) diverges from Postgres. A line in the configuration docs next to "single-writer limits apply" would cover it.

internal/storage/v2/dialect/sqlite/auth_attempt.go
```diff
@@ -0,0 +1,387 @@
...
+func (as authAttemptStatements) HandoffAuthAttempt(ctx context.Context, attempt *domain.AuthAttempt) error {
+	if attempt.HandoffToken == nil {
+		return fmt.Errorf("failed to handoff auth attempt: handoff token is required")
+	}
+	id, err := parseIdentity(attempt.ID)
+	if err != nil {
+		return err
+	}
+	now := time.Now().UTC()
+	if _, err := as.client.Exec(ctx, handoffAuthAttemptStmt,
```
fforootd: **Parity gap: silent success on a missing attempt.** Postgres does `QueryRow` + `RETURNING handed_off_at`, so handing off a nonexistent attempt surfaces `NoRowFoundError`; this `Exec` ignores the affected count, "succeeds", and stamps `HandedOffAt` client-side.

SQLite supports `RETURNING` (this file already uses it elsewhere), so the Postgres shape ports directly — or `execAffected` + `database.NewNoRowFoundError(sql.ErrNoRows)` on `n == 0`. The stale-challenge paths below (`ChallengeSucceeded`/`ChallengeFailed`) got exactly this right, which makes the asymmetry here easy to trip over later.

internal/storage/v2/dialect/sqlite/migration/sql/000001_init.sql
```diff
@@ -0,0 +1,510 @@
...
+    scope           TEXT    NOT NULL DEFAULT ('[]'),
+    expires_at      INTEGER,
+    created_at      INTEGER NOT NULL,
+    CONSTRAINT fk_tokens_project
+        FOREIGN KEY (project_id) REFERENCES projects (id) ON DELETE CASCADE
+);
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+CREATE TABLE sessions (
```
fforootd, line 210: **Schema parity: `sessions` and `tokens` have no FK to `users`.** Postgres and Spanner cascade both on user deletion — `tokens` has `FOREIGN KEY (project_id, user_id) REFERENCES users ... ON DELETE CASCADE` (000009_tokens.sql), and main just added the same `fk_sessions_user ... ON DELETE CASCADE` to `sessions.user_id` on both dialects for delete-user (#637). Here neither FK exists, so deleting a user on SQLite leaves their sessions and session tokens alive and resolvable — a security-relevant divergence in the default local dialect once this merges with main.

Both FKs port directly (same shape as `fk_user_passwords_user` further down). Also absent from `tokens`, lower priority: the `chk_tokens_type_identifiers` CHECK and the `sessions → user_agents` FK.

Probably easiest to fold into the merge-main round this branch needs anyway — ideally with a stmttest case asserting sessions/tokens are gone after `DeleteUserByID`, so all three dialects are held to it.

internal/storage/v2/dialect/sqlite/compiler.go
```diff
@@ -0,0 +1,256 @@
...
+	c.Builder.WriteString(s)
+}
+
+func (c *statementCompiler) WriteArg(arg any) {
+	writeArg(c, arg)
+}
+
+// writeArg appends a ? placeholder and converts time.Time / time.Duration to
+// the integer representations used by SQLite (Unix nanoseconds).
+func writeArg(c *statementCompiler, arg any) {
```
fforootd, line 259: Nit: worth extending this comment with the representable range — INTEGER Unix-nanos covers ~1678–2262, and a zero `time.Time{}` passed **by value** would store an overflowed number that round-trips as year 1754 rather than zero. All current write paths use `now` or `*time.Time`, so this is just a guard rail for future statements.

packages/testing/README.md
```diff
@@ -139,9 +139,9 @@ Measured on an arm64 macBook (dev build, July 2026):
 
 | Operation | Time |
 | --- | --- |
-| Cold boot (fresh data dir: initdb + migrations + health) | ~20–27s |
-| Warm restart (existing data dir) | ~15s |
-| Stop incl. embedded-Postgres reap | ~12s |
+| Cold boot (fresh data dir: SQLite migrate + health) | typically under a few seconds |
```
fforootd, line 236: Nit: the table header above still says "Measured on an arm64 macBook (dev build, July 2026)" but these rows are now estimates. Either re-measure (a cold boot with the SQLite default would be a great number to publish) or soften the header — as-is it presents guesses as measurements.

AGENTS.md
```diff
@@ -358,8 +358,10 @@ example via nvm) before running any `corepack` or `pnpm` command.
 ### System dependencies for Go tests
 
 `libicu-dev` and `libssl-dev` are required at runtime by `embedded-postgres`
-(the Go test helper that auto-downloads PostgreSQL, currently 18). If Go
-tests fail with missing-library errors, install them once with
+when intentionally running Postgres-backed tests or an explicit
```
fforootd: This paragraph is stale in a useful way: the Postgres-backed test path no longer uses `embedded-postgres` at all — `dbtest.Postgres` goes through `testdb.PostgresDSN` (`ZITADEL_TEST_POSTGRES_URL` or a testcontainer). And after this PR removes the `cmd/server` call site, `dialect/postgres/embedded` has zero importers, so `libicu-dev`/`libssl-dev` aren't needed on any in-repo path anymore.

Suggest dropping the section (or scoping it to "only if you run the embedded-postgres helper manually"). Whether to remove the package + `fergusstrange/embedded-postgres` dependency entirely is follow-up material, not this PR.

fforootd (changes requested):
Really solid port overall — this has the right shape, and the discipline shows: the multi-write rule is respected everywhere I traced (`CreateAuthAttempt`, the `InsertSession` → `CreateSessionToken` chain, `ExchangeSession`), `withTransaction` deliberately no-op-nests like Spanner (the weakest common guarantee — correct for cross-dialect code), the single-connection pool + `_txlock=immediate` is the right SQLite concurrency answer, and the 8-way concurrent exchange test targets exactly the failure mode that motivated it. On the schema side, the generated `expires_at (updated_at + time_to_live)` mirrors the Postgres trigger precisely, and the generated-column trick for the single-active-DEK unique index is a clean emulation of a partial expression index. stmttest registration follows ADR 041 to the letter, and docs/ADR/AGENTS land in the same PR. I ran `go build ./...`, `go vet`, the unit tests, and `go test -tags=sqlite_integration ./internal/storage/v2/stmttest/ ./internal/storage/v2/dialect/sqlite/` locally on the pre-merge head (fcfca5f3) — all green, sub-second suites, so the speed story holds up.

Requesting changes for two correctness bugs in the string-filter path (inline: missing `ESCAPE` clause, `LIKE` case-insensitivity) and one structural item:

**The freshly pushed merge head (4ddebc2a) brings in latest main but doesn't compile yet** — main's #637/#649 grew the statement interfaces, and the sqlite side hasn't been ported:

```
sqlite/crypto_keys.go:131: *cryptoKeyStatements does not implement service.CryptoKeyStatements (missing method CreateSigningKey)
sqlite/user.go:312: *userStatements does not implement service.UserStatements (missing method UserExists)
```

So the port round still needed: `UserExists`, `GetSigningKey`/`CreateSigningKey` (+ signing-key columns/tables in the init migration, mirroring what #649 did to postgres/spanner `000012_crypto_keys.sql`), and the user-FK parity from #637 — see the inline comment on the init migration for that last one.

One follow-up worth tracking separately (pre-existing gap, not this PR's fault): stmttest has no string-filter contract suite for any dialect — which is exactly where the engines genuinely differ. Both LIKE bugs here would have been caught by one.

internal/storage/v2/dialect/sqlite/compiler.go
```diff
@@ -0,0 +1,262 @@
...
+			c.WriteString(")")
+			value = strings.ToLower(filter.Value)
+		} else {
+			c.WriteString(col)
+		}
+		c.WriteString(" LIKE ")
+		pattern.CompileLikePattern(c, filter.Match, value)
+		// SQLite has no default LIKE escape character; EscapeLikePattern uses
+		// backslash, so ESCAPE must be stated explicitly.
+		c.WriteString(` ESCAPE '\'`)
```
fforootd, line 218: **Bug: ignore-case LIKE filters miss exact same-case non-ASCII matches.**

The column is folded with SQLite's `LOWER()` (ASCII-only), but the value with Go's Unicode-aware `strings.ToLower`, so the two sides disagree on any non-ASCII rune. Repro on this branch, project named `Übung`:

| filter | result |
| --- | --- |
| `StringEqualFold(name, "Übung")` | ✅ match |
| `StringContains(name, "Übung")` | ✅ match |
| `StringContainsFold(name, "Übung")` | ❌ 0 rows |

`strings.ToLower("Übung")` → `übung`, while `LOWER('Übung')` keeps `Übung` — so the *ignore-case* filter is stricter than the case-sensitive one, and searching for exactly what's stored finds nothing.

Suggested fix: fold the value on the SQL side too, so both sides use the same ASCII fold (consistent with the `StringMatchEqual` branch above): keep `value = filter.Value` and wrap the compiled pattern in SQL, e.g. `LIKE LOWER('%' || ? || '%') ESCAPE '\'` (the escape chars `\`, `%`, `_` are caseless, so escaping is unaffected).

Per the new AGENTS.md rule this contract belongs in `stmttest` — a portable case with a non-ASCII fixture asserting `ContainsFold(sameCase)` matches would pin it for all three dialects. The docs caveat about ASCII-only `LOWER()` covers divergence from Postgres, but not same-case misses.

---

internal/storage/v2/dialect/sqlite/dialect.go
```diff
@@ -0,0 +1,113 @@
...
+		_ = sqlDB.Close()
+		return nil, err
+	}
+	return newPool(sqlDB), nil
+}
+
+func (c Config) dsn() (string, error) {
+	if c.DSN != "" {
+		return c.DSN, nil
+	}
```
fforootd, line 72: **`Config.DSN` silently bypasses the required pragmas.**

When `DSN` is set, none of `sqlitePragmaQuery` is applied unless the caller includes it by hand. SQLite defaults `foreign_keys` to OFF, so a raw DSN silently disables every FK/cascade the schema and the typed `ForeignKeyError` mapping rely on; missing `case_sensitive_like(1)` flips LIKE semantics; missing `_txlock=immediate` reintroduces exactly the `SQLITE_BUSY_SNAPSHOT` class this PR fixes.

Suggest appending the default pragmas when the DSN doesn't already carry them — or at minimum a loud doc comment on `DSN` stating they're required for correct behavior.

---

internal/storage/v2/dialect/sqlite/dialect.go
```diff
@@ -0,0 +1,113 @@
...
+	if c.Path == "" {
+		return "", fmt.Errorf("sqlite: path or dsn is required")
+	}
+	if c.Path == ":memory:" {
+		// Shared cache so multiple pool connections see the same in-memory DB.
+		return "file:zitadel?mode=memory&cache=shared&" + sqlitePragmaQuery, nil
+	}
```
fforootd, line 81: **`:memory:` resolves to one shared DB per process.**

Every `Config{Path: ":memory:"}` maps to the same shared-cache name (`file:zitadel`), so two independently connected pools in one process share a single database (and one pool closing last keeps it alive). Fine for the single-server path, surprising for tests/embedders — a unique name per `Connect` (counter/random suffix) or a doc note on the field would remove the trap.

---

internal/storage/v2/dialect/sqlite/dialect.go
```diff
@@ -0,0 +1,113 @@
...
+	dir := filepath.Dir(c.Path)
+	if dir != "" && dir != "." {
+		if err := os.MkdirAll(dir, 0o700); err != nil {
+			return "", fmt.Errorf("sqlite: create data directory: %w", err)
+		}
+	}
```
fforootd, line 87: Hardening nit: the parent directory is created `0o700`, but the DB file itself is created by the driver with default (umask-dependent) permissions — and it stores encryption/signing key material and token hashes. Worth tightening the file to `0600` (pre-create it, or chmod after first open), including the WAL/SHM siblings if practical.

---

internal/storage/v2/dialect/sqlite/token.go
```diff
@@ -0,0 +1,319 @@
...
+	},
+	domain.TokenFieldExpiresAt: {
+		SQLName: "expires_at",
+		Accessor: func(t *domain.Token) any {
+			if t.ExpiresAt == nil {
+				return time.Time{}
+			}
+			return *t.ExpiresAt
+		},
+		Coerce: database.CoerceTime,
```
fforootd, line 317: Edge case: this accessor returns `time.Time{}` for a nil `ExpiresAt`, and `ValuesFrom` feeds accessor output into pagination cursors when ordering by `expires_at` — which lands the overflowed `UnixNano()` sentinel that `writeArg`'s own comment warns about ("round-trips near year 1754") in a cursor. Consider returning `(*time.Time)(nil)` here instead; `writeArg` already maps that to SQL NULL, keeping the footgun out of cursors.

---

internal/storage/v2/dialect/sqlite/migration/sql/000001_init.sql
```diff
@@ -0,0 +1,605 @@
...
+    WHERE handoff_token IS NOT NULL;
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+CREATE TABLE user_agents (
+    id          INTEGER PRIMARY KEY AUTOINCREMENT,
+    project_id  TEXT    NOT NULL,
+    info        TEXT    NOT NULL DEFAULT ('{}'),
+    created_at  INTEGER,
+    updated_at  INTEGER,
```
fforootd, line 170: Nit: `user_agents.created_at` / `updated_at` are always NULL on SQLite — `InsertSession` omits them and there's no DEFAULT, while Postgres stamps `NOW()`. Nothing reads them today, but stamping them in the insert (like the other tables) keeps the schemas honestly parallel.

---

fforootd (commented):
Reviewed the full diff and verified at head `dde40a1`:

- `go build ./cmd/... ./internal/storage/v2/...` ✅
- `go test ./internal/storage/v2/dialect/sqlite/...` ✅
- `go test -tags sqlite_integration ./internal/storage/v2/stmttest/ ./internal/storage/v2/dialect/sqlite/` ✅ (contract suite in ~0.1s vs ~20s+ embedded-Postgres bring-up — lovely dev-loop win)

**Overall: solid.** The dialect mirrors postgres/spanner structurally, the multi-write `withTransaction` rule is respected everywhere it applies (`CreateSessionToken` matches the postgres shape — only reachable from wrapped paths), the Spanner `THEN RETURN` handoff fix is correct (`collectOneRow` wraps `ErrRowNotFound` → `NoRowFoundError` parity), and the postgres/spanner schemas already carry the user FKs the new cascade suite asserts, so those CI legs should stay green. The stmttest requirement added to AGENTS.md is a welcome ratchet.

**Before merge, I'd address:**

1. The confirmed `*Fold` LIKE bug (inline in `compiler.go`, with repro) — ignore-case contains/starts/ends misses exact same-case non-ASCII matches.
2. A decision on `Config.DSN` bypassing the required pragmas (inline in `dialect.go`) — a raw DSN silently runs with `foreign_keys` OFF.

The remaining inline comments (`:memory:` shared name, zero-time cursor edge in `token.go`, DB file perms, `user_agents` timestamps) are nits/follow-ups.

**Also noted:**

- Branch is behind `main`; the missing commits (#718/#719) touch only the TS SDKs — plain update-branch, no semantic overlap.
- `zitadel stop --json` drops `data.runtime.embedded_postgres` from the envelope — strictly a shipped CLI behavior change with no `@zitadel/cli` changeset. Defensible since the runtime it described is gone, but worth an explicit call against the decision table.

---

internal/storage/v2/dialect/sqlite/token.go
```diff
@@ -0,0 +1,324 @@
...
+	domain.TokenFieldScope: {
+		SQLName:  "scope",
+		Accessor: func(t *domain.Token) any { return t.Scope },
+		Coerce:   database.CoerceSliceAsAny(database.CoerceStringValue),
+	},
+	domain.TokenFieldExpiresAt: {
+		SQLName: "expires_at",
+		Accessor: func(t *domain.Token) any {
+			if t.ExpiresAt == nil {
+				return (*time.Time)(nil)
```
fforootd, line 313: Could we cover the full cursor round-trip before resolving this one? Returning a typed nil avoids the zero-time sentinel, but `Cursor.Marshal` serializes it as JSON `null`, and `database.CoerceTime` rejects `nil` on the next request. I reproduced it with two non-expiring tokens ordered by `(expires_at, token_id)`: page 1 returns a cursor, then page 2 fails with `The provided cursor is invalid.` Since the same accessor change landed in Postgres and Spanner too, it would be great to add one portable `stmttest` and handle nullable cursor coercion/comparison explicitly.

internal/storage/v2/dialect/sqlite/migration/sql/000001_init.sql
```diff
@@ -0,0 +1,605 @@
...
+    team_id     TEXT    NOT NULL,
+    user_id     TEXT    NOT NULL,
+    status      TEXT    NOT NULL DEFAULT ('active'),
+    created_at  INTEGER NOT NULL,
+    updated_at  INTEGER NOT NULL,
+    PRIMARY KEY (project_id, team_id, user_id),
+    CONSTRAINT fk_team_memberships_project
+        FOREIGN KEY (project_id) REFERENCES projects (id) ON DELETE CASCADE,
+    CONSTRAINT fk_team_memberships_team
+        FOREIGN KEY (project_id, team_id) REFERENCES teams (project_id, id) ON DELETE CASCADE
```
fforootd: Could we align these lifecycle FKs with the Postgres/Spanner schema before making SQLite the default? This table currently has no `(project_id, user_id) -> users` FK, so `CreateTeamMembership` accepts a membership for a nonexistent user—I reproduced a nil error with a real project/team and `missing-user`. The team FK also cascades here, while the other dialects intentionally use RESTRICT/NO ACTION per ADR 024 so lifecycle services own that behavior. A portable negative `stmttest` for the missing-user case would keep all three dialects honest.

internal/storage/v2/dialect/sqlite/dialect.go
```diff
@@ -0,0 +1,195 @@
...
+			return "", "", fmt.Errorf("sqlite: create data directory: %w", err)
+		}
+	}
+	// Absolute path in URI form; query params apply on every new connection
+	// (connection-scoped pragmas like foreign_keys default OFF otherwise).
+	abs, err := filepath.Abs(c.Path)
+	if err != nil {
+		return "", "", fmt.Errorf("sqlite: resolve path: %w", err)
+	}
+	return "file:" + filepath.ToSlash(abs) + "?" + sqlitePragmaQuery, abs, nil
```
fforootd: Would you mind building this `file:` DSN with URI path escaping instead of concatenation? A configured path containing `?` is parsed as a query boundary: with `nested?mark/zitadel.db`, connect + migrate succeeds but no database is created at the configured path. `#` and percent escapes carry the same class of risk. A small table test for URI metacharacters would make the path contract safer.

internal/storage/v2/dialect/sqlite/main_test.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+var testPool *Pool
+
+func TestMain(m *testing.M) {
+	dir, err := os.MkdirTemp("", "zitadel-sqlite-integration-*")
+	if err != nil {
+		panic(err)
+	}
+	defer os.RemoveAll(dir)
```
fforootd, line 28: Small cleanup follow-up: could this use a `runTests` helper like the Postgres suite? `os.Exit` skips deferred functions, so this `defer os.RemoveAll(dir)` never runs and each SQLite integration run leaves its temporary DB/WAL/SHM directory behind. Returning `m.Run()` from a helper would let the close/removal defers execute before `os.Exit`.

fforootd (commented):
Thanks for the thorough follow-up pass here. I exercised a few SQLite parity and failure paths and found four remaining edge cases that seem worth tightening before merge:

fforootd (approved):

PR #672, opened 2026-07-30
feat(api): add the team update endpoint

### LINKED ISSUE #619: Team: Update
## Context

Part of #598. Team CRUD is missing Update entirely: only Create and Get exist today. Team's only mutable field is `name`, tracked in #618.

## Scope / Tasks

- Add storage support for updating a team's name, on both postgres and spanner.
- Add a service-level Update operation with the same validation conventions used elsewhere: empty name is rejected, an unknown team is a not-found error.
- Add a `PATCH /teams/{team_id}` endpoint and its request schema. `name` is required in the request body; there's no null/clear semantics since it's the only mutable field and can never be legitimately cleared.
- Gate the endpoint the same way `CreateTeam` is gated (project-secret scope, write access). No per-member team authorization exists yet; this stays on the project-secret model like every other team operation.
- Map the new validation error to a 400 response.

## Acceptance Criteria

- [x] `PATCH /teams/{team_id}` updates a team's name and returns the updated team state.
- [x] Empty `name` returns 400.
- [x] Unknown team id returns 404.
- [x] Request without write access to the project returns 403.
- [x] Update is implemented and tested on both postgres and spanner dialects.

## Depends on

- #618 (Team: add name field)

grvijayan:
## Summary

Closes #619.

- `PATCH /teams/{team_id}` renames a team. The name is trimmed and must be 1 to 200 characters, matching `createTeam`.
- The name is optional in the body, so a second patchable field can be added later without forcing callers to resend it. #619 asked for it to be required; that would have had to be loosened as soon as a second field landed. Omitting it today still returns 400 `team.name_invalid`, since it is the only field.
- Names are unique per project ignoring case, so a taken name returns 409.
- Only active teams can be renamed. A deactivated team is indistinguishable from a missing one and returns 404.
- `createTeam` now declares its 403 response. It was missing, so the generated client decoded it as an untyped error while `updateTeam` decoded it as a typed one.
- `ensureTeam` in the user bootstrap no longer treats every `UniqueError` as "the team already exists". Since #646 added the per-project unique name, the violation can come from the name index, leaving the team missing and a later `team_memberships` insert to fail on its foreign key. It now looks the team up first, because `UniqueError` does not carry the constraint name on every dialect.

## Validation

- `moon run server:generate && moon run server:check-generate`
- `moon run server:openapi`
- `go build ./...`, `gofmt -l internal/`, `go vet` on both integration tags
- `go test ./internal/domain/... ./internal/service/... ./internal/api/...`
- `go test -tags postgres_integration ./internal/bootstrap/... ./internal/storage/v2/dialect/postgres/... ./internal/api/integration_test/...`
- `go test -tags spanner_integration ./internal/storage/v2/dialect/spanner/... ./internal/api/integration_test/...`
- `corepack pnpm exec changeset status --since origin/main`

## Release notes / changeset

- Changeset: `.changeset/team-update.md` — team update endpoint, `@zitadel/server` minor.

## Notes

- Storage adds a single `UpdateTeam` statement per dialect with `RETURNING` / `THEN RETURN`, gated on `status = 'active'`, so the 404 for a deactivated team comes from the write itself. No migration.
- `UpdateTeamInput.Name` is a `*string` so the service can tell an absent name from an empty one. The handler no longer flattens the two, which would return the wrong error once a second field exists.


grvijayan:
to be merged after https://github.com/zitadel/nextgen/pull/674

internal/service/team_test.go
```diff
@@ -133,7 +133,102 @@ func TestTeamService_GetTeam(t *testing.T) {
...
+		name      string
+		input     service.UpdateTeamInput
+		setupStmt func(*servicemocks.MockAllStatements)
+		wantErr   error
+		check     func(t *testing.T, got *domain.Team)
+	}{
+		{
+			name:  "ok",
+			input: service.UpdateTeamInput{ProjectID: "proj_1", TeamID: "team_1", Name: new("renamed")},
+			setupStmt: func(s *servicemocks.MockAllStatements) {
```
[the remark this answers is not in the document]
grvijayan: it compiles fine and the tests pass

.changeset/team-update.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zitadel/server": minor
```
fforootd, line 2: This PR changes the published TS client too: `packages/api` regenerates from `api/openapi/**`, so the client gains `updateTeam` and — via the `get-team-response.yaml` → `team-response.yaml` rename — loses the exported `GetTeamResponse` type (plus its zod/msw derivatives) in favor of `TeamResponse`. Without a bump here `@zitadel/api` won't republish, so SDK consumers never see `updateTeam`, and the rename lands silently in whatever later release does bump it. Precedent: `delete-user-endpoint.md` bumps both packages for the same shape of change.

```suggestion
"@zitadel/server": minor
"@zitadel/api": minor
```

Worth a clause in the note for the `GetTeamResponse` → `TeamResponse` rename as well — no in-repo consumer references the old name, so it's only the published surface. While here: the file is missing its trailing newline.

api/openapi/endpoints/teams/methods.yaml
```diff
@@ -33,6 +33,12 @@ post:
         application/json:
           schema:
             $ref: ../../components/error-details.yaml
+    '403':
```
fforootd, line 36: Good catch on the missing 403 — and its rationale ("the generated client decoded it as an untyped error while `updateTeam` decoded it as a typed one") applies verbatim to 404: a foreign-project secret on create returns 404 `team.project_not_found` (asserted in `authz_test.go`), and `createTeam` declares no 404. Since you're touching this file anyway, closing both gaps would leave the two write ops symmetric:

```yaml
    '404':
      description: Project not found.
      content:
        application/json:
          schema:
            $ref: ../../components/error-details.yaml
```

api/openapi/endpoints/teams/by_id/methods.yaml
```diff
@@ -40,3 +40,71 @@ get:
...
+          schema:
+            $ref: ../../../components/error-details.yaml
+    '403':
+      description: Insufficient permissions.
+      content:
+        application/json:
+          schema:
+            $ref: ../../../components/error-details.yaml
+    '404':
+      description: Team not found, or not active.
```
fforootd: This status also carries `team.project_not_found` when the bearer isn't bound to the project — the new authz test asserts exactly that — so the description undersells it slightly.

```suggestion
      description: Team not found or not active, or the project does not exist.
```

api/openapi/endpoints/teams/update-team-request.yaml
```diff
@@ -0,0 +1,7 @@
+type: object
```
fforootd, line 1: Optional: `minProperties: 1` would encode "PATCH needs at least one field" at the contract level, and it stays correct once a second field lands. Even if ogen doesn't enforce it server-side, the contract would document the requirement instead of leaving `{}` looking valid while the service 400s it. Your service-level choice is defensible and well-explained in the PR body, so take or leave. Also: missing trailing newline.

internal/bootstrap/users/ensure.go
```diff
@@ -43,17 +43,23 @@ func ensureProject(ctx context.Context, stmts service.AllStatements, projectID s
...
+		return nil
+	}
+	// Only a missing row means we still have to create it.
+	if _, ok := errors.AsType[*database.NoRowFoundError](err); !ok {
+		return fmt.Errorf("get team %q: %w", teamID, err)
+	}
 	// The bootstrap header carries no team name, so derive a placeholder
 	// name from the team ID to satisfy the NOT NULL name column.
-	err := stmts.CreateTeam(ctx, &domain.Team{
+	if err := stmts.CreateTeam(ctx, &domain.Team{
```
fforootd, line 58: The old over-broad `UniqueError` swallow also masked the one benign case: two replicas bootstrapping the same files concurrently now hard-fail on the PK collision that used to read as "exists". In-process import is sequential, so this is cross-replica-only and theoretical — but a re-`GetTeamByID` on `UniqueError` (return nil if the team is there now) would make it airtight for one extra branch. Non-blocking.

internal/storage/v2/dialect/postgres/team_test.go
```diff
@@ -149,6 +149,101 @@ func TestTeamStatements_Get(t *testing.T) {
 	})
 }
 
+func TestTeamStatements_UpdateTeam(t *testing.T) {
```
fforootd, line 152: Convention note, explicitly non-blocking: this suite and its Spanner twin are ~95 near-identical lines of pure behavioral parity — the shape `stmttest/` exists for per ADR 041 / `internal/storage/v2/AGENTS.md` (shared suites asserting through `service.AllStatements`; dialect packages keep engine-specific tests). Teams predate stmttest, so you're following the existing placement — fine to land as-is, but porting the team suites over would be a good follow-up.

fforootd (commented):
Thanks — this is a tight PR. The layering came out clean: `ValidateTeamName` shared between create and update, `*string` on `UpdateTeamInput` keeping absent ≠ empty, and the status-gated `UPDATE … RETURNING` so the 404 for a deactivated team comes from the write itself. Test coverage is genuinely thorough — both dialects, typed-response assertions, the raw-request test to reach the server-side validators the generated client short-circuits, and the regression test pinning the `ensureTeam` name-index failure mode. The `createTeam` 403 fix and the bootstrap fix are both good catches.

Verified locally on the head commit: `go build`, `gofmt -l`, and the domain/service/bootstrap/api unit suites pass; the branch is current with main. I didn't rerun the integration tags.

One item I'd fix before merge — the changeset misses `@zitadel/api` — the rest is contract polish and follow-up material. Details inline.

🤖 Review drafted with [Claude Code](https://claude.com/claude-code)

internal/storage/v2/dialect/postgres/team_test.go
grvijayan: Will do this in a follow-up PR to keep the scope of this PR small.

fforootd: works for me

api/openapi/endpoints/teams/update-team-request.yaml
grvijayan: `{}` would move from 400 `team.name_invalid` to a generic 400 `req.invalid`, because the failure happens in the body decoder before the handler runs. And the check is looser than it reads: ogen increments its property count for every key in the object, unknown ones included, so `{"foo": 1}` passes `minProperties` and still lands on `team.name_invalid`. We would end up with two codes for the same mistake, and the specific one only on the less obvious input.

PR #666, opened 2026-07-30
docs(authz): Wave 0 permission storage DDL spike

adlerhurst:
## Summary

Adds the Wave 0 design spike for portable FGA relational storage (ADR 032–033 / epic #419): locked decisions D1–D14, Postgres/Spanner-oriented DDL strawman, dual-write membership edges, check/list SQL sketches, end-to-end narrative, and a #333 cross-project grant depiction (foreign user/team principal on the protected `project_id`).

Links the doc from the API design index and points ADR 032/033 follow-ups at it so #422 can implement migrations without re-litigating the schema shape.

## Validation

- Doc review against plan acceptance criteria (D1–D14 present, dual-write + #333 depiction, deferrals for Leopard / partitioning / app grants)
- Addressed thermo-nuclear review P1–P8: one check SQL shape with `$principal_home_project_id`, relations+closure only (D14), tighter DDL CHECKs, `authz.md` / ADR follow-up alignment
- Docs-only change; no runtime or migration code

## Release notes / changeset

No changeset required — no shipped behavior changed (docs only).

## Notes

- Addressed thermo-nuclear review P1–P8 on this branch
- Out of scope: Goose migrations, `internal/authz/*`, resolver (#423), OpenFGA compiler (#421), implementing #333 product/identity
- Related: #419, #422, #333

<div><a href="https://cursor.com/agents/bc-11b6c032-d3d0-4f51-b3d7-a13dae256e44"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-11b6c032-d3d0-4f51-b3d7-a13dae256e44"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

Issue #423, opened 2026-06-26
Permission Resolver & SQL List Predicate Injection


muhlemmer:
## Context

[ADR 033 §4](https://github.com/zitadel/nextgen/blob/main/docs/adrs/033-internal-permission-management.md#4-resolver-and-list-filtering-for-internal-resources) defines the runtime authorization path. The resolver lives in `internal/authz/resolver` and must support both single-resource checks and list endpoint predicate injection.

The request pipeline is:
```
credential → resolve path/body/query scope → permission resolver → scope-bound DAL
```

## Single-Resource Checks

Indexed semi-joins over:
- principal's assignments
- relation/permission closure (pre-computed at policy upload)
- credential class constraints
- resolved scope

## List Endpoint Predicate Injection

Every protected list endpoint declares `resource_kind`, `operation`, `scope_source`, and `required_permissions`. The resolver returns a SQL predicate that is injected into the resource query. Repositories expose which columns carry `project_id`, `team_id`, and owner/resource identifiers. The database filters rows in a single query — no O(n) per-row checks in application code.

## Tasks

- [x] Implement `internal/authz/resolver`: single-resource check using indexed semi-join
- [x] Implement list predicate builder: generates a scope-bound SQL predicate from the principal's assignments and resolved scope
- [x] Define the repository metadata interface (which columns map to `project_id`, `team_id`, etc.)
- [x] Wire the resolver into the API request pipeline (after `resource_scope_index` lookup)
- [x] Conformance tests comparing single-resource checks and list predicates across PostgreSQL and Spanner
- [x] Ensure decisions are cached only within a single request; no cross-request decision cache at MVP
- [x] **L4 check/list oracle tests (required with resolver)** — per [`docs/design/api/authz-testing.md`](docs/design/api/authz-testing.md) §L4 / rollout slice C:
  - Generate random profile-valid models (reuse `authztest.GenerateValidModel`) plus random tuples (assignments, membership edges, RSI)
  - Compare an in-memory graph oracle to resolver `Check` / `ListObjects` — they must agree
  - Oracle rules for the MVP profile: direct assignment, computed-userset via closure, bounded TTU, team principals via `authz_membership_edges` only
  - Run as dialect-backed tests (Postgres + Spanner at minimum); do not treat OpenFGA check as source of truth (optional differential later)

## References

[...]

Issue #422, opened 2026-06-26
Relational Schema & Migrations for Authorization


muhlemmer:
## Context

[ADR 033 §2](https://github.com/zitadel/nextgen/blob/main/docs/adrs/033-internal-permission-management.md#2-resource_scope_index-and-canonical-storage) requires all authorization facts to be first-class local rows in the same database as the resources they protect — no sidecar, no synchronization.

Schema must be portable across PostgreSQL and Spanner per the constraint in [internal/storage/AGENTS.md](https://github.com/zitadel/nextgen/blob/main/internal/storage/AGENTS.md).

## Tables / Entities to Design

- `resource_scope_index` — maps globally-addressable resource IDs to project/team scope (already referenced in [URL Architecture](https://github.com/zitadel/nextgen/blob/main/docs/design/api/url-architecture.md))
- Catalog tables: permission/relation definitions, expression edges, optional bundle aliases, relation-implication closure — versioned per catalog
- `assignments` — binds principals to permissions/relations at explicit scopes
- `app_grants` — binds users or teams to apps/app groups for token claims
- Agent delegation rows: grantor, delegation ID, expiry/revocation state, scope, audit provenance

## Constraints

- Relation implication closure is stored on policy/schema update, not computed per membership or grant write
- Authorization rows must carry the same residency/partition metadata as the resources they protect
- All cross-project access (staff, operator, agency, support) must be explicit grants with scope, expiry, grantor, and audit provenance — no implicit inheritance

## Tasks

- Design schema for all tables above
- Write portable SQL migrations (PostgreSQL + Spanner compatible)
- Add indexes sufficient for the semi-join patterns used by the resolver (see #421 and the resolver issue)
- Ensure `team_memberships` remains separate from lifecycle ownership per [ADR 024](https://github.com/zitadel/nextgen/blob/main/docs/adrs/024-user-team-lifecycle-ownership.md)

## References

- [ADR 033 §2 — `resource_scope_index` and canonical storage](https://github.com/zitadel/nextgen/blob/main/docs/adrs/033-internal-permission-management.md#2-resource_scope_index-and-canonical-storage)
- [ADR 024: User/Team Lifecycle & Ownership](https://github.com/zitadel/nextgen/blob/main/docs/adrs/024-user-team-lifecycle-ownership.md)
- [URL Architecture — resource_scope_index](https://github.com/zitadel/nextgen/blob/main/docs/design/api/url-architecture.md)
[...]

PR #677, opened 2026-07-31
feat(database): Wave 1 authz MVP storage (#422)

adlerhurst:
## Summary

Wave 1 authz MVP storage on top of Wave 0 (#666): Goose migrations for the six tables, dual-write resource-scope identifiers + membership edges from existing project/team/user/membership statements, and `PersistCatalogVersion` for compiler `CatalogMutations` (#720). SQLite gets the same schema and dual-write path for local parity.

Review / maintainability follow-ups:
- Membership-edge deletes use one `DeleteAuthzMembershipEdges(filter)` API (kept `ForTeamDeactivate` for the subquery).
- Shared `dialect/authz` holds multi-write user helpers, `MembershipEdgeSchema`, and catalog row builders (engines keep SQL). Removed the three identical dual-write adapter files and identity wrappers.
- Typed authz discriminators (including DDL-aligned principal types), assignment scope constructors, and `AuthzMembershipEdgeKey`.
- Spanner `activeUniqueKey` lives in Spanner storage (not domain).
- Spanner membership-edge upsert uses a `created_at` no-op update (PK updates rejected).
- Authz enum SQL parameters bind via `.String()` (matching `MembershipStatus`), including sibling ResourceKind/CatalogKind binds.
- Spanner nullable authz params bind via `spannerNullString` / `spannerNullTime` (not Go pointers).
- `CreateAuthzAssignment` mints empty IDs via `ensureManagedID` (`PrefixAuthzAssignment` / `asgn_`).
- `GetAuthzCatalog` loads a persisted catalog + projected rows; deep Persist round-trip asserts live in `stmttest` (all dialects). Seeded `viewer→admin` closure depth is 2; Spanner relation-reference columns have DEFAULTs.
- `authz_membership_edges` has Wave 0 set FK `(project_id, set_id) → teams` + MVP `CHECK (user∈team)` with `ON DELETE CASCADE` on **all** dialects (Spanner matched to Postgres/SQLite). `stmttest` `DeleteTeamCascadesEdges` asserts hard-deleting a team removes edges on every engine.

## Validation

```sh
go generate ./internal/service/
go test ./internal/domain/ ./internal/service/ ./internal/storage/v2/dialect/authz/ ./internal/storage/v2/dialect/spanner/ -run ActiveUniqueKey
go test -tags sqlite_integration ./internal/storage/v2/stmttest/ -run 'Authz|PersistCatalog|DualWrite'
go test -tags postgres_integration ./internal/storage/v2/stmttest/ ./internal/storage/v2/dialect/postgres/ -run 'Authz|PersistCatalog|CatalogSeed|DualWrite'
```

Postgres + SQLite green (including `DeleteTeamCascadesEdges`). Spanner compile-checked + ActiveUniqueKey unit tests; full emulator coverage via CI `server:test-spanner`.

## Release notes / changeset

Changeset already present: `.changeset/authz-mvp-storage.md` (`@zitadel/server` minor). Review/maintainability follow-ups are Go storage/API-shape refactors — no additional changeset.

## Notes

- Stacked on Wave 0 (`cursor/permission-storage-wave0-6e44` / #666).
- Re-synced with rewritten Wave 0 tip after main (`7f696931`): kept Wave 1 shipped schema; folded Wave 0 RSI MATCH SIMPLE + delete notes and `system-permission-catalog` see-also link. Skipped Wave 0 `grantor_type` null CHECK (not in Wave 1 migrations).
- Out of scope: resolver (#423), Leopard, filling bundles, #420 auto-compile, Filter support for team-deactivate subqueries, splitting a second `dialect/authz*` package.

<div><a href="https://cursor.com/agents/bc-9a29c97f-a069-4f09-9bec-244a599e3354?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9a29c97f-a069-4f09-9bec-244a599e3354&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


internal/service/statement.go
```diff
@@ -279,3 +282,31 @@ type BrandingStatements interface {
...
+// AuthzMembershipEdgeStatements stores the authz projection of team membership.
+type AuthzMembershipEdgeStatements interface {
+	Statements
+	UpsertAuthzMembershipEdge(ctx context.Context, edge *domain.AuthzMembershipEdge) error
+	GetAuthzMembershipEdge(ctx context.Context, projectID, setType, setID, memberType, memberID string) (*domain.AuthzMembershipEdge, error)
+	ListAuthzMembershipEdgesByMember(ctx context.Context, projectID, memberType, memberID string) ([]*domain.AuthzMembershipEdge, error)
+	DeleteAuthzMembershipEdge(ctx context.Context, projectID, setType, setID, memberType, memberID string) error
+	DeleteAuthzMembershipEdgesByMember(ctx context.Context, projectID, memberType, memberID string) error
+	DeleteAuthzMembershipEdgesBySet(ctx context.Context, projectID, setType, setID string) error
+}
```
adlerhurst, line 384: these statements feels very resource oriented. I would rather have more use case oriented statements.

internal/service/statement.go
```diff
@@ -318,3 +323,61 @@ type ClaimStatements interface {
...
+//     DeleteAuthzMembershipEdgesForTeamDeactivate helper (team set + lifecycle-owned users).
+//   - Get/ListByMember: resolver / dual-write test reads.
+type AuthzMembershipEdgeStatements interface {
+	Statements
+	UpsertAuthzMembershipEdge(ctx context.Context, edge *domain.AuthzMembershipEdge) error
+	GetAuthzMembershipEdge(ctx context.Context, projectID, setType, setID, memberType, memberID string) (*domain.AuthzMembershipEdge, error)
+	ListAuthzMembershipEdgesByMember(ctx context.Context, projectID, memberType, memberID string) ([]*domain.AuthzMembershipEdge, error)
+	DeleteAuthzMembershipEdge(ctx context.Context, projectID, setType, setID, memberType, memberID string) error
+	DeleteAuthzMembershipEdgesByMember(ctx context.Context, projectID, memberType, memberID string) error
+	DeleteAuthzMembershipEdgesBySet(ctx context.Context, projectID, setType, setID string) error
```
adlerhurst: can't we combine the Delete* methods into a single one and use `database.Filter` to define the condition?

internal/authz/compiler/output.go
```diff
@@ -52,6 +56,22 @@ const (
...
+	case TermComputedUserset:
+		return "computed_userset", nil
+	case TermTupleToUserset:
+		return "tuple_to_userset", nil
+	case TermInvalid:
+		return "", fmt.Errorf("invalid expression edge kind")
+	default:
+		return "", fmt.Errorf("unknown expression edge kind %d", k)
+	}
+}
```
adlerhurst: this is IMO at the completely wrong place. There are already dialect independent sub-packages in the dialect package.

internal/storage/v2/authzmembershipedge/schema.go
```diff
@@ -0,0 +1,40 @@
+package authzmembershipedge
```
adlerhurst: why is this not part of the storage/v2/dialect/authz package like the others?

internal/storage/v2/dialect/postgres/authz_catalog_persist_test.go
```diff
@@ -0,0 +1,214 @@
+//go:build postgres_integration
+
+package postgres
```
adlerhurst: shouldn't all tests in this file have the same result in all dialects and can therefore be moved to the stmttest package?

livio-a (commented):
**Review: Wave 1 authz MVP storage** — reviewed as part of the stack (#666 → #677 → #758).

Overall this is a carefully built, high-consistency change. I verified the riskiest pieces by hand: the dual-write (`team_memberships` + `authz_membership_edges` + `resource_scope_index`) is wrapped in the same `withTransaction` on every create/update/delete/deactivate path across all three dialects, so those rows can't diverge; placeholder/column counts match the DDL exactly; everything is parameterized (no injection surface); table/index parity across Postgres/Spanner/SQLite is clean; and the Spanner `active_unique_key` stand-in for the Postgres partial-unique index is genuinely equivalent (same tuple order, `\x1f`-joined, nulled on revoke). No blocker- or high-severity defect.

### Worth fixing before merge

**[MEDIUM] `CreateAuthzAssignment` doesn't mint `id` via `ensureManagedID`.**
`internal/storage/v2/dialect/{postgres,spanner,sqlite}/authz_assignment.go` insert `a.ID` verbatim, whereas `CreateProject`/`CreateTeam`/`CreateUser` all open with `ensureManagedID(&x.ID, domain.PrefixX)`. There's no `PrefixAuthzAssignment` in `internal/domain`. This deviates from ADR 047 / `internal/storage/v2/AGENTS.md` ("assign empty IDs with dialect Ensure/ensureManagedID before INSERT; don't invent IDs in domain/service/handlers"). A future grants-API caller that follows the established pattern (leave ID empty, let the dialect mint it) will hit `CHECK (id <> '')` instead of a minted ID. Suggest adding `PrefixAuthzAssignment` and minting in the dialect statement.

**[MEDIUM] Catalog-persistence correctness is deep-tested only on Postgres.**
Only `postgres/authz_catalog_persist_test.go` reads back the projected `authz_relations` / `authz_relation_references` / `authz_expression_edges` / `authz_relation_closure` rows and verifies retire-previous. There's no `spanner/` or `sqlite/` equivalent (confirmed absent), and the shared `stmttest` persist test only asserts the call succeeds. A Spanner/SQLite bug that mis-writes closure depth or drops references would pass CI. Given how meticulously the DDL is mirrored across dialects, the correctness verification should be too.

### Lower severity

- **[LOW] Seed closure depth drift.** The hand-seeded `viewer→admin` closure row is written at depth **1**, but the model chain is viewer→editor→admin, so the compiler BFS emits depth **2**. No Wave 1 runtime impact (resolver joins from/to only, ignores depth), but it's a silent landmine for when #420 recompiles `cat_sys_1` through `PersistCatalogVersion`. Note the depth-invariant test in #758 can't catch this — its oracle re-implements the same BFS.
- **[LOW] Spanner `authz_relation_references` omits DEFAULTs.** `ref_relation`/`wildcard`/`condition` are `NOT NULL` with no `DEFAULT`, whereas Postgres/SQLite give `DEFAULT ''`/`DEFAULT false`. An INSERT omitting those columns succeeds on PG/SQLite but hits NOT-NULL on Spanner. Latent (mapper always supplies all columns) but a real DDL parity gap.
- **[LOW] `authz_relations.kind` hardcoded to `'relation'`** in `dialect/authz/catalog_rows.go` (the compiler carries no kind), so `'permission'` — allowed by the CHECK — can never be produced. Fine for MVP; worth a comment/follow-up.
- **[LOW] Team FKs use `ON DELETE CASCADE`** on `resource_scope_index`/`authz_assignments`, diverging from the deliberate `RESTRICT`/`NO ACTION` on `team_memberships` (ADR 024: deletion via lifecycle service, not raw cascade). Internally consistent across dialects — flagging so it's a conscious choice, not drift.
- **[LOW] SQLite project-delete cascade depends on `PRAGMA foreign_keys = ON`.** `DeleteProjectByID` now relies on FK cascade to clean RSI/edges/assignments; correct on PG/Spanner, silently orphans on SQLite without the pragma. Pre-existing global assumption, now extended to the authz tables — worth confirming the pool enforces it.
- **[NIT]** PG/SQLite membership-edge upsert uses a no-op PK self-assignment (`DO UPDATE SET project_id = EXCLUDED.project_id`) to preserve `created_at`; the Spanner variant documents why (`created_at = created_at`) but PG/SQLite don't.

### Test coverage (relocation verified, minor gaps)

The removed `team_membership_test.go` coverage was **relocated and expanded** into the shared `stmttest` suite (CRUD/FK/not-found now run on all three dialects; dual-write status variants moved to `authz_dual_write_test.go` and grew `Deactivate_clears...` cases) — no net loss. Remaining gaps, all LOW: membership-edge/dual-write tests only exercise `user`+`team` (no agent-member or app-set cases); and nothing asserts `authz_assignments`/edges are cascade-cleaned on project delete (only RSI is checked).

livio-a (commented):
**`authz_membership_edges` — doc/migration mismatch on the set FK.** #666's Wave 0 DDL now declares `FOREIGN KEY (project_id, set_id) REFERENCES teams (project_id, id)` on `authz_membership_edges`, but the migration here (`000015`/`000018`/`000002`) only has the `project_id → projects` FK. Since Wave 0 is the frozen schema, the migration should match it — and adding this FK is the DB backstop that otherwise leaves `set_id` free-floating (an edge could reference a non-existent team, caught only by dual-write discipline). Worth adding across all three dialects (Postgres/SQLite `REFERENCES teams(project_id, id)`, Spanner `... ON DELETE NO ACTION` per convention). If instead this is meant to stay app-enforced for MVP, drop the line from the Wave 0 doc so the two agree.

PR #685, opened 2026-08-01
feat: verify and restore scaffolded app files in doctor

fforootd:
## Summary

`zitadel doctor` reported `ok: true` with `proxy.ts` — the request boundary that makes `/__nextgen` proxying and route protection work — deleted, and `doctor --fix` did not restore it, despite SKILLS.md documenting both behaviors. Surfaced by the 2026-07-30 agent-driven CLI evaluation ("Wicklore" friction log, finding #1).

This PR settles the file-ownership contract and makes doctor tell the truth:

- **Scaffold manifest**: `setup` records what it actually wrote in a `scaffold` section of `.zitadel/state.json` — per-file sha256 + ownership class (`infrastructure` / `presentation`), plus `scaffolded_framework` and `dev_port` (closing the dead `dev_port` fallback in `patch-context`).
- **`managed-files` check**: missing infrastructure file (boundary, provider, `custom-elements.d.ts`) ⇒ **fail**; missing generated page ⇒ **warn**; edited (marker kept) and adopted (marker removed) files pass, classified in `details`. Pre-manifest apps degrade to template-derived, presence-only expectations with conditionally-scaffolded files (the framework home page) excluded.
- **`doctor --fix` is now restore-missing-only, globally**: repairs replay content ops only for absent targets (`missingOnly` + a new `overwrites` marker on the homepage edit op) and can never replace an existing scaffolded app file (additive repairs still append). This deliberately narrows `DependencyCheck.fix`, whose `force: true` repair could clobber a user-edited homepage once `scaffolded_framework` became restorable. The fix loop now also runs for warned checks so a deleted page is restorable.
- **ADR 042** records the contract, its states, and the known fallback-mode limitation; SKILLS.md's doctor entry is now accurate as written.

## Validation

- `moon run cli:test` — 818/818 across 103 files, including new suites: manifest/template modes, edited/adopted classification, restore-while-edited-stays-byte-identical, dry-run, homepage restoration gated on `scaffolded_framework`, `loadPatchContext` restoration, setup manifest assertions.
- `moon run cli-journey-e2e:test` — 14/14 script tests with the updated mock sequencing.
- `moon run workspace:journey -- --framework next` — 4/4 passed end to end against packed tarballs through Verdaccio, including the new `doctor detects and repairs managed-file drift` spec (delete boundary → doctor fails → `--fix` restores → passes) plus the password and passkey full loops.
- `moon run cli:lint cli:typecheck` and `node scripts/check-adrs-index.mjs` green.

## Release notes / changeset

`@zitadel/cli` **minor** — `.changeset/doctor-managed-files.md` (changes shipped product behavior per the decision table).

## Notes

- Per-file lifecycle states: `pristine` / `edited` / `adopted` / `managed` (template mode, no hash) / `missing`.
- The `--fix` semantic narrowing is the one behavior change to existing checks; it matches the flag's documented text and is recorded in ADR 042.
- Fallback mode judges pre-manifest apps by current templates — a Next major upgrade across the `middleware.ts`→`proxy.ts` boundary can misjudge them until the next setup rewrites the manifest (ADR 042 "Consequences").
- Guidance merge-blocks in README/AGENTS.md are out of scope (sections, not files).
- Groundwork intentionally in place for the follow-up envelope-hygiene PR (typed `files_written` rows, package.json key-order preservation).


docs/adrs/042-scaffolded-file-ownership-and-drift-detection.md
```diff
@@ -0,0 +1,97 @@
+# ADR 042: Scaffolded File Ownership and Drift Detection
+
+## Status
+
+Accepted
+
+## Context
+
```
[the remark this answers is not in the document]
fforootd: Reformatted to the blockquoted **Status/Date/Context** header used by ADR 039–041 (plus a Relates-to line for ADR 004); `check-adrs-index` stays green. The line-68 occurrence was the same `## Status` section, now gone.

apps/cli/SKILLS.md
```diff
@@ -102,9 +102,14 @@ the CLI's help layer, not the envelope.
...
-  Docker checks apply only when using `--runtime docker` or `--image`.
-  `--fix` re-applies missing managed files.
+  exists. The `managed-files` check compares the scaffolded app files against
+  the manifest setup recorded in `.zitadel/state.json`: a missing
+  infrastructure file (the request boundary, `custom-elements.d.ts`) fails,
+  a missing generated page warns, and files you edited (marker kept) or
+  replaced (marker removed) pass as `edited`/`adopted`. The default local
+  runtime is the `@zitadel/server` npm binary; Docker checks apply only when
+  using `--runtime docker` or `--image`. `--fix` restores missing managed
+  files and never overwrites an existing file.
```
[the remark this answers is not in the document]
fforootd: Agreed — tightened to "never replaces an existing scaffolded app file" with the additive appends (gitignore/env/dependency) called out explicitly, and aligned the same overstated phrase in the changeset and PR body.

apps/cli/src/lib/orca/patchers/rule/reclaim.ts
```diff
@@ -25,3 +28,48 @@ export function reclaimableOps(plan: ScaffoldPlan): FileOp[] {
...
+      const candidates = typeof op.path === "string" ? [op.path] : op.path;
+      const present = await Promise.all(
+        candidates.map((candidate) => exists(join(cwd, candidate))),
+      );
+      if (!present.includes(true)) {
+        kept.push(op);
+      }
+      continue;
+    }
+    kept.push(op);
```
fforootd, line 75: Could we make `add-dep` conditional on the dependency actually being absent in missing-only mode? It currently always survives this filter, and `addDependency` replaces any different existing version. That means `doctor --fix` for an unrelated missing page can silently rewrite a user-pinned `@zitadel/sdk-*` range and leave the lockfile out of sync, even though the dependency check was already passing. A preservation test for an existing SDK version during managed-file repair would help lock this down.

apps/cli/src/lib/orca/patchers/rule/base.ts
```diff
@@ -41,26 +41,60 @@ export abstract class AbstractRulePatcher implements Patcher {
...
   }
 
+  /**
+   * The subset of {@link routeFiles} that is load-bearing for the integration
+   * (request boundary, provider, type declarations). The `doctor`
+   * managed-files check fails when one is missing; everything else is a
+   * presentation starting point and only warns. Defaults to none.
+   */
+  protected infrastructureFiles(_view: PatchView): ReadonlyArray<string> {
+    return [];
```
fforootd, line 168: Could each patcher classify its load-bearing files instead of inheriting the empty default? Only Next overrides this today, so files such as Angular's `proxy.conf.cjs` and Nuxt's auth/components plugins are recorded as `presentation`. Deleting one therefore leaves `doctor` at a warning with top-level `ok: true`, even though the auth/proxy path is broken. Since the journey proof is Next-only, a couple of non-Next classification tests would catch this gap.

apps/cli/src/lib/scaffold-manifest.ts
```diff
@@ -0,0 +1,88 @@
...
+  try {
+    const state = await readState(cwd);
+    const scaffold = state.scaffold;
+    if (!scaffold || typeof scaffold !== "object") {
+      return undefined;
+    }
+    if (!scaffold.files || typeof scaffold.files !== "object") {
+      return undefined;
+    }
+    return scaffold;
```
fforootd: Could we validate the entries inside `files` before returning the manifest? The current check accepts values such as `{"proxy.ts": null}`, after which `evaluateManifest` dereferences `entry.class` and `doctor` exits through an exception instead of degrading to template mode as the comment promises. Checking that each entry has a string hash and an `infrastructure`/`presentation` class should keep corrupt or future-shaped state diagnosable.

apps/cli/src/commands/doctor/checks/managed-files.ts
```diff
@@ -0,0 +1,228 @@
...
+    const manifest = await readScaffoldManifest(ctx.cwd);
+    const missingBefore = manifest ? await missingManifestPaths(ctx.cwd, manifest) : [];
+    const patchCtx = await loadPatchContext(ctx.cwd, ctx.orca, ctx.cliVersion);
+    await ctx.orca.patcherFor(patchCtx.framework.id).repair(patchCtx, {
+      cwd: ctx.cwd,
+      dryRun: ctx.dryRun,
+      force: false,
+      missingOnly: true,
+    });
+    if (ctx.dryRun || !manifest || missingBefore.length === 0) {
```
fforootd: Could template mode materialize a manifest after a successful repair, or should ADR 042 be adjusted? When `manifest` is undefined, this return means `--fix` never writes one; meanwhile `setup` skips already-initialized projects. Pre-manifest apps therefore remain in template fallback permanently, including the documented template-growth / Next-major misclassification risk, although the ADR says the first `--fix` resolves it. A migration assertion here would make the intended behavior explicit.

fforootd (commented):
Thanks for tackling the managed-file ownership gap — I found four edge cases that seem worth tightening before this contract becomes public.

apps/cli/src/lib/orca/patchers/rule/reclaim.ts
fforootd: Done — `withoutExistingTargets` now drops `add-dep` whenever the dependency is declared at any version (a differing pin is an overwrite, not a restore); it survives only when absent, so `DependencyCheck.fix` still re-adds a genuinely missing SDK. New test: repairing a missing boundary leaves `"@zitadel/sdk-next": "user-pinned-0.0.1"` untouched. ADR 042 + SKILLS.md updated to say so.

apps/cli/src/lib/orca/patchers/rule/base.ts
fforootd: Classified — Angular marks `proxy.conf.cjs` and Nuxt marks both plugins (`auth.server.ts`, `zitadel-components.client.ts`) as infrastructure, so deleting one now fails doctor. The Vite SPAs deliberately stay presentation-only: their single marked file is the root component (the customization surface), and their plumbing lives in `vite.config.*` config-edit merges the file check can't verify — noted in the base hook's docstring as the known boundary. A new per-patcher classification suite pins all eight maps so a future patcher can't silently inherit the empty default.

apps/cli/src/lib/scaffold-manifest.ts
fforootd: Validated — `readScaffoldManifest` now checks every entry for a string `hash` and a known `class` and returns `undefined` otherwise, so `{"proxy.ts": null}` (or a future-shaped manifest) degrades to template mode as documented instead of throwing through doctor. Test covers the malformed-entry case.

apps/cli/src/commands/doctor/checks/managed-files.ts
fforootd: Implemented the materialization (rather than adjusting the ADR): a successful template-mode `--fix` now records the marker-bearing files on disk as the manifest — adopted (marker-less) and conditionally-scaffolded files stay out — so the next run verifies in exact manifest mode and the fallback's template-growth risk ends at the first repair, as ADR 042 promised. The migration test asserts template → fix → manifest mode with the adopted page excluded. Best-effort: an unreadable state.json leaves the app in template mode as before.

fforootd:
Addressed the codex review (all four findings) in 11eb7b15, on top of a plain merge from main:

1. **Config-wiring drift (P1)** — implemented the structural checks rather than narrowing the guarantee. `Patcher.verify` probes each labelled config-edit wiring through the merge transform's own idempotency: the transforms only add what is missing, so a changed output against current content means the wiring is detached. Vite/Nuxt proxy merges, `angular.json` `proxyConfig`, and the Angular auth routes are `infrastructure` (detached ⇒ fail); the Angular `dev` script is `convenience` (⇒ warn); a transform that throws on restructured content yields no verdict. `--fix` re-applies wirings by replaying the same edit ops. The exact repro is now a test: an Angular project with only the `angular.json` wiring removed fails doctor with `detached managed config wiring: angular.json`, and `--fix` restores it.
2. **Next 15→16 migration (P1)** — manifest-mode `--fix` now reconciles against the current template set before rewriting: entries whose file is gone and which the current plan no longer writes are dropped; template paths the repair introduced are adopted with class + on-disk hash. Regression test covers `middleware.ts` manifest → upgrade → `--fix` creates `proxy.ts`, drops the retired entry, doctor passes.
3. **Array manifest (P2)** — `readScaffoldManifest` validates via `isObject` (rejects arrays) at every level; `scaffold.files: []` degrades to template mode with a test.
4. **Changeset wording (P2)** — the migration path is stated as `doctor --fix` (setup skips initialized projects); ADR 042 fixed the same claim and documents the wiring verification and reconcile semantics.

Validation: 904/904 CLI tests (new: wiring detach/repair, convenience-warn, 15→16 reconcile, array degrade), lint/typecheck/ADR-index green.

fforootd:
Codex round 2 addressed in 3f24d3c7 — both P1s were real and are now closed:

1. **Missing config → false green** — `verify()` verdicts are now total. A missing host config file reports `detached` (the wiring cannot exist in a file that doesn't — infrastructure ⇒ fail), and a transform that throws on restructured content reports `unknown` (⇒ warn, e.g. a multi-project `angular.json` without `defaultProject`, where failing outright would false-alarm legitimate manual setups). Nothing is silently dropped anymore. `ConfigWiringStatus.applied` became `state: applied | detached | unknown`. Codex's exact repro is a test: `rm angular.json` ⇒ `fail: detached managed config wiring: angular.json`; the restructured case is a second test asserting the `unverifiable` warning.
2. **15→16 with retained `middleware.ts`** — new `retiredAlternates` mapping (Next declares the boundary pair mutually exclusive) drives the migration in `--fix`: a **hash-proven pristine** leftover is deleted and `proxy.ts` installed, with the manifest swapping entries — the one deletion `--fix` performs, gated on the manifest proving the bytes are the CLI's own. Anything less than pristine (edited, adopted, or unhashable in template mode) is a **conflict**: nothing is deleted, `proxy.ts` is excluded from the repair via the new `excludePaths` option (unrelated missing files still restore), and doctor fails with manual-migration guidance. Both repro shapes are tests: pristine-retained ⇒ swap converges to pass; edited-retained + unrelated missing page ⇒ page restored, no `proxy.ts`, `middleware.ts` byte-identical, explicit conflict failure.

ADR 042, SKILLS.md, and the changeset document the total-verdict semantics and the pristine-only deletion gate. 926/926 CLI tests, lint/typecheck green.

fforootd:
Codex round 3 addressed in 8dba6c17 (on top of the latest main your Update-branch pulled in):

1. **No manifest during an unresolved conflict (P1)** — materialization is now gated on every current *infrastructure* file being present and marker-recordable. During a boundary conflict (or over an adopted boundary) no manifest is written; the template fallback — whose presence checks keep demanding the boundary — stays in charge, so deleting `middleware.ts` afterwards still fails instead of passing against a boundary-less manifest. The full lifecycle is a test: pre-manifest 15→16 with edited `middleware.ts` → `--fix` creates no `proxy.ts` and writes no manifest, doctor keeps failing → user migrates their edits (rename with marker) → next `--fix` materializes with `proxy.ts` recorded → pass. The adopted-boundary variant is covered too (stays on fallback, still passes).
2. **One-way alternate mapping (P2)** — `retiredAlternates` is declared only in the ≥16 direction; on Next 15 `proxy.ts` was not a reserved convention, so a user-owned root `proxy.ts` no longer trips the conflict logic (test: Next 15 fixture + user `proxy.ts` ⇒ pass, no conflict message).

On the "scope out Next 15" question (also discussed with Florian directly): recommendation is no — the one-way mapping delivers the intended scoping for the conflict machinery, the migration code exists precisely *for* 15→16 upgraders, and a hard floor without an explicit setup/detect error would leave a scaffolded `proxy.ts` silently inert on 15 (unprotected routes). If the support matrix should shrink, that's a deliberate versioned error in a follow-up, not a side effect here.

929/929 CLI tests, lint/typecheck green. ADR 042 documents the materialization gate and the one-directional mapping.

PR #686, opened 2026-08-01
feat: embeddable session surface, neutral email copy, shipped JSX types

fforootd:
## Summary

- **Shared surface base** — new `packages/components/src/orchestrator/surface.ts` (`ZitadelSurface`) owns the `variant`/`theme` properties, the `ThemeController`, and the per-update surface application (base tokens → branding overrides → font links → `data-theme`/`data-theme-dark` host stamps). `<zitadel-login>` is refactored onto it with zero behavior change (its existing `theme.browser.spec.ts` / `embedding.browser.spec.ts` suites are the guard, all green); its `updated()`-time `data-widget` re-stamp loop stays because its `zl-page-shell` lives in unsafeHTML-parsed markup.
- **`<zitadel-session>` becomes embeddable** — extends the surface base: default `variant="widget"` is content-sized and transparent through every layer; `variant="page"` claims the viewport with the page paint on its internal `zl-page-shell` (the host no longer paints `100vh` black). The hardcoded `dataset.theme = "dark"` connect-time stamp is gone; `theme` resolves explicit → branding (none yet) → variant fallback (`dark` for page, `auto` for widget). Widget polarity is a plain template binding (`?data-widget`) — no querySelector loop needed for a static template.
- **`<zitadel-logout>` gets real theming** — adopts `ThemeController` with a new `theme` property (fallback `auto`), drops the hardcoded dark stamp, and applies `applyBrandingTokens` beside `applyBaseTokens` so the token pipeline matches the other surfaces. It stays content-sized (`inline-block`), no `variant`.
- **Neutral email copy + business overlay** — builtin locales now ship "Email" / `you@example.com` (de placeholders `du@beispiel.de`, it "E-mail" / `tu@esempio.com`). The previous business strings move to a new `businessLocales` overlay (`locales/business.ts`, exported from the package root and orchestrator barrels; re-exported from `@zitadel/sdk-next/client`). Also fixes the latent gap where `it` was missing from the top-level barrel.
- **Shipped JSX declarations** — hand-authored `src/jsx.d.ts` with React JSX declarations for all three elements' full public prop surface (types indexed off the element classes), copied verbatim into `dist/` (tsdown `copy`; dts-bundling ambient declare-module files is fragile) and exposed as types-only `exports["./jsx"]`. `@zitadel/sdk-next` gains a `./jsx` shim (`/// <reference types="@zitadel/components/jsx" />`) copied into its dist. A new `jsx-declarations.spec.ts` drift guard asserts every non-state `elementProperties` entry (attribute name where declared, property name otherwise) appears in the declaration text.
- **Demo fixed-surface fix (integrated from #687)** — both demo login pages and both demo admin pages center widgets on a hardcoded light backdrop while the dev mock's tenant branding pins dark, which painted near-invisible dark-mode text/buttons onto the light pages. The demos now declare their fixed surface on the widgets (`theme="light"` on `<zitadel-login>` and `<zitadel-logout>` in demo-next and demo-nuxt), which is the documented embedding contract — element-level `theme` outranks tenant branding precisely because the embedding app knows its own surface. The demos now actively demonstrate that precedence against a mock tenant pinning the opposite mode. (The logout pins are possible only in this PR, which introduces logout's `theme` property — hence the integration instead of a standalone follow-up.)
- **Conformance + browser specs** — `orchestrator-surface.spec.ts` (jsdom) pins the shared `variant`/`theme` contract, defaults, and explicit-theme `data-theme` stamping across login/session/logout; `session-surface.browser.spec.ts` (opt-in Chromium suite) mirrors the login embedding patterns: widget = content-sized + transparent through every layer + no font injection, page = viewport-claiming + painted shell + brand font, `theme="light"` reaches real pixels.
- **Ripples** — `liquid.spec.ts` builtin-locale expectation ("Email is required."), console-e2e login label (`getByLabel("Email")`), `apps/demo-next/src/custom-elements.d.ts` replaced by the `@zitadel/sdk-next/jsx` reference (it was a stale hand-copy missing `zitadel-session`, `variant`, `theme` and still listing the retired `session-exchange-path`), and the session Storybook fullscreen story pins `variant="page"`.

- **Review round (2026-08-02, c43e8af3)** — the five review findings are addressed atomically: sdk-core's `ZitadelSessionConfig` gains `variant`/`theme` and `ZitadelLogoutConfig` gains `theme`, forwarded by all six framework wrappers (React was type-only) with a forwarding spec each; the JSX declarations intersect `React.RefAttributes<Element>` so `ref`/`key` type against the concrete element; `locales` accepts (readonly) partial overlays with an undefined-safe merge, making `businessLocales` directly assignable end to end; the document-level font links are collectively owned per surface (most-recent registration wins, disconnected owners pruned lazily) so one surface can no longer strip another's fonts; and the CLI scaffolds pin `variant="page"` on the generated `/profile` pages (Next + Nuxt) while the generated `custom-elements.d.ts` references the shipped `@zitadel/sdk-next/jsx` declarations instead of a hand-copy. A compile-time consumer fixture in demo-next exercises `ref`/`className`/`key`, the closed prop unions, and `businessLocales` (placed there because sdk-next's own `tsc --noEmit` proved vacuous for spec files — its solution-style config checks nothing; worth a separate look). Branch merged with `main` (absorbing the `@zitadel/testing` fixed-group addition and the new add-user e2e, whose sign-in label follows the neutral copy).

## Validation

- `moon run components:test` — 391 unit tests / 32 files green (conformance + drift specs, updated liquid spec, font-link ownership + two-surface specs).
- `moon run components:test-browser` — 67 tests / 11 files green in real Chromium (existing login theme/embedding guards + the new session surface suite).
- `moon ci :lint :typecheck :build :test` — 107 affected tasks green after the review round + main merge (components, all SDK wrappers incl. their new forwarding specs, cli with the scaffold assertions, demo typechecks incl. the JSX fixture — whose `@ts-expect-error` lines double as proof the augmentation is actually checked).
- `corepack pnpm exec changeset status --since origin/main` — components + sdk-next minor, fixed group rides along.
- JSX chain proven semantically: `demo-next:typecheck` fails with `TS2339: Property 'zitadel-login' does not exist` when the augmentation is absent and passes with it; a probe confirmed closed unions are enforced (`variant="fullscreen"` rejected). The demo login widget's `theme="light"` now typechecks through the shipped declaration chain. Note: TS's JSX rule always permits hyphenated attribute names on intrinsic elements (same mechanism as `data-*`), so typo'd *kebab-case* attributes can't be rejected — identifier-ish props are fully checked.
- Visual sanity (built dist, api-mock + demo-next dev): login page shows the neutral "Email" label/placeholder and — under an emulated **dark** OS preference plus the mock's dark tenant branding — resolves `data-theme="light"` with fully legible dark title/label text and a dark primary button. A full sign-in through the mock (identifier → password → exchange → `/admin`) succeeded on the refactored orchestrator; the admin logout avatar renders as a dark circle on the light page and its dropdown opens with the light surface (`#ffffff`, dark text). Storybook session story renders the dark page card, light page card, and the content-sized transparent widget correctly.
- Skipped: `demo-next-e2e:e2e` (opt-in, not in CI) — its web server hardcodes port 8080, which is held by a long-running local `zitadel` dev server on this machine. The layers it covers (proxy, middleware, `maybeCompleteFlow`/`post-sign-in-url`) are untouched by this PR, and the manual mock sign-in above exercised the same handover path.

## Release notes / changeset

Real changesets (decision table row 1 — shipped component behavior changed): `session-widget-variant.md`, `neutral-email-copy.md`, `components-jsx-types.md` (all `@zitadel/components` minor), `sdk-next-jsx-shim.md` (`@zitadel/sdk-next` minor), and `sdk-surface-props.md` (sdk-core + the six framework SDKs + cli, minor) for the wrapper surface props and scaffold pin. The demo-app changes need no changeset (both demos are in the changesets `ignore` list).

## Notes

- **Deliberate alpha break**: `<zitadel-session>` no longer renders page-like by default — dedicated routes must opt in with `variant="page"`. The CLI scaffold templates pin `variant="page"` in a follow-up PR (`apps/cli/src/lib/orca/**` deliberately untouched here).
- **Supersedes #687**: the demo fixed-surface fix was merged into this branch (single trivial conflict on `custom-elements.d.ts`, resolved to this branch's shipped-declarations reference) because this PR both types `theme` in TSX and introduces the logout `theme` property the admin pages need.
- The session/logout "dark-only" token references turned out to already be theme-flipping semantic aliases (`--zl-color-text-primary-white` & friends flip under `[data-theme="light"]`); the only true dark pins were the hardcoded `data-theme` stamps, now removed. One dark-only leftover remains: the logout dropdown's `box-shadow: 0 12px 32px rgba(0,0,0,.32)` — there is no shadow token in `@zitadel/design-tokens` yet, so it stays hardcoded (readable in both modes) rather than inventing a token value here.
- `@types/react` is added to components' devDependencies (types-only, no runtime react dependency): the shipped `jsx.d.ts` augments `declare module "react"`, and module augmentation binds to the react instance resolved from the declaring file — without it, workspace consumers resolved no react from `packages/components` and the augmentation silently applied to nothing. `src/jsx.d.ts` is excluded from the components tsconfig programs (no react types needed there); the drift spec covers names, demo-next typecheck covers semantics.
- `businessLocales` entries are `Partial<Locale>` and deliberately exempt from the en/de/it key-parity spec (verified it only covers the full dicts). The de overlay uses "Geschäftliche E-Mail" as specced (today's de label was already neutral; only its placeholder was business-flavored).


fforootd (commented):
Thanks for pulling these related surface improvements together — the component and browser coverage is thoughtful. I found a few integration and typing edges that look worth tightening before this ships; details are inline.

packages/components/src/orchestrator/surface.ts
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * Colour mode: `light`, `dark`, or `auto` (follow `prefers-color-scheme`).
+   * Empty means "not stated", and resolution falls through to the tenant's
+   * `branding.theme.mode`, then to a variant-derived default — `dark` for
+   * `page` (the hosted design system surface) and `auto` for `widget`, so an
+   * embedded widget matches the visitor's preference instead of forcing a
+   * dark card onto a light page. Set it explicitly when your app's surface
+   * is fixed: `<zitadel-login theme="light">`.
+   */
+  @property({ type: String }) accessor theme: "" | ThemeMode = "";
```
fforootd, line 41: Would you mind wiring these new surface props through the shared SDK contract and framework adapters in this PR as well? `ZitadelSessionConfig` / `ZitadelLogoutConfig` still omit `variant` / `theme`, and the Angular, Vue, Svelte, Qwik, and Solid wrappers explicitly forward only the old fields. I also reproduced the React side as a compile error (`Property 'variant'/'theme' does not exist`). Since the fixed Changesets group releases those SDKs alongside components, the new surface contract would otherwise be unavailable through the published wrappers. A small wrapper-level type test would help keep this aligned.

packages/components/src/orchestrator/surface.ts
```diff
@@ -0,0 +1,77 @@
...
+export class ZitadelSurface extends LitElement {
+  /**
+   * Sizing/chrome mode. Components size to content; pages compose them —
+   * so the default is `widget`: content-sized, transparent host, no
+   * document-level default-font injection. Dedicated routes — the hosted
+   * shell and the scaffolded pages — opt into `page`, which claims the
+   * viewport, paints the surface background, and ships the brand font.
+   * Fine-grained height override in both modes: `--zl-page-min-height`.
+   */
+  @property({ type: String, reflect: true }) accessor variant: "widget" | "page" = "widget";
```
fforootd, line 30: Could we update the CLI's generated Next and Nuxt `/profile` templates in this PR rather than leave that to a follow-up? They currently render `<zitadel-session>` without `variant="page"`, so this new default makes freshly generated profile pages content-sized and removes the page-shell centering/default font as soon as this package is released. The green journey does not catch that because it asserts navigation and logout rather than surface geometry. Adding the attribute plus a scaffold assertion here would keep the default change atomic.

packages/components/src/jsx.d.ts
```diff
@@ -0,0 +1,59 @@
...
+        "flow-name"?: string;
+        "project-id"?: string;
+        "proxy-path"?: string;
+        url?: string;
+        "post-sign-in-url"?: string;
+        "resume-flow-id"?: string;
+        lang?: string;
+        locales?: ZitadelLogin["locales"];
+      };
+      "zitadel-session": React.HTMLAttributes<HTMLElement> & {
```
fforootd: Could these intrinsic element types include React's ref attributes as well? `React.HTMLAttributes<HTMLElement>` does not carry `ref`, so a normal `<zitadel-session ref={sessionRef} />` currently fails typechecking (`Property 'ref' does not exist`). Using `DetailedHTMLProps` or intersecting each entry with `React.RefAttributes<ZitadelSession>` would preserve the concrete element type. It may also be worth adding a small consumer type fixture because the textual property-drift test cannot exercise React's standard props.

packages/components/src/orchestrator/locales/business.ts
```diff
@@ -0,0 +1,51 @@
...
+ * loginElement.locales = { en: { ...en, ...businessLocales.en } };
+ * ```
+ *
+ * Each entry is a partial dictionary carrying only the affected keys —
+ * deliberately exempt from the en/de/it key-parity check that guards the
+ * full built-in dictionaries.
+ */
+import type { Locale } from "./en.js";
+
+export const businessLocales: Readonly<Record<string, Partial<Locale>>> = {
```
fforootd, line 20: Would it make sense to align the `locales` property type with this partial overlay? The runtime already merges each custom dictionary over the builtin one, but `ZitadelLogin["locales"]` is still `Record<string, Locale>`. As a result, `<zitadel-login locales={businessLocales} />` fails typechecking because these entries are `Partial<Locale>`. Allowing `Record<string, Partial<Locale>>` (and updating the shared SDK type) would make the exported preset directly usable and match the documented runtime behavior.

packages/components/src/orchestrator/surface.ts
```diff
@@ -0,0 +1,77 @@
...
+      applyBrandingTokens(root, branding, this.themeController.theme);
+      // Ship the design-system brand face by default on dedicated pages;
+      // drop it when a tenant font takes over so we don't fire a redundant
+      // request. Widget mode never injects the default into the host
+      // document — the embedding app owns its typography (and its visitors'
+      // font-CDN connections). Tenant `font_url` is explicit server-side
+      // branding state, so it applies in both modes.
+      const tenantFontUrl = branding?.font_url ?? null;
+      applyDefaultFont(root, this.variant === "page" && !tenantFontUrl ? undefined : null);
+      applyFontUrl(root, tenantFontUrl);
```
fforootd, line 72: Could we give the document-level font links ownership or reference counting before every surface calls these helpers? `font-loader.ts` stores one global link per id and removes it whenever `href` is null. I reproduced a page login loading `zl-default-font-link`, then a widget session mounting and removing that link here; the session also clears any tenant font because it passes `null`. This is new for session and can make one embedded surface silently change another. A two-surface test plus per-document ownership would cover it.

packages/components/src/orchestrator/surface.ts
fforootd: Done in c43e8af3 — `ZitadelSessionConfig` gains `variant`/`theme` and `ZitadelLogoutConfig` gains `theme` in sdk-core, and all six framework wrappers forward them (React was type-only via the shared props), each with a wrapper-level forwarding spec.

packages/components/src/orchestrator/surface.ts
fforootd: Done in c43e8af3 — both generated /profile templates pin `variant="page"` (Next + Nuxt), the scaffolded custom-elements.d.ts now references the shipped `@zitadel/sdk-next/jsx` declarations instead of a hand-copy, and the setup-next integration test asserts both.

packages/components/src/jsx.d.ts
fforootd: Done in c43e8af3 — every entry now intersects `React.RefAttributes<Element>` (`ref`/`key`, concrete element type), and a compile-time consumer fixture in demo-next covers ref/className/key, the closed unions, and `businessLocales` (it lives there because sdk-next's own `tsc --noEmit` turned out to check nothing for spec files — flagged in the PR notes).

packages/components/src/orchestrator/locales/business.ts
fforootd: Done in c43e8af3 — the element property is now `Readonly<Record<string, Partial<Locale>>>` with an undefined-safe merge, sdk-core's `locales` widened to match, and the demo fixture asserts `locales={businessLocales}` compiles.

packages/components/src/orchestrator/surface.ts
fforootd: Done in c43e8af3 — the document-level links are now collectively owned: each surface registers its desired href, `null` withdraws only that surface's vote (most recent registration wins, disconnected owners pruned lazily, withdrawal never resurrects an externally removed link). Covered by ownership unit specs plus a two-surface test.

fforootd:
Heads-up for the next update-with-main: main refactored the console-e2e `signIn` helpers into `apps/console-e2e/src-real/support.ts` (#700-era), which conflicts with this branch's neutral-copy ripple fix in the two spec-local copies. The resolution is to take main's structure and move the fix into `support.ts` (`getByLabel("Work email")` → `getByLabel("Email")`) — otherwise console-e2e-real types into a label that no longer exists once the neutral copy is live. Already resolved that way on the stacked B2 branch `feat/use-case-scaffold-copy` (both merged and journey-verified), so it can be cherry-picked or just redone identically here.

PR #688, opened 2026-08-01
fix: deduplicate setup file reporting and preserve package.json key order

fforootd:
## Summary

Two reporting/formatting defects from the 2026-07-30 CLI evaluation (Wicklore friction log, findings #7 and #8):

- **`setup` reordered an existing `package.json`.** The `add-dep` writer re-serialized the whole document through `stableStringify`, alphabetizing the user's top-level keys (`name`/`version`/`scripts` moved below `dependencies`). New `updateJsonPreservingOrder` in `lib/json.ts` splices the mutation in while preserving the user's key order, indentation style (including tabs and compact documents), and trailing newline; only the touched dependency map is name-sorted, matching what package managers write. The Angular `dev`-script merge uses the same path. `stableStringify` stays for CLI-owned files, per its docstring's determinism rationale.
- **`files_written` contained duplicates and directories.** The scaffold accumulator now records typed, path-deduplicated rows (`{path, kind: file|dir, action: create|update}`). The envelope's `files_written` becomes deduplicated file paths only; a new `data.files` carries the typed rows so agents can tell what setup created versus merged into (your `package.json` is an `update` row). Pre-existing directories now report as skipped instead of written, and `describeWrittenFile`'s hardcoded directory skip-list is gone — the type system replaced it.

## Validation

- `moon run cli:test` — 812/812 across 103 files, including new suites: accumulator dedupe (both env-file double-merges), net-action semantics, create-vs-update classification, dir rows, add-dep order/indent/newline preservation, and five `updateJsonPreservingOrder` cases (tab indent, compact stays compact, no invented trailing newline).
- Integration: setup on the pre-existing Next fixture asserts no duplicate `files_written` entries, no directory entries, typed rows for `proxy.ts` (create) / `package.json` (update) / `.zitadel` (dir), and that the user's `name`/`private` key order survives.
- `moon run cli:lint cli:typecheck` green.

## Release notes / changeset

`@zitadel/cli` **patch** — `.changeset/setup-file-reporting.md`.

## Notes

- `data.files` covers the scaffolded artifacts; the `.zitadel/` sync resources continue to report through `files_written`. SKILLS.md documents the split.
- The idempotent-rerun report changes shape deliberately: directories move from `files_written` to `files_skipped` on reruns, making the "Patched N files" count honest.
- Sibling PR #685 (doctor managed-files) touches neighboring `setup`/SKILLS.md lines; whichever lands second takes a trivial merge.


apps/cli/src/lib/json.ts
```diff
@@ -13,6 +13,39 @@ export function stableStringify(value: unknown): string {
...
+ */
+export function updateJsonPreservingOrder(
+  source: string,
+  path: string,
+  mutate: (value: Record<string, unknown>) => void,
+): string {
+  const value = parseJsonObject(source, path);
+  mutate(value);
+  const body = JSON.stringify(value, null, detectIndent(source));
+  return source.endsWith("\n") ? `${body}\n` : body;
```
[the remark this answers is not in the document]
fforootd: Fixed — the helper now re-joins the stringified body with the source's own EOL (`\r\n` when present) and appends the trailing newline in that style, so CRLF documents stay CRLF end to end. Covered by a new CRLF unit case asserting no stray bare-LF remains.

apps/cli/src/lib/orca/patchers/rule/file-writer/index.ts
```diff
@@ -144,15 +180,29 @@ async function ensureDir(
   dryRun: boolean,
   result: ScaffoldAccumulator,
 ): Promise<void> {
+  // A directory that already exists is a skip, not a write — but the mode is
+  // still (re-)applied below so a rerun heals loosened permissions.
+  const existed = await stat(path).then(
+    (info) => info.isDirectory(),
+    () => false,
```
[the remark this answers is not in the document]
fforootd: Agreed — a healed mode is a real change. `ensureDir` now compares the existing mode against the requested one: drifted permissions report as a `{kind: dir, action: update}` row (dry-run previews the same), and only a genuinely untouched directory lands in `files_skipped`. Unit case covers 0755→0700 healing and the settled rerun.

apps/cli/src/lib/json.ts
```diff
@@ -13,6 +13,42 @@ export function stableStringify(value: unknown): string {
...
+  source: string,
+  path: string,
+  mutate: (value: Record<string, unknown>) => void,
+): string {
+  const value = parseJsonObject(source, path);
+  mutate(value);
+  // JSON.stringify emits LF; re-join with the source's own line endings so a
+  // CRLF document stays CRLF instead of getting a whole-file EOL rewrite.
+  const eol = source.includes("\r\n") ? "\r\n" : "\n";
+  const body = JSON.stringify(value, null, detectIndent(source)).split("\n").join(eol);
```
fforootd: Could we make this a targeted edit rather than reserializing the entire document? `JSON.stringify` keeps the detected indent width and key order, but it still removes blank lines and expands unrelated inline objects. For example, a one-line `scripts` or `engines` object is rewritten even though only `dependencies` changed, so this can still produce the whole-file diff the PR is trying to avoid. A byte-level test with mixed formatting (for example, a blank line plus an inline nested object) would make the preservation contract explicit.

apps/cli/src/lib/orca/patchers/rule/file-writer/index.ts
```diff
@@ -144,15 +180,33 @@ async function ensureDir(
   dryRun: boolean,
   result: ScaffoldAccumulator,
 ): Promise<void> {
+  // An already-existing directory is a skip — unless its permissions drifted
+  // from the requested mode, in which case the chmod below repairs them and
+  // the report says so (an "update" row, not a silent skip).
+  const pre = await stat(path).catch(() => undefined);
+  const existed = pre?.isDirectory() ?? false;
+  const healsMode = existed && mode !== undefined && (pre!.mode & 0o777) !== mode;
```
fforootd: Could we make this mode-repair check platform-aware? Node documents that on Windows only the write permission is changeable and owner/group/other distinctions are not implemented, so comparing `stat.mode & 0o777` to exactly `0o700` is not reliable there: https://nodejs.org/docs/latest-v24.x/api/fs.html#file-modes. Since the CLI ships on Windows, `.zitadel` could keep being reported as an `update` on reruns even after `chmod`. Gating exact POSIX mode healing/reporting off on `win32` (with a platform-shaped test) seems safer. The current `full-pr` runner is Ubuntu-only, so it would not expose this.

fforootd (commented):
Two focused follow-ups on preserving user-owned files and keeping the new reporting contract cross-platform.

apps/cli/src/lib/json.ts
fforootd: Reworked as asked — `setTopLevelJsonKey` now scans the document with a string-aware tokenizer, locates the one top-level value, and splices it in place: blank lines, inline nested objects, and every byte outside the touched key survive verbatim. A parse-back verification guard throws rather than writing a bad splice. The new byte-level tests assert exact expected documents for the mixed-formatting case (blank line + inline `engines`/`scripts`), CRLF, tabs, compact, key-insertion, and escaped key-looking strings inside values.

apps/cli/src/lib/orca/patchers/rule/file-writer/index.ts
fforootd: Gated — `healsMode` now requires `process.platform !== "win32"`, so on Windows an existing directory always reports as skipped (the chmod attempt stays, harmlessly). Platform-shaped test stubs `process.platform` to win32 and asserts the rerun skips instead of flagging an update.

Issue #699, opened 2026-08-02
API: implement listSessions — GET /sessions returns 501


bastionstack:
## Context

`GET /sessions` is routed, documented and generated, but the service behind it is a stub:

```go
// internal/service/session.go
func (s *sessionService) List(ctx context.Context, input ListSessionInput) ([]*domain.Session, error) {
	//TODO implement me
	return nil, domain.ErrNotImplemented()
}
```

## Scope

- Implement `sessionService.List` against the v2 statements, scoped by `project_id` the way `ListUsers` is.
- Handle the parameters the handler currently drops — `internal/api/session.go` carries a `// TODO: handle params`.
- Ship with cursor pagination rather than offset/limit: [ADR 027](https://github.com/zitadel/nextgen/blob/main/docs/adrs/027-cursor-based-pagination.md) cites this endpoint as its example.

## Why now

The console's Sessions screen is built against it and is parked until this lands, along with its nav entry. `RevokeSession` is already implemented, so listing is the only missing piece.

## Reference

`internal/service/session.go` — `Delete` alongside it shows the statement pattern.

Issue #667, opened 2026-07-30
Prepare Cloud customers for billing and paid tiers


elinashoko:
## Summary

When a project is successfully claimed, the associated customer account should be prepared for Zitadel Cloud billing.

The customer should start on the Free tier and have a Stripe customer linked to their account, so they can upgrade to a paid Cloud tier later without additional account setup.

This ticket only prepares the customer for future billing. It does not introduce subscriptions, payments or entitlement enforcement.

## User story

As a Zitadel Cloud customer, I want my account to be ready for Cloud tiers so that I can use the Free tier immediately and upgrade to a paid tier later.

## How it works

After a project is successfully claimed:

1. The customer account is assigned the Free Cloud tier.
2. A Stripe customer is created for the customer account.
3. The Stripe customer ID is stored against the customer account.
4. The customer can start using the claimed project without waiting for Stripe setup to complete.

The Stripe customer represents the customer account, not an individual owner or project. This allows one customer account to own multiple projects without creating separate billing customers.

### Cloud tier

- Every newly claimed Cloud customer starts on the Free tier.
- The current Cloud tier is recorded against the customer account.
- The customer account remains the source of truth for the customer’s tier.
- Assigning the Free tier does not create a Stripe subscription.

### Stripe customer

- A Stripe customer is created for every newly claimed Cloud customer.
- The Stripe customer is associated with the customer account.
- The Zitadel customer account ID is included in Stripe as a stable reference.
- The Stripe customer ID is stored against the customer account.
- Only the minimum customer information required for future billing is sent to Stripe.

### Reliability

- Stripe setup does not block or roll back a successful project claim.
- Failed Stripe customer creation can be retried.
- Retrying does not create duplicate Stripe customers.
- Stripe setup failures are visible to the team operating Zitadel Cloud.

## Acceptance criteria

- [ ] A newly claimed Cloud customer is assigned the Free tier
- [ ] The assigned tier is stored against the customer account
- [ ] A Stripe customer is created for the customer account
- [ ] The Zitadel customer account ID is included in the Stripe customer metadata
- [ ] The Stripe customer ID is stored against the customer account
- [ ] A Stripe customer is not created separately for each project or owner
- [ ] Stripe customer creation does not block access to the claimed project
- [ ] Failed requests can be retried without creating duplicate Stripe customers
- [ ] No Stripe subscription, payment method or invoice is created
- [ ] The functionality applies only to Zitadel Cloud

## Out of scope

- Paid subscriptions
[...]

Issue #620, opened 2026-07-23
Team: List/Query


grvijayan:
## Context

Part of #598. Team CRUD is missing List: only Create and Get exist today. Teams carry a lifecycle status (`active`/`deactivated`/`pending_purge`), which affects what a list of teams should return by default.

## Scope / Tasks

- Add storage support for querying teams scoped to a project, using the same cursor-pagination approach already used by Project's list, on both postgres and spanner.
- Add a service-level List operation with the same request/response conventions used by Project: limit, page token, sorting, filters.
- Add a `POST /teams/query` endpoint and its request/response schemas, with `project_id` as a required query parameter (matching Create and Get).
- Gate the endpoint the same way `GetTeam` is gated (project-secret scope, read access).
- Support filtering and sorting by `createdAt`, matching what's available today.
- Return `status` on each listed team, matching the field exposed by Get and Create (#621).

## Acceptance Criteria

- [ ] `POST /teams/query` returns teams belonging to that project, paginated with a cursor.
- [ ] Results can be filtered and sorted by `createdAt`.
- [ ] Each team in the query result includes its `status`.
- [ ] Request without read access to the project returns 403.
- [ ] Query is implemented and tested on both postgres and spanner dialects.

## Open Questions

- Should the default query exclude deactivated and pending-purge teams, or return all statuses with status exposed as an explicit filter?

## Resources

- [ADR 027: Cursor-Based Pagination](https://github.com/zitadel/nextgen/blob/main/docs/adrs/027-cursor-based-pagination.md)
- [ADR 031: OpenAPI Querying](https://github.com/zitadel/nextgen/blob/main/docs/adrs/031-openapi-querying.md)

Issue #419, opened 2026-06-26
Permission Management — ADR 032–034 Implementation


muhlemmer:
## Overview

This epic tracks the full implementation of [ADR 032–034: Permission Management](https://github.com/zitadel/nextgen/blob/main/docs/adrs/).

The goal is a **portable relational FGA core** that covers both internal Zitadel resource authorization and external app/API permission grants using the same primitives. OpenFGA is used as the authoring and interchange format for app-group policies; the runtime contract is Zitadel-owned relational state with dialect-portable query plans (PostgreSQL and Spanner).

## Key Design Decisions

- One model, two catalogs: **system catalog** (Zitadel-owned) and **app-group catalog** (customer-owned)
- OpenFGA language package for parsing/syntax; Zitadel owns semantic validation, compilation, and SQL plan generation
- Canonical relational storage — no sidecar, no synchronization lag
- Relation closure computed at policy-upload time, not on every membership/grant write
- List endpoints receive an injected SQL predicate — no O(n) per-row checks in application code; multi-scope list queries derive authorized scopes first
- Authorization facts live in the same DB as the resources they protect (strong consistency)
- Every action under a scoped grant or agent delegation records full grant context in event actor metadata — no hidden access path
- Resolver enforces strict 403 vs 404 denial semantics to prevent scope-boundary information leakage
- Staff/support grants and break-glass operator access are out of scope for this epic — tracked by [issue #333](https://github.com/zitadel/nextgen/issues/333)

## References

- [ADR 032: Permission Catalogs — A Portable Relational FGA Core](https://github.com/zitadel/nextgen/blob/main/docs/adrs/032-permission-catalogs.md)
- [ADR 033: Internal Permission Management](https://github.com/zitadel/nextgen/blob/main/docs/adrs/033-internal-permission-management.md)
- [ADR 034: External Permission Management](https://github.com/zitadel/nextgen/blob/main/docs/adrs/034-external-permission-management.md)
- [ADR 024: User/Team Lifecycle & Ownership](https://github.com/zitadel/nextgen/blob/main/docs/adrs/024-user-team-lifecycle-ownership.md)
- [Authorization Design](https://github.com/zitadel/nextgen/blob/main/docs/design/api/authz.md)
- [URL Architecture](https://github.com/zitadel/nextgen/blob/main/docs/design/api/url-architecture.md)
[...]

PR #702, opened 2026-08-02
fix(console): stop offering navigation and controls that do not work

bastionstack:
# Which Problems Are Solved

Most of the console's chrome was not connected to anything, and it was no longer
possible to tell by looking which parts were real.

- **The Sessions link could not load.** `GET /sessions` is routed, documented and
  generated, but `sessionService.List` is a `//TODO implement me` returning
  `ErrNotImplemented`, so every visit answered 501 and landed on the error
  boundary.
- **The organisation switcher listed placeholder organisations** — "Acme Inc /
  Free", "Clearwater Labs / Pro", "Benimac LTD / Enterprise", "Horizons Studio /
  Pro" — with a hardcoded current selection and a plan badge naming a tier
  nothing sells. It sat above every screen, so the whole console looked scoped to
  a real organisation on a real plan.
- **Four sidebar entries were permanently disabled.** App groups, Applications,
  Analytics and Activity Log rendered as `aria-disabled` rows to match the mock.
  A disabled row still advertises a feature: it reads as "your account cannot do
  this" rather than "this does not exist".
- **Search and Documentation did nothing.** Both rendered as ordinary enabled
  buttons with no handler, which is worse than being disabled.

# How the Problems Are Solved

Each surface is commented out in place with the endpoint or decision that brings
it back, rather than deleted, since all of them return unchanged once unblocked:

- Sessions loses its screen and its nav entry until listing sessions exists
  (#699). `revokeSession` is already implemented, so the parked table is close to
  complete.
- The organisation switcher waits on a team list endpoint (#620) and on billing
  for the plan badge (#667). The project switcher beside it is real and stays.
- `DESIGN_ONLY_NAV` is emptied, with each entry's blocker recorded against it.
- Search waits on a cross-resource query endpoint; Documentation waits on a
  published docs URL.

Projects keeps its route and loader but loses its sidebar entry. Its data is
real, but it has had no design hand-off — the layout is a developer's key/value
table — and beside Users it read as a finished feature. `/projects` still answers
for anyone who wants it, as `/system` and `/flow-definitions` already did.

That leaves one nav entry, Users, which is the honest count of what has been
designed and built.

# Additional Changes

- The three real-instance e2e suites shared `signIn` by copy-paste; it is now a
  single helper, along with `expectNoErrorBoundary`. Byte-identical copies of the
  step every suite starts with could have drifted while looking the same.
- The shell spec now asserts the inverse of what it did: every nav row is a link,
  and the unbuilt labels appear nowhere.

# Additional Context

- Raised while sweeping: #699 (`GET /sessions` returns 501 — no issue tracked it)
- Related: #620 (team list), #667 (billing tiers), #350 (audit log), #419
  (app groups)

### DIFF
--- a/.changeset/console-remove-unbacked-ui.md
+++ b/.changeset/console-remove-unbacked-ui.md
@@ -1,0 +1,5 @@
+---
+"@zitadel/server": patch
+---
+
+The console no longer offers navigation and controls that had nothing behind them. The Sessions link, which could not load because listing sessions is not implemented yet, is hidden along with an organisation switcher that listed placeholder organisations, four permanently disabled sidebar entries, and Search and Documentation buttons that did nothing when clicked.
--- a/.changeset/console-user-list-columns.md
+++ b/.changeset/console-user-list-columns.md
@@ -1,0 +1,5 @@
+---
+"@zitadel/server": minor
+---
+
+The console's users list now takes its columns from the user schema, so it shows the attributes your schema actually defines instead of a fixed set. Users whose schema omits a column render it as empty rather than blank.
--- a/apps/console-e2e/src-real/add-user.spec.ts
+++ b/apps/console-e2e/src-real/add-user.spec.ts
@@ -1,6 +1,8 @@
 import type { Locator, Page } from "@playwright/test";
 import { expect, test } from "@zitadel/testing/playwright";
 
+import { expectNoErrorBoundary, signIn } from "./support";
+
 /**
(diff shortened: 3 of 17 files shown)


bastionstack:
Folded into #700. The delete-user dialog came first and the sweep grew out of it, so it belongs there as additional work rather than as a separate PR — splitting it left this one mapped to no issue at all.

Branch deleted; the commits are unchanged on `feat/console-delete-user-dialog-632`.

PR #707, opened 2026-08-02
feat: use-case-aware scaffold copy and embedding guidance

fforootd:
## Summary

Closes out the scaffold-template half of the Wicklore friction-log remediation (B2), building on #686's surface contract:

- **Use-case-aware copy**: `RendererTemplates.authPage` now receives the recorded use case; `--use-case business` scaffolds wire the SDK's `businessLocales` overlay (`locales={businessLocales}` from `@zitadel/sdk-next/client`) into the generated login/register pages, restoring work-email wording on top of the widgets' now-neutral built-in copy. Minimal/consumer scaffolds stay plain. The context field is required-but-optional-valued so every caller states what it knows — `doctor --fix` regenerates the same markup from `zitadel.json`'s recorded use case rather than assuming setup defaults.
- **Chrome to the widget**: the generated Next and Nuxt profile pages drop their hardcoded `minHeight`/`background: #0f0f11` main chrome — the session card's `variant="page"` surface (#686) paints the full-page chrome from design tokens; the wrapper now only pins the color scheme, matching the auth pages.
- **Embedding discoverability** (friction-log finding 6): every scaffolded page names the `variant="widget"` alternative in an emitted comment, and the generated AGENTS.md guidance gains one paragraph pointing at the `variant`/`theme` knobs — plus, on Next only, the SDK-shipped JSX types (`custom-elements.d.ts` → `@zitadel/sdk-next/jsx`).
- **Contract docs**: SKILLS.md's `--use-case` entry now names the copy-overlay effect; `PatchContext.useCase`'s docstring no longer claims the value never shapes output.

## Validation

- `moon run cli:lint cli:typecheck cli:test` — 934/934, including a new unit matrix (business/consumer/minimal templates, profile chrome, guidance knob sentence with its non-Next gate) and a new integration scenario: business scaffold → delete the login page → `doctor --fix` restores the business-flavored markup from the recorded use case (the #685×B2 seam, exercised through the real CLI).
- Local journeys `run-local.mjs --framework next` (4/4, incl. the drift probe) and `--framework nuxt` (3/3) — the published-consumer path with the changed templates.
- `changeset status --since origin/main` — `@zitadel/cli` minor.

## Release notes

`zitadel setup --use-case business` now overlays work-email copy on the generated auth pages via the SDK's `businessLocales`; generated profile pages lean on the session card's `variant="page"` chrome instead of hardcoding it; scaffolded pages and guidance name the `variant`/`theme` embedding knobs and the shipped JSX types.

## Notes

- Developed stacked on #686 and opened after it merged, so the diff is B2-only; the branch history keeps #686's original commits plus reconciliation merges (never rebased).
- Nuxt scaffolds intentionally do **not** get the business overlay yet: their pages import from `@zitadel/sdk-nuxt`, which doesn't re-export `businessLocales` — wiring it cleanly needs that re-export first (follow-up-sized, not template-sized).
- The `web-component` renderer placeholder ignores the new context param (structurally valid); it never scaffolds.


fforootd:
Codex round 1 (relayed) — both findings addressed on 0c656c26.

**1. [P1] `businessLocales` ignored on React 18 — confirmed, fixed via ref assignment.** Verified the exact failure shape: on React <19 the JSX prop decays to an attribute, and `locales` is `@property({ attribute: false })`, so the overlay silently vanished. Notably the pre-existing `project={project}` prop carries the same React-19 reliance but is backstopped by design — `resolveApi`'s precedence is prop → `configureZitadel()` global → attributes, so on React 18 the widget still works and `locales` was the only *silent* loss. The template now assigns the overlay through the ref (`ref={(element) => { if (element) { element.locales = businessLocales; } }}`), which binds as a property on 18 and 19 alike; the emitted comment states why, so the pattern doesn't get "simplified" back into a prop. Unit + integration tests now assert the ref form and reject the prop form, and the shipped-types chain is proven by a real `tsc --noEmit` of the exact pattern inside demo-next (`RefAttributes<ZitadelLogin>` ships in the jsx declarations).
On the alternative (require React 19): same shape as the Next-15 question on #685 — a real floor should be an explicit setup/doctor version error, never a silent narrowing, and that's a product decision beyond this PR. The ref makes the template correct at the declared floor instead.

**2. [P2] Overlay promised for every framework, delivered on Next — contract scoped, parity tracked.** SKILLS.md and the changeset now name Next scaffolds explicitly ("other frameworks keep the neutral copy until their SDK packages re-export the overlay"). Full wiring is deliberately out: the SPA scaffolds render SDK wrapper components (property-safe internally, so no React-18 caveat there), but none of sdk-react/vue/angular/svelte/solid/qwik/nuxt re-export `businessLocales` — the honest unit of work is per-SDK re-exports + template wiring + the test matrix, tracked as a follow-up task.

full-pr is rerunning on 0c656c26.

PR #663, opened 2026-07-30
feat(database): unify resource ID generation in storage dialects

adlerhurst:
## TL;DR

Every resource PK is a dialect-minted `prefix_<opaque>` string (**Postgres ULID**, **SQLite ULID**, **Spanner UUID v4**). SQL has no `IDENTITY` / uuid / AUTOINCREMENT defaults. Create APIs reject client resource PKs. Documented as **ADR 047**.

Also cleaned review leftovers: `sessions.token_id` is **nullable** (retired `"0"` sentinel), redundant Create-supplied DEFAULTs dropped across dialects, dead `project_secret`/`preview_secret` removed.

Follow-up review pass: prefix selection rules + keep-any ceremony note in ADR 047; statements-only ID minting in root/storage `AGENTS.md`; `ResourcePrefix` comment softened; `v2database` aliases dropped; `usr_`/`htok_` fixtures aligned to `user_`/`handoff_`.

**Synced `main`:** ListTeams + team statement suites in `stmttest` (#728/#733); dialect Spanner `team_test.go` removed; empty-ID Create asserted in stmttest (dialect minting).

## How to review (order)

1. **Contract** — `docs/adrs/047-dialect-id-generation.md` (+ amended 011/012/028)
2. **Generators** — `internal/storage/v2/dialect/idgen/` + dialect `managed_id.go` (postgres/sqlite ULID, spanner UUID)
3. **DDL** — goose SQL under `postgres/`, `spanner/`, `sqlite/` (TEXT PKs; nullable `sessions.token_id`; no redundant array/JSON/TTL DEFAULTs)
4. **Create paths** — mint-before-insert; session insert binds NULL `token_id` until token mint
5. **API** — OpenAPI `sess_`/`att_`/`ch_` patterns; KEK pre-mint for JWE `kid`
6. **Agent contract** — root + `internal/storage/v2/AGENTS.md` identifier minting
7. **Skim** — `api/generated/**`, stmttest HasPrefix fixtures

## Behavior changes

- Alpha-breaking: wipe local/CI DBs.
- Wire shape: always `prefix_…`; Spanner bodies are UUID v4.
- Local default dialect is SQLite; ID rules match Postgres (ULID).
- `sessions.token_id` is NULL until the real token is linked (no `"0"` placeholder).
- Secrets (handoff / session token material) unchanged — not row PKs.

## Validation

- `go test ./internal/service/ ./internal/domain/` — PASS
[...]

PR #627, opened 2026-07-24
docs: claim lifecycle v2 ADR (supersedes ADR 003) (#606)

IAM-marco:
## Summary

Implements Claim A1 (#606, part of epic #96): the ADR that supersedes the Withdrawn ADR 003 and records, as a single source of truth, the claim-lifecycle decisions made across the [epic #96 breakdown discussion](https://github.com/zitadel/nextgen/issues/96#issuecomment-4923654787), so downstream tickets (A2 contract #609, E1 service #611, etc.) point at one document.

- **New** `docs/adrs/041-claim-lifecycle-v2.md` (Status: Accepted). Records the six settled decisions: claim state is a permission-engine project-to-team grant (no `project_claims` table, no status field); `claim/complete` authed by the `__nextgen_session` platform session cookie; `challenge_id` is an ephemeral, handoff-token-style hashed random value; the personal team is created at platform registration (#527), not at claim; email-only identity for MVP; and the descopes/prerequisites (#527/#605, #419) with their accepted risks.
- Marks ADR 003 as superseded (header pointer + README index annotation) and adds the 041 index row.
- Rewrites only the top "current implementation" notes in `docs/design/platform/claim-flow.md` and `secret.md` so they stop framing secret rotation at claim, domain-based team matching, team creation at claim, and 14-day expiry as MVP, and point at ADR 041. The fuller target-design bodies are left intact.

Split into two commits: the ADR + its ADR-directory cross-references, then the design-doc note updates.

## Validation

- Documentation-only; no tests/build to run.
- Verified all internal ADR links resolve and that the two code claims cited in the ADR are accurate: `SecurityHandler.HandleOAuth2` (`internal/api/security.go`) and the `projects.project_secret` column.

## Release notes / changeset

- No changeset required: no shipped behavior changed (documentation only).

## Notes

- Closes #606.
- A1 has no dependencies; it unblocks A2 (#609), which encodes this ADR's semantics into the OpenAPI contract.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #710, opened 2026-08-02
docs: propose ADR 050 — the Dev Inbox capability

fforootd:
## Why

Flows that deliver a secret out of band — email verification, email-OTP login, password recovery (ADR 038) — cannot be tested end-to-end, and won't be when they arrive, unless tests can observe what the server sent. The server has **no email delivery subsystem at all yet**, which makes this the cheapest possible moment to fix the contract. And the platform overview already *narrates* the capability this ADR specifies: the scratch dashboard's dev inbox and the "Dev Inbox / Bring Your Own / Managed" delivery picker. This ADR is the server-side contract behind that story.

## What changed

**ADR 050 (Proposed): The Dev Inbox — Captured Outbound Messages.** Rewritten from the initial "capture transport" draft after two review inputs — a competitor scan (Supabase, Resend, Clerk, Firebase emulator, Stytch, Better Auth) and a codex review — into a product-capability framing:

- Composition/delivery split with per-environment **delivery modes**; `capture` stores messages in the project/environment-scoped **Dev Inbox** (ADR 035 scoping, ADR 036 operator plane).
- Messages are a **versioned record contract** (`schema_version`) carrying typed, purpose-discriminated **`artifacts`** (code/link — populated by composition from the flow engine's own state, never parsed from templates; `variables` stays a non-API record of render inputs) plus `rendered` forms for humans. Two timestamps with distinct meanings: `expires_at` = artifact/challenge validity (guaranteed shared per record), `retained_until` = store retention.
- **One cursor-paginated API, three consumers**: inbox UI, `zitadel dev-inbox` JSON commands, and `@zitadel/testing` (`cursor()` + `waitForMessage({after, to, purpose, artifact})` — the artifact-key predicate makes `waitForCode()` wait for a *code-bearing* record, since a purpose may produce link-only messages; keys, never values, appear in list metadata) — no test-level `clear()`, which would be cross-worker destructive under the kit's shared-instance parallelism. Waits are bounded and cancellable (`timeoutMs`, `AbortSignal`, typed timeout error echoing the resolved filters), and every consumer call binds an **explicit environment** (`InstanceHandle.environmentId`; CLI `--env` with deterministic resolution, echoed in the envelope).
- **Zero-config at the front doors, explicit in the binary**: `zitadel start` / `startLocalZitadel()` enable it, `connectZitadel()` discovers it; no tee mode.
- **Production-class environments require a real provider, enforced by a `provider_required` environment policy** (ADR 035 defers env-classes, so the ADR defines the enforceable bit instead of a classification) — enforced at **deploy, on every provider/policy mutation, at startup, and at send time** (a send without a working provider fails loudly, never falls back to capture); one deliberate divergence from the overview's current "Dev Inbox (default)" for claimed production, flagged in the ADR.
- **Browser inbox surfaces never hold the operator credential — and must earn their session** (ADR 036): BFF with a scoped HTTP-only session minted only through an authenticated exchange — human auth on claimed surfaces, or a one-time handoff minted by the explicit `zitadel dev-inbox open`, with the transfer specified: fragment-carried token (never in access logs), POST exchange, `history.replaceState`, `Referrer-Policy: no-referrer`, log redaction. `zitadel start` never emits a bearer-bearing URL — it reports `available` vs `configured` capability state and points at `open` only via a human-facing `next_actions` hint, never `next_commands` (which agents execute by contract, and which would also fail pre-`setup` when no secret exists). The scratch dashboard's anonymous first-visit cookie is called out as insufficient for inbox content — a second overview amendment.
- **List responses are metadata-only** — masking `variables` alone would be theater while `rendered.*` embeds the same secret; the full message is a separately authorized read. **Any durable backend encrypts message content at rest** (ADR 029); only the ephemeral in-memory store is exempt.
- **The kit cursor is a high-water mark, not an ADR 027 page token**: defined on empty inboxes, ordered by the store's monotonic per-environment sequence, purge-stable, carrying a **store epoch** (a restarted in-memory store fails old cursors fast instead of hanging waits), with distinct stale-token errors. The **rendered preview is sandboxed** (iframe, no scripts, restrictive CSP) per the flow-engine template-security guidance.
- Alternatives rejected with reasoning: external SMTP sink (Supabase's own layering as evidence), log capture, **fixed test codes on the real server** (Clerk-style — that experience belongs in `@zitadel/api-mock`), **admin-mint** (hash-only storage + flow-binding), kit-embedded SMTP listener.
- Staged build order: contracts → store/API → kit → CLI → UI → real providers.

Also in this branch's history: renumbered as `main` claimed numbers in flight — 042 → 043 → 044 → 046 → 047 → 048 → **050** (042 scaffolded-file ownership, 043 framework version floors #713, 044+045 scaffold posture and copy overlays #714, 046 claim lifecycle v2 #627, 047 dialect-owned identifier generation #663, 048+049 wide events #459) — with `main` merged in at each collision; the file is now `050-dev-inbox.md`.

## Validation

Docs-only.

## Review notes

Status is **Proposed** — a conversation starter. The first review round (livio) recorded a lean on every open question, now folded into the ADR: composition via a typed flow intent + async notification boundary (v4-shaped enqueue → compose → capture|deliver, wide-events/River as the bus); durable store only on multi-replica / shared scratch inbox / unread-survives-restart (ADR 029 bar applies equally to queue payloads); default-deny support access until #333; optional explicit `consumed_at` over auto-consume; one operator port with an optional path-prefix kill-switch. The same round added a normative purpose × artifact-key table, the static email-dependence detection signal for the `provider_required` gates, and an "On acceptance" checklist carrying both platform-overview amendments (claimed-production delivery default; scratch-inbox session acquisition). The leans harden into decisions when the ADR moves to Accepted.

A later `main` merge brought #722's ADR 036 amendment (credential exposure contracts), which governs exactly the fragment-carried handoff token this ADR specifies — the BFF section now carries an explicit conformance passage instead of a silent divergence: the exchange behaves as a credential-establishing operation (browser-attested `Origin` enforced, so an exfiltrated token can't be exchanged from a foreign context), the minted session is inbox-scoped and can never become a login session, and the endpoint adopts ADR 036's PKCE-style proof binding for URL-transiting handoffs once that machinery ships.

The #459 merge (wide events, ADRs 048/049) touches the composition lean, which named "wide-events/River" as the natural bus: ADR 048 defines wide events as an *audit* primitive (deny-by-default PII, export-only), so the lean now cross-references it and folds "can the audit stream double as the dispatch bus, or does the boundary ride a job queue with its own payloads" into the already-open event-shape question. ADR 049's events API stays on plain ADR 027 keyset cursors, so it does not collide with this ADR's high-water-mark cursor contract.


docs/adrs/043-test-visible-email-otp-capture.md
```diff
@@ -0,0 +1,100 @@
...
+today and will not be testable when they arrive unless tests can observe the
+code or link the server sent. The test kit's roadmap names
+`email.waitForCode` as the missing surface; it is blocked on a server-side
+story, not on kit design.
+
+Two facts shape the options:
+
+- **Greenfield:** the server currently has no SMTP client, no notification
+  module, and no flow step that sends email. Nothing needs retrofitting.
+- **The kit's seed principle** (see `packages/testing/AGENTS.md`): every kit
```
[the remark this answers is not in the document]
fforootd: Fair catch — that file lands in #709 (in review alongside this PR). The citation now says so explicitly: `packages/testing/AGENTS.md`, added by #709 (`39be90c0`); the principle itself is also stated inline in the same bullet, so the ADR reads standalone either way.

docs/adrs/047-dev-inbox.md
```diff
@@ -0,0 +1,355 @@
...
+  "retained_until": "…"
+}
+```
+
+Humans inspect `rendered`; agents and tests read `artifacts.code` /
+`artifacts.link` — **typed, purpose-discriminated fields the composition
+layer populates from the flow engine's own state** (it generated the
+secret), never parsed back out of templates. Which artifacts a purpose
+requires is part of the versioned contract (`email_verification` carries a
+`code` and/or `link`; a magic-link purpose requires `link`). `variables`
```
livio-a: Non-blocking: worth a small normative table (purpose → required artifact keys) in the versioned contract. Prose “code and/or link” will get ambiguous once kit types and `artifact:` filters freeze.

docs/adrs/050-dev-inbox.md
```diff
@@ -0,0 +1,355 @@
...
+does not classify environments either. It defines the signal the server
+enforces instead: a `provider_required` policy on the environment record.
+Provisioning sets it (the claim flow marks production environments;
+operators can set it anywhere), and while it is set, deploying a release
+that enables email-dependent flows without a configured provider
+**fails** — a hard failure, not a warning. Deployment is not the only
+gate, because delivery mode and provider configuration are mutable
+environment state: **every provider or policy mutation validates the
+invariant** (removing the provider, or enabling `provider_required`
+without one, is rejected while email-dependent flows are deployed),
```
livio-a, line 155: Non-blocking: how does the server know a release “enables email-dependent flows”? One sentence on the detection signal (flow step kinds / notification intents / capability bit) would help the deploy and mutation gates stay implementable without inventing policy later.

docs/adrs/050-dev-inbox.md
```diff
@@ -0,0 +1,355 @@
...
+startup re-validates against drift, and at send time a
+`provider_required` environment without a working provider **fails the
+send loudly — it never falls back to capture**. When env-classes arrive,
+they map classes to this policy's default. This is one deliberate divergence
+from the overview's current narrative (which lists "Dev Inbox (default)"
+for claimed production): customer-facing password-reset mail must not
+silently land in a developer inbox. The overview gets amended to match if
+this ADR is accepted.
```
livio-a, line 163: +1 on both overview amendments (production not defaulting to capture; scratch cookie insufficient for inbox — also called out below). Suggest linking follow-up issues (or a checklist under Consequences) so Accept doesn’t leave narrative and contract disagreeing.

docs/adrs/050-dev-inbox.md
```diff
@@ -0,0 +1,355 @@
...
+a human-facing `next_actions` hint, never through `next_commands`: the
+CLI contract tells agents to prefer and *execute* `next_commands`, and
+`open` both spawns a browser and requires a configured environment secret
+that does not exist yet on the fresh golden path (`start` runs before
+`setup`). The durable operator credential never appears in a URL or
+reaches the browser. The scratch dashboard's current
+anonymous first-visit cookie (platform overview) is **not** sufficient
+authorization for inbox content — pre-claim, its inbox view rides the
+same secret-mediated handoff. That is a second overview amendment,
+alongside the delivery-mode default.
```
livio-a, line 228: +1 — second overview amendment. Same follow-up ask as on the delivery-mode default: link an issue or checklist item when moving to Accepted.

docs/adrs/047-dev-inbox.md
```diff
@@ -0,0 +1,355 @@
...
+  of silent capture — which requires one amendment to the platform
+  overview's delivery-mode default.
+- The scratch dashboard's inbox, the local inbox UI, the CLI, and the kit
+  stay one contract; none can drift into a private side channel.
+
+## Open questions
+
+- Composition locus: a `send` effect in the flow engine, or a notification
+  service consuming events (and how that aligns with the wide-events
+  direction)?
```
livio-a: Lean: notification/async boundary after the flow emits an intent (v4-shaped: enqueue → compose → capture|deliver). Flow should not own SMTP/HTML. Artifacts still come from flow/challenge state as specified above. Wide-events/River fit naturally as that bus; exact event shape can stay open.

docs/adrs/047-dev-inbox.md
```diff
@@ -0,0 +1,355 @@
...
+- Durable-store trigger: at what point (shared cloud dev instances,
+  multi-replica) the in-memory store must be replaced.
```
livio-a: Lean: stay in-memory for local single-process (`zitadel start` / `startLocalZitadel`). Switch to a durable store when any of: multi-replica, shared cloud/scratch inbox, or unread messages must survive restart. At-rest encryption per ADR 029 applies at that point (same bar whether the durable home is the inbox store or a job queue payload).

docs/adrs/047-dev-inbox.md
```diff
@@ -0,0 +1,355 @@
...
+- Support-access overlap: a support agent reading a tenant's dev inbox
+  reads user secrets — same policy surface as the support-access model
+  discussion.
```
livio-a: Agree this is #333 territory. Default deny for staff/support until that model exists; inbox content is bearer-secret class (break-glass + audit, not a normal grant). No inbox-specific policy design needed in this ADR.

docs/adrs/047-dev-inbox.md
```diff
@@ -0,0 +1,355 @@
...
+- Consumed-state: should verifying a code mark the message consumed in the
+  inbox (needs flow cooperation), or is expiry display enough?
```
livio-a: Lean: expiry display (`expires_at`) is enough for v1 correctness — kit waits stay cursor/filter based and don’t need consumed.

Prefer an **optional explicit** mark-consumed API on the operator plane (`consumed_at`, idempotent; message stays readable until `retained_until`) over auto-consume on successful verify. Auto-consume couples inbox writes to the check path and gets messy under resend / parallel workers. A later read-only `challenge_status` derived from auth-attempt state is observation, not an inbox write.

docs/adrs/047-dev-inbox.md
```diff
@@ -0,0 +1,355 @@
...
+- Does the read API belong on the existing operator port or a separate
+  debug listener that deployments can firewall wholesale?
```
livio-a: Lean: same operator port and API. No separate debug listener.

Optionally later: runtime flag / config (or gateway rule) that disables or rejects the inbox path prefix wholesale. Keeps discovery (`available` vs `configured`) and the three consumers on one contract; network isolation stays deploy/config, not a second surface.

livio-a (approved):
## Review

**Approve as Proposed.** Strong capability ADR — composition/delivery split, one operator API for UI/CLI/kit, parallel-safe cursor semantics, and the security/handoff rules are in good shape. Greenfield timing is right.

I love that this solves several pain points we have with Zitadel v4 in one contract: no SMTP-sink / log-scraping for E2E, structured artifacts instead of parsing rendered mail, a real inbox for humans and agents over the same API, parallel-safe waits without destructive `clear()`, capture as a first-class delivery mode instead of a test-only side channel, and cloud/dev deployments that no longer need a default SMTP just to get a working environment — `capture` is the zero-config path until a real provider is required.

### Accept gate
The two platform-overview amendments already named in the ADR should be scheduled (or linked as follow-ups) when this moves to Accepted:

1. Claimed production must not default to Dev Inbox (`provider_required`).
2. Scratch dashboard anonymous cookie is insufficient for inbox secrets — secret-mediated handoff.

### Open questions — provisional leans (happy to discuss)
Recording these so build step 2 doesn’t thrash; fine to keep as open questions with a “lean” note:

1. **Composition locus** — prefer an async notification path (flow emits intent; worker composes/delivers or captures). Thin enqueue-from-flow is fine; flow-owned mailer is not. Artifacts still sourced from flow/challenge state as the ADR says.
2. **Durable store** — in-memory until shared/cloud/multi-replica (or restart must preserve unread). ADR 029 applies once durable.
3. **Support access** — inherit #333; default deny for staff until that model exists. Don’t invent inbox-specific policy here.
4. **Consumed-state** — expiry display is enough for correctness. Prefer an optional explicit mark-consumed API (operator plane) over auto-consume on verify (avoids flow coupling / resend races). Kit ignores it.
5. **Port** — same operator port/API. Later optional runtime flag (or gateway rule) to disable/block the inbox path prefix wholesale — no separate debug listener.

### Nice-to-haves (non-blocking)
- Normative purpose → required-artifacts table (prose “code and/or link” is soft for kit types).
- One line on how “email-dependent flows” are detected at deploy/mutation time.

docs/adrs/047-dev-inbox.md
fforootd: Added the normative table (purpose × `code`/`link`, with an at-least-one invariant where both are optional) in `75bc4d22` — kit types and `artifact:` filters freeze against rows, not prose; a new purpose lands with its row, and new keys extend `schema_version`.

docs/adrs/050-dev-inbox.md
fforootd: Added the detection signal in `75bc4d22`: email-dependence is read statically from the release's pinned flow definitions — a step kind that declares an outbound-message effect (the same declaration composition consumes at runtime) marks it; no separate capability bit to drift. One signal serves the deploy and mutation gates alike.

docs/adrs/050-dev-inbox.md
fforootd: Added an "On acceptance" checklist under Consequences in `75bc4d22` carrying both overview amendments, so Accept can't leave narrative and contract disagreeing. Kept it a checklist rather than issues while the ADR is still Proposed; they convert to issues at Accept.

docs/adrs/050-dev-inbox.md
fforootd: Covered by the new "On acceptance" checklist under Consequences (`75bc4d22`) — its second item is exactly this amendment (anonymous first-visit cookie → secret-mediated handoff).

docs/adrs/047-dev-inbox.md
fforootd: Recorded as the lean on the open question in `75bc4d22`: flow emits a typed intent; an async notification boundary composes then captures-or-delivers (v4-shaped enqueue → compose → capture|deliver); the flow never owns SMTP/HTML; artifacts still come from flow/challenge state; wide-events/River named as the natural bus with the exact event shape left open.

docs/adrs/047-dev-inbox.md
fforootd: Recorded in `75bc4d22`: in-memory for the local single-process instance; durable when any of multi-replica / shared cloud-scratch inbox / unread-survives-restart applies — with the ADR 029 bar applying equally whether the durable home is the inbox store or an async job-queue payload.

docs/adrs/047-dev-inbox.md
fforootd: Recorded in `75bc4d22`: default deny for staff/support until #333 lands; inbox content classed bearer-secret (break-glass plus audit, never a normal grant); the policy design stays in #333, not this ADR.

docs/adrs/047-dev-inbox.md
fforootd: Recorded in `75bc4d22`: v1 ships expiry display only and kit waits stay cursor/filter-based; if marking arrives it's an explicit idempotent operator-plane `consumed_at` with the message readable until `retained_until` — never auto-consume on verify; a derived read-only `challenge_status` stays observation, not an inbox write.

docs/adrs/047-dev-inbox.md
fforootd: Recorded in `75bc4d22`: one operator port and API, no second listener; optionally later a config/gateway kill-switch rejecting the inbox path prefix wholesale; network isolation stays deploy/config, not a second surface.

PR #718, opened 2026-08-02
fix: verify the tunnelled session token and keep it out of client payloads

fforootd:
Was stacked on #717 (merged — now targeting `main`). Closes the two sdk-next hardening findings from the getSession session-state work, plus a Codex review round (provider client-boundary leak, `opaqueTokenTimeoutMs`).

## Summary

- **`auth()` verifies instead of trusts.** It previously `decodeJwt()`ed the `x-nextgen-auth-token` header without signature verification and treated *any* other non-empty value as a middleware-validated opaque token — so on any route outside the middleware `matcher` (the scaffolded matcher covers only the proxy path and protected routes), a client-supplied header spoofed any identity, and even `x-nextgen-auth-token: garbage` returned `isAuthenticated: true`. `auth()` now re-verifies every value it reads: JWT-shaped tokens get full `verifyJwt` (JWKS signature, `iss`, `exp`, alg/typ allow-lists — same rules and defaults as the middleware, in-process key cache), opaque tokens are validated against `GET /sessions/me` (React `cache()`-deduplicated per render pass, fail-closed with a `console.warn` that distinguishes forged headers from JWKS/backend reachability problems). Opaque sessions now also return the real identity from the backend response instead of `userId: "unknown"`. New optional `AuthOptions` (a `Pick` of the middleware options) for apps that customise verification; the docstring now states the matcher precondition instead of claiming the middleware always validated the header.
- **`NextgenProvider` strips the token server-side — and is server-only.** The README's documented pattern (`await auth()` into the provider) serialised the raw session token into the RSC flight payload, readable by any client script. The provider now converts to the client-safe `ClientAuthResult` (from #717, explicit field pick) **before** the value crosses the server→client boundary. Per the Codex P1 finding, the provider is additionally guarded with `server-only` and exported from `@zitadel/sdk-next/server` (+ root), *not* `/react`: wrapping it in a `"use client"` file (the common `providers.tsx` pattern) would make the wrapper the boundary and serialise the still-unstripped prop — that wrapper is now a build error instead of a silent leak. Client-seeded trees (e.g. from `getSession()`) use the new `AuthContextProvider` on `/react`, which only accepts the token-less shape. `useAuth()` returns `ClientAuthResult`, matching sdk-nuxt and `getSession()`.
- **Client entry + root-barrel guard.** New `@zitadel/sdk-next/react` entry (`useAuth`, `AuthContextProvider`, client-safe types); `auth.ts` imports `server-only` (already a dependency, previously unwired), so pulling `auth()` into a `"use client"` graph fails the build with an import trace. README gains an entry-point table; `sideEffects` declared (`client.js` for element registration, `auth.js`/`provider.js` to keep the guard imports unprunable).
- **`opaqueTokenTimeoutMs` honoured (Codex P2 — ran deeper than reported):** the option already existed in sdk-core's `NextgenMiddlewareOptions` and sdk-nuxt honours it, but *both* sdk-next validation passes coupled the `GET /sessions/me` timeout to `jwksTimeoutMs` — the middleware's opaque fallback (pre-existing bug) and the new `auth()` (which copied it faithfully). Both now thread `opaqueTokenTimeoutMs` (default 5000 ms), it's part of `AuthOptions`, and the README options table documents it.
- **Build correctness (found while wiring the above):** tsup's chunk splitting hoisted the compiled `createContext` module into a shared chunk *without* its `"use client"` directive — the published provider/hook entries could never have worked from a real consumer (nothing in-repo exercised them, which is how it survived). The package now builds with per-file `tsc` emit like sdk-core (directives preserved per module, dep types resolved via built dist by neutralising the `@zitadel/source` condition), relative imports carry `.js` extensions, and demo-next now renders `NextgenProvider` + a `useAuth` badge so the documented pattern compiles in CI forever.
- `isJwtShaped` moves to `@zitadel/sdk-core` (shared JWS/JWE structural detection; the middleware imports it from there).

## Validation

- `moon run sdk-core:{lint,typecheck,build,test} sdk-next:{lint,typecheck,build,test} demo-next:{lint,typecheck,build} demo-next-e2e:lint` — all green (sdk-next: 132 tests at head; earlier counts in review threads reflect older heads).
- New regression tests use real RSA-signed JWTs against a mocked JWKS: forged/unsigned JWT rejected, signed-without-`sub` rejected (previously *authenticated* via the opaque fallthrough), expired rejected, garbage opaque header rejected via mocked 401, backend-confirmed opaque token authenticates with real identity, anonymous/500/unreachable fail closed, and `opaqueTokenTimeoutMs` proven with a signal-respecting fetch mock (abort actually reaches the request). Provider tests assert the context value contains no `token` key for every accepted input shape; `AuthContextProvider` covered for client seeding.
- Compile-guard checks in demo-next (temporary routes, since removed): importing `auth` from the root barrel in a `"use client"` page fails the build with the full import trace into `dist/auth.js`; the Codex P1 repro — a `"use client"` wrapper around `NextgenProvider` — fails with `'server-only' cannot be imported from a Client Component module`; the committed subpath-based provider/badge usage builds clean.
- `demo-next-e2e/src/auth.spec.ts` now asserts the raw session-cookie value appears **nowhere** in the server's `/admin` response (HTML + inlined RSC flight payload) — a durable leak detector for this class. Lint/typecheck-verified; not executed locally (port 8080 is held by a live local `nextgen` server and the e2e web-server config pins that port; the suite is opt-in local). Worth a run wherever 8080 is free.
- `dist/` inspected: `"use client"` heads `context.js`/`useAuth.js`, `provider.js` opens with `import "server-only"` (shared-render deliberately removed), `auth.js` likewise.

## Release notes / changeset

`.changeset/verify-auth-header-strip-token.md` — `@zitadel/sdk-next` minor, `@zitadel/sdk-core` minor. Security hardening plus surface changes (`useAuth()` return shape, new `/react` entry, `NextgenProvider` server-only, root barrel build-errors in client graphs, opaque timeout no longer follows `jwksTimeoutMs`).

## Notes

- Defense-in-depth cost: on matcher-covered routes the middleware has already validated the token, so `auth()`'s re-verification adds ~one cached-JWKS signature check for JWTs, or one `/sessions/me` round-trip per render pass for opaque tokens (per-render `cache()` dedupe). If that round-trip ever matters, an HMAC attestation minted by the middleware (e.g. keyed off `ZITADEL_PROJECT_SECRET`) could skip it — deliberately out of scope here.
- A further evolution would be `auth()` reading the `__nextgen_session` cookie directly and dropping the matcher precondition entirely — but that contradicts the just-landed #717 guidance (which routes app chrome to `getSession()`), so it's flagged rather than done.
- Root-barrel client imports of *client* exports (`useAuth`) happen to compile under Turbopack's per-export tree shaking; that's bundler-dependent, so the README documents `/react` and `/session` as the supported client imports rather than blessing it.


fforootd:
Addressed the Codex review round in c4e7f50:

**P1 (provider leak through a client wrapper) — accepted, fixed structurally.** The shared-component strip was only safe when the provider rendered from a server module; a consumer's `"use client"` wrapper made the *wrapper* the RSC boundary and serialised the raw `session` prop before the strip ran. `NextgenProvider` now imports `server-only` directly and lives on `@zitadel/sdk-next/server` (+ root) — the reproduced wrapper pattern now fails the build with `'server-only' cannot be imported from a Client Component module` (verified against demo-next). `/react` keeps `useAuth` and gains `AuthContextProvider` (accepts only the token-less `ClientAuthResult`) for client-seeded trees. Per the test request, `demo-next-e2e/src/auth.spec.ts` now asserts the raw session-cookie value appears nowhere in the `/admin` response (HTML + inlined flight payload); it's lint/typecheck-verified but awaits an e2e run on a machine where port 8080 is free.

**P2 (`opaqueTokenTimeoutMs` ignored) — accepted; it ran one layer deeper.** The option already exists in sdk-core and sdk-nuxt honours it, but the sdk-next *middleware's* opaque fallback also coupled `GET /sessions/me` to `jwksTimeoutMs` (pre-existing; `auth()` had copied it). Both sdk-next passes now thread `opaqueTokenTimeoutMs` (default 5000 ms), it's part of `AuthOptions`, documented in the README options table, and covered by a signal-respecting fetch-mock test that proves the abort actually reaches the request.

Also refreshed the PR body (test count is 132 at head — the 127 was a pre-merge-round number, as flagged).

PR #720, opened 2026-08-03
refactor(authz): OpenFGA parser, profile validator, and compiler

### LINKED ISSUE #421: OpenFGA Parser, IR & Profile Compiler
## Context

[ADR 032 §2](https://github.com/zitadel/nextgen/blob/main/docs/adrs/032-permission-catalogs.md#2-openfga-parser-and-profile-compiler) defines the compiler pipeline for converting OpenFGA DSL/JSON into Zitadel's internal relational representation.

The pipeline is:
```
OpenFGA DSL / JSON
  → upstream OpenFGA language parser / transformer
  → Zitadel profile validator
  → Zitadel authz IR
  → relational catalog rows + query plans
```

Zitadel must **not** build its own DSL parser — use [`github.com/openfga/language/pkg/go`](https://pkg.go.dev/github.com/openfga/language/pkg/go) for parsing and syntax. Zitadel owns semantic validation, profile enforcement, and SQL plan generation.

## Package Boundaries

| Package | Responsibility |
|---|---|
| `internal/authz/openfga` | Parse DSL/JSON with upstream package; normalize to Zitadel IR |
| `internal/authz/profile` | Reject unsupported constructs; enforce bounded, portable rules |
| `internal/authz/compiler` | Compute relation/permission closure; produce relational catalog mutations and query-plan metadata |

## Supported Profile (MVP)

- Direct relations
- Relation implication / computed usersets
- Union
- Bounded tuple-to-userset through known Zitadel hierarchy edges (`project`, `team`, `app_group`, `app`, `user`)
- Userset references for team membership and assignment

## Rejected Constructs (fail at upload time with explicit diagnostics)

- Unbounded recursion
- Contextual tuples on request hot paths
- Wildcard expansion for list endpoints
- Arbitrary caveats/conditions requiring non-indexed attribute evaluation

## Tasks

- Add `github.com/openfga/language/pkg/go` as a dependency
- Implement `internal/authz/openfga`: parse DSL/JSON → IR
- Implement `internal/authz/profile`: semantic validation against supported profile; return structured diagnostics for unsupported constructs
- Implement `internal/authz/compiler`: compute relation closure; produce catalog row mutations and query-plan metadata
- Unit tests covering supported constructs, rejection diagnostics, and round-trip DSL/JSON

## References

- [ADR 032 §2 — OpenFGA parser and profile compiler](https://github.com/zitadel/nextgen/blob/main/docs/adrs/032-permission-catalogs.md#2-openfga-parser-and-profile-compiler)
- [openfga/language Go package](https://pkg.go.dev/github.com/openfga/language/pkg/go)

livio-a:
## Summary

- Implements #421: OpenFGA DSL/JSON → portable authz IR (`internal/authz/openfga`), MVP profile validation with structured diagnostics (`internal/authz/profile`), and compilation to storage-neutral catalog mutations + query-plan metadata (`internal/authz/compiler`).
- Uses upstream `github.com/openfga/language/pkg/go` for parsing; Zitadel owns profile enforcement and plan output. Persistence/SQL mapping stays with #422.

## Validation

- `go test ./internal/authz/...`

## Release notes / changeset

- Changeset: `.changeset/openfga-authz-compiler.md` — OpenFGA authz parser/profile/compiler foundation (`@zitadel/server` minor)

## Notes

- Output deliberately has no catalog IDs, SQL, or dialect types — #422 maps rows to Postgres/Spanner.
- Bundles / flat `{resource}.{verb}` catalog alignment and Leopard flattening are deferred.
- Closes #421


.changeset/openfga-authz-compiler.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zitadel/server": minor
+---
+
+Add OpenFGA authorization model parser, profile validator, and compiler foundation for portable permission catalogs.
```
fforootd: Nothing imports `internal/authz` yet (I checked on the branch), so the shipped binary is byte-identical — per our own [decision table](https://github.com/zitadel/nextgen/blob/main/.changeset/README.md#decision-table) that's the "touches a publishable path but ships nothing" row: **empty changeset**, and per the [title ladder](https://github.com/zitadel/nextgen/blob/main/CONTRIBUTING.md#title-format) not `feat` (step 1: nobody can reach this yet) — `refactor(authz)` or `chore(authz)`.

The summary text is also nearly the style guide's own bad example ("Implement ADR 030 errreport foundation in internal/errreport") — "foundation" is on the banned-word list there precisely because it means nothing to a release-notes reader.

Suggestion: make this an empty changeset and spend the real release note on the PR that wires the upload/check surface (#422 / ADR 034), where there's an outcome a customer can actually read about. The Semantic PR check passed only because it verifies changeset *presence*, so this won't get caught mechanically.

internal/authz/profile/profile.go
```diff
@@ -0,0 +1,230 @@
...
+			diagnostics = append(diagnostics, Diagnostic{
+				Code:     CodeRelationNotFound,
+				Type:     objectType,
+				Relation: relation.Name,
+				Message:  fmt.Sprintf("implies undefined relation %q", rewrite.Relation),
+			})
+		}
+	case authz.RewriteTupleToUserset:
+		diagnostics = append(diagnostics, v.validateTupleToUserset(index, objectType, relation, rewrite)...)
+	case authz.RewriteUnion:
```
fforootd, line 140: The JSON entry path can smuggle shapes the DSL grammar can't produce — `LoadJSONStringToProto` only deserializes; OpenFGA's typesystem validation lives in openfga/openfga, not the language package, so this validator is the only semantic gate for JSON input. One shape gets through silently — an empty union:

```json
{"schema_version": "1.1", "type_definitions": [
  {"type": "user"},
  {"type": "document", "relations": {"viewer": {"union": {"child": []}}}}
]}
```

validates clean and compiles to a `document#viewer` plan with **zero terms**: present in the catalog, never satisfiable, no diagnostic (I verified this on the branch). It fails safe (deny), but it's exactly the "degrading silently" that ADR 032 says upload time should fail loud on. Rejecting unions with no children here — or emitting a diagnostic for any relation whose flattened term list ends up empty, which also covers future gaps — would close it.

internal/authz/compiler/compiler.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func flattenRewrite(
+	model authz.Model,
+	target Relation,
+	directReferences []authz.RelationReference,
+	rewrite authz.Rewrite,
+) []QueryTerm {
+	switch rewrite.Kind {
+	case authz.RewriteDirect:
```
fforootd, line 97: Same crafted-JSON vector as the empty-union case: a union with two `this` children

```json
{"union": {"child": [{"this": {}}, {"this": {}}]}}
```

passes the profile and flattens to two identical `TermDirect` entries — duplicate expression edges at positions 0 and 1 (verified on the branch). Harmless as an OR today, but if #422 puts a uniqueness constraint on expression-edge rows (target, kind, source, tupleset), this insert collides at persist time with a much worse error than a profile diagnostic. Deduping repeated terms during flattening would keep the compiled catalog canonical regardless of input shape.

internal/authz/profile/profile.go
```diff
@@ -0,0 +1,230 @@
...
+		})
+	}
+
+	return diagnostics
+}
+
+// validateTupleToUserset checks that "<relation> from <tupleset>" resolves
+// through a directly assignable tupleset whose target types are known
+// hierarchy edges and actually define the computed relation.
+func (v Validator) validateTupleToUserset(index modelIndex, objectType string, relation authz.Relation, rewrite authz.Rewrite) Diagnostics {
```
fforootd, line 183: Three of the branches in this function have no test coverage: the undefined-tupleset case, `must restrict to plain types` (userset/wildcard restriction on the tupleset), and `type %q does not define relation %q`. Worth adding rows to the `TestValidateRejectsUnsupportedConstructs` table — and same for `missing_type_restriction`, which is interesting because it's reachable *only* via JSON (the DSL can't write a direct relation without a bracket list), so a test doubles as documentation of that asymmetry.

Nit while here: line 201 renders the hierarchy list with `%v` → `[app app_group project team user]` — Go slice syntax in a customer-facing diagnostic. A quoted, comma-joined list reads better.

internal/authz/profile/profile.go
```diff
@@ -0,0 +1,230 @@
...
+// DefaultHierarchyTypes are the object types a tuple-to-userset may traverse.
+// Bounding traversal to known edges keeps a check's join depth predictable on
+// both PostgreSQL and Spanner.
+var DefaultHierarchyTypes = []string{"app", "app_group", "project", "team", "user"}
+
+// Validator checks a model against the supported profile.
+type Validator struct {
+	// HierarchyTypes bounds tuple-to-userset traversal. Empty means
+	// [DefaultHierarchyTypes].
+	HierarchyTypes []string
```
fforootd, line 37: `HierarchyTypes: []string{}` silently falls back to the defaults, so "no TTU allowed at all" is inexpressible. If that's intended, one more sentence here pins it down; if not, distinguishing nil (defaults) from empty (deny all TTU) is cheap now and a breaking change later.

internal/authz/profile/profile.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (v Validator) validateTypeRestrictions(index modelIndex, objectType string, relation authz.Relation) Diagnostics {
+	diagnostics := make(Diagnostics, 0)
+	for _, reference := range relation.DirectlyRelatedTypes {
+		diagnostic := Diagnostic{Type: objectType, Relation: relation.Name}
+
+		switch {
+		case reference.Wildcard:
+			diagnostic.Code = CodeUnsupportedWildcard
+			diagnostic.Message = fmt.Sprintf("wildcard restriction %q:* cannot be planned for list endpoints", reference.Type)
```
fforootd: nit: this renders as `wildcard restriction "user":*` — quoting the whole token (`"user:*"`) reads more naturally. The exact string is pinned in `profile_test.go` and `compiler_test.go`, so it's a three-site change.

fforootd (commented):
Solid implementation — this tracks ADR 032 §2 closely: the package boundary is verbatim from the ADR table, the hierarchy edges and unsupported-construct set match, and the output types stay clean of catalog IDs/SQL so #422 keeps full freedom on the relational shape.

What I did beyond reading: checked out the branch, ran `go vet` + `go test ./internal/authz/...` (green, matching CI), hand-checked the closure BFS and its deterministic catalog-order emission against the test expectations, and threw four adversarial models at it. Recursion detection held up well — a self-type TTU through a *valid* hierarchy edge (`define parent: [project]` + `define viewer: [user] or viewer from parent`) is rejected as `unbounded_recursion`, and a cross-type team↔project TTU cycle rejects both participants with stable diagnostics. The two shapes that slipped through are both crafted-JSON-only and deny-safe; repros are in the inline comments.

Things I want to call out as particularly good:

- The diagnostics design: comparable `Diagnostic` + `Diagnostics.Unwrap() []error` makes a single violation `errors.Is`-matchable without scanning, and reusing OpenFGA's own error vocabulary (with the documented rationale for *not* reusing 2002) is exactly the right customer-empathy call.
- Determinism is handled at every layer (sorted normalization, positions, catalog-order closure rows) — the future catalog-diff story will lean on this.
- The oneof-discriminator switch in `normalizeRewrite` instead of the unreliable `GetThis()` is a subtle trap avoided, and the comment explains it for the next reader.
- The closure/plan split (computed usersets → closure; TTU and userset refs → plan terms) is the correct reading of ADR 032 §3's write-cheap model, and there's a test pinning that TTU is not treated as a global implication.

Two notes that don't belong on any one line:

- **Upload-path bounds:** recursion checking is O(V·E) per relation, the closure is O(V²) rows worst case, and ANTLR parses untrusted tenant input. All fine at schema scale, but #422's upload API should cap input size and type/relation counts — worth carrying over as a note on that issue so it doesn't get lost.
- **ADR bookkeeping:** ADR 032 is still marked *Proposed* and its follow-up item 4 is exactly this PR — worth ticking off (or revisiting the status) when this lands.

The one thing I'd change before merge is the changeset/title pairing — see the comment on the changeset file.

adlerhurst (commented):
No additional comments to ffo's review.

internal/authz/openfga/parse.go
```diff
@@ -0,0 +1,230 @@
...
+		if err != nil {
+			return authz.Type{}, fmt.Errorf("normalize OpenFGA relation %q.%q: %w", result.Name, name, err)
+		}
+
+		relation := authz.Relation{
+			Name:    name,
+			Rewrite: normalizedRewrite,
+		}
+		// Direct type restrictions live in relation metadata, not in the
+		// This userset rewrite itself.
```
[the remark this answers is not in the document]
livio-a: Fixed — comment now refers to the DirectUserset ("this") rewrite.

internal/authz/openfga/parse.go
```diff
@@ -0,0 +1,230 @@
...
+		if err != nil {
+			return authz.Rewrite{}, fmt.Errorf("normalize difference subtract: %w", err)
+		}
+		return authz.Rewrite{
+			Kind:     authz.RewriteDifference,
+			Children: []authz.Rewrite{base, subtract},
+		}, nil
+	default:
+		return authz.Rewrite{}, fmt.Errorf("unsupported userset rewrite")
+	}
```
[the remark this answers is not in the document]
livio-a: Fixed — fallback includes `%T` of the userset oneof.

.changeset/openfga-authz-compiler.md
livio-a: Done in `32f10f57` — switched to an empty changeset and renamed the PR to `refactor(authz): …`.

internal/authz/profile/profile.go
livio-a: Done — empty unions (and unions with no assignable leaves) now fail with `empty_relation_definition`.

internal/authz/compiler/compiler.go
livio-a: Done — union flatten dedupes identical terms before emitting expression edges / plan terms.

internal/authz/profile/profile.go
livio-a: Done — added coverage for undefined tupleset, plain-type tupleset restriction, missing computed relation on the target type, and JSON-only `missing_type_restriction`. Hierarchy list is now quoted/comma-joined.

internal/authz/profile/profile.go
livio-a: Done — `nil` keeps the defaults; non-nil empty denies all TTU. Documented on the field.

internal/authz/profile/profile.go
livio-a: Done — message is now `wildcard restriction "user:*"`.

fforootd (approved):

internal/authz/compiler/closure.go
```diff
@@ -0,0 +1,54 @@
...
+		for _, implied := range direct[current] {
+			if depth, seen := depths[implied]; seen && depth <= nextDepth {
+				continue
+			}
+			depths[implied] = nextDepth
+			queue = append(queue, implied)
+		}
+	}
+	return depths
+}
```
adlerhurst: Looks sloppy to me. wouldn't a `for _, current := range queue` loop express the same?

internal/authz/compiler/output.go
```diff
@@ -0,0 +1,108 @@
+// Package compiler turns a profile-valid authz model into storage-neutral
+// catalog mutations and query-plan metadata.
+//
+// The output deliberately carries no catalog ID, version ID, SQL, or dialect
+// types. Issue #422 owns the relational schema and maps these records into
+// PostgreSQL and Spanner rows.
```
adlerhurst: This description makes sense but move it to a doc.go file. it is also outdated (sqlite is missing)

internal/authz/compiler/closure.go
livio-a: Swapped the `queue[0]` / `queue[1:]` dequeue for `for i := 0; i < len(queue); i++`. Didn't use `range queue` because append-during-iteration wouldn't visit new nodes in Go; index loop is the standard BFS pattern here.

adlerhurst (approved):

PR #721, opened 2026-08-03
feat: user metadata

### LINKED ISSUE #631: Frontend: User detail — overview, authentication, and project permissions

### LINKED ISSUE #661: Backend: Use page-tokens for list users endpoint
Currently the list users endpoint uses an `offset` and `limit` to do pagination. However, ADR027 describes to use cursor based pagination instead. This is already implemented on other endpoint. 

Because the list users endpoint was created before this ADR was created, it still uses an older form of pagination. This should be refactored, preferably before the first version of the application since this would break the contract.

Sorting is currently also not implemented. These can both be addressed at the same time.

wim07101993:
## Summary

- added metadata to user schema and responses
- refactored user attributes a bit
- added cursor pagination to list users

## Validation

- `go test ./...`

## Release notes / changeset

- Changeset: .changeset/user-list-cursor-pagination-and-metadata.md

## Notes

-


internal/domain/attribute.go
```diff
@@ -47,57 +141,41 @@ func NewCreateAttribute(key string, value any, unique AttributeUniqueness) (*Cre
...
-		tree[keyNodes[0]] = value
-		return
+// Get returns the value for the given key.
+//
+// TODO(go v27): make this method generic
+func (attrs CreateAttributes) Get(key AttributeKey) (value *CreateAttribute, ok bool) {
+	for _, attr := range attrs {
+		if attr.Key == key {
+			return new(attr), true
+		}
```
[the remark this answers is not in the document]
wim07101993: not true

internal/api/user.go
```diff
@@ -170,17 +171,37 @@ func (h *Handler) GetMyUser(ctx context.Context) (api.GetMyUserRes, error) {
...
+func domainUserToApiUser(user *domain.User) (*api.User, error) {
+	userData, err := user.Attributes.ToMap()
+	if err != nil {
+		return nil, domain.ErrInternal(err).WithMessage("failed to parse user attributes")
+	}
+
+	props, err := convertUsingJson[api.UserAdditional](userData)
 	if err != nil {
 		return nil, err
 	}
```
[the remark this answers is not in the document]
wim07101993: this will only be true while old data is still in circulation.

bastionstack:
`full-pr` is failing on `testing:test-integration` — `tests/integration/kit.test.ts > startLocalZitadel (real instance)`.

Cause is the new list shape. `packages/testing/tests/integration/kit.test.ts:31` still treats the response as an array:

```ts
const users = await zitadel.api.listUsers();
const ids = users.map((user) => user.id);
```

`listUsers` now resolves to `{ users, next_page_token }`, so `.map` is not a function. One line:

```ts
const { users } = await zitadel.api.listUsers();
const ids = users.map((user) => user.id);
```

Same class of change hits the console — I've updated its list screen and specs on #726, which stacks on this branch and is waiting for it to land.

internal/service/flow_create_user_for_passkey_handler_test.go
```diff
@@ -144,14 +133,14 @@ func TestFlowCreateUserForPasskey_PersistsAllCollectedSchemaFields(t *testing.T)
 	require.NoError(t, err)
 	require.NotNil(t, captured)
 
-	emailAttr := attributeByKey(t, captured.Attributes, "email")
+	emailAttr, _ := captured.Attributes.Get("email")
```
grvijayan: `attributeByKey` (removed above) [called t.Fatalf on a missing attribute](https://github.com/zitadel/nextgen/pull/721/changes#diff-a1faf39fbc4901523487f0aa79e45935e19ce24633b862f5c3b18e9c80cd2aa7L101), which is being ignored by the replacement.

internal/domain/attribute.go
```diff
@@ -24,14 +36,96 @@ const (
...
+			// not a leaf node, traverse down the tree
+
+			var m map[string]any
+			v, ok := subTree[keyNodes[0]]
+			if !ok {
+				// nested map does not yet exist, create a new one
+				m = make(map[string]any)
+			} else if m, ok = v.(map[string]any); !ok {
+				// if the key overlaps with another value which is not an object, error to be sure
+				return nil, errors.New("the given key already exists in the map with a value which is not a map")
```
grvijayan: It would be helpful to include the `key` in the error message.

api/openapi/endpoints/users/user-metadata.yaml
grvijayan, line 1: just wondering if the "user-metadata" that's defined here could be conflated with the ones users define as metadata in their schemas 🤔

grvijayan (commented):
looks good; just minor comments

internal/service/flow_create_user_for_passkey_handler_test.go
wim07101993: good catch

api/openapi/endpoints/users/user-metadata.yaml
wim07101993: good point, I added validation checks in the domain layer

wim07101993: The reason I kept the metadata inside the object, is because frontend will probably want to sort or filter by fields inside the metadata like creation date or status.

grvijayan (approved):

PR #725, opened 2026-08-03
ci: shard full-pr into parallel resource jobs behind the aggregated gate

fforootd:
## Summary

Closes the wall-clock half of the CI cost discussion: `full-pr` spent ~8 of its 10–13 minutes in the serial `env -u CI` tail, whose tasks boot real instances, carry `cache: false`, and can never cache-skip — while the moon graph itself is already 0.6–0.8 min on the Depot remote cache. Parallelism was the only remaining lever, so the single job becomes shard jobs split by resource conflict:

- `graph` — PR-title check, CI-mode detection, `server:check-generate`, `moon ci :lint :typecheck :build :test :test-browser :check-adrs`, `server:test`, and the Postgres/Spanner integration suites (keeps the WIF `id-token` permission to itself).
- `snapshot` — `release:snapshot -- --skip-container` (or the version-PR validation path in version-only mode), uploading `dist/release` as the `release-snapshot` artifact.
- `journey-fresh-app` / `journey-presets` — fan out from `snapshot` via the artifact; the journey tasks dep only on `cli:build`/`testing:build`, so nothing rebuilds the release. `journey-presets` pairs the passkey-first preset with the test-kit consumer journey to balance the fresh-app shard (~2.7 vs ~2.5 min).
- `browser-e2e` — `testing:test-integration` + `demo-next-e2e:e2e-real`, then `console-e2e:e2e-real`; the console suite still runs after the demo suite on the same machine (embedded-Postgres contention), but no longer behind the journeys.
- `full-pr` — the unchanged required branch-protection context, now an aggregator: `scripts/verify-shard-results.mjs` fails it on any failed, cancelled, or wrongly-skipped shard, and on drift between the `needs:` list and the script's expectations (both directions, loud). Journey shards gate on the snapshot shard's detected mode, so version-only PRs skip them legitimately.

Expected critical path: setup + snapshot → fresh-app journey ≈ 5–6 min against the 10–13 min baseline. A flaky shard now fails alone and "re-run failed jobs" reruns only that shard. Fork PRs keep the GH-hosted-runner ternary on every shard; the aggregator runs on a plain runner and needs no install (script chain is node-builtins only).

Docs updated where the single-job layout was described (CONTRIBUTING `#what-ci-runs` as the canonical list — also fixing its stale claim that Spanner doesn't run in CI — plus pointer-level wording in README, AGENTS.md, and packages/testing/AGENTS.md).

## Validation

- `node --test scripts/verify-shard-results.test.mjs` — 9/9 pass.
- `moon run workspace:test` — the new task runs the same tests green through moon (it's matched by `:test` in the graph shard's `moon ci`).
- Fixture-ran `scripts/verify-shard-results.mjs` with full-green (exit 0) and failed/cancelled/skipped-journey fixtures (exit 1, named problems).
- YAML parse of `.github/workflows/ci.yml` and `moon.yml`.
- This PR's own CI run is the real proof: shard wall-clock vs the ~10–13 min baseline, and the `full-pr` aggregate going green. Timings will be posted in a comment after the run.
- Not run: a deliberately failing shard on this branch; the eject path is covered by the unit fixtures above.

## Release notes / changeset

- No changeset required — no shipped behavior changed (CI workflow, CI-glue scripts, and docs only).

## Notes

- The `merge_group` trigger, per-ref concurrency, and single required context are deliberately preserved: this is the groundwork for enabling the merge queue and dropping the strict up-to-date requirement afterwards.
- All shards stay on `depot-ubuntu-24.04-16`; right-sizing individual shards (the journeys likely don't need 16 cores) is a follow-up cost knob, kept out of this PR to avoid mixing a perf change into the split.
- Version-only PRs now spin up `graph`/`browser-e2e` runners that skip their steps (~2–3 wasted runner-minutes per changesets-PR update); acceptable until a cheap job-level gate exists that doesn't serialize the full-mode critical path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,23 +12,31 @@ concurrency:
   group: ci-${{ github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
+env:
+  # Moon marks failed tasks by color alone, so a failed `moon ci` can end at
+  # "Tasks: 1 failed" with no named target. The summary names them in plain
+  # text and replays each failed task's own output. Green runs pay only a
+  # pass/fail list; the replay section is empty when nothing failed.
+  MOON_SUMMARY: detailed
+
+# The work is split into shard jobs by resource conflict, not by category:
+# the moon graph and Go suites share one machine, the release snapshot feeds
+# the journey shards through an artifact, and the real-instance browser
+# suites get machines of their own so their embedded-Postgres instances
+# cannot contend with anything else. Branch protection requires the single
+# context `full-pr`; that job aggregates the shard results, so the required
+# check's name survives the split.
 jobs:
-  full-pr:
+  graph:
     # Fork PRs execute untrusted repository code. Keep them off Depot runners,
     # which inject DEPOT_CACHE_TOKEN independently of GitHub secret filtering.
     runs-on: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork && 'ubuntu-24.04' || 'depot-ubuntu-24.04-16' }}
-    timeout-minutes: 120
+    timeout-minutes: 60
     permissions:
       contents: read
       # Required to mint the OIDC token for Workload Identity Federation when
       # authenticating to the Spanner test instance (trusted runs only).
       id-token: write
-    env:
-      # Moon marks failed tasks by color alone, so a failed `moon ci` can end at
-      # "Tasks: 1 failed" with no named target. The summary names them in plain
-      # text and replays each failed task's own output. Green runs pay only a
-      # pass/fail list; the replay section is empty when nothing failed.
-      MOON_SUMMARY: detailed
     steps:
       - uses: actions/checkout@v6
         with:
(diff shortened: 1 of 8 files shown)


fforootd:
Measured timings from this PR's own run ([30807333125](https://github.com/zitadel/nextgen/actions/runs/30807333125)), all shards green:

| job | duration | window |
|---|---|---|
| `snapshot` | 2.8 min | t+0.0 → t+2.8 |
| `graph` | 1.7 min | t+0.0 → t+1.7 |
| `browser-e2e` | 2.8 min | t+0.0 → t+2.8 |
| `journey-fresh-app` | 3.6 min | t+2.8 → t+6.5 |
| `journey-presets` | 4.1 min | t+2.8 → t+7.0 |
| `full-pr` (aggregate) | 6 s | t+7.0 → t+7.1 |

**Wall clock: 7.1 min** against the 10–13 min single-job baseline. The critical path is `snapshot` → `journey-presets` → aggregate, as designed. Two calibration notes:

- This diff touches no product code, so `graph` ran cache-fast (1.7 min); on server/UI-touching PRs expect it nearer its ~3 min historical share — still off the critical path.
- `journey-presets` carries two journeys plus ~1.1 min of per-job setup, which is why it lands at 4.1 min rather than the naive 2.7. Splitting it into two shards would put the wall at ≈5.4 min for the cost of one more runner spin — left as a follow-up knob.

Also verified in the run: `workspace:test` was selected by `:test` and marked affected by `scripts/verify-shard-results.test.mjs` in the graph shard's `moon ci`, so the aggregator's judgment logic is executed CI-side, not just locally.

fforootd:
Closing unmerged. The measured shard win (7.1 min vs 10–13) is real, but review surfaced that it solves the problem at the Actions layer while Moon has the primitives to own the DAG itself (multi-target parallel graphs, task extends for arg-baked variants, mutex for real resource conflicts, deps for snapshot→journey ordering) — and the layer rule says Moon owns what/order/scope. Superseded by a moon-native single-graph redesign, planned on the assumption #665 (SQLite local default) lands and shrinks the instance-boot cost the shards were routing around. The measured timings in this PR remain the baseline reference.

PR #732, opened 2026-08-03
docs: affected-gate canary (do not merge)

fforootd:
Draft canary stacked on #731: the diff against its base is one docs file, so the gated lanes (go suites, snapshot, journeys, real-instance suites, browser install) should all skip and the run should finish in ~2–3 min. Will be closed after the timing is recorded.


fforootd:
Purpose served: the docs-only run skipped all 16 gated steps and finished in 57s (details on #731). Deleting the branch.

PR #733, opened 2026-08-03
test: port the team statement suites to stmttest

grvijayan:
## Summary

Follow-up to #672. #664 moved the statement
contract suites for projects, branding, users, passkeys, passwords, recovery
codes and TOTP into `internal/storage/v2/stmttest` per ADR 041, and left teams
and team memberships behind. This closes the teams half.

`Create`, `Get`, `UpdateTeam` and `Deactivate` were asserted twice, once per
dialect — They now run once through `service.AllStatements`, per dialect, via `forEachDialect`.

Rebased onto #665, which landed the SQLite dialect and its own
`stmttest/team_test.go` covering `GetByID` and `UpdateTeam`. This branch's suite
is a superset, so the merge keeps it and folds in what main had: the
`sqlite_integration` build tag, `ensureProject` in place of a local helper, and
main's `GetByID` naming and `UpdatedAt` assertion. All suites now run on three
dialects.

Changes:

1. **The port.** Four suites move to `stmttest/team_test.go`;
   `dialect/spanner/team_test.go` is deleted; `dialect/postgres/team_test.go`
   drops to the two tests that need pgx.
2. **The cascade.** `TestTeamStatements_Deactivate_CascadesMembershipsAndOwnedUsers`
   was postgres-only because it seeded fixtures with raw `testPool.pool.Exec`,
   not because the behavior is engine-specific. Spanner and SQLite run the same
   four DMLs and had **no coverage for them at all**. Rewritten against
   `AllStatements` — `CreateUser` carries `LifecycleOwnerTeamID`, `GetUser` and
   `GetTeamMembership` replace the raw reads — so all three cover it now.
3. **A SQLite schema fix.** Running the shared suite against SQLite surfaced a real gap in #665.
`teams.name` was declared plain `TEXT`, so SQLite accepted a 201 character name
that Postgres rejects via `CHECK (length(name) <= 200)` and Spanner via
`STRING(200)`. `000001_init.sql` gains the same constraint Postgres uses. SQLite cannot
`ALTER TABLE ADD CONSTRAINT`, and the dialect shipped with a single init
migration, so it goes there — **existing local `zitadel.db` files keep the old
schema until recreated**. Carries a patch changeset.

Three assertions tightened while porting, each verified on every engine:

- duplicate `(project_id, id)` expected a bare error; now `database.UniqueError`
- an over-length name now expects `database.CheckError`
- `empty id returns error` moved to the shared suite rather than staying
  postgres-local. Every `CreateTeam` implementation rejects it with the same Go
  guard, so it never reaches the postgres `CHECK (id <> '')` that the others lack

Staying dialect-local per ADR 041: `TestDeactivateTeam_rollsBackWhenSecondWriteFails`
(injects a failing `pgx.Tx`) and `team_tx_test.go`. Team memberships are still
duplicated and are the next PR.

## Validation

All three dialects, since a single-dialect run proves nothing for a parity change.

- `go test -tags postgres_integration ./internal/storage/...` — ok
- `go test -tags spanner_integration -parallel 1 -p 1 ./internal/storage/...` — ok
- `go test -tags sqlite_integration ./internal/storage/...` — ok
- Every team suite reports once per brought-up dialect
- `gofmt -l internal/` and `go vet` under each tag — clean

The new cascade coverage was mutation-checked rather than trusted for going
green. Breaking `deactivateTeamOwnedUsersStmt` fails on the user status
(`deactivated` vs `active`); breaking `deactivateOwnedUsersMembershipsStmt`
fails on the `otherTeamID` membership (`removed` vs `active`), the indirect path
where a membership in a *different* team is removed because the user itself was
deactivated. Both reverted.

The SQLite constraint was confirmed by watching the over-length case fail before
the fix and pass after, with the skip removed.

## Release notes / changeset

`.changeset/sqlite-team-name-length.md` — patch, for the SQLite schema fix. The
test port itself needs none.

## Notes

SQLite and Spanner have no counterpart to either Postgres transaction test —
nothing asserts that their `DeactivateTeam` routes the four writes through
`withTransaction` or that they roll back together. That gap predates this PR;
Spanner RW transactions cannot nest and the emulator is single-transaction, so
the pgx fault-injection trick has no direct analogue. Flagging, not fixing.


fforootd:
Besides the conflict looks good

Issue #527, opened 2026-07-14
Platform Project Bootstrapping


IAM-marco:
## Context

Zitadel cloud manages its own customers with Zitadel: one designated **platform project** is where the developers who use our cloud register, hold their personal teams, and (later) administer the projects they claim. Today nothing of this exists: no project is marked as "platform", there is no way for a developer to register an account on our cloud, and no team is created for them.

This epic is the **absolute blocker for [Claim: Ownership (#96)](https://github.com/zitadel/nextgen/issues/96)**: claiming is cloud-only and reduces to associating a project with the claiming user's personal team, so without a bootstrapped platform (users, teams, registration) there is no concept of claiming at all. It is equally foundational for any future cloud console and billing surface. See the [claim breakdown](https://github.com/zitadel/nextgen/issues/96#issuecomment-4923654787) (sections 3 and 6) for how claiming depends on this work.

## What is needed

1. **Platform project provisioning.** A designated platform project per environment: how it is created (config-driven ensure at startup, migration-seeded, or manually provisioned), how its ID reaches the server and the console (config), and what it is seeded with (user schema, login/registration flow definitions). A narrow, flag-gated slice of this (just ensuring a platform project row exists at backend startup, off by default) is filed as #605, split out so local development and testing of claim (#96) is not blocked on the rest of this epic.
2. **Registration and login on the platform project.** Developers sign up or sign in with **email only** for the first iteration (no SSO; IdP/SSO registration and management is its own epic, required by public launch). Designs exist (see the [Figma designs](https://www.figma.com/design/xkvBjkOJ8ENuHdTGZHXezK/Zitadel---UI-Design---External?node-id=6979-2914) linked from #96). The flow engine already supports password/passkey registration and session exchange; this epic wires it up for the platform project as a real, hosted surface.
3. **Automatic "Personal Team" creation at registration.** When a user registers **on the platform project**, a team named "Personal Team" is created automatically and the user is associated with it (membership). Constraints:
   - Must be **restricted to platform-project registrations**: users registering on customer projects must NOT get auto-created teams. This needs a mechanism to attach project-specific side effects to registration (hook / rule / action, or a platform-specific API layer wrapping registration).
   - Team names are not unique; renaming comes later with the console surface.
[...]

PR #736, opened 2026-08-03
feat(setup): ensure the platform project at startup behind a flag

### LINKED ISSUE #605: Platform project bootstrap: ensure a platform project exists at backend startup (flag-gated)
Part of #527.

## Goal

At server startup, when explicitly enabled via a flag, idempotently ensure a designated platform project row exists, so local development and testing of the claim flow (epic #96) has something to point at without waiting on #527's full scope (registration UI, auto team creation, infrastructure, deployment).

## Current state

There is no concept of a "platform project" anywhere in the server. `internal/bootstrap/users/ensure.go` already has the idempotent-insert pattern this ticket follows, but for a different purpose: it backfills `projects`, `teams`, and `json_schemas` rows so a bootstrap user import's foreign keys resolve (`ensureProject`, `ensureTeam`, `ensureJSONSchema`, all `INSERT ... ON CONFLICT DO NOTHING`). Nothing currently designates any project as "the platform project", and nothing runs this kind of ensure at normal server startup; today it only runs on the standalone user-bootstrap command path.

- Off by default. Only runs when a flag or config value is explicitly set, so no environment gets a platform project created silently.
- Config supplies the intended project ID.
- Scope is provisioning only: no registration or login flows, no team auto-creation hook, no infrastructure or deployment work. All of that remains #527's broader scope.

## Dependencies

- None. Foundation task, can start now.
- Unblocks Claim A2's dependents Claim E1 (#611) and Claim H1 (#615), which both need a real platform project to develop and test against.

## Test first (TDD)

Write a test first that runs the ensure logic twice against the same database and asserts: the project exists after the first run, and the second run is a no-op (no error, no duplicate row, same ID). Also test that the ensure logic does not run at all when the flag is unset. Then implement.

## Tasks

- Add a config field or flag (for example `PLATFORM_PROJECT_ID`, or a `platform.bootstrap` config block) that is empty or false by default.
- When set, at startup, idempotently ensure a project row exists with that ID, mirroring `ensureProject` in `internal/bootstrap/users/ensure.go` (`INSERT ... ON CONFLICT (id) DO NOTHING`).
- Wire the check into the server startup path (`cmd/`) so it runs once, after the database connection is established, before the server starts accepting requests.
- Log clearly whether the platform project is being bootstrapped, so it is obvious from server logs whether this is active.

## Reference

`internal/bootstrap/users/ensure.go` for the idempotent-insert pattern to mirror; `cmd/` for where server startup wiring lives today.

IAM-marco:
## Summary

Closes #605 (part of epic #527).

- New `internal/bootstrap/platform` package: `Ensure` idempotently creates the platform project row (`Name: "Platform"`) when explicitly enabled, treating a duplicate-key `*database.UniqueError` as the success path — mirroring `internal/bootstrap/users/ensure.go`. Disabled (the default) is a no-op with no statement calls.
- New config flag `platform.bootstrap_project` (default `false`, env `NEXTGEN_PLATFORM_BOOTSTRAP_PROJECT`): when true, the server ensures the project pinned by the existing `platform.project_id` exists at startup — after migrations, before user import and default-project resolution. Enabling it without a pinned id fails config validation.
- Console ADR 0004's "the server creates no project" fact now records this one explicit opt-in exception; standalone default semantics are unchanged.

## Validation

- `go build ./...`, `go vet ./internal/bootstrap/platform/ ./cmd/server/`
- `go test -count=1 -race ./internal/bootstrap/... ./cmd/server/...` (gomock unit tests: flag off = zero statement calls, twice-run idempotency, empty-id error, error propagation; config validation table)
- `go test -tags sqlite_integration ./internal/bootstrap/platform/` (runs `Ensure` twice against a real SQLite DB; row exists after run 1, unchanged after run 2)
- End-to-end against one SQLite database: run 1 with the flag logs "created platform project" and the pinned id resolves as default; run 2 logs "already exists"; a run without the flag emits no bootstrap lines; the flag without `platform.project_id` exits with `platform.bootstrap_project requires platform.project_id`. Exactly one `proj_platform | Platform` row in the DB.
- `moon run workspace:check -- --only release`

## Release notes / changeset

- Changeset: `.changeset/platform-project-bootstrap.md` — flag-gated platform project bootstrap at startup via `platform.bootstrap_project` (`@zitadel/server`, minor).

## Notes

- The bootstrap reuses the existing `CreateProject` statement, so no new stmttest suite is needed; the duplicate-id → `UniqueError` contract is already covered by `stmttest/project_test.go`.
- The row is a bare project (no keys, schemas, or flows), which is #605's provisioning-only scope; full platform-mode provisioning stays with #527.
- Config shape (bool + reuse of `platform.project_id`) was chosen over a separate bootstrap id to keep "the platform project" and "the pinned default" from diverging, and to preserve ADR 0004's fail-fast on a mistyped pin when the flag is off.


internal/bootstrap/platform/ensure_test.go
```diff
@@ -0,0 +1,121 @@
...
+			wantErr: storageErr,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			ctrl := gomock.NewController(t)
+			pool := servicemocks.NewMockStatementPool(ctrl)
+			stmts := servicemocks.NewMockAllStatements(ctrl)
```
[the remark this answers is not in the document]
IAM-marco: This is no longer valid in go 1.26

cmd/server/config.go
```diff
@@ -29,14 +30,28 @@ type PlatformConfig struct {
 	// ProjectID pins the deployment's default project to an existing
 	// project. When empty (the default), a standalone deployment tracks its
 	// first-created project — the one the customer's `zitadel setup`
-	// creates. The server never creates a project itself; a configured id
-	// that does not exist is a startup error.
+	// creates. The server never creates a project itself, unless
+	// BootstrapProject explicitly opts in (#605); a configured id that does
+	// not exist is otherwise a startup error.
 	ProjectID string `mapstructure:"project_id"`
```
vitorbari, line 38: Do we have to document and/or validate the format of the `project_id`?

IAM-marco: Good point, let me take a look

IAM-marco: Went with owning the id server-side instead of validating a hand-typed one.

The body doesn't have to be a UUID/ULID: dialects mint `proj_<ULID|UUID>` at insert time (#663), but ADR 047 §3.3 allows readable bodies for client-visible ids and the `projects` PK is a plain TEXT column, so `proj_platform` is a valid, non-colliding id.

So rather than ask operators to author a formatted id (and get the format wrong), the platform project is now a server-owned singleton with the built-in id `domain.PlatformProjectID` (`proj_platform`). `platform.bootstrap_project` needs no `platform.project_id`.

`platform.project_id` stays as the standalone pin to an existing project; when set it is now validated to carry the `proj_` prefix (via a new `ResourcePrefix.Matches` helper), and with bootstrap on it must match the built-in id. Documented on the config fields and in Console ADR 0004.

Pushed in 94696085 / e9b25f82 / 753bcd6e.

PR #737, opened 2026-08-03
feat(api): add team delete endpoint

### LINKED ISSUE #622: Team: Delete
## Context

Part of #598. Team CRUD is missing Delete. Per ADR 024's lifecycle matrix, deleting a team is a deactivate/tombstone operation, not a hard delete, and the full cascade (team, its memberships, its lifecycle-owned users, and those users' memberships) is already implemented in storage on both postgres and spanner. This ticket is wiring: service, API, and OpenAPI layers on top of that existing cascade.

## Scope / Tasks

- Add a service-level Delete operation that invokes the existing deactivation cascade.
- Add a `DELETE /teams/{team_id}` endpoint and its response schema, with `project_id` as a required query parameter (matching Get).
- Add a `team.delete` scope, separate from `team.write`, matching the pattern already used for flow definitions and sessions. Gate the endpoint on it.
- Return 404 for an unknown team id, matching the convention used for Update (#619).
- Return the deactivated team's state in the response.
- Treat deleting an already-deactivated team as idempotent, not an error.

## Acceptance Criteria

- [x] `DELETE /teams/{team_id}` deactivates the team and returns its updated state.
- ~~[ ] Unknown team id returns 404.~~ An unknown team id returns 204 as well. A 404 would let a credential
holding only `team.delete` enumerate team ids that `GET /team/{team_id}` would refuse
show it. Delete now answers the same whether or not the team exists.
- [x] Request without `team.delete` access to the project returns 403.
- [x] Deleting an already-deactivated team succeeds without error.
- [x] Team memberships and lifecycle-owned users are deactivated as part of the same request, matching ADR 024's cascade.

## Open Questions

- `domain.TeamStatus` includes `pending_purge`, allowed by the teams table's `CHECK` constraint, but no code path sets it today. Should this Delete cascade (or a follow-on hard-purge flow, per ADR 024's retention/audit/recovery policy) be what wires it up, or should the unused status be removed instead?

## Out of scope

- Revoking team-scoped API keys as part of the cascade. ADR 024 calls this out as canonical delete behavior, but no team-scoped API key resource exists in the codebase yet.

## Resources

- [ADR 024: User, Team, and Lifecycle Ownership](https://github.com/zitadel/nextgen/blob/main/docs/adrs/024-user-team-lifecycle-ownership.md)

grvijayan:
Closes #622.

## Summary

Adds `DELETE /teams/{team_id}`, the last missing team CRUD operation.

Per ADR 024's lifecycle matrix, deleting a team is a deactivation rather than an
erasure. The team row survives with `status = deactivated`, stays readable
through `getTeam`, and its memberships plus lifecycle-owned users deactivate
with it. The cascade already existed in storage for all three dialects and is
transactional; this PR wires the service, authz, handler, and contract on top of
it, and makes the cascade itself idempotent.

The endpoint answers 204 and is idempotent. `DeactivateTeam` now guards its team
`UPDATE` on `status = 'active'` and skips the rest of the cascade when that
affects no rows, so `updated_at` records when the team was deactivated rather
than when delete was last called. The guard is what makes this hold under
concurrency: a racing call re-evaluates the predicate after the winner commits
and matches nothing. An earlier revision did the check in the service between a
read and the write, which two concurrent deletes could both pass.

### Deviation from the acceptance criteria

The issue asks for 404 on an unknown team id, "matching Update conventions".
This PR answers **204** instead, deliberately.

That convention does not transfer. `team.write` is in the read scope list, so
`updateTeam`'s 404 tells a caller nothing they could not already learn from
`getTeam`. `team.delete` grants no read, so a 404 here would let a credential
holding only `team.delete` enumerate team ids that `getTeam` would refuse to
show it. Delete now answers the same whether or not the team exists.

The endpoint therefore declares no 404 at all. Its failure modes are 400 (the
id fails the contract's pattern), 401, and 403. Since an unknown team answers
204, the only thing left that can fail is the project-binding check in
`requireProjectAccess`, which compares the `project_id` parameter against the
token's project without querying the database — so `requireTeamDelete` refuses
it as the permission failure it is. A project the credentials are not bound to
answers the same whether or not it exists.

The issue's other open question, `pending_purge`, is left as is: delete writes
`deactivated` only, and the unused status stays as the placeholder for a future
hard-purge flow.

## Validation

```sh
go build ./... && go test ./internal/... -count=1                # pass
go test -tags sqlite_integration -run TestTeamStatements \
  -count=1 ./internal/storage/v2/stmttest/                        # pass
go test -tags postgres_integration -run Team -count=1 \
  ./internal/api/integration_test/ ./internal/storage/v2/stmttest/ \
  ./internal/storage/v2/dialect/postgres/                         # pass
moon run server:generate && moon run server:check-generate        # clean
gofmt -l internal/ api/generated/                                 # empty
corepack pnpm exec changeset status --since origin/main           # @zitadel/server minor
node scripts/check-pr-title.mjs --title "feat(api): add team delete endpoint"
```

## Release notes / changeset

Real changeset added: `.changeset/add-team-delete-endpoint.md`, bumping
`@zitadel/server` at minor. A new endpoint is shipped product behavior.

## Notes
- Open question, deliberately not answered here: `requireProjectAccess` answers
  a project-binding failure with the resource's not-found code, justified by
  matching an in-project miss. An idempotent delete does not satisfy that
  premise, so `deleteTeam` refuses with `team.permission_denied` through
  `requireTeamDelete` rather than borrowing the read's 404.
  `deleteFlowDefinition` is in the same position — idempotent by its own
  contract, and declaring neither 403 nor 404 today — and the argument extends
  to `DeleteUserByID` and `revokeSession`. Whether the rule should move into
  `requireProjectAccess` for every resource the guard serves is worth deciding,
  but not in a team PR. The guard never queries the database, so this is a
  semantics question rather than a disclosure one.
- Team-scoped API key revocation stays out of scope, as the issue states, since
  no such resource exists yet.


internal/api/integration_test/team_test.go
```diff
@@ -523,6 +523,121 @@ func TestUpdateTeam(t *testing.T) {
...
+		deleteTeam(t, "does-not-exist")
+	})
+
+	// The cascade itself is proven per dialect by the storage suite; this only
+	// shows the endpoint reaches it rather than writing the team status alone.
+	t.Run("cascades to lifecycle-owned users", func(t *testing.T) {
+		t.Parallel()
+
+		harness.CreateUserSchema(t, project, harness.EnsureTestData(t).Schemas.CreateSchemaRequestUserSchema)
+		userSchemaURL := "https://raw.githubusercontent.com/zitadel/nextgen/refs/heads/main/api/openapi/endpoints/schemas/examples/user-schema-example.yaml"
```
wim07101993: Referencing the github repo adds a dependency which is not needed. You can use the `test_data` package to fetch a valid schema and generate users for the default schemas

internal/service/team.go
```diff
@@ -211,3 +211,16 @@ func (s *TeamService) Update(ctx context.Context, input UpdateTeamInput) (*domai
 	}
 	return team, nil
 }
+
+// Delete deactivates the team and cascades to its memberships and the users
```
wim07101993, line 215: How does the cascade of status happen? I don't see anything about it in code.

grvijayan: the cascade is handled in the storage layer.
this is in postgres: https://github.com/zitadel/nextgen/blob/8859fc77663cc66e6326b05e4f7e195278503691/internal/storage/v2/dialect/postgres/team.go#L149-L165

wim07101993 (approved):

PR #738, opened 2026-08-03
refactor: drive the account and pill gradients from Figma tokens

bastionstack:
# Which Problems Are Solved

The console account swatch and the `<zl-pill>` sheen hand-copy the hex values of Figma's `Gradient Colors` collection, so they cannot follow the theme and drift whenever the collection changes.

# How the Problems Are Solved

- `AppShell.tsx` and `pill.css` reference `--zl-gradient-red-start`, `--zl-gradient-neutral-start` and `--zl-gradient-base-end` instead of literals.
- `pill.css` applies the design's 0.6 alpha with `color-mix`, since `rgb()` cannot take a hex custom property. First use of `color-mix` in this repo.

# Additional Changes

None. `layout-chrome.css` keeps its `--zl-color-icon-default-purple` / `-pink` references: the gradient collection has no equivalent pair, so swapping would change colours rather than tokenise them.

# Additional Context

Dark mode is unchanged — the resolved stops are byte-identical to the literals they replace (`#f25543` / `#0f0f11`, and `rgb(72 74 87 / .6)` / `rgb(15 15 17 / .6)`).

Light mode needs a design decision before this can merge. `--zl-gradient-base-end` is `#fafafa` in light, which is exactly `--zl-sidebar` (`#fafafa`), so the account swatch dissolves into the sidebar at its 75% stop. The pill sheen ends on `#fafafa` at 60% over a `#ffffff` pill surface, which is effectively invisible. Those light values read as a page-background fade for a large surface, not stops for a small chip that has to hold an edge.

Options: keep both second stops dark in all themes and leave them as literals, pair `--zl-gradient-red-start` with a stop that retains contrast in light, or accept the fade as intended.

Depends on #727, which is merged.

### DIFF
--- a/apps/console/src/components/app-shell/AppShell.tsx
+++ b/apps/console/src/components/app-shell/AppShell.tsx
@@ -185,7 +185,7 @@ function UserMenuItem({ user, onSignOut }: { user?: ShellUser; onSignOut?: () =>
           >
             <span
               aria-hidden
-              className="size-8 shrink-0 rounded-lg bg-[linear-gradient(232deg,#f25543_17%,#0f0f11_75%)]"
+              className="size-8 shrink-0 rounded-lg bg-[linear-gradient(232deg,var(--zl-gradient-red-start)_17%,var(--zl-gradient-base-end)_75%)]"
             />
             <span className="flex min-w-0 flex-1 flex-col gap-0.5 leading-none">
               <span className="truncate text-[14px] text-sidebar-foreground">{displayName}</span>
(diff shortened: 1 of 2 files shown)


bastionstack:
Closing — the premise was wrong.

`Gradient/Red` is a Figma **style**, not a pair of stops:

```
linear-gradient(232.14259181541067deg, rgb(242, 85, 67) 17.263%, rgb(15, 15, 17) 74.94%)
```

The hardcode in `AppShell.tsx` is already a faithful (rounded) transcription of it. `#0f0f11` is a fixed stop inside the style, not the theme-responsive `base/end` this PR substituted — which is why light mode dissolved into the sidebar.

The `Gradient Colors` variables are the colour ingredients designers compose styles from, not per-call-site tokens. Styles don't export, so there is currently no correct token for these sites. Tracked in #741.

Issue #615, opened 2026-07-23
Claim H1: claim page in console


IAM-marco:
Part of #96.

## Goal

Add a route in the console that drives email registration or login against the platform project and completes the claim.

## Current state

`apps/console/src/routes/` has one directory per resource (`projects`, `users`, `sessions`, `schemas`, `flow-definitions`, `system`); there is no `claim` route yet. Visual design for this page is being refined separately.

## Decisions that shape this work

- Reuse the existing login widgets (flow API and LiquidJS), customized via their variables for the claim context, wrapped in custom React around them, per the UI meeting's decision.
- The claim step itself is a plain `claim/complete` API call after authentication, cookie-authenticated, not itself a flow, possibly with no UI element of its own.
- Building this page does not need #527's full scope, only PB0 (#605) to have something to develop and test locally against. Real production usage additionally needs #527's deployment and infrastructure work.

## Dependencies

- Claim A2 (#609). The contract this calls.
- Claim E2 (#612). Needed to exercise the real `claim/complete` path, not just the api-mock.
- PB0 (#605). A platform project to develop and test against.

## Test first (TDD)

Write component and route tests first against the api-mock: an unauthenticated visit prompts login or registration, a signed-in visit calls `claim/complete` and shows success, and a 409 or 410 response from the API surfaces a clear error state. Then implement.

## Tasks

- Add a new route under `apps/console/src/routes/claim/` (or wherever the claim URL's project ID parameter routes to), following the directory structure of an existing route such as `apps/console/src/routes/projects/`.
- Embed or link to the existing login widgets for registration or login against the platform project, customized via their variables for the claim context.
- On successful authentication, call `claim/complete` and render a success state; surface 409 (already claimed, show the team) and 410 (expired, prompt to restart) distinctly.

## Reference

`apps/console/src/routes/projects/` (an existing route's structure to mirror); the existing login widget components, wherever they are consumed today in the console or login-ui apps.

## Current state update (2026-08-20)

All three dependencies are now satisfied: E2 (#612) merged 2026-08-20 via PR #912, so the three claim endpoints run on a real server and this page can be verified end to end, not just against the api-mock. Two things make the implementation path concrete:

- `claim/init` already returns `claim_url = <console base URL>/claim?challenge_id=...&project_id=...` (`internal/service/claim.go`), so the route this ticket adds is the URL the CLI opens today.
[...]

Issue #613, opened 2026-07-23
Claim G1: CLI zitadel claim command


IAM-marco:
Part of #96.

## Goal

Add the `zitadel claim` CLI command: call `claim/init`, open the browser, poll until completion, and persist the result into `.zitadel/secret`.

## Current state

`apps/cli/src/lib/project.ts` has `ZitadelSecret` (`{project_id, project_secret, preview_secret, preview_origins, created_at}`) and `readZitadelSecret`, but no write function yet; a write helper needs to be added. `apps/cli/src/lib/server.ts` already has the shared `resolveServer` helper (precedence: `--server` flag, `ZITADEL_API_BASE`, environment config, default), including a `local` value resolved by `apps/cli/src/lib/local-server/runtime.ts`'s `resolveLocalServer`. `setup` already uses this (see `apps/cli/src/commands/setup/index.ts`'s `--server local` handling). `zitadel claim` should reuse the same helper rather than inventing its own server-targeting logic.

## Decisions that shape this work

- Support the same `--server` flag and `local` value as `setup`, via `resolveServer`, so the command works against a local backend and console out of the box.
- Friendly, distinct CLI output for 409 (already claimed, print the team) and 410 (expired, tell the user to re-run).
- Buildable and largely finishable against `packages/api-mock` as soon as A2 lands; full end-to-end verification against the real server needs E2.

## Dependencies

- Claim A2 (#609). The contract this calls.
- Its full end-to-end verification needs Claim E2 (#612), but implementation can start once A2 lands.

## Test first (TDD)

Write CLI command tests first against the api-mock, following how `setup`'s tests are structured, covering: a successful claim (secret file updated with `claimed_at` and `team_id`), 409 handling, 410 handling, and the `--server local` path. Then implement.

## Tasks

- Add a write helper to `apps/cli/src/lib/project.ts`, extending `ZitadelSecret` with `claimed_at` and `team_id`, writing the file back atomically (write-temp-then-rename) and preserving the existing fields.
- Add `apps/cli/src/commands/claim.ts`: read the current secret via `readZitadelSecret`, resolve the server via `resolveServer` (the same as `setup`), call `claim/init`, open `claim_url` in the browser, poll `claim/status` with backoff until completed, 410, or a timeout, then call the new write helper.
- Handle 409 (print the existing team, exit cleanly) and 410 (tell the user to re-run `zitadel claim`) with clear messages.
- Register the command with the CLI's command loader, next to `setup` and `status`.

## Reference

[...]

Issue #611, opened 2026-07-23
Claim E1: claim service (init/status/complete)


IAM-marco:
Part of #96.

## Goal

Implement the claim service backing `claim/init`, `claim/status`, and `claim/complete`, wiring the project-secret verification, the new B2 statements, the C2 session check, and the permission-management grant write into one coherent flow.

## Current state

`internal/api/handler.go`'s `Handler` struct composes one service per resource (`projectService`, `teamService`, `sessionService`, `tokenService`, and so on), constructed in `NewHandler`. Project secrets are not stored server-side as plaintext or hashes: `Project.ProjectSecret(encrypter)` and `PreviewSecret(encrypter)` mint JWE tokens carrying `{ProjectID, Scope}` on demand, and `TokenService.VerifyToken` decrypts a presented bearer back into a `*domain.Token`. This is what `SecurityHandler.HandleOAuth2` already does for every project-secret-authenticated endpoint. There is no `claimService` yet.

## Decisions that shape this work

- `claim/init`: authenticate via the existing project-secret flow (the same as any other project-secret endpoint), create a challenge (via B2), hashing the presented bearer to store as `initiating_secret_hash`. 409 if a grant already exists for the project.
- `claim/status`: proof of possession, hash the presented bearer and compare it to the stored hash on the challenge; a mismatch is 403. 410 once expired.
- `claim/complete`: authenticate via C2 (session cookie), resolve the personal team via B2's new statement, write the grant via #419's integration point, mark the challenge completed, all in one transaction.
- `team_memberships` already exists (#287), so this does not need #527's full scope, only a bootstrapped platform project (PB0, #605) to resolve a team against. Until #527's registration and auto-team-creation lands in production, tests and local development can manually seed a personal team and membership for test users against the project PB0 bootstraps.

## Dependencies

- Claim A2 (#609). The contract this implements.
- Claim B2 (#610). Challenge statements and the personal-team lookup.
- Claim C2 (#608). Platform-session verification.
- #419 (permission management). Writes the grant; without it there is nothing to write the association into.
- PB0 (#605). A platform project must exist to resolve a team against.
- Unblocks Claim E2 (#612).

## Test first (TDD)

[...]

Issue #96, opened 2026-05-18
Claim: Ownership


elinashoko:
Clarifying comment: https://github.com/zitadel/nextgen/issues/96#issuecomment-4923654787

---

# TL;DR

A developer can start using Zitadel immediately without creating an account.

When the developer runs the command, Zitadel creates an unclaimed project automatically. The project works immediately: developers can authenticate users, test flows locally, and share preview deployments without first creating accounts or teams. 

Unclaimed projects are intentionally temporary. They are designed for experimentation, fast prototyping, and evaluation rather than long-term ownership.

Once the project becomes meaningful - shared with teammates or moves closer to expiry - the developer claims it. Claiming attaches the Project to a Team (an automatically created personal team) and grants administrative access without breaking the existing setup.

Nothing about the application itself should need to change during claim. The issuer remains stable, users continue working, passkeys remain valid, and the SDK config does not need to be rewritten.

The core principle is simple:

> Start building first. Attach ownership later.

---

## Goals

- Let developers start using Zitadel immediately without signup friction.
- Preserve continuity when moving from experimentation to ownership.
- Projects should be creatable without requiring immediate human ownership.
- Introduce ownership only once the project becomes valuable or collaborative.
- Keep the lifecycle simple and understandable for both humans and AI-assisted workflows.

---

<details><summary><h1>Ownership & Lifecycle</h1></summary>

This section explains the core concepts behind the claiming model and how projects evolve from experimentation to ownership.

## Lifecycle

Lifecycle describes whether a project has ownership attached to it.

| Lifecycle | Behaviour |
|---|---|
| Unclaimed | Temporary project intended for experimentation and preview sharing |
| Claimed | Persistent project attached to a human and organization |

### Unclaimed projects

Unclaimed projects are designed to optimise for speed and low friction. A developer should be able to start instantly, test authentication flows, share previews, evaluate the product, and experiment with AI-generated apps without first creating accounts, resources, or billing relationships.

At the same time, unclaimed projects intentionally come with weaker guarantees. They are temporary by design and are not intended for long-term ownership, collaboration, recovery, or production usage.

The experience should consistently communicate a simple idea:

> This works immediately, but it is temporary until claimed.

### Claimed projects

Claimed projects are persistent and owned. Once a Project is claimed, it becomes attached to an owning Team and gains long-term persistence, collaboration capabilities, recovery options, and access to production-ready workflows.

[...]

PR #739, opened 2026-08-03
feat(api): add claim lifecycle endpoints to the OpenAPI contract

### LINKED ISSUE #609: Claim A2: OpenAPI contract for claim/init, claim/status, claim/complete
Part of #96.

## Goal

Add the three claim endpoints to the OpenAPI spec and regenerate the ogen client and server code, following the same per-operation file layout the Project API contract established.

## Current state

`api/openapi/endpoints/projects/` already has the layout to mirror: a top-level `methods.yaml` plus per-operation request and response schema files, and sub-directories per path segment (`by_id/` for `PATCH /projects/{id}`, `query/` for `GET /projects`). `docs/design/platform/api/claim-api.yaml` is the original design draft for `claim/init`, `claim/status`, `claim/complete`, plus `verified_identity` and `team_choice` fields the epic has since dropped: session auth replaces `verified_identity`, and the personal team already exists by the time claim runs, so there is no `team_choice`.

- Endpoints: `POST /projects/{id}/claim/init`, `GET /projects/{id}/claim/status`, `POST /projects/{id}/claim/complete`.
- `claim/init` and `claim/status` use the existing bearer, project-secret security scheme. `claim/complete` uses the existing `nextgenSession` cookie scheme (see Claim C2, #608).
- No `verified_identity` object, no `team_choice`. Error responses follow ADR 030's canonical codes.
- What each endpoint does:
  - `claim/init`: called by the CLI with the project secret. Creates a challenge, returns `claim_url`, `challenge_id`, `expires_at` (10-minute TTL). 409 if already claimed.
  - `claim/status`: polled by the CLI with the same project secret that called init. Reports pending or completed, with `team_id`, `claimed_at`, `dashboard_url` on completion. 403 on a bearer mismatch, 410 once expired.
  - `claim/complete`: called by the browser, cookie-authenticated. Body carries `challenge_id`. A minimal success response is enough; the browser renders its own success state.

## Dependencies

- Claim A1 (#606). The ADR settles the exact semantics this contract encodes.
- Unblocks Claim E1 (#611), Claim G1 (#613), and Claim H1 (#615), all of which consume the generated types.

## Test first (TDD)

Add OpenAPI conformance fixtures for the new paths and schemas before, or while, editing the YAML, following the same pattern used for the Project API's fixtures.

## Tasks

- Create `api/openapi/endpoints/projects/by_id/claim/` with a `methods.yaml` for the three operations and one request or response schema file per operation, mirroring the `by_id/` and `query/` sibling directories.
- Define the response schemas: init returns `{ claim_url, challenge_id, expires_at }`; status returns `{ status: pending | completed, team_id?, claimed_at?, dashboard_url? }`; complete returns a minimal success response.
- Define the 409 (already claimed, with `team_id` and `dashboard_url`), 403, and 410 error responses using ADR 030's error schema.
- Register the new paths in `openapi/openapi-spec.yaml`, wiring `claim/init` and `claim/status` to the existing bearer security scheme and `claim/complete` to the existing `nextgenSession` scheme.
- Run `go generate ./...` inside `api/` so `api/generated/` has the new types.

## Reference

`api/openapi/endpoints/projects/` (especially `by_id/methods.yaml` and `query/methods.yaml` for the per-operation-file convention), `docs/design/platform/api/claim-api.yaml` (the original draft, useful for field names, simplified per the decisions above), `api/openapi/openapi-spec.yaml` (where to register the paths).

IAM-marco:
Closes #609. Part of #96.

## Summary

Adds the three claim endpoints settled by ADR 046 to the OpenAPI contract and regenerates the ogen server/client code:

- `POST /projects/{project_id}/claim/init` — CLI, project secret (`oauth2: [project.write]`); 201 with `claim_url` / `challenge_id` / `expires_at`; 409 `proj.already_claimed` with `team_id` + `dashboard_url` in `details`.
- `GET /projects/{project_id}/claim/status?challenge_id=...` — CLI poll, same secret that initiated; 200 `pending`/`completed` (with `team_id`, `claimed_at`, `dashboard_url`); 403 `proj.permission_denied` on a foreign secret; 410 `proj.claim_expired`.
- `POST /projects/{project_id}/claim/complete` — browser, `nextgenSession` cookie; body `{ challenge_id }`; minimal 200 with `project_id`, `team_id`, `claimed_at`.

Endpoint files follow the per-path directory convention (`endpoints/projects/by_id/claim/{init,status,complete}/methods.yaml` + sibling schemas, mirroring `users/by_id/passkeys/`). The `proj.already_claimed` and `proj.claim_expired` codes are new domain sentinels in `internal/domain/project.go`, so their const-code schemas under `components/schemas/errors/` are generated per ADR 030.

Contract only: no service implementation (that is Claim E1, #611), so the generated operations respond `501` via `api.UnimplementedHandler`. Unblocks #611, #613, #615.

## Validation

- New "claim lifecycle" block in `packages/api-mock/src/spec-conformance.spec.ts` (11 tests, written first per the ticket's TDD note) exercises all three endpoints against the Express mock, validating bodies with the orval-generated zod schemas: init/status/complete happy path incl. session-cookie leg, 409/403/410/401/404/400 envelopes. `moon run api-mock:test`: 53 passed.
- `moon run server:check-generate` — generated output committed in sync.
- `moon ci :lint :typecheck :build :test` — 22 tasks, 0 failures.

## Release notes / changeset

`.changeset/claim-openapi-contract.md`: `@zitadel/server` minor, `@zitadel/api` minor. `@zitadel/api-mock` is private and in the changeset ignore list.

## Notes

- The 409 response layers a typed `details` object (`team_id`, `dashboard_url`) over the generated `proj-already_claimed` schema via `allOf`; ogen merges it cleanly.
- `status` completion fields are optional-and-absent while pending (not nullable), avoiding the ogen `type: [X, 'null']` limitation documented in `api/openapi/readme.md`.
- `claim/complete` is implemented in the mock as an Express route (it needs the module-scoped session store), mirroring the `sessions/me` cookie validation.


packages/api-mock/src/platform-handlers.ts
```diff
@@ -405,6 +510,106 @@ export function setupPlatformHandlers() {
...
+        return HttpResponse.json(
+          errorBody("proj.already_claimed", "the project is already claimed by a team", {
+            team_id: claim.teamId,
+            dashboard_url: claim.dashboardUrl,
+          }),
+          { status: 409 },
+        );
+      }
+
+      const challengeId = shortId();
```
fforootd: Could we reconcile this generator with the current `main` challenge-ID contract before merging? `main` now requires `^ch_[a-zA-Z0-9_-]+$`, while `shortId()` returns 12 unprefixed hex characters. I merged this head with current `main` (`de02bfcc`) and `moon run api-mock:test` dropped from 53/53 to 47/53: six claim tests return 400. Since this value authorizes a project claim, could we also mint the 128 bits of cryptographic randomness required by ADR 046 rather than only prefixing the current 48-bit value? The existing `full-pr` ran before #663 landed, which explains why it stayed green.

packages/api-mock/src/server.ts
```diff
@@ -317,6 +321,36 @@ export function createMockApp(options: { issuer: string }): express.Express {
...
+  // with the module-scoped __nextgen_session cookie, exactly like GET
+  // /sessions/me. The challenge_id from the body is its browser-safe authorization.
+  app.post(
+    "/projects/:project_id/claim/complete",
+    jsonBodyParser,
+    (req: express.Request, res: express.Response) => {
+      const token = (req.cookies as Record<string, string>).__nextgen_session;
+      const session = token ? sessionStore.get(token) : undefined;
+      if (!token || !session || new Date(session.expires_at) < new Date()) {
+        res.status(401).json(errorBody("auth.unauthorized", "missing or invalid session token"));
```
fforootd, line 345: Could we model ADR 046’s session-eligibility boundary here rather than treating any unexpired mock session as sufficient? This currently accepts sessions from any project and never requires a verified factor; the conformance helper creates a customer-project session with `factors: []`. That means the advertised full flow proves a session the real endpoint must reject. An explicit platform-project, active-session, verified-factor fixture/check would keep the downstream console and CLI tests honest.

packages/api-mock/src/platform-handlers.ts
```diff
@@ -303,12 +348,72 @@ export function snapshotPlatformStore(): PlatformStoreSnapshot {
...
+
+export function completeClaimChallenge(
+  challengeId: string,
+): { status: number; body: CompleteClaim200 | ErrorBody } {
+  const challenge = store.claimChallenges.get(challengeId);
+  if (!challenge) {
+    return { status: 404, body: errorBody("not_found", "claim challenge not found") };
+  }
+  if (
+    challenge.status !== "completed" &&
```
fforootd: Could we reject a challenge once its status is `completed` and enforce the TTL independently of status? In the current shape, a completed challenge skips expiry, reuses the existing claim, and returns 200 on every subsequent completion; the status endpoint likewise remains readable forever. ADR 046 describes completion as single-use and says expired challenges return 410. Tests for a second completion and for a completed-but-expired challenge would make that security property explicit.

internal/domain/project.go
```diff
@@ -29,6 +29,14 @@ func ErrProjectPermissionDenied() Error {
 	return newError(PrefixProject.ErrorCodePrefix("permission_denied"), "the project management API requires the project secret", nil, nil)
 }
 
+func ErrProjectAlreadyClaimed() Error {
```
fforootd, line 32: Could we wire both new public sentinels into `projectErrorResponse` in this PR? ADR 030 requires each stable public code to have a resource-mapper entry, but the current project mapper only handles not-found, permission, name, and missing-ID errors. A future #611 handler returning these values would therefore produce 500 instead of the documented 409/410. Adding the two mappings and focused tests here would make the generated contract genuinely ready for that follow-up.

api/openapi/endpoints/projects/by_id/claim/status/claim-status-response.yaml
```diff
@@ -0,0 +1,23 @@
...
+required:
+  - status
+properties:
+  status:
+    type: string
+    enum:
+      - pending
+      - completed
+    description: Whether the claim has completed.
+  team_id:
```
fforootd: Could we express this as pending/completed variants so generated clients can rely on the completion payload? With only `status` required, `{ "status": "completed" }` is contract-valid and a pending response may also carry stale completion fields. A `oneOf` with `status: const` variants can keep these fields absent while pending and require `team_id`, `claimed_at`, and `dashboard_url` when completed, without making anything nullable.

fforootd (commented):
Thanks for laying down this contract and adding end-to-end mock coverage. I found a few additional places where tightening the contract and mock now should keep the follow-up server, CLI, and console work on a reliable path. I’ve kept these separate from the three existing Copilot threads.

api/openapi/endpoints/projects/by_id/claim/init/methods.yaml
```diff
@@ -0,0 +1,53 @@
+post:
+  operationId: initClaim
+  summary: Initialize a project claim
+  description: |
+    Starts a claim challenge for an unclaimed project. Authenticated by the
+    project secret, this mints a single-use, 10-minute challenge and returns
```
vitorbari: 💅 
better just to say it is an expirable challenge instead of 10-minutes. If we ever change the duration, the docs could become obsolete.

api/openapi/endpoints/projects/by_id/claim/status/methods.yaml
```diff
@@ -0,0 +1,65 @@
...
+      description: The challenge minted by the matching claim init.
+      schema:
+        $ref: ../../../../../components/schemas/challenge-id.yaml
+  security:
+    - oauth2:
+        # project.write, not project.read: project.read is carried by the
+        # browser-plane preview secret, so it cannot gate an operator read
+        # until ADR 036 splits the credential planes into distinct schemes.
+        - project.write
+  responses:
```
vitorbari, line 29: For pooling endpoints, it is important to provide a HTTP status code 429 Too Many Requests

api/openapi/endpoints/projects/by_id/claim/complete/methods.yaml
```diff
@@ -0,0 +1,61 @@
+post:
+  operationId: completeClaim
+  summary: Complete a project claim
+  description: |
+    Called by the browser after the developer authenticates on the claim page.
+    Authenticated by the `__nextgen_session` cookie, it attaches the project to
+    the developer's personal team using the `challenge_id` from the claim URL as
```
vitorbari, line 7: Would it be possible to receive a `team_id` in this request in case the user wants to pick a different team?

packages/api-mock/src/platform-handlers.ts
```diff
@@ -405,6 +510,106 @@ export function setupPlatformHandlers() {
...
+      const authed = [...store.projects.values()].find((p) => p.projectSecret === token);
+      if (!authed) {
+        return HttpResponse.json(errorBody("auth.unauthorized", "missing or invalid credentials"), {
+          status: 401,
+        });
+      }
+      const project = store.projects.get(path.data.project_id);
+      if (!project) {
+        return HttpResponse.json(errorBody("proj.not_found", "project not found"), { status: 404 });
+      }
```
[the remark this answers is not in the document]
IAM-marco: Fixed in f2879fd6. \`claim/init\` now requires the bearer to be the path project's own secret: a secret belonging to a different project resolves to 404 rather than authorizing the claim. Added a conformance test (`init with another project's secret returns 404`).

packages/api-mock/src/platform-handlers.ts
```diff
@@ -405,6 +510,106 @@ export function setupPlatformHandlers() {
...
+        return HttpResponse.json(errorBody("not_found", "claim challenge not found"), {
+          status: 404,
+        });
+      }
+      if (challenge.initiatingSecret !== bearerToken(request)) {
+        return HttpResponse.json(
+          errorBody("proj.permission_denied", "the presented project secret did not initiate this challenge"),
+          { status: 403 },
+        );
+      }
```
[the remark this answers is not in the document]
IAM-marco: Fixed in f2879fd6. Added a 401 gate before the 403: a missing, empty, or unknown bearer is now 401 `auth.unauthorized`, and only a valid project secret that did not initiate the challenge returns 403 `proj.permission_denied`. Added a `status without a bearer token returns 401` test.

packages/api-mock/src/server.ts
```diff
@@ -317,6 +321,36 @@ export function createMockApp(options: { issuer: string }): express.Express {
...
+          .status(400)
+          .json(
+            errorBody("invalid_request", "request does not conform to spec", {
+              issues: parsed.error.issues,
+            }),
+          );
+        return;
+      }
+      const result = completeClaimChallenge(parsed.data.challenge_id);
+      res.status(result.status).json(result.body);
```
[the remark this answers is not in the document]
IAM-marco: Fixed in f2879fd6. `completeClaimChallenge` now takes `project_id` and returns 404 when it does not match the challenge's project, so a valid `challenge_id` cannot be spent through a different project's path. Added a `complete on a mismatched project returns 404` test.

packages/api-mock/src/platform-handlers.ts
IAM-marco: Done in f2879fd6. Merged current `main` and reconciled the generator: `challenge_id` is now minted as `ch_` + 128 bits of `crypto.randomBytes` (base64url), matching the `^ch_` contract from #663 and ADR 046's entropy requirement. The six failing tests are green again (now 59/59 with the added cases).

packages/api-mock/src/server.ts
IAM-marco: Modelled in f2879fd6. `claim/complete` now requires a platform-project session that is active and carries a verified factor (`PLATFORM_PROJECT_ID`, `state === "active"`, `factors.length > 0`); anything else is rejected 401. Exchanged sessions now carry a verified factor (the contract defines `active` as having one), the conformance helper mints a platform session, and a new test asserts a customer-project session is rejected 401.

packages/api-mock/src/platform-handlers.ts
IAM-marco: Done in f2879fd6. TTL is now enforced independently of status on both `status` and `complete`, so a completed-but-expired challenge returns 410. Completion is single-use and first-claim-wins: a second completion (or a second challenge once the project is claimed) returns 409 `proj.already_claimed` with the owning team and dashboard. Added tests for the second completion and the completed-but-expired case.

internal/domain/project.go
IAM-marco: Done in f2879fd6. Wired both sentinels into `projectErrorResponse`: `proj.already_claimed` maps to 409 and `proj.claim_expired` to 410, so a #611 handler returning them produces the documented statuses instead of 500. Added a table test (`TestProjectErrorResponse`) covering all project codes.

api/openapi/endpoints/projects/by_id/claim/status/claim-status-response.yaml
IAM-marco: Done in f2879fd6. Rewrote the status response as discriminated `pending`/`completed` variants (`oneOf` with a `status` discriminator). `{ "status": "completed" }` is no longer contract-valid, and a pending response cannot carry completion fields. Regenerated ogen and orval; the zod union now rejects a completed payload missing `team_id`/`claimed_at`/`dashboard_url`.

api/openapi/endpoints/projects/by_id/claim/init/methods.yaml
IAM-marco: Done in f2879fd6. Reworded to "single-use, short-lived challenge" and pointed at `expires_at` for the exact expiry, so the duration is not baked into prose.

api/openapi/endpoints/projects/by_id/claim/status/methods.yaml
IAM-marco: Done in f2879fd6. Added a 429 response to `claim/status`, and also to `claim/complete` since ADR 046 requires completion to be rate-limited against `challenge_id` guessing.

api/openapi/endpoints/projects/by_id/claim/complete/methods.yaml
IAM-marco: Good question, but this is intentionally out of scope for the claim MVP. ADR 046 §4 (and its Consequences) records that the developer's personal team is created once at platform registration, and claim only writes the project-to-team grant, so there is exactly one destination team and nothing to choose. That is why the earlier draft's `team_choice` was dropped from this contract. Picking a different or existing team (work-domain team matching, join requests) is explicitly deferred in the epic (#96, "Existing company team match" is marked out of scope for the MVP, covered in a later iteration). When that lands it will extend the contract rather than change this one, so I have left the request body as just `challenge_id`.

IAM-marco:
Thanks for the thorough review. Pushed `f2879fd6` addressing all eleven threads, on top of a merge with current `main`.

**Contract**
- Reconciled `challenge_id` with `main`'s `^ch_` contract (#663): minted as `ch_` + 128 bits of crypto randomness per ADR 046.
- Status response is now discriminated `pending`/`completed` variants, so `{ "status": "completed" }` alone is no longer valid and pending responses can't carry completion fields.
- Added `429` to `claim/status` and `claim/complete` (completion is rate-limited per ADR 046). A `409` `proj.already_claimed` is now declared on `complete` for first-claim-wins.
- Reworded the "10-minute" prose to defer to `expires_at`.

**Server**
- Wired `proj.already_claimed` → 409 and `proj.claim_expired` → 410 into `projectErrorResponse`, with a table test, so the #611 handlers return the documented statuses.

**Mock fidelity (ADR 046)**
- `init` requires the bearer to be the path project's own secret (foreign secret → 404).
- `status` distinguishes 401 (no/invalid bearer) from 403 (valid but not the initiating secret).
- `complete` checks `project_id`, enforces the TTL independently of status, is single-use/first-claim-wins, and requires a platform-project, active, verified-factor session.

Declined one item with rationale: a `team_id` in the complete body (the personal team pre-exists and is the sole destination; multi-team selection is out of scope for the MVP per ADR 046 §4 and epic #96).

Validation: `moon run api-mock:test` 59/59, `server:check-generate` in sync, `moon ci :lint :typecheck :build :test` green (22 tasks).

fforootd (approved):

vitorbari (approved):

Issue #607, opened 2026-07-23
Claim B1: migration for claim_challenges


IAM-marco:
Part of #96.

## Goal

Add the `claim_challenges` table: one ephemeral row per in-flight claim attempt, tracking the challenge from `claim/init` through `claim/complete` or expiry.

## Current state

Migrations run up to `000013_branding.sql` on both Postgres and the Spanner mirror (Spanner is missing `000010`, an existing, unrelated gap). `team_memberships` already exists as of `000011_user_team_lifecycle.sql`. Nothing claim-specific exists yet. This is the only new table the claim epic itself owns.

## Decisions that shape this work

- `claim_challenges` needs: an identifier, the target `project_id`, a hash of the initiating secret (for `claim/status`'s proof-of-possession check, see Claim E1, #611), a status (pending or completed), an expiry, and a created timestamp. No stashed-secret column: secret rotation is descoped, so `claim/status` never hands back a new secret.
- No team or membership columns here: `team_memberships` already exists (#287), and `teams.name` plus full CRUD is tracked separately in #598.

## Dependencies

- None. Foundation task, can start now.
- Unblocks Claim B2 (#610), which adds the storage v2 statements operating on this table.

## Test first (TDD)

Not applicable in the usual sense (a schema migration), but confirm the migration applies cleanly on both dialects and that its down migration reverses it, following however existing migrations in this repo are exercised.

## Tasks

- Add `internal/storage/database/dialect/postgres/migration/sql/000014_claim_challenges.sql` with an `-- +goose Up` and `-- +goose Down` pair creating `zitadel_nextgen.claim_challenges (id, project_id, initiating_secret_hash, status, expires_at, created_at)`, foreign-keyed to `projects(id)`.
- Add the Spanner mirror at `internal/storage/database/dialect/spanner/migration/sql/000014_claim_challenges.sql`, following the same numbering and the dialect differences already established between the two trees (compare `000011_user_team_lifecycle.sql` between the two dialects for the pattern).
- Pick the identifier format for `id` (see Claim A1's ADR, #606, for the ephemeral, handoff-token-style decision) and add an index useful for a future expiry sweep, even though automated expiry enforcement itself stays out of scope for this epic.

## Reference

`internal/storage/database/dialect/postgres/migration/sql/000011_user_team_lifecycle.sql` and its Spanner counterpart, both for the dialect-pair pattern; `internal/storage/AGENTS.md` for the portability constraints between Postgres and Spanner.

PR #740, opened 2026-08-03
refactor(database): add claim challenge statements and the personal-team lookup

### LINKED ISSUE #610: Claim B2: storage v2 statements, grant-write integration, and personal-team lookup
Part of #96.

## Goal

Add the storage v2 statements Claim E1 needs: challenge CRUD, the integration point for writing the grant, and a new read that resolves a user's personal team through the already-existing `team_memberships` table.

## Current state

`internal/service/statement.go` defines `AllStatements` as a composition of per-resource statement interfaces (`ProjectStatements`, `TeamStatements`, and so on), each implemented per dialect under `internal/storage/v2/dialect/{postgres,spanner}/`. `TeamStatements` today only has `CreateTeam`, `GetTeamByID`, and `DeactivateTeam`; it has no membership methods. `team_memberships` itself is only reachable through the legacy repository layer (`internal/storage/database/repository/team_membership.go`, backing `domain.TeamMembershipRepository`, with `Create`, `Get`, `ListByUser`, `ListByTeam`, `UpdateStatus`). There is no v2 equivalent yet, so a claim transaction cannot read `team_memberships` and write the challenge or the grant on the same connection without one.

- Everything the claim transaction touches must be reachable from storage v2, so the whole `claim/complete` transaction (validate challenge, resolve team, write grant, mark completed) can run on one connection.
- The grant write itself calls into whatever API the permission-management work (#419) exposes; this ticket only provides the integration point, not the grant's own schema.
- `teams.name` and full team CRUD are out of scope, tracked in #598. This ticket only needs to resolve a team ID, not display a name.

## Dependencies

- Claim B1 (#607). The `claim_challenges` table this operates on.
- Unblocks Claim E1 (#611), which calls these statements inside its transaction.

## Test first (TDD)

Write statement-level tests first, following the existing Postgres and Spanner test pattern (for example `internal/storage/v2/dialect/postgres/team_test.go`), for: creating a challenge, getting it by ID, marking it completed and rejecting a second completion, and resolving a user's personal team via `team_memberships` (present, absent, multiple memberships). Then implement.

## Tasks

- Add a `ClaimStatements` interface to `internal/service/statement.go`, alongside `ProjectStatements` and `TeamStatements`, included in `AllStatements`, with `CreateChallenge`, `GetChallengeByID`, `MarkChallengeCompleted`, and a personal-team lookup method (for example `GetPersonalTeamForUser(ctx, projectID, userID) (*domain.Team, error)`, reading `team_memberships` and joining `teams`).
- Implement the Postgres statements under `internal/storage/v2/dialect/postgres/` (new `claim.go`), following the file-per-resource convention (`project.go`, `team.go`).
- Implement the Spanner equivalents under `internal/storage/v2/dialect/spanner/`.
- Regenerate the statement mocks (`internal/service/mocks/statement.mock.go`, per the `go:generate` directive at the top of `statement.go`).
- Add `internal/domain/claim.go` with the `ClaimChallenge` domain type and any new field enums, following the `Project` and `Team` pattern (`NewClaimChallenge`, a `ClaimChallengeField` enum if listing is ever needed).

## Reference

`internal/service/statement.go` (the interface to extend), `internal/storage/v2/dialect/postgres/team.go` and `project.go` (the pattern to mirror), `internal/storage/database/repository/team_membership.go` (the existing legacy-layer membership queries to port to v2), `internal/domain/team.go` (the domain-type pattern).

IAM-marco:
## Summary

- Claim B2 (#610): the storage v2 statements Claim E1 (#611) needs to run the whole `claim/complete` transaction on one connection — challenge CRUD over `claim_challenges` (from #607) plus resolving a user's personal team through `team_memberships`. Closes #610.
- New `ClaimStatements` interface (`CreateChallenge`, `GetChallengeByID`, `MarkChallengeCompleted`, `GetPersonalTeamForUser`), embedded in `AllStatements` and implemented on all three dialects. Challenge reads/writes are project-scoped so a cross-project challenge id behaves like a nonexistent one (ADR 046 §3); `MarkChallengeCompleted` enforces single-use and project scope atomically in one `UPDATE` guarded on `status = 'pending'`. `GetChallengeByID` returns the row regardless of expiry — TTL validation stays read-time service logic.
- The personal team resolves as the user's earliest membership (`ORDER BY created_at, team_id LIMIT 1`), returned only when that membership and its team are both active — correct by construction, since the personal team is created at registration (ADR 046 §4). There is no fallback: a deactivated personal team (or removed earliest membership) yields not-found rather than silently resolving a later, possibly shared, team (review feedback). No `ClaimChallengeField` enum or list method: challenges are ephemeral and never listed.
- The sqlite dialect predates #607, so `claim_challenges` is folded into its squashed `000001_init.sql` (pre-release migrations are modified in place).
- The grant-write integration point is structural: `ClaimStatements` being part of `AllStatements` under `pool.Transaction` lets E1 combine these with the permission-engine grant API (#419) on one connection. No grant code lands here.

## Validation

- `moon run server:test` — passes (incl. `check-generate`; mocks and OpenAPI error schemas regenerated via `go generate ./...`).
- `moon run server:test-sqlite` — passes; all 15 new claim subtests green.
- `moon run server:test-postgres` — passes; all 15 new claim subtests green.
- All four lanes rerun after merging `main` (#663's sqlite ID rewrite included) and after the no-fallback rework: same results, with `GetPersonalTeamForUser` now at 5 subtests per dialect.
- `moon run server:test-spanner` — all new claim subtests green. Pre-existing emulator failures remain (`TestJSONSchemaStatements_CRUD`, `TestProjectStatements_Update`/`_List`, `TestTeamStatements_UpdateTeam` timestamp-equality assertions); verified they fail identically on clean main with these changes stashed.

## Release notes / changeset

- Changeset: `.changeset/claim-storage-statements.md` — "Add claim challenge storage statements (create, get, mark completed) and the personal-team lookup across Postgres, Spanner, and SQLite, groundwork for the claim lifecycle (ADR 046)." (`@zitadel/server`, patch)

## Notes

- The ticket text predates two changes: the legacy repository layer it references has been removed (membership already lives in storage v2), and the sqlite dialect now exists as a third target — both handled here.
- The sqlite migration comments cite "ADR 041" for the claim lifecycle; that is copied verbatim from the existing postgres/spanner `claim_challenges` migrations (pre-renumber, now ADR 046) to stay consistent across the three files. A follow-up could sweep all three.
- Behavioral coverage lives entirely in `stmttest/` per ADR 041; no per-dialect claim tests were needed (nothing engine-specific surfaced).


internal/service/statement.go
```diff
@@ -284,3 +285,23 @@ type BrandingStatements interface {
...
+	// CreateChallenge inserts a pending claim challenge; entity.ID is the
+	// SHA-256 hash of the challenge token, minted by the caller.
+	CreateChallenge(ctx context.Context, entity *domain.ClaimChallenge) error
+	GetChallengeByID(ctx context.Context, projectID, id string) (*domain.ClaimChallenge, error)
+	// MarkChallengeCompleted flips pending -> completed; a challenge that is
+	// absent, in another project, or already completed returns NoRowFoundError.
+	MarkChallengeCompleted(ctx context.Context, projectID, id string) error
+	// GetPersonalTeamForUser resolves the user's personal team: the earliest
+	// active team membership joined to an active team.
+	GetPersonalTeamForUser(ctx context.Context, projectID, userID string) (*domain.Team, error)
```
fforootd: Could we make this contract avoid falling through to a different team? Each dialect applies the active membership/team filters before `ORDER BY`, so if the earliest personal team is deactivated, this returns the user’s next active membership; the new test currently locks that fallback in. Once Claim E1 writes the grant, that could attach the project to a shared team even though ADR 046 deliberately defers target-team selection. I think we should either resolve an explicit personal-team association, or select the earliest membership first and return not found when that membership or team is inactive.

fforootd (commented):
Thanks for adding the portable statement suite and keeping the challenge transition atomic across all three dialects. I found one ownership edge that looks important to resolve before the follow-on claim service uses this contract; details inline.

Two smaller follow-ups:

- GitHub currently reports branch conflicts with `main` in `internal/service/statement.go`, the generated statement mocks, and SQLite’s squashed migration. Could you merge `main`, regenerate, and rerun the integration lanes after resolving those?
- Could we classify the title as `refactor(database): add claim challenge statements and personal-team lookup`? Per the repository’s title rules, this is shipped storage groundwork but not yet a user-reachable capability. The `@zitadel/server` patch changeset can still stay.

IAM-marco:
> GitHub currently reports branch conflicts with main in internal/service/statement.go, the generated statement mocks, and SQLite’s squashed migration. Could you merge main, regenerate, and rerun the integration lanes after resolving those?

Yes, I'll regen the mocks

> Could we classify the title as refactor(database): add claim challenge statements and personal-team lookup? 

I was not aware these were the rules. I can rename to refactor, despite being new functionality being introduced

internal/service/statement.go
IAM-marco: Agreed, the fallback was wrong. Fixed in 2825d9cd with your second option: the statement now selects the user's earliest membership unconditionally (ORDER BY created_at, team_id LIMIT 1) and only returns it when both that membership and its team are active; otherwise it is NoRowFoundError, never a later membership. The stmttest suite locks this in both directions: deactivating the earliest team now expects not-found instead of the next team, and a new case proves a removed earliest membership does not fall back even when a later active membership exists.

An explicit personal-team association would be the cleaner long-term contract, but there is no marker to resolve today (ADR 024 models the personal team as a normal team, and registration bootstrapping in #527 owns its creation), so that would be a schema decision beyond this PR's scope. Happy to file a follow-up if you want to pursue it.

IAM-marco:
Review items are addressed as of 2825d9cd:

- Merged main (including #663's managed-ID work and its sqlite ID rewrite), resolved the three conflicts, and regenerated the statement mocks. All four test lanes rerun: unit, sqlite, and postgres pass; spanner shows only the four known pre-existing emulator failures, nothing new.
- Personal-team lookup reworked to the no-fallback contract, details in the inline thread.
- Title reclassified to refactor(database); the @zitadel/server patch changeset stays as-is.

PR #744, opened 2026-08-04
ci: per-variant journey gates and an affected-scoped framework matrix

fforootd:
## Summary

Rungs 2+3 on the affected-gating probe from #731: the three journey steps stop being all-or-nothing, and the fresh-app framework matrix scales to what actually moved.

**Per-variant journey gating (rung 2).** The variant↔surface map lives as task-space constants in `scripts/ci-mode.mjs` — never file globs, moon's graph stays the authority:

| variant | triggers |
|---|---|
| fresh-app | shared surface ∪ any `sdk-*:build` ∪ journey project |
| passkey-preset, test-kit | shared surface ∪ journey project |
| *(shared surface)* | `server:test`, `cli:build`, `api:build`, `config:build`, `testing:build`, `login-ui:build`, `components:build`, `sdk-core:build`, `sdk-next:build` |

The key discriminator is `server:test` as the go-side signal instead of `server:build`: console embeds into the server binary, so console-only changes mark `server:build` affected — but the journeys never render the console, and embedding breakage is `server:build`'s own job inside the `moon ci` graph. `server:test` moves only when Go/API sources move. The snapshot runs iff any journey does (tarball handoff stays a filesystem contract moon cannot see).

Net effect, live-probed against real moon queries: **console-only PRs run the e2e suites but zero journeys and no snapshot** (~6.4 min off that class); **single-SDK PRs run only the fresh-app matrix** (passkey/testkit skip, −2.7 min); login-ui/testing-kit/journey changes still run all three.

**Affected-scoped matrix (rung 3).** `journey_matrix=single` collapses the fresh-app journey to `--framework next` unless an SDK build, `cli:build` (owns the scaffold templates), or the journey project moved — the scaffolded flow widget is framework-independent, so a server-only change proves as much on one framework as on five (~1.3 min off every go-side PR).

**Query-failure observability.** Investigating why the docs-only #710 ran the full 9.6-min tail post-#731: its moon query failed transiently (658 ms step, silent) and correctly failed open — but nothing said so. The initial theory (a moved base tip poisons moon's two-dot diff) was reproduced against and **disproved** — moon merge-bases internally — so no base-handling change ships; instead, query failure exit codes and non-JSON stdout now print to the step log, and the forced-full reason prints to stdout as well as the step summary.

## Validation

- `node --test scripts/ci-mode.test.mjs` — 19/19; fixtures are live `moon query` captures from current main (new sdk-vue class; console/login-ui/testing refreshed 2026-08-04).
- `moon run workspace:test` — same suite through moon (`:test` selection).
- Live probes against real queries on committed diffs: console-only → journeys/snapshot false, suites true; go-only → all gates true with `journey_matrix=single`; aging-branch repro (fork 3 commits back, current-tip base) → correctly gated, disproving the base-staleness theory.
- YAML parse of ci.yml.
- This PR's own run force-fulls (touches `scripts/` + `.github/`) — the safety rule, as designed.

## Release notes / changeset

- No changeset required — no shipped behavior changed (CI workflow + CI-glue script + docs).

## Notes

- Console-only PRs lose the `release-snapshot` artifact (their journeys skip). Console review happens on the Vercel preview, and any code path that genuinely feeds the snapshot re-enables it via the surface map — flag it if you want a console-triggered snapshot anyway.
- passkey-preset and test-kit currently share one trigger surface (both scaffold Next and drive the widget); they have separate gates so the map can diverge later without workflow changes.
- The go-side probe commit history from the investigation is in the branch history; the merge-base machinery was deliberately built, disproved, and removed before this PR.


.github/workflows/ci.yml
```diff
@@ -130,26 +132,38 @@ jobs:
...
+      # journey does (the tarball handoff is a filesystem contract moon
+      # cannot see).
       - name: Build release snapshot without container
         if: ${{ steps.ci-mode.outputs.mode == 'full' && steps.ci-mode.outputs.snapshot == 'true' }}
         run: env -u CI -u GITHUB_ACTIONS moon run release:snapshot -- --skip-container
 
+      # journey_matrix=single collapses the framework matrix to Next when no
+      # SDK, CLI, or journey surface moved — the flow widget is
+      # framework-independent, so a server-only change proves as much on one
+      # framework as on five.
```
fforootd: Could we say “the full framework matrix” here instead of “five”? `apps/cli-journey-e2e/scripts/frameworks.mjs` currently lists eight frameworks, so the hard-coded count is already stale.

scripts/ci-mode.test.mjs
```diff
@@ -54,58 +86,86 @@ test("docs-only change gates everything off", () => {
...
+test("components affectedness triggers journeys via the shared surface", () => {
+  const { gates, matrix } = resolveGates({
     mode: "full",
     files: ["packages/components/src/atoms/index.ts"],
     targets: COMPONENTS_SLICE,
   });
+  assert.deepEqual(journeys(gates), [true, true, true]);
   assert.equal(gates.browsers, true);
-  assert.equal(gates.journeys, false);
+  assert.equal(matrix, "single");
```
fforootd: Could we use the full affected capture for this expectation (or rename it so it does not read like the real components class)? I probed a one-file `packages/components` change against this PR’s current Moon graph; `--downstream deep` also returns every `sdk-*:build` plus `cli-journey-e2e:*`, so this code resolves `journey_matrix=full`, not `single`. The trimmed slice currently tests a target combination Moon does not emit for that change class, which could make the expected runtime reduction look broader than it is.

CONTRIBUTING.md
```diff
@@ -276,16 +276,22 @@ dependencies, then picks one of two modes via `scripts/ci-mode.mjs`:
...
-frontend-only PR skips the Go suites. The gates fail open: known repo-wide
+provably cannot reach its
+tasks — a docs-only PR runs almost nothing, a frontend-only PR skips the Go
+suites, a console-only PR runs the e2e suites but no journeys. The journey
+variants gate per surface (the map lives as task-space constants in
+`scripts/ci-mode.mjs`): the fresh-app journey answers to every SDK plus the
+shared login surface, the passkey-preset and test-kit journeys to the shared
+surface only, and the fresh-app framework matrix collapses to a single
+framework unless an SDK, the CLI, or the journey project itself moved. The
+snapshot runs iff any journey does — the tarball handoff between them is a
```
fforootd, line 287: Could we update the artifact sentence below while we are changing this contract? It still says “the release snapshot always,” but this new `snapshot: anyJourney` gate intentionally leaves docs-only and console-only PRs without one. Otherwise the same section gives contributors two different expectations for a missing snapshot artifact.

.github/workflows/ci.yml
fforootd: Reworded in 49ef07ab: both the workflow comment and the ci-mode.mjs header now say "the full matrix" and point at `apps/cli-journey-e2e/scripts/frameworks.mjs` as the owner of the list — no hard-coded count to go stale.

scripts/ci-mode.test.mjs
fforootd: You're right — the real class resolves `journey_matrix=full`. In 49ef07ab the test uses the full 91-task capture (re-probed on this branch, matching yours: all 8 `sdk-*:build` plus `cli-journey-e2e:*`) and asserts reality: every journey gated on, full matrix, and `go_tests` false (`server:test` never enters the set). The trimmed fixture survives only as a clearly-labeled synthetic `TEST_BROWSER_SLICE` isolating the `:test-browser`-alone browsers trigger, which no real class produces in isolation.

CONTRIBUTING.md
fforootd: Updated in 49ef07ab: the artifacts sentence now reads "the release snapshot whenever a journey lane runs — a run whose journeys are gated off, such as a docs-only or console-only PR, uploads no snapshot", so the section carries one consistent expectation.

adlerhurst (approved):

Issue #678, opened 2026-07-31
Customise the authentication experience


elinashoko:
## Vision

Authentication should feel like a natural part of the customer’s product rather than a separate Zitadel experience.

Zitadel provides complete, secure authentication capabilities and multiple ways to present them. Customers can embed Zitadel login components in their application, use a standalone login served by Zitadel, or build and operate their own frontend.

Within each model, customers choose the level of customization that fits their product and use case.

## Mental model

> **One authentication foundation, multiple presentation choices, and the freedom to take more control over time.**

| Model | Zitadel provides | Customer controls |
|---|---|---|
| **Customer-embedded login** | Login components, authentication journeys and security controls | The application around the components and supported component customization |
| **Zitadel-served login** | A complete standalone login experience | The supported branding, translations, and authentication configuration |
| **Fully custom frontend** | Authentication APIs, policies and backend security enforcement | The complete frontend experience |

These models share the same authentication foundation, but they are not one presentation or customization surface.

Where applicable, customers should be able to reuse branding, translations, and authentication configuration across embedded and Zitadel-served login. The exact customization available for each model is defined separately.

These are choices rather than a required progression. Customers can select the model that fits their needs and change their approach over time.

Placement map: [customization strategy](https://github.com/zitadel/nextgen/blob/main/docs/design/branding/customization-strategy.md), [ADR 057](https://github.com/zitadel/nextgen/blob/main/docs/adrs/057-login-customization-categories.md).

### Target users

The people creating, reviewing or operating the authentication experience include:

- **Developers** who want to launch and maintain authentication without building every part of the UI themselves.
- **Product and design teams** who want authentication to match the rest of their product.
- **Platform and identity teams** who manage authentication consistently across applications, products or brands.
- **Security, legal and compliance teams** who need the experience to meet organizational and regulatory requirements.

### Common use cases

Across these groups, customers need to:

1. Launch a professional authentication experience with minimal effort.
2. Match authentication to their product’s branding, translations, and interaction patterns.
3. Preview and safely publish changes across complete authentication journeys.
4. Maintain consistency while adapting the experience across applications, products, brands or customers.
5. Meet localization, accessibility, legal, security and compliance requirements.
6. Take full ownership of the frontend when the Zitadel-provided experience is insufficient.

### Customer-embedded login

Customers embed the Zitadel login components into their own application.

[...]

Issue #676, opened 2026-07-31
Add basic auth customisation to CLI/ onboarding flow


elinashoko:
## Context

Zitadel already provides five starter designs for the hosted authentication experience:

- `centered`
- `split`
- `split-right`
- `hero`
- `minimal`

A design can currently be selected during setup using:

```sh
zitadel setup --design <name>
```

However, design selection is not included in the interactive CLI setup. Developers who follow the standard onboarding flow receive the default `centered` design without being shown the available alternatives.

The existing branding configuration also supports a logo URL and hero image URL, but these options are not exposed during onboarding.

Full Liquid template customisation will be handled separately after the project has been claimed. This ticket only covers the basic customisation needed during initial setup.

## Intended experience

During CLI onboarding, developers can:

1. Choose one of the existing starter designs.
2. Optionally add the supported brand assets.
3. Review their choices.
4. Complete setup with the selected design and assets already applied.

Developers can skip customisation and continue with the default design. They do not need to understand or edit Liquid templates during onboarding.

## Position in the existing setup flow

The customisation section should be the final configuration step before the project is created.

The updated setup order should be:

1. Choose a framework, if starting in an empty directory.
2. Confirm the detected framework.
3. Choose a Zitadel server.
4. Confirm the development port.
5. Choose who will sign in to the app—the user schema.
6. Choose how users should sign in—the sign-in preset.
7. Customise the authentication experience.
8. Review the captured configuration.
9. Create and configure the project.

The new customisation prompts should be added immediately after `SignInPresetPrompt` and before the CLI displays `Configuration captured` and starts creating the project.

This keeps the questions in a logical order:

- The user schema defines what information is collected.
- The sign-in preset defines how users authenticate.
- The design defines how that experience looks.

### 1. Introduce authentication customisation

After the sign-in preset has been selected, the CLI introduces the final configuration section:

```text
Customise your authentication experience

Choose a starter design and optionally add your branding.
You can fully customise the template after setup.
```

### 2. Choose a starter design

```text
Choose a starter design:

❯ Centered           Form centred in a card
  Split              Branding left, form right
  Split (reversed)   Form left, branding right
  Hero               Large branding area with supporting content
  Minimal            Simple layout without card styling

View examples: <documentation URL>
```

Behaviour:

- `Centered` is preselected.
- The developer selects a design using the existing CLI prompt controls.
- The CLI uses readable names in the prompt while storing the existing design values.

| Display name | Stored value |
|---|---|
[...]

PR #750, opened 2026-08-04
feat: ask for the login design in the setup wizard

fforootd:
## Summary

- The setup wizard now ends with a design question (#676): keep the preselected **Built-in** template (writes nothing), or pick one of the five starter designs — `centered`, `split`, `split-right`, `hero`, `minimal` — which ejects the design into `.zitadel/branding/` and publishes it as branding revision 1 during setup, exactly like the existing `--design` flag (which now skips the question). Template ownership stays an explicit opt-in: the built-in default is preselected and the option labels say that a design writes files.
- The chosen design is now observable everywhere agents and humans look: the "Zitadel is ready" summary box (`Login design` row), the JSON envelope (`data.design`, `null` for built-in), setup telemetry (`design` dimension, allow-listed enum), and the local-server retry hint (an interactively chosen design survives into the suggested retry command).
- `next_actions` now covers the branding journey in both directions: with a design, it points at `.zitadel/branding/login.liquid` + `branding.json` and the plan/apply loop; without one, it makes `branding eject` discoverable.
- New docs page **Customize the login design** (`docs/cli/customize-design`): the five starter designs with per-design asset usage (only `split`/`split-right` render `hero_url`; `minimal` uses no assets), choosing at setup vs later, the `https://`-only asset rule, the revision model, and the ownership trade-off of ejecting. Slot for visual examples per design is a follow-up once we have screenshots.
- The envelope-contract scanner now understands topic commands: `publicCliCommand("branding eject")` maps to the oclif id `branding:eject` instead of the bare topic `branding`.

## Validation

- `moon run cli:test` — 962/962 (includes new DesignPrompt unit tests and a `setup --design split` integration test asserting the ejected files, the inlined wire body, state pinning, and an empty first plan)
- `moon run cli:lint cli:typecheck` — clean (24 pre-existing warnings, 0 errors)
- `moon run docs:build` — clean

## Release notes / changeset

- Changeset: .changeset/setup-design-question.md — `@zitadel/cli` minor: the setup wizard asks how the login should look; the choice is reported in the summary, JSON envelope, and retry hint.

## Notes

- Design-selection UX intentionally differs from the flow #676 sketches in two ways, both grounded in the current branding model (ADR 037/040): the wizard offers **Built-in** as the preselected sixth option (choosing a design forks the template into the repo — that should be opt-in, not the default path), and there are no logo/hero URL prompts (URL-typing in a wizard is hostile, and the assets only render on 2 of 5 designs — the scaffolded `branding.json` plus `next_actions` cover it). Config-only basic branding (layout + assets without template ownership) is wire-reserved (`CreateBrandingBody.layout`) but the master template doesn't implement it yet — that's the proposed opener for #678 rather than scope here.
- The platform api-mock still has no branding routes; the new integration test carries its own POST/GET `*/branding` handlers. Adding branding to `setupPlatformHandlers` is a reasonable follow-up.
- Ejected templates have no upgrade path (`--force` overwrites local edits) — tracked as its own concern, worth a ticket.


apps/cli/tests/integration/contract.test.ts
```diff
@@ -68,8 +68,15 @@ async function typescriptFiles(dir: string): Promise<string[]> {
 }
 
 function commandIdFromArgs(args: string): string | undefined {
-  const id = args.trim().split(/\s+/)[0];
-  return id && /^[a-z][\w:-]*$/.test(id) ? id : undefined;
+  // Leading bare words form the command id; topic commands are suggested with
+  // a space ("branding eject") but registered with oclif's colon id
+  // ("branding:eject"). Flags and free text end the id.
+  const words: string[] = [];
```
fforootd, line 74: This drops the previous support for colon-form command IDs. For example, `publicCliCommand("branding:eject", ...)` now produces no ID here and is silently omitted from the visibility assertion. Could we accept an already-colonized first token as well as space-separated topic words, and add focused cases for `branding eject`, `branding:eject`, and `setup --force`?

apps/cli/src/commands/setup/index.ts
```diff
@@ -427,7 +439,13 @@ export default class Setup extends BaseCommand {
         })),
         files_skipped: result.filesSkipped.map((file) => relativeDisplay(cwd, file)),
         install: installOutcome.install,
-        next_actions: installOutcome.nextActions,
+        // The chosen login design, or null for the built-in template — so
+        // agents can verify what setup published without diffing the repo.
```
fforootd, line 466: Could we update `apps/cli/SKILLS.md` with this new setup envelope field and the interactive/`--design` behavior? The root `AGENTS.md` explicitly requires that canonical agent contract to stay aligned with CLI behavior; its setup flag list still omits `--design`, and its envelope section does not explain `data.design` or the `null` built-in value.

fforootd (commented):
Two additional findings are inline.

Separate current blocker: `full-pr` fails because the `cli` scope is not allowed by `.github/semantic.yml`. Please rename this to an accepted title such as `feat: ask for the login design in the setup wizard` and rerun CI. The job stopped at the title check, so every build/test/journey step was skipped; the green Vercel previews do not validate the CLI test claims on this head.

Issue #612, opened 2026-07-23
Claim E2: claim handlers and integration tests


IAM-marco:
Part of #96.

## Goal

Wire the ogen-generated handlers to the claim service and write integration tests covering the full set of documented failure modes.

## Current state

`internal/api/integration_test/` already has comparable coverage to model this on: `project_test.go`, `team_test.go`, and `session_me_test.go` are the closest analogues, covering project-secret-authenticated CRUD and cookie-authenticated session endpoints respectively. Error mapping in handlers follows the pattern in `internal/api/session.go`'s `sessionErrorResponse` (a switch over `domain.Error` codes to HTTP status codes), which claim's own error mapper should mirror.

## Decisions that shape this work

- Error mapping follows ADR 030's canonical codes; each claim-specific error (already claimed, expired, bearer mismatch) needs a distinct, documented status code (409, 410, 403 respectively).
- Tests must exercise a real flow-engine login in the test harness for the `claim/complete` happy path, not a mocked session, so the C2 session check is exercised end to end.

## Dependencies

- Claim E1 (#611). The service this wires up and tests.
- Unblocks the end-to-end verification of Claim G1 (#613) and Claim H1 (#615); both can be built earlier against the api-mock, but need this ticket for real end-to-end testing.

## Test first (TDD)

This ticket is the test-first work for the claim service. Write integration tests before, or while, wiring the handlers: the happy path (a real flow-engine login through to a completed claim), an already-claimed project (409), an expired challenge (410 on both status and complete), a mismatched bearer on `claim/status` (403), and a concurrent-complete race (two simultaneous completions, only one succeeds).

## Tasks

- Add `internal/api/claim.go` implementing the three ogen handler methods that A2's codegen produces for `claim/init`, `claim/status`, and `claim/complete`, calling into `claimService`.
- Add a claim-specific error mapper, following `sessionErrorResponse`'s pattern in `internal/api/session.go`.
- Add `internal/api/integration_test/claim_test.go` with the test cases listed above, using the harness's existing flow-engine login helpers (see `registration_flow_test.go` or `password_flow_test.go` for how a real login is driven in tests).

## Reference

`internal/api/integration_test/project_test.go`, `team_test.go`, and `session_me_test.go` (closest existing analogues); `internal/api/session.go` (`sessionErrorResponse`, the error-mapping pattern); `internal/api/integration_test/registration_flow_test.go` or `password_flow_test.go` (driving a real login in a test).

## Current state update (2026-08-14)

[...]

PR #751, opened 2026-08-04
feat(api): verify platform session for claim/complete

### LINKED ISSUE #608: Claim C2: platform-session verification for claim/complete
Part of #96.

## Goal

Authenticate the human calling `claim/complete`, reusing the platform's existing session-cookie security scheme rather than building a new one, and add the claim-specific checks it does not already perform.

## Current state

Cookie-based session authentication already exists: `SecurityHandler.HandleNextgenSession` (`internal/api/security.go`) verifies the `__nextgen_session` cookie via `tokenService.VerifyToken`, checks `domain.ValidateSessionToken` (token type plus session ID presence), and stashes the parsed token in the request context via `sessionTokenFromContext`. It already secures `GetMySession`, `RevokeMySession`, and `GetMyUser` (see `sessionCookieOperations` in `security.go`). What it does not check: that the session belongs to a specific project, or that the session has at least one verified factor. `GetMySession`'s own extra check, `validateSessionToken` in `session.go`, only compares `TokenID` and expiry, not factor count. `Session.State()` already encodes "has factors and not expired" as `SessionStateActive` (`internal/domain/session.go`).

- Cookie only, no token-in-body variant.
- CSRF: console and API share the same domain, so `SameSite=Lax` (already the cookie's attribute, see `sessionCookie` in `session.go`) is sufficient for the MVP; no separate CSRF token.
- The session must belong to the platform project (config, see PB0, #605) and be `SessionStateActive`, so anonymous, pre-login sessions can never claim.

## Dependencies

- None. Buildable and unit-testable against the existing session primitives now.
- Its end-to-end verification needs a real platform project (PB0, #605). Consumed by Claim E1 (#611).

## Test first (TDD)

Write unit tests first for the new check: a session belonging to a different project is rejected, a `SessionStateBuilding` session (no factors yet) is rejected, an expired session is rejected, and a valid active platform-project session succeeds and returns the correct user ID. Then implement.

## Tasks

- Add the claim/complete operation to `sessionCookieOperations` in `internal/api/security.go`, so the OpenAPI security scheme wires it to `HandleNextgenSession` the same way `GetMySession` is wired.
- Add a small helper, likely in `internal/api/claim.go`, that: reads the token via `sessionTokenFromContext(ctx)`, loads the full session via the session service, checks `session.ProjectID` against the configured platform project ID, checks `session.State() == domain.SessionStateActive`, and returns the session's user ID.
- Return a distinct error for each failure mode (wrong project, not active, expired), so Claim E2 (#612) can write targeted tests against them.

## Reference

`internal/api/security.go` (`HandleNextgenSession`, `sessionCookieOperations`), `internal/api/session.go` (`GetMySession`, `validateSessionToken`, `sessionTokenFromContext`), `internal/domain/session.go` (`Session.State()`, `SessionStateActive`).

IAM-marco:
## Summary

Closes #608 (Claim C2, part of epic #96).

- New `verifyClaimSession` helper in `internal/api/claim.go`: authenticates the human behind `claim/complete` via the existing `__nextgen_session` cookie scheme and adds the claim-specific checks it lacked (ADR 046 §2). The session must belong to the configured platform project and be `SessionStateActive` (≥1 verified factor); the helper returns the session's user id. Production caller lands with Claim E1 (#611).
- New internal sentinels `claim.session_wrong_project`, `claim.session_expired`, `claim.session_not_active` in `internal/domain/claim.go`, so Claim E2 (#612) can write targeted tests per failure mode. On the wire every rejection stays the contract's plain `auth.unauthorized` 401; the sentinel is kept in the parent chain (log-only per ADR 030).
- `CompleteClaim` added to `sessionCookieOperations` so the pre-handler 401 carries the documented "Missing or invalid session token." message (the scheme wiring itself was already generated from the spec, #739).
- Platform project id threaded onto `api.Handler` from `platform.project_id` (the pin #736 builds on); empty fails closed, so unpinned standalone deployments reject every claim session.
- Added the missing mockgen directive for `SessionService` and its generated mock, used by the new unit tests.

Encoded trade-offs: session not-found is masked as 401 (no existence leak on an auth path); wrong-project wins over lifecycle state (a foreign caller learns nothing about the platform session); expiry is checked explicitly before `State()`, whose factor-first ordering would report an expired factor-less session as `Building`.

## Validation

- TDD: `internal/api/claim_internal_test.go` written first (red: build failure on the missing helper), then implementation (green).
- `go test ./internal/api/ -run 'VerifyClaimSession|SessionCookieOperations' -v`: 11 table-driven subtests (missing token, not-found, service passthrough, rotated token, wrong project, unpinned platform, expired, expired+factor-less quirk pin, building, nil user id, happy path) plus the `sessionCookieOperations` guard, all pass.
- `go build ./...`, `go vet ./internal/... ./cmd/...`, `gofmt -l` clean.
- `go test ./internal/api/... ./internal/domain/... ./cmd/server/... ./internal/service/...` all ok.
- `moon run server:test` passes (includes `server:check-generate`; the three new `claim-session_*.yaml` error schemas are its generator output and are committed).

## Release notes / changeset

No changeset: Go-only internal helper with no user-visible behavior yet; the endpoint that consumes it ships with #611.

## Notes

- Deliberately out of scope: the `CompleteClaim` endpoint handler (#611), a `claim.*` branch in `errorResponse` (the helper always wraps, so the existing `auth.unauthorized` 401 mapping applies), changes to `validateSessionToken`/`HandleNextgenSession`, and harness platform-project support (#612).
- No dependency on PR #736; only the `platform.project_id` config field already on main is consumed. Trivial overlap in `cmd/server/server.go` if #736 lands first.


internal/api/claim.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		return "", err
+	}
+
+	if session.TokenID != sessionToken.TokenID {
+		return "", invalidSessionCredential(domain.ErrSessionTokenInvalid())
+	}
+	// Checked before the lifecycle so a foreign caller learns nothing about the
+	// platform session's state.
+	if h.platformProjectID == "" || session.ProjectID != h.platformProjectID {
```
grvijayan: we could do this check before the `GET` call with `sessionToken.ProjectID`

internal/api/claim.go
```diff
@@ -0,0 +1,57 @@
...
+	// session as Building.
+	if time.Now().After(session.ExpiresAt) {
+		return "", invalidSessionCredential(domain.ErrClaimSessionExpired())
+	}
+	if session.State() != domain.SessionStateActive {
+		return "", invalidSessionCredential(domain.ErrClaimSessionNotActive())
+	}
+	if session.UserID == nil || *session.UserID == "" {
+		return "", invalidSessionCredential(domain.ErrClaimSessionNotActive())
+	}
```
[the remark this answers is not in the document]
IAM-marco: Addressed in 3df0fac by broadening the message: "The session is not active: it has no verified factor or no resolved user." Kept a single sentinel rather than adding a distinct one: the nil-user branch is a defensive invariant check (a session with a verified factor always carries a user id), and a dedicated code would add a fourth wire schema for a case that should never occur. The broadened message now accurately covers both branches that return it.

internal/api/claim.go
grvijayan: I see that `h.platformProjectID` wouldn't be configured in the case of self-hosters - in this case, `invalid session` feels misleading. If it's for a cloud-setup, not setting this config would have to fail during startup (probably), right?

IAM-marco: Valid

IAM-marco: This method will be called only when claiming happens. So self-hoster are not concerned.

IAM-marco: Done in 1e665fc: the platform-project check now compares sessionToken.ProjectID before the Get, so a foreign-project session is never loaded. The tests' gomock expectations also prove no lookup happens for the wrong-project verdicts.

grvijayan: if it's only applicable for the cloud setup, the platformProjectID still shouldn't be empty, right? Or am I missing something? 🤔

grvijayan: besides this, everything else looks fine - please feel free to merge it.

IAM-marco: > if it's only applicable for the cloud setup, the platformProjectID still shouldn't be empty, right? Or am I missing something? 🤔

It should not. The project ID is set [via config](https://github.com/zitadel/nextgen/blob/d4198416f65fc5ac7182a5ccf9cb247bf07b4922/cmd/server/server.go#L369): you set both `project_id` and `bootstrap_project=true`. [Validation](https://github.com/zitadel/nextgen/blob/d4198416f65fc5ac7182a5ccf9cb247bf07b4922/cmd/server/config.go#L71) then makes sure if you set the bool to true, project ID is set as well.

Then that value is forwarded to `h.platformProjectID`.

I guess this is a second layer of validation, independent from the other one

PR #752, opened 2026-08-04
feat: Added endpoint to list a users teams

### LINKED ISSUE #735: API: expose a user's teams on the user read endpoints
## Context

Now that #620 has landed `POST /teams/query`, the console can list and name teams. What it still cannot do is say which teams a user belongs to — nothing on the user read responses carries that.

`domain.TeamMembership` is described as "the canonical N:N roster between users and teams", and `ListTeamMemberships` exists in the storage layer with filters by user and by team. `domain.User` also carries `LifecycleOwnerTeamID`. Neither reaches the API: `GET /users` and `GET /users/{user_id}` return the schema attribute tree plus `id` (and `metadata`, once #721 lands).

## Scope

- Expose a user's team membership on the user read responses, or as a dedicated endpoint if the roster is better read separately.
- Decide which of the two concepts the console should render. They are different:
  - **`LifecycleOwnerTeamID`** — the single team that owns the user's lifecycle (ADR 024), or none when self-owned.
  - **team memberships** — the N:N roster. Design decision **D1** says a user can belong to multiple teams, with a role per team, so the Team column is plural.

  If the column is memberships, one team id is not enough.
- Include enough to render without an N+1: a name alongside each id, or a shape the console can resolve in one `POST /teams/query`.

## Why now

The users list design has a **Team** column, and it is the last piece of #630 that has no path. The list endpoint exists; the link from user to team does not.

## Reference

- `internal/domain/team_membership.go`, `internal/domain/user.go` (`LifecycleOwnerTeamID`)
- `internal/service/user.go` — `GetUserByID`, `ListUsers`
- Related: #620 (team list, done) · #703 (the same pattern for status and timestamps) · #630

wim07101993:
## Summary

- Added endpoint to list a users teams

## Validation

- `go test ./...`

## Release notes / changeset

- Changeset: .changeset/list-user-teams-endpoint.md - Added endpoint to list a users teams
     

## Notes

-


internal/api/user.go
```diff
@@ -126,6 +126,44 @@ func (h *Handler) ListUserPasskeys(ctx context.Context, params api.ListUserPassk
 	return res, nil
 }
 
+// ListUserTeams serves the user's team roster — the N:N membership list, one
+// page at a time, each entry carrying the team's name so a client renders the
+// page without resolving ids one by one. Lifecycle ownership is a different
+// question and stays on the user itself (ADR 024).
+func (h *Handler) ListUserTeams(ctx context.Context, params api.ListUserTeamsParams) (api.ListUserTeamsRes, error) {
+	if err := requireProjectAccess(ctx, string(params.ProjectID), userAccess, opRead); err != nil {
```
grvijayan: regarding using `user.read`:
we're querying a user's teams, but the response contains information about the teams, which would normally require `team.read`

should we use both `user.read` and `team.read` scopes?

internal/service/user.go
```diff
@@ -186,6 +198,57 @@ func (s *UserService) ListPasskeys(ctx context.Context, input ListPasskeysInput)
...
+		Filter: database.And(
+			database.Equal(database.Col(domain.UserTeamFieldProjectID), input.ProjectID),
+			database.Equal(database.Col(domain.UserTeamFieldUserID), input.UserID),
+			database.Or(onRoster...),
+		),
+		Pagination: database.Page[domain.UserTeamField]{
+			Limit:  uint32(normalizeLimit(input.Limit)),
+			Cursor: []byte(input.PageToken),
+			OrderBy: database.OrderBy[domain.UserTeamField]{
+				Columns: []database.Column[domain.UserTeamField]{
```
grvijayan, line 223: Claude:

`created_at` is a redundant sort key

Order is (`team_name`, `created_at`, `team_id`). For a fixed (`project_id`, `user_id`), `team_id` is already unique (it's part of the PK), so (`team_name`, `team_id`) totally orders the roster. The middle column just widens the cursor and adds a comparison term to every keyset query for nothing. Drop it.

internal/api/user.go
wim07101993: good point, I will add both. We can relax the authz later if needed

grvijayan (approved):

Issue #748, opened 2026-08-04
Sessions: Soft-revoke + `revoked` state


vitorbari:
**Context:** Revoke currently hard-deletes the row, so revoked sessions vanish and `state: revoked` is unreachable.

**Specifications (approach):**
- Move revocation from hard-delete to soft-revoke (persist a marker on the session)
- Add `revoked` as a first-class state in the session lifecycle model
- Propagate the new state through the read/query API and the status filter

**Acceptance criteria:**
- [ ] Revoking a session keeps the record instead of deleting it
- [ ] A revoked session reads as `state: revoked` on get and query
- [ ] Filtering by status (including `revoked`) works
- [ ] Tests cover revoke, revoked read, and status filter

vitorbari:
Decide to keep the hard-deleted of sessions, and make the DELETE endpoints idempotent.

PR #757, opened 2026-08-05
feat(api): replace GET /sessions with POST /sessions/query

### LINKED ISSUE #746: Contract: adapt OpenAPI for `POST /sessions/query`
**Context:** The list endpoint is currently defined as `GET /sessions` (and was never implemented). We are replacing it with `POST /sessions/query` to match the [ADR 031 query specification](https://github.com/zitadel/nextgen/blob/main/docs/adrs/031-openapi-querying.md) already used by `/teams/query` and `/projects/query`.

**Full path / contract:**

```
POST /sessions/query?project_id={project_id}
Auth: oauth2 — session.read
```

Request body:

```json
{
  "filter": [
    { "field": "state",   "operation": "equals", "value": "active" },
    { "field": "user_id", "operation": "equals", "value": "user_123" }
  ],
  "sorting": { "field": "createdAt", "direction": "desc" },
  "limit": 20,
  "page_token": null
}
```

Response `200`:

```json
{
  "sessions": [
    {
      "session_id": "sess_2ab...",
      "project_id": "proj_1x...",
      "state": "active",
      "user_id": "user_123",
      "factors": [ { "type": "password", "verified_at": "2026-08-04T09:59:00Z" } ],
      "created_at": "2026-08-04T10:00:00Z",
      "expires_at": "2026-08-05T10:00:00Z"
    }
  ],
  "next_page_token": "eyJ..."
}
```

**Acceptance criteria:**
- [ ] `POST /sessions/query` defined per the shape above; `GET /sessions` list removed
- [ ] `session-filter-field.yaml` exposes `createdAt`, `user_id`, `state`
  - filtering by `team_id` will be introduced as part of https://github.com/zitadel/nextgen/issues/749
- [ ] Spec bundles/lints and ogen types regenerate cleanly
- [ ] No behavior change (handler still returns not-implemented)

grvijayan:
Moves session listing onto the ADR 031 query shape, replacing `GET /sessions` with `POST /sessions/query?project_id=`

Closes #746.

## Sorting

`sorting.field` gets its own `session-sort-field.yaml` with `created_at` and `user_id`. Filtering keeps all three fields (`created_at`, `user_id`, `state`).

`state` is computed, not stored:

* `building` — no verified factors (`last_verified_at IS NULL`)
* `expired` — has a verified factor and `expires_at < now()`
* `active` — has a verified factor and `expires_at >= now()`
* `revoked` — depends on #748 (soft-revoke)

Its value moves as `now()` advances or a factor is verified, so it cannot be a sort key. The cursor holds the last row's sort values, so rows get skipped or duplicated between pages.

ADR 031's template shares one enum between `sorting.field` and `filter[].field`. This is the first endpoint to split them.

## Validation

   - `npx @redocly/cli lint api/openapi/openapi-spec.yaml` — valid
   - `node scripts/check-go-generate.mjs` — exit 0
   - `go test ./...` — pass
   - `node scripts/check-adrs-index.mjs` — ok (47 records)
   - `moon run api:generate` — client exports `querySessions`, no `listSessions`; `QuerySessionsBodySortingField` is `created_at` and `user_id`, filter keeps all three
   - `moon run docs:generate` — bundle carries `querySessions`; nav no longer links a removed page

## Release notes / changeset

`.changeset/session-query-endpoint.md`


api/openapi/endpoints/sessions/query/session-filter-field.yaml
```diff
@@ -0,0 +1,10 @@
+type: string
+description: |
+  Field to filter or sort sessions by:
+  - `created_at`: RFC3339 timestamp
+  - `user_id`: user id
+  - `state`: one of `building`, `active`, `expired`, `revoked`
+enum:
+  - created_at
+  - user_id
+  - state
```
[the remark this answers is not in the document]
grvijayan: will be refactored in a follow-up PR

api/openapi/endpoints/sessions/query/session-sort-field.yaml
```diff
@@ -0,0 +1,7 @@
+type: string
```
wim07101993, line 1: In other calls, the filter field and sort fields are the same.   I don't mind keeping them separate, but it will create a descrepancey.
I also don't think we should dismiss sorting by other fields because it would skip duplicate rows between pages. Can't we solve that by adding the created-at field as a backup sorter?

grvijayan: There is no `state` column to order by. It is computed from `expires_at` and the verified factors, and its value changes as `now()` changes, so a backup sorter doesn't help. 

We can add `user_id` though. 
Claude noted a bug which would not support it atm as `user_id` can be null (we can still add it as this API is still a WIP). 

> Sorting on `user_id` is broken today. It is NULL until a factor is verified, and ordering by a nullable column silently drops those rows on all three dialects. Filing that separately. Once it is fixed `user_id` can join the sort enum, with `id` appended as the tiebreaker the way `teamOrderBy` and `projectOrderBy` already do.

wim07101993 (approved):

api/openapi/endpoints/sessions/query/session-sort-field.yaml
grvijayan: added `user_id` as a sort field in https://github.com/zitadel/nextgen/pull/757/commits/7c8afc09327b4ed1b4665e19eac9207980b203e6

wim07101993 (approved):

PR #758, opened 2026-08-05
test(authz): compiler property fuzz and generative persist suite

adlerhurst:
## Summary

Confidence-building tests for the authz stack **before** the grants/check API / resolver (#423).

### Design
`docs/design/api/authz-testing.md` — L1 crash fuzz → L2 compiler invariants → L3 DB persist → L4 resolver oracle.

### L1/L2
- Shared `authztest` generator with **named recipe tables** (`GenerateModel` mixed / `GenerateValidModel` valid-only)
- Seeded invariant suite + `FuzzCompileInvariants` / parse fuzz
- Full shortest-path transitive closure oracle
- `TestGenerateValidModelAlwaysCompiles` (500 seeds)

### L3
- Generative persist via `GenerateValidModel` (no soft-skip)
- `compiler.PersistedCatalog` + `LoadCatalogMutations` (thin wrappers over `GetAuthzCatalog` + `PersistedCatalogFromDomain`)
- `stmttest.TestPersistCatalogVersion_Generative`

Rebuilt onto `main` after Wave 0/1 squash merges so duplicate Wave 1 history no longer conflicts.

## Validation

- `go test ./internal/authz/...`
- `go test -tags sqlite_integration ./internal/storage/v2/stmttest/ -run 'Generative|Authz|Catalog' -count=1`

## Release notes / changeset

No changeset required — docs/tests (+ test-oriented catalog read unused by product paths yet).

## Notes

- L4 check oracle remains with #423.
- Dual-write sequence fuzz deferred.

<div><a href="https://cursor.com/agents/bc-015bbf42-4acd-415a-87f9-976a7e6a54c4?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-015bbf42-4acd-415a-87f9-976a7e6a54c4&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/docs/design/api/README.md
+++ b/docs/design/api/README.md
@@ -31,8 +31,9 @@ Individual decisions are marked inline in each doc:
 5. [`authn-and-auth-flows.md`](authn-and-auth-flows.md) — auth_attempts state machine, OIDC adapter, SSR handoff.
 6. [`authz.md`](authz.md) — credential × scope × permission.
 7. [`permission-storage.md`](permission-storage.md) — Wave 0 relational DDL for catalogs, assignments, membership edges, and `resource_scope_index` (feeds issue #422).
-8. [`security-and-origins.md`](security-and-origins.md) — environment-gated origin wildcards, CORS, CSRF.
-9. [`resource-map.md`](resource-map.md) — the full endpoint surface grouped by concern.
+8. [`authz-testing.md`](authz-testing.md) — L1–L4 fuzzy/property strategy for compiler, persist, and (later) resolver oracles.
+9. [`security-and-origins.md`](security-and-origins.md) — environment-gated origin wildcards, CORS, CSRF.
+10. [`resource-map.md`](resource-map.md) — the full endpoint surface grouped by concern.
 
 ## Sibling doc sets
 
--- a/docs/design/api/authz-testing.md
+++ b/docs/design/api/authz-testing.md
@@ -184,7 +184,7 @@ integration tags. Slice C is the long-term gate for authz behavior.
 go test ./internal/authz/...
 
 # Generative persist round-trip (needs a dialect tag)
-go test -tags sqlite_integration ./internal/storage/v2/stmttest/ -run Generative -count=1
+go test -tags sqlite_integration ./internal/storage/stmttest/ -run Generative -count=1
 
 # Longer fuzz (local / overnight)
 go test ./internal/authz/compiler/ -fuzz=FuzzCompileInvariants -fuzztime=30s
--- a/docs/design/api/authz-testing.md
+++ b/docs/design/api/authz-testing.md
@@ -1,0 +1,207 @@
+# Authz testing without an API
+
+> **Status:** RECOMMENDED exploration (not locked)
+> **Context:** Wave 1 storage ([#422](https://github.com/zitadel/nextgen/issues/422) /
+> [PR #677](https://github.com/zitadel/nextgen/pull/677)) and the OpenFGA
+> compiler ([#421](https://github.com/zitadel/nextgen/issues/421) /
+> [PR #720](https://github.com/zitadel/nextgen/pull/720)) land before a public
+> grants/check API and before the resolver ([#423](https://github.com/zitadel/nextgen/issues/423)).
+> This note maps how to gain confidence in compiler + database behavior anyway.
+>
+> Sibling: [`permission-storage.md`](permission-storage.md),
+> [`authz.md`](authz.md), [ADR 032 §2](../../adrs/032-permission-catalogs.md#2-openfga-parser-and-profile-compiler).
+
+## Problem
+
+The interesting surface for authorization is eventually:
+
+```
+credential × assignment × membership × catalog closure/edges → check/list
+```
+
+Today we have the left-hand pieces (parser → profile → compiler →
+`PersistCatalogVersion`, dual-write RSI / membership edges) but not the
+resolver or HTTP API. Waiting for those leaves schema and compile bugs to
+surface late. Fuzzy / property testing can exercise the shipped layers now.
+
+## What "fuzzy testing" means here
+
+Three related techniques, not one tool:
+
+| Technique | Input | Oracle | Fits |
+| --- | --- | --- | --- |
+| **Crash fuzz** | Arbitrary bytes (DSL/JSON) | No panic; bounded runtime | `openfga.Parse*` |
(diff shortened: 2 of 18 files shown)


livio-a (commented):
**Review: compiler property fuzz + generative persist suite** — reviewed as top of the #666 → #677 → #758 stack.

These are genuine, well-oracled tests — not vacuous. The invariant suite asserts several independent structural properties (reflexive/same-object closure, dense edge/reference positions, plan↔edge consistency, no dangling targets), and determinism is handled carefully throughout: FNV-seeded generators, a reproducible fuzz corpus via `Seed(i)`, catalog-order closure emission, and order-independent `ElementsMatch` in persist assertions. I found nothing that would flake in CI. Three things worth addressing:

**[MEDIUM] The closure-depth oracle is a line-for-line copy of the production algorithm.**
`invariants_test.go` `allShortestDepths` is byte-for-byte identical to production `internal/authz/compiler/closure.go` `shortestDepths` (same init, same queue, same relaxation), so it cannot falsify a depth bug — a shared BFS error passes both. Concretely, this is why nothing in the suite catches the #677 seed discrepancy where `viewer→admin` is hand-seeded at depth 1 but the compiler computes depth 2. Suggest an independent oracle (e.g. hand-written expected depths for the fixture models).

**[MEDIUM] Generative persist round-trip never exercises the non-trivial readback columns.**
`GenerateValidModel` restricts to `refPlain` (valid recipes never emit TTU/wildcard/condition), so the round-trip in `authz_generative_test.go` never persists a `tuple_to_userset` edge or a `wildcard`/`condition` reference. The new `LoadCatalogMutations` code that scans `ref_relation`/`wildcard`/`condition` and reconstructs TTU `Source`/`Tupleset` is therefore never verified end-to-end — a swapped/broken `wildcard` or `condition` scan would pass. (Computed-userset `Source` round-trip *is* covered.)

**[MEDIUM / scope] ~440 lines of production readback code ship in a test-labeled PR with no product caller.**
`LoadCatalogMutations` (postgres/spanner/sqlite), `compiler.PersistedCatalog` + `Persisted()`, the new interface method on `statement.go`, `ParseExpressionEdgeKind`, and the regenerated mocks are all non-test production surface. There's no product caller (the doc + `statement.go` acknowledge #423 brings it), and its only exercise is the `*_integration`-gated generative test — so `go test ./...` gives it zero coverage. The code itself looks correct, but consider whether it should land with its #423 caller instead of riding here.

**Lower severity**
- **[LOW]** `openfga/fuzz_test.go` crash-fuzz discards both returns (`_, _ = ParseDSL(input)`) — pure no-panic oracle. Matches the doc's L1 intent, but the equivalence property the doc (`authz-testing.md` L1) also advertises isn't implemented.
- **[LOW]** PG/SQLite catalog-exists checks use `err == pgx.ErrNoRows` / `err == sql.ErrNoRows` rather than `errors.Is` — brittle if the sentinel is ever wrapped. (Spanner maps `codes.NotFound`, so it's fine.)
- **[NIT]** `assertPlansMatchEdges` compares `edgesByTarget[target][i]` positionally without re-sorting by `Position` (unlike `assertDenseEdgePositions`) — deterministic today, fragile if emission order ever diverges from plan-term order.

Generator bias and determinism otherwise check out — `TestGenerateValidModelAlwaysCompiles` (500 seeds) guards against trivially-empty catalogs, and the rich invariants still run on real shapes via the curated fixtures.

livio-a (commented):
**`assertShortestClosureDepths` self-grades the depth values.** In `internal/authz/compiler/invariants_test.go`, `allShortestDepths` is a line-for-line copy of production `shortestDepths` (`compiler/closure.go`), and the depth assertion compares the copy's output against the compiler's:

```go
want := shortestDepth(implication.Source, implication.Implied, direct)
require.Equal(t, want.depth, implication.Depth, "closure depth for %#v", implication)
```

Both sides run identical BFS on identical input, so they always agree — a bug in the depth computation (`<` vs `<=`, `+1`, …) shifts both and the test still passes. The reachability/presence checks in the same function are fine; it's only the depth *value* that isn't independently verified.

`compiler_test.go`'s `TestCompileComputesShortestReflexiveTransitiveClosure` (#720) already does this the right way — it hardcodes expected `Depth` values for a fixed model and compares to `output.Catalog.Closure`. Suggest the same here: assert hand-written depths for one or two fixture models instead of re-deriving them with the production algorithm.

---

Minor, non-blocking (fine to fold into #423 when this code gets a real caller):

- **Generative round-trip only emits `refPlain`.** `validOnlyRefRecipes` is `{ refPlain }`, so the generative persist suite never exercises wildcard / condition / TTU reference readback. It's covered by the curated `TestPersistCatalogVersion_DeepRoundTrip`, so not a hole — but the generative path adds no coverage there. Worth a `refWildcard` (kept acyclic) in the valid set if cheap.
- **`openfga` crash-fuzz is no-panic-only** (`_, _ = ParseDSL(input)` / `ParseJSON`). Matches the doc's L1 intent, but the equivalence property `authz-testing.md` advertises ("normalized model equal on equivalent encodings") isn't implemented anywhere yet.

livio-a (approved):
Approving. The self-graded closure-depth oracle is resolved — `invariants_test.go` no longer re-derives depths, and `compiler_test.go`'s new `TestCompileComputesClosureDepthsAlongChain` asserts depths 2 and 3 by hand on a longer chain, so a depth-computation bug would now be caught. Nice touch extending the fixture past depth 1.

Two minor, non-blocking follow-ups (fine to fold into #423 when this readback path gets a real caller):

- Generative valid recipes are still `refPlain`-only, so the generative persist suite doesn't exercise wildcard/condition/TTU reference readback (the curated `TestPersistCatalogVersion_DeepRoundTrip` does cover those columns).
- `openfga` crash-fuzz is no-panic-only; the equivalence property `authz-testing.md` advertises isn't implemented yet.

PR #759, opened 2026-08-05
feat(api): soft-revoke sessions and expose the revoked state

### LINKED ISSUE #748: Sessions: Soft-revoke + `revoked` state

vitorbari:
## Summary

Revoking a session now **soft-deletes** it: the row is kept and stamped with `revoked_at` instead of being hard-deleted, so revoked sessions stay visible and report `state: revoked`.

A new nullable `revoked_at` column was added to the `sessions` table.

Closes #748.


vitorbari:
Decided to keep hard-delete + idempotent endpoints.

PR #760, opened 2026-08-05
feat: implemented token revocation

### LINKED ISSUE #511: Crypto: Token revocation via jti / signature storage (not hashing)
ADR requirement (§Hashing → Tokens): Generated tokens must not be stored; if revocation is needed, store the token id (`jti`) or the signature — no hashing required.

Current state (NOT FOUND): No revocation store / denylist; no `jti` persistence. Opaque tokens are stateless-encrypted and cannot be individually revoked.

Tasks:
- Decide revocation model (store `jti` or signature for revocable token types).
- Add a revocation store keyed by `jti`/signature with expiry-based cleanup.
- Enforce revocation checks in the token verifier for applicable token types.

Acceptance: A revoked token id is rejected at verification; no full token or hashed token is stored.

wim07101993:
## Summary

- implemented token revocation

## Validation

- `go test ./...`

## Release notes / changeset

- Changeset: .changeset/token-revocation-jti.md — implemented token revocation

## Notes

- GH issue: https://github.com/zitadel/nextgen/issues/511


cmd/server/console_runtime.go
```diff
@@ -64,11 +67,8 @@ func standaloneRuntimeResolver(projects service.ProjectService, keys service.Key
 		}
 		meta.ConsoleProjectID = project.ID
 
-		tokenCrypter, err := keys.GetProjectCrypter(ctx, project.ID, domain.EncryptionKeyPurposeToken)
-		if err != nil {
-			return consoleRuntime{}, err
-		}
-		publishableKey, err := project.PreviewSecret(tokenCrypter)
+		tkn := project.PreviewToken()
+		publishableKey, err := tokens.GenerateJWE(ctx, tkn)
```
IAM-marco: This line now writes a database row on every HTTP request, which I don't think is intended.

Before this PR, building the publishable key was just encryption: `project.PreviewSecret(crypter)` serialized a token and encrypted it, touching no storage. `tokenService.GenerateJWE` inserts a row into `tokens` first, because that insert is what mints the `jti` that makes the secret revocable.

The problem is where this particular call sits. This resolver runs once per request to `GET /console/runtime.json` (see the comment on [line 46](https://github.com/zitadel/nextgen/blob/1089273563db11ddff70d8682a1dc509cd4f01ce/cmd/server/console_runtime.go#L46-L49): the document is deliberately recomputed per request so a newly created default project shows up without a restart). That route is registered directly on the `http.ServeMux` at [server.go line 460](https://github.com/zitadel/nextgen/blob/1089273563db11ddff70d8682a1dc509cd4f01ce/cmd/server/server.go#L457-L461), above the catch-all mount that carries the API middleware, so it is reachable without any credentials.

So every console page load, and every anonymous request to that URL, adds a `project_preview` row. Nothing removes them: `project.PreviewToken()` sets no `ExpiresAt` ([project.go line 79](https://github.com/zitadel/nextgen/blob/1089273563db11ddff70d8682a1dc509cd4f01ce/internal/domain/project.go#L79-L85)), so even the `expires_at` sweeper the changeset describes as follow-up work would never collect them.

There is a second effect worth calling out: because each request mints a fresh `jti`, every previously issued publishable key stays independently valid. Revoking a leaked one only retires the single key from that one request, not the project's publishable key.

Suggested fix: mint the publishable key once per project and reuse it (resolve it lazily and cache it, or store it on the project), keeping the per-request refresh for `ConsoleProjectID` only. That keeps the key revocable, which is the point of the change, without an insert per request.

One note on test coverage: [console_runtime_test.go](https://github.com/zitadel/nextgen/blob/1089273563db11ddff70d8682a1dc509cd4f01ce/cmd/server/console_runtime_test.go#L197-L203) mocks `GenerateJWE`, so no existing test would notice the extra writes.

internal/storage/v2/dialect/sqlite/migration/sql/000002_token_revocation.sql
```diff
@@ -0,0 +1,135 @@
...
+-- Project and preview secrets become storable so a leaked one can be revoked
+-- (ADR 036); they authenticate software, so they carry no user or session id.
+--
+-- SQLite cannot alter a CHECK constraint, so the type/identifier check is
+-- widened by rebuilding the table. Nothing references `tokens`, so the
+-- drop-and-rename needs no foreign-key juggling.
+
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE tokens_new (
```
IAM-marco: Pre-release migrations are modified in place in this repo rather than appended to, so these three new files probably belong folded into the migrations that created `tokens`: `000009_tokens.sql` for postgres/spanner and `000001_init.sql` for sqlite.

Precedent: #740 (`04e77a71`) folded `claim_challenges` into sqlite's squashed init, stating "pre-release migrations are modified in place"; #637 (`7ea32f82`) edited `000007_user_agents_and_sessions.sql`, #649 (`4b984afb`) edited `000012_crypto_keys.sql`, and #646 (`fd31b20c`) edited `000002_teams.sql`.

That matters here because all three files only widen the existing CHECK constraint and add one index. Folding removes this 135-line table rebuild entirely, along with the `ALTER TYPE ... ADD VALUE` / `NO TRANSACTION` sequencing in [000015_token_revocation.sql#L6-L17](https://github.com/zitadel/nextgen/blob/1089273563db11ddff70d8682a1dc509cd4f01ce/internal/storage/v2/dialect/postgres/migration/sql/000015_token_revocation.sql#L6-L17) and both `Down` sections.

internal/service/session.go
```diff
@@ -121,14 +121,25 @@ func (s *sessionService) List(ctx context.Context, input ListSessionInput) ([]*d
...
 
 func (s *sessionService) Delete(ctx context.Context, input DeleteSessionInput) error {
-	err := s.v2Pool.Statements().DeleteSessionByID(ctx, input.ProjectID, input.SessionID)
-	if err != nil {
-		if errors.Is(err, domain.ErrSessionNotFound()) {
-			return nil
+	return s.v2Pool.Transaction(ctx, func(ctx context.Context, tx Statementer[AllStatements]) error {
+		err := tx.Statements().DeleteSessionByID(ctx, input.ProjectID, input.SessionID)
+		if err != nil {
+			if errors.Is(err, domain.ErrSessionNotFound()) {
```
IAM-marco: This early `return nil` fires before `DeleteTokensBySessionID` on [line 133](https://github.com/zitadel/nextgen/blob/1089273563db11ddff70d8682a1dc509cd4f01ce/internal/service/session.go#L132-L134), so a session whose row is already gone keeps its token records, and with revocation now modeled as record deletion, those tokens stay valid. Falling through to the token delete preserves the idempotency added in #768 (`fa907c22`), since deleting by session id is already a no-op on zero rows.

Separately, the sentinel check on [line 135](https://github.com/zitadel/nextgen/blob/1089273563db11ddff70d8682a1dc509cd4f01ce/internal/service/session.go#L134-L138) looks like copy-paste: `DeleteTokensBySessionID` never returns `ErrSessionNotFound`, and if it ever did, a genuine failure to revoke would be swallowed as success.

.changeset/token-revocation-jti.md
```diff
@@ -0,0 +1,36 @@
...
+- **Deleting a session revokes the tokens it issued**, in one transaction, so no
+  token record outlives the session it authenticates. Rotating a session token
+  likewise revokes its predecessor, so a rotated-out token stops working the
+  moment its successor is issued.
+- **Project and preview secrets are revocable.** They are now issued with a
+  stored `jti`, so a leaked secret can be retired instead of living forever —
+  these credentials have no expiry of their own.
+
+Expired records are never honoured — verification checks `expires_at` too. There
+is no background sweeper yet, so records that expired without being revoked
```
IAM-marco, line 31: [#511](https://github.com/zitadel/nextgen/issues/511) lists "add a revocation store keyed by `jti`/signature **with expiry-based cleanup**" as a task, so deferring the sweeper leaves that acceptance item open. The rest of #511 is met: `jti` is the revocation key, nothing hashed or full-token is stored, and `IntrospectToken` rejects a revoked id.

Worth filing the follow-up and linking it here so #511 can be closed against something concrete. Note that an `expires_at`-based sweeper alone would not bound the table given the per-request preview tokens flagged in `cmd/server/console_runtime.go`. Those carry no `expires_at` at all.

internal/api/error_handler_test.go
```diff
@@ -60,7 +60,7 @@ func TestOgenErrorHandlerInvalidSessionCookie(t *testing.T) {
 	t.Parallel()
 	mock := gomock.NewController(t)
 	tokenService := mocks.NewMockTokenService(mock)
-	tokenService.EXPECT().VerifyToken(gomock.Any(), gomock.Any()).Return(nil, errors.New("bad token"))
+	tokenService.EXPECT().IntrospectToken(gomock.Any(), gomock.Any()).Return(nil, errors.New("bad token"))
```
IAM-marco: Might be worth switching `gomock.Any()` on the second argument to `http.Cookie.Value` to make a stronger verification

internal/api/security_test.go
```diff
@@ -27,7 +27,7 @@ func TestHandleNextgenSession(t *testing.T) {
 
 		mock := gomock.NewController(t)
 		tokenService := mocks.NewMockTokenService(mock)
-		tokenService.EXPECT().VerifyToken(gomock.Any(), gomock.Any()).Return(token, nil)
+		tokenService.EXPECT().IntrospectToken(gomock.Any(), gomock.Any()).Return(token, nil)
```
IAM-marco: Same comment as before for all the calls to IntrospectToken: can change second argument `gomock.Any()` to the actual expected value

IAM-marco (changes requested):

internal/storage/v2/dialect/sqlite/migration/sql/000002_token_revocation.sql
IAM-marco: Ok what I would do is:

  - Extract the SQL migration from `init.sql` to a separate file
  - Modify that extracted code to match the new structure of the tokens table

cmd/server/console_runtime.go
wim07101993: Good catch, don't know how I missed that

internal/storage/v2/dialect/sqlite/migration/sql/000002_token_revocation.sql
wim07101993: fixed

internal/service/session.go
wim07101993: fixed

internal/service/token.go
```diff
@@ -14,35 +18,115 @@ import (
...
+			},
+		},
+	}
+	result, err := s.v2Pool.Statements().ListTokens(ctx, opts)
+	if err != nil {
+		return nil, domain.ErrInternal(err).WithMessage("failed to get token")
+	}
+	now := time.Now()
+	for token, err := range result.Iterate(func(cursor []byte) (*database.ListResult[*domain.Token], error) {
+		result.NextCursor = cursor
```
IAM-marco: The cursor is assigned to the wrong place, so paging past the first 100 rows cannot work.

`Iterate` hands the callback the previous page's cursor and expects the callback to return the *next* page ([list.go#L30-L52](https://github.com/zitadel/nextgen/blob/cf14d54f45f2521c48c4a4e49d1f3741a9d4686b/internal/storage/database/list.go#L30-L52)). The cursor is an input to the query, so it belongs on the options passed to `ListTokens`, which reads it from `Pagination.Cursor`. Writing it to `result.NextCursor` sets an output field on the page `Iterate` has already consumed, and the `ListTokens` call on the next line still runs with `opts` unchanged, cursor nil.

The effect is that the callback re-fetches page one forever: same 100 rows, same non-empty `NextCursor`, so the loop never terminates. On this path that would hang a request to the unauthenticated `/console/runtime.json`.

`rotateEncryptionKeys` has the shape to copy ([keys.go#L172-L175](https://github.com/zitadel/nextgen/blob/cf14d54f45f2521c48c4a4e49d1f3741a9d4686b/internal/service/keys.go#L172-L175)):

```go
opts.Pagination.Cursor = cursor
return s.v2Pool.Statements().ListTokens(ctx, opts)
```

Worth saying that nothing reaches this today: `PreviewToken()` sets no `ExpiresAt` ([project.go#L81-L87](https://github.com/zitadel/nextgen/blob/cf14d54f45f2521c48c4a4e49d1f3741a9d4686b/internal/domain/project.go#L81-L87)), so `Active(now)` is true for every preview record and the first row always returns before the second page is requested. It only bites if preview credentials ever gain an expiry.

Which raises the simpler option: since a preview record is active by construction, `Limit: 1` and reading `result.Items[0]` would drop the pagination entirely. If you keep the loop for the expiry case, the cursor fix above is what makes it terminate.

IAM-marco (changes requested):

internal/service/token.go
wim07101993: As for now, no pagination would work. But it felt brittle. If somebody changes that behavior of how tokens are stored/expired, it might introduce bugs.

IAM-marco (approved):

PR #772, opened 2026-08-06
feat: capture the user agent for flow-created sessions

### LINKED ISSUE #763: Capture the user agent for flow-created sessions
**Context:** Sessions created by the flow engine are persisted with no user-agent 
POST /sessions` captures it, but flow-originated sessions — the common
case — don't. As a result the operator session list and detail (#745) show blank
device info for most sessions, and per-device session management is degraded.

**Specifications (approach):**
- Source the user agent at flow start the way `POST /sessions` does (request body and/or derived request metadata such as the `User-Agent` header and IP), thread it through `StartFlowRequest` into `resolveFlowSession`, and pass it to `domain.NewSession` so the anonymous session is created with it.
- The reuse path (client-supplied session, incl. step-up) must not overwrite the existing session's user agent.

**Acceptance criteria:**
- [ ] A session created by a flow records the request's user agent (device/IP), not `NULL`
- [ ] The user agent survives exchange — `building → active` keeps it on the same row
- [ ] Supplying an existing session reuses it without overwriting its user agent
- [ ] Session detail/list surfaces the device info for flow-originated sessions
- [ ] Tests cover the flow-start capture and the exchange-preserves-it path

vitorbari:
## Context:

Sessions created via the flow engine are lacking user agent / device information.

## Changes

This PR adds a middleware to extract user-agent header and client IP address from the request and injects it to the context (as an instance of `domain.UserAgent`).
The Flow service then pass this object when calling `domain.NewSession`.

Note: 

1. Data such as device fingerprint was supposed to be sent via [`POST /flow/{id}/event`](https://github.com/zitadel/nextgen/blob/main/api/openapi/components/flows/flow-event-request.yaml) which has not been implemented yet.
2. [`POST /sessions`](https://github.com/zitadel/nextgen/blob/main/api/openapi/endpoints/sessions/create-session-request.yaml#L14) expects user-agent data in the payload, however, the `POST /flows` will just extract them from the headers.

## Test

Session data after authentication:

Screenshot 2026-08-07 at 09 43 46


internal/api/middleware/user_agent.go
```diff
@@ -0,0 +1,57 @@
...
+	ua := &domain.UserAgent{IP: ip}
+	if header != "" {
+		ua.Info = map[string]any{"user_agent": header}
+	}
+	return ua
+}
+
+// clientIP is the first valid hop of X-Forwarded-For when set (the proxy is
+// trusted, as for X-Forwarded-Host), else RemoteAddr without its port.
+func clientIP(r *http.Request) string {
```
vitorbari, line 50: This function is available on `chi`, but since it is not currently a direct dependency, I duplicated it.

internal/api/middleware/user_agent.go
```diff
@@ -0,0 +1,57 @@
...
+	if xff := r.Header.Get("X-Forwarded-For"); xff != "" {
+		first, _, _ := strings.Cut(xff, ",")
+		if ip := strings.TrimSpace(first); net.ParseIP(ip) != nil {
+			return ip
+		}
```
adlerhurst, line 55: That topic looks more complicated as I expected. 😄

After looking up how others did it I found multiple libs which check for multiple headers. Therefore it might be worth making the header configurable as list and provide a sane default. See https://github.com/chaunsin/goip/blob/master/goip.go#L42-L60 as an example.

adlerhurst (commented):
Nice work. Getting the correct IP from the headers seems more complex than i would have expected :D

internal/api/middleware/user_agent.go
vitorbari: Yes, this topic can get complicated.

The correct thing to do is to introduce a list of trusted proxies, which influences the logic used for `WithRequestHostMiddleware` and `WithUserAgentMiddleware` (clientIP). 

However that goes a bit beyond the scope of this ticket. I would propose to create a follow up issue, here is the draft: https://github.com/zitadel/nextgen/issues/802

Wdyt?

adlerhurst (approved):

internal/api/middleware/user_agent.go
adlerhurst: I would then add a TOOD to make it clear that this implementation might not work for all cases.

vitorbari: Done.

Issue #774, opened 2026-08-06
Claim follow-up: gate `zitadel claim` on a cloud server, and the cross-machine staleness gap


IAM-marco:
Part of #96. Follow-up from #614.

## Context

Claiming is a cloud concept: self-hosters manage their own infrastructure and have nothing
to attach. #614 gates its `setup` / `status` / `doctor` nudges accordingly, on the `server`
origin recorded in `zitadel.json` via `apps/cli/src/lib/oclif/server-kind.ts`.

The `zitadel claim` command itself (shipped in #754) has no such gate. Run against a local
or self-hosted server it will call `initClaim` and fail with whatever the backend answers,
rather than saying up front that there is nothing to claim.

## Two open items

**1. Gate the command.** Decide whether `zitadel claim` should refuse a non-cloud target
with a clear message, or keep its current behaviour and let the backend answer. If it
should refuse, reuse the same helper #614 introduces so the command and the nudges cannot
disagree.

**2. Cross-machine staleness.** #614's nudges read claim state from the local
`.zitadel/secret` only, because A2 (#609) exposes no ambient claim-state read: `getClaimStatus`
requires an active challenge id, and the [breakdown](https://github.com/zitadel/nextgen/issues/96#issuecomment-4923654787)
§3.5 deliberately dropped claim fields from `GET /projects/{id}` as attributes of the grant,
owned by the permission-management track (#419).

So a project claimed from another machine, or a `.zitadel/secret` restored from a backup,
reads as unattached locally. Today the impact is only a stale suggestion: `zitadel claim`
turns the resulting `409` into a clean skip. If that becomes a real annoyance, the fix is
an ambient claim-state read, which is a permission-surface decision rather than a CLI one.

## Update (2026-08-20)

Two developments sharpen both items:

1. E2 (#612, PR #912) shipped, so the claim endpoints answer for real instead of 501. That makes the ungated command degrade **worse** on a non-cloud target, not better: `claim/init` succeeds against any server holding the secret, the CLI opens `<console base URL>/claim?...` (a page that does not exist until #615), and then polls until the 10-minute challenge expires. Previously it failed fast. This strengthens the case for gating via #614's server-kind helper.
2. The ambient claim-state read from item 2 now has a ratified home: ADR 053 §6 (merged Proposed via PR #876) defines the authorized-project discovery query, "projects on which this principal has at least the requested relation". When that query ships, it is the natural cross-machine claim-state read; nothing changes on the CLI side until then.

Issue #598, opened 2026-07-22
Resource model - Teams


elinashoko:
# Team

## Key rules

- A Team always belongs to exactly one Project.
- The meaning of a Team depends on which Project it belongs to.
- In the platform Project, a Team represents a customer account.
- In a customer Project, a Team represents one of the customer's tenants (for example, a company, workspace, merchant, or department).

## Purpose

A Team represents a logical grouping within a Project. It is used to organise users, applications, and other resources.

The same Team resource is used in different contexts:

- Platform Project → customer account
- Customer Project → customer tenant

For more information on how Teams fit into the overall identity model, see:
- https://github.com/zitadel/nextgen/issues/249

## Lifecycle

A Team can be in one of the following states:

| Status | Description |
|--------|-------------|
| ACTIVE | The Team is available for use. |
| DEACTIVATED | The Team is no longer active but is retained for historical and audit purposes. |

See ADR 024 for lifecycle ownership details.

## Operations

A Team supports the following operations:

- Create
- List
- Get
- Update
- Delete

### Resource Identifier

```text
team_id
```

Globally unique and immutable.

### Parent

```text
project_id
```

Identifies the Project that contains the Team.

## Schema

| Field | Type | Description |
|------|------|-------------|
| team_id | UUID | Unique Team identifier. |
| project_id | UUID | Parent Project identifier. |
| name | String | Human-readable Team name. |
| status | Enum | Current lifecycle status. |
| created_at | Timestamp | Time the Team was created. |
| updated_at | Timestamp | Time the Team was last updated. |

## Related documents

- Resource Map - https://github.com/zitadel/nextgen/blob/main/docs/design/api/resource-map.md
- Platform architecture (Issue #249)
- ADR 024 – User/Team lifecycle ownership - https://github.com/zitadel/nextgen/blob/main/docs/adrs/024-user-team-lifecycle-ownership.md

PR #776, opened 2026-08-06
feat: report team attachment in setup, status, and doctor

### LINKED ISSUE #614: Claim G2: CLI team-attachment nudges
Part of #96.

## Goal

Surface team-attachment state in the CLI's existing output: a "temporary until you attach
it to a team" nudge for projects that have not been claimed yet, and the owning team once
they have.

## Current state

G1 (#613) shipped in #754, which changes the shape of this ticket:

- `apps/cli/src/commands/claim.ts` is the `zitadel claim` command.
- `.zitadel/secret` now carries optional `claimed_at` and `team_id`, typed on
  `ZitadelSecret` in `apps/cli/src/lib/project.ts`. `writeZitadelSecret` records them.
- `claim.ts` already treats that local record as authoritative: it skips the platform
  round trip entirely when both fields are present, because first-claim-wins is enforced
  server side anyway.

The three nudge surfaces still know nothing about it: `apps/cli/src/commands/setup/index.ts`
(via `buildSummary`, rendered by `setup/summary.ts`), `apps/cli/src/commands/status.ts`,
and the checks under `apps/cli/src/commands/doctor/checks/`.

**Read the local file, not the API.** A2 (#609) shipped exactly three operations:
`initClaim`, `getClaimStatus`, `completeClaim`. `getClaimStatus` requires an *active
challenge id* and returns 403 unless the presented secret minted that challenge, so it
cannot answer "is this project attached?" outside a live claim. There is deliberately no
ambient claim-state read: the [implementation breakdown](https://github.com/zitadel/nextgen/issues/96#issuecomment-4923654787)
§3.5 dropped claim fields from `GET /projects/{id}` because they are attributes of the
grant, owned by the permission-management track. So these nudges read `.zitadel/secret`
through the existing `readZitadelSecret`, and cost no network at all. That also keeps
`status` and `doctor` fast and usable offline.

**Show the team id, not a team name.** Teams have no `name` column yet (#598), and the
claim responses carry only `team_id` and `dashboard_url`.

**Cloud only.** Claiming is a cloud concept: self-hosters manage their own infrastructure
and have nothing to attach. `zitadel setup --server local` is the documented dev loop, so
an ungated nudge would permanently advertise an impossible action during local
development. Gate on the `server` origin recorded in `zitadel.json`, bucketed by the
existing `serverKind` helper in `apps/cli/src/lib/oclif/server-kind.ts`. Note this must
come from `zitadel.json`, not `meta.source`: `doctor` hardcodes its source to the local
server URL, and `status --server local` overrides it.

**Copy follows G1's vocabulary.** Say "attach this project to a team so it becomes
permanent", matching the shipped `claim` command. `apps/cli/tests/integration/vocabulary.test.ts`
bans the word "unclaimed" in `README.md` and `SKILLS.md`, so it must not reach either.

**No deletion promise.** The epic's 14-day lifetime is unenforced, since there is no
scheduled-task infrastructure (breakdown §3.2). Messaging invites attaching a team without
implying projects will actually be deleted.

## Dependencies

- Claim A2 (#609), closed. The contract `zitadel claim` speaks.
- Claim G1 (#613), closed. The command being nudged toward, and the source of the local
  `claimed_at` / `team_id` fields.

## Test first (TDD)

Write tests first for each surface (setup summary, status, doctor check), asserting the
nudge appears for a cloud project with no `claimed_at` / `team_id`, the team id appears
once both are present, and neither appears for a local or self-hosted server.

## Tasks

- Add a shared helper that decides the state once (attached / not attached / not
  applicable) and owns the strings, so the three surfaces cannot drift. It reads
  `.zitadel/secret` via `readZitadelSecret` and gates on `serverKind`. Add a reader for
  `zitadel.json`'s top-level `server` alongside the existing `readDevelopmentIssuer` in
  `apps/cli/src/lib/project.ts`.
- Setup: add an ownership row to the `Project` section built by `buildSummary` in
  `apps/cli/src/commands/setup/index.ts`, and append the claim action to the box actions
  and to `next_actions` / `next_commands` in the JSON envelope. A freshly created project
  is always unattached, so the only branch here is cloud versus not.
- Status: `projectStatus()` already reads both `zitadel.json` and `.zitadel/secret`, so no
  extra I/O is needed. Carry the state on the `configured` variant of `ProjectStatus`, and
  extend `nextActionsFor` / `nextCommandsFor`. The nudge is additive: keep the existing
  user-presence staging intact.
- Doctor: add a new check under `apps/cli/src/commands/doctor/checks/` and register it in
  `checks/index.ts`. It must emit `warn`, never `fail`: an unattached project is not a
  defect, and `doctor` turns any `fail` into a thrown `E_VALIDATION`. Follow
  `checks/schema.ts`, which overrides `run()` to return a warning and delegates to
  `super.run()` otherwise; the `AbstractSanityCheck`-only checks like `secret.ts` and
  `project-match.ts` can emit only pass or fail. Leave `fix()` as the inherited no-op:
  `doctor --fix` calls it on every non-passing check, and a claim needs a human in a
  browser. Add a branch to `advisoryForWarnings` so the warning surfaces `zitadel claim`,
  the way the `docker-cli` and `managed-runtime-processes` warnings already do.
- Document the new envelope fields in `apps/cli/SKILLS.md`, next to the existing `claim`
  entry, so agents can branch on attachment state.
- Add a changeset for `@zitadel/cli`, following `.changeset/cli-claim-command.md` from G1.

## Known limitation

The local file is the only source, so a project claimed from another machine, or a
`.zitadel/secret` restored from a backup, reads as unattached here. The nudge is advisory
and `zitadel claim` itself handles the real answer (a `409` becomes a clean skip), so the
worst case is a stale suggestion. Revisit if and when the permission-management track
exposes an ambient claim-state read.

## Reference

`apps/cli/src/commands/claim.ts` and `apps/cli/src/lib/project.ts` (the local state and
the predicate to mirror), `apps/cli/src/commands/doctor/checks/schema.ts` (the
warning-capable check pattern), `apps/cli/src/commands/setup/index.ts` and
`setup/summary.ts`, `apps/cli/src/commands/status.ts`,
`apps/cli/src/lib/oclif/server-kind.ts`.

IAM-marco:
## Summary

Closes #614. Part of #96.

- `setup`, `status`, and `doctor` now say whether a project is attached to a team, so a developer meets the temporary nature of a fresh project without first knowing `zitadel claim` exists.
- State comes from `.zitadel/secret` alone, which G1 (#754) already writes. **There is no ambient claim-state endpoint to call:** `getClaimStatus` answers for one live challenge and rejects a secret that did not mint it, and claim attributes were deliberately kept off `GET /projects/{id}` because they belong to the grant, not the project ([breakdown §3.5](https://github.com/zitadel/nextgen/issues/96#issuecomment-4923654787)). Reading locally is also what keeps `status` and `doctor` fast and working offline.
- **Advisory, never a failure.** An unattached project works exactly like an attached one, so `doctor` warns and still exits 0. Failing would have broken every scripted `zitadel doctor` run against a project nobody had claimed yet.
- **Cloud only.** Local and self-hosted projects have no team to attach to, and `zitadel setup --server local` is the documented dev loop, so an ungated nudge would advertise an impossible action for the life of every local project. The gate reads the `server` in `zitadel.json`, not the server a command is pointed at: `doctor` pins its source to the local runtime URL and `status --server local` overrides it.
- Copy shows the **team id**, not a name: teams have no `name` column yet (#598), and the claim responses carry only `team_id`. It frames projects as temporary **without promising deletion**, since the 14-day lifetime is unenforced.

The ticket was written before G1 shipped and asked for an API call that does not exist and a team name that cannot be resolved; it has been updated to match, and #774 tracks the two follow-ups.

## Two judgement calls worth a look

**An existing test's expectation changed.** `status.test.ts` asserted `next_actions` is empty when the platform is unreachable. The claim nudge legitimately fires there, because it is a local read, which is the point of reading locally. The assertion now distinguishes probe-dependent journey guidance (still withheld) from the claim nudge (still shown), rather than suppressing the nudge to keep the old line green.

**The setup summary renders every row with a green `✓`**, so the nudge reads as `✓ Ownership   temporary until you attach it to a team`. The value is dimmed. If that checkmark reads wrong, the alternative is dropping the row and keeping only the action line below the box.

## Validation

```
moon run cli:test        # 1011 passed (109 files)
moon run cli:lint        # clean
moon run cli:typecheck   # clean (KNOWN-VACUOUS per moon.yml, so weak signal)
```

Smoke-tested by hand against the built CLI, covering every state:

- detached cloud project: `doctor` warns and exits 0, `status` and the setup box carry the nudge
- attached: `[ok] claim: Project is attached to team team-abc`, no nudge anywhere
- project whose `zitadel.json` says `http://localhost:8080`: silent **even when invoked with `--server https://api.zitadel.cloud`**, confirming the gate reads the project's recorded server rather than the flag
- `doctor --fix` on a detached project: warning persists, `.zitadel/secret` byte-identical

## Notes

`moon run cli:readme` produces a ~130-line diff that is pure terminal-width reflow, unrelated to this change, so it is not included. Worth knowing that task is not width-stable.

## Release notes / changeset

- `.changeset/cli-claim-nudges.md` — minor bump for `@zitadel/cli`.


apps/cli/src/commands/doctor/index.ts
```diff
@@ -259,6 +260,11 @@ function advisoryForWarnings(
     nextCommands.push(...advice.nextCommands);
   }
 
+  if (warnings.some((check) => check.name === "claim")) {
+    nextActions.push(claimAction(cliVersion));
+    nextCommands.push(claimCommand(cliVersion));
+  }
```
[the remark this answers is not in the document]
IAM-marco: Already handled: `advisoryForWarnings` ends with `return { nextActions: unique(nextActions), nextCommands: unique(nextCommands) }` (doctor/index.ts:279, helper at :490), so every branch including this one is de-duplicated before returning. That is also why the existing `docker-cli` and `managed-runtime-processes` branches push unconditionally. No change needed.

apps/cli/src/commands/doctor/checks/claim.ts
```diff
@@ -0,0 +1,81 @@
...
+      };
+    }
+
+    if (state.kind === "detached") {
+      return {
+        name: this.name,
+        status: "warn",
+        message:
+          "This project is temporary until you attach it to a team. Run `zitadel claim` " +
+          "to make it permanent; nothing about the project changes.",
```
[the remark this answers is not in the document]
IAM-marco: Keeping this one as-is, deliberately. Check *messages* in this battery quote the bare command: `schema.ts:37` says "New projects scaffold editable defaults via `zitadel setup`". The `publicCliCommand` form belongs to `next_actions`/`next_commands`, which is exactly where this check's advisory already puts it via `claimAction(cliVersion)`/`claimCommand(cliVersion)` (doctor/index.ts:263-266), so a `doctor` run already prints the runnable `npx @zitadel/cli@<version> claim` right below the warning.

Inlining the versioned npx string into the check message too would duplicate it on every run and make the one-line warning considerably longer, without telling the reader anything the advisory does not.

apps/cli/src/commands/setup/index.ts
```diff
@@ -743,6 +761,14 @@ function buildSummary(opts: {
     { label: "App will run", value: styleUrl(issuer) },
   ];
 
+  // A project this command just created is never attached to a team yet, so the
+  // only question here is whether attaching is a thing at all on this server.
+  // `claimSummary` returns undefined off the cloud, which drops the row.
+  const ownership = claimSummary(claimState({ secret: {}, server }));
+  if (ownership) {
+    projectRows.push({ label: "Ownership", value: styleDim(ownership) });
```
bastionstack: Agreed the checkmark reads wrong. `renderSummary` hardcodes `pc.green("✓")` on every row (`summary.ts:45`), so it's a done-marker on the one row that isn't done, and dimming the value doesn't soften the glyph. The action line underneath already says the same thing, so the box says it twice. I'd drop the row.

apps/cli/src/commands/setup/index.ts
```diff
@@ -398,7 +412,11 @@ export default class Setup extends BaseCommand {
       // pre-coloured rows (path/url/id helpers) survive intact.
       consola.box({
         title: "Zitadel is ready",
-        message: [renderSummary(sections), "", installOutcome.boxActions.join("\n")].join("\n"),
+        message: [
+          renderSummary(sections),
+          "",
+          [...installOutcome.boxActions, ...claimNudge.actions].join("\n"),
```
bastionstack, line 423: `boxActions` is documented as a staged subset: customize/publish is held back until login works, and `nextActions` carries the full journey for agents (`install.ts:28-35`). The staging there is omission, not ordering, so landing this last doesn't make it post-verify. Either put it in `nextActions` only, or update the rationale in `install.ts`.

`status` gets this right, additive with no staging to respect.

apps/cli/src/commands/setup/index.ts
```diff
@@ -379,6 +381,18 @@ export default class Setup extends BaseCommand {
     });
 
     const writtenRel = allFilesWritten.map((file) => relativeDisplay(cwd, file));
+    // The nudge rides both surfaces from one decision: the box for humans, the
+    // envelope for agents. It lands after the install/verify actions because
+    // attaching a team is the step *after* the app demonstrably works, not a
+    // precondition for trying it. Empty off the cloud, where nothing can be
+    // attached.
+    const claimNudge =
```
bastionstack, line 394: Fires under `--dry-run` too, and the new test asserts it there. Dry run creates no project and writes no secret, so the nudge is about a project that doesn't exist. Gate on `!dryRun`, or say in the test that dry run previews the real output verbatim.

apps/cli/src/lib/claim-state.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Decides the attachment state from local state alone.
+ *
+ * **Why no network call.** The contract exposes no ambient claim-state read:
+ * `getClaimStatus` answers for one live challenge and rejects a secret that did
+ * not mint it, and claim attributes were deliberately kept off
+ * `GET /projects/{id}` because they belong to the grant, not the project. So
+ * `.zitadel/secret` is the only source, which is also why `status` and `doctor`
+ * stay fast and work offline. `zitadel claim` reaches the same conclusion the
```
bastionstack, line 42: It reaches it by copy: `claim.ts:101` still has `if (secret.claimed_at && secret.team_id)` inline. Point it at `claimState()` now the predicate has a home, otherwise the two drift the first time that record grows a field.

apps/cli/src/lib/project.ts
```diff
@@ -156,6 +157,20 @@ export function readUseCase(config: Record<string, unknown>): string | undefined
   return typeof config.useCase === "string" ? config.useCase : undefined;
 }
 
+/**
+ * Reads the top-level `server` origin from a parsed `zitadel.json`, falling
+ * back to {@link DEFAULT_SERVER} exactly as {@link resolveServer} does when the
```
bastionstack: Nit: `resolveServer` checks `environments.<env>.server` first (`server.ts:56`), so not quite "exactly as". Top-level-only is the right behaviour here, it's the parity claim I'd soften. No impact today since the patcher always writes a concrete top-level `server` (`base.ts:324`).

apps/cli/src/commands/doctor/checks/claim.ts
```diff
@@ -0,0 +1,81 @@
...
+    } catch {
+      // A missing or unparseable secret/config is already reported, loudly and
+      // with a repair path, by the `secret` and `config` checks. Implementing
+      // `SanityCheck` directly means nothing wraps a throw here, so swallowing
+      // it is what keeps a broken project reporting its real problem instead of
+      // crashing the whole battery on a nudge.
+      return {
+        name: this.name,
+        status: "pass",
+        message: "Skipped: could not read the project files that record the owning team",
```
bastionstack, line 44: Non-blocking: a `pass` reading "Skipped" means anything consuming these as data can't tell "couldn't work it out" from "attached". Fine for humans, the enum has no third state, and `secret`/`config` carry the real failure.

apps/cli/src/commands/doctor/checks/index.ts
```diff
@@ -15,6 +15,7 @@ import { DependencyCheck } from "./dependency";
 import { ManagedFilesCheck } from "./managed-files";
 import { ProjectMatchCheck } from "./project-match";
 import { SchemaCheck } from "./schema";
+import { ClaimCheck } from "./claim";
```
bastionstack, line 18: Nit: imports and re-exports here are alphabetical; this goes after `SchemaCheck` in all three spots.

IAM-marco: Checked this one and I think the premise is off: the file is in registry order, not alphabetical. Exports run Config, Secret, SecretPermissions, Gitignore, EnvExample, Framework, Schema, Dependency, ManagedFiles, ProjectMatch, which matches `SANITY_CHECKS` exactly (imports match too, bar Schema/Dependency). Alphabetical would put `ClaimCheck` first, before `ConfigCheck`, rather than after `SchemaCheck`.

The file header says to add a check "by appending an instance to the registry below", and appending last in all three spots is what keeps the export list and the registry in step, so I have left it where it is. Happy to reorder if you would rather the display order changed, but that is a different call from a sorting nit.

apps/cli/src/commands/doctor/checks/claim.ts
IAM-marco: Agreed, and leaving it. `CheckOutcome` has no "not applicable" state, and inventing one would touch every consumer of the enum to describe a case only this check can hit. As you say, `secret`/`config` carry the real failure with the repair path, so nothing is lost except precision in a field that is already advisory.

Worth noting the same `pass` also covers the off-cloud case ("not on a server where teams apply"), so a data consumer already cannot read `pass` as "attached" here. `data.project.claim` in `status` is the field to branch on: it is absent off the cloud and carries the explicit `kind` otherwise.

apps/cli/src/commands/setup/index.ts
IAM-marco: Took the second option: the test now says why. Dry run already previews the real output verbatim, so the install, start, and verify actions all name a project it never created (`project_id` is literally `dry-run-0000`). Gating only the claim nudge on `!dryRun` would make it the one action suppressed, which makes dry run a dishonest preview of exactly the surface the test covers.

Note this is the opposite call from `zitadel claim --dry-run`, which does stop early, because there the dry run would otherwise mint a real challenge on the platform. Setup dry run writes nothing, so there is nothing to protect against.

apps/cli/src/commands/setup/index.ts
IAM-marco: Dropped in 8e88e51e. The action line below the box already carries it, so the box now says it once.

apps/cli/src/commands/setup/index.ts
IAM-marco: Rationale replaced in 8e88e51e. You are right that the staging is omission rather than ordering, so appending was never a way of deferring it. Kept in `boxActions` (it is the only human surface now the row is gone) and the comment now gives the actual reason: attaching a team is orthogonal to the login journey, same as in `status`. Left `install.ts` alone, since what it returns is still the staged subset; the claim nudge is appended by the caller.

apps/cli/src/lib/claim-state.ts
IAM-marco: Good catch, that was the real drift risk. Both now call a shared `isAttached` in 8e88e51e.

Kept it as its own predicate rather than pointing `claim.ts` at `claimState()`: the command has to answer this on any server, but `claimState()` returns `not-applicable` off the cloud, so a local project with a recorded team would have stopped skipping and gone on to mint a challenge.

apps/cli/src/lib/project.ts
IAM-marco: Softened in 8e88e51e. The doc now states the difference instead of claiming parity, and says why top-level-only is what this caller wants: an environment override says where one invocation talks, not where the project lives.

PR #777, opened 2026-08-07
feat: streamline error code over oas

wim07101993:
## Summary

Some places used generated error codes, other specified them directly. This pr aims to stream line that by only using the generated codes.

## Validation

- `go generate ./...`
- `go test ./...`

## Release notes / changeset

- Changeset: .changeset/error-codes-in-oas.md — streamline error code in oas (list @zitadel/server for server changes)

## Notes

- I (Claude) had to change the error code generator.


api/internal/erroranalysis/flow.go
```diff
@@ -0,0 +1,927 @@
...
+func (f *frame) assign(s *ast.AssignStmt) bool {
+	changed := false
+
+	// errors.AsType[T](err) / errors.As(err, &t) hand the underlying error to
+	// their target. Without this the transaction-unwrap idiom
+	//   if de, ok := errors.AsType[domain.Error](err); ok { return de }
+	// would drop every error raised inside the transaction.
+	if len(s.Rhs) == 1 {
+		if call, ok := unparen(s.Rhs[0]).(*ast.CallExpr); ok {
+			if src, ok := f.errorUnwrapSource(call); ok && len(s.Lhs) > 0 {
```
livio-a, line 330: This binds the unwrapped codes to `s.Lhs[0]`, which is right for `errors.AsType[T](err)` (target is the first result) but wrong for `errors.As(err, &target)` — there the first LHS is the `bool`, and the real target is `call.Args[1]`'s pointee, which gets nothing. Also, `if errors.As(err, &de) { return de }` with no assignment isn't seen at all, since if-conditions aren't walked for calls.

The codebase idiom is `AsType`, so nothing breaks today — but one stray `errors.As` in a service path would silently drop codes from the spec, and the failure mode is exactly the client decode failure this PR exists to fix. Worth handling `errors.As`'s second arg explicitly, or at least documenting it as unsupported so a future refactor doesn't walk into it.

api/internal/erroranalysis/flow.go
```diff
@@ -0,0 +1,927 @@
...
+		// No body in the analyzed module: standard library, generated code, or
+		// an interface method that dispatch failed to resolve. Opaque.
+		return nil
+	}
+
+	key := fn.FullName() + "#" + b.key()
+	if cached, ok := a.cache[key]; ok {
+		return cached
+	}
+	if a.active[key] {
```
livio-a, line 37: Mutual recursion can poison the cache: while A is active, B's call to A contributes nil, and B's under-approximated result is then cached permanently — the intra-procedural fixpoint never re-evaluates it, so every future caller of B gets the truncated set. Probably unreachable in current service code, but cheap to guard: don't cache a result computed while any cycle member was active (track whether the walk hit an `active` function), or add a comment stating mutual recursion is out of scope.

api/cmd/gen_openapi_errors/main.go
```diff
@@ -579,6 +645,106 @@ type serviceMethod struct {
...
+//
+// Omitting one is not cosmetic: the generated client discriminates the error
+// response on `code`, so a code missing from an operation's oneOf makes a real
+// response fail to decode instead of surfacing as the error it is.
+//
+// Unlike boundaryErrors these are added per operation rather than per service
+// method, because they are the transport's to raise, not the service's. Folding
+// them into the service sets would report every hand-written "errors:" comment
+// as drifted for omitting something no service can return.
+var transportErrors = []string{
```
livio-a, line 704: This also lands on operations with no security requirement — `getHealth`, `getLive`, `getReady`, `getOpenIDConfiguration` now advertise `auth.unauthorized` they can never return. Over-advertising is harmless by your own (correct) argument, but it dilutes the precision that makes the typed models attractive. Gating `domain.ErrAuthUnauthorized` on whether the operation declares a security requirement would be cheap.

api/cmd/gen_openapi_errors/main.go
```diff
@@ -579,6 +645,106 @@ type serviceMethod struct {
...
+	keys := make([]string, 0, len(declared))
+	for key := range declared {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+
+	drifted := 0
+	for _, key := range keys {
+		want := declared[key].errorFuncs
+		got := inferred[key].errorFuncs
```
livio-a: `inferred[key]` is read without an existence check: a method skipped as `Unimplemented` yields a zero value, so every code its comment declares gets reported as "comment claims but code cannot return". Skip keys missing from `inferred` (or say "not analyzed") to keep the report trustworthy.

api/internal/erroranalysis/erroranalysis_test.go
```diff
@@ -0,0 +1,131 @@
...
+func moduleRoot(t *testing.T) string {
+	t.Helper()
+	_, file, _, ok := runtime.Caller(0)
+	if !ok {
+		t.Fatal("cannot resolve test file path")
+	}
+	return filepath.Join(filepath.Dir(file), "../../..")
+}
+
+func analyzeRepo(t *testing.T) map[string]Method {
```
livio-a, line 26: Each test runs a full `packages.Load` + whole-module analysis, so the suite pays that cost four times. A package-level `sync.Once` (or `TestMain`) sharing one result would cut it to one. Separately, `TestDumpAnalysis` is a logging aid that runs on every `go test ./...` — consider gating it behind `testing.Verbose()` or an env var.

internal/service/user.go
```diff
@@ -90,15 +104,15 @@ func NewUserService(
 	v2Pool StatementPool,
 	schemaStore domain.JSONSchemaStore,
 	hasher crypto.Hasher,
-) *UserService {
-	return &UserService{
+) UserService {
```
livio-a, line 113: Deliberate deviation from "accept interfaces, return structs", and justified — the generator's reflected floor only matches interface-typed handler fields. Add a one-line comment saying so, or someone will helpfully revert it and quietly shrink the reflected floor.

api/internal/erroranalysis/erroranalysis.go
```diff
@@ -0,0 +1,564 @@
+// Package erroranalysis infers, from the Go source itself, which domain error
```
livio-a, line 1: The doc comment is excellent. One addition worth making explicit for future maintainers: the analysis *requires generated code to exist* (it type-checks the module), which makes generation ordering a real constraint — see the overall review comment re: bootstrap.

livio-a (changes requested):
Really nice work — the analysis is careful where it matters (call-site-sensitive dispatch, inspected-vs-raised, the transaction-callback handling), the union-with-floor design fails in the safe direction, and the drift report already caught real rot in the `AuthAttemptService` comments. Two things before merge, a few smaller ones inline:

1. **Title/type:** CI flagged it too — this ships typed per-endpoint error models to `@zitadel/server` and `@zitadel/api` (the changeset says `minor`), so it's `feat:`, not `chore:`.

2. **Generation bootstrap coupling:** `gen_openapi_errors` used to only *parse* source; it now needs the whole module to *type-check* (`packages.Load` + hard exit on load errors). That means generation no longer works from a tree without `*_enumer.go` files. Today CI only regenerates over the committed tree so this is invisible, but it collides with the in-flight `refactor/faster-codegen` work, which runs the `api` generators before enumer and adds a pruned-bootstrap CI gate. Whichever lands second breaks. I'll adjust the phase ordering on my branch; flagging it here so we sequence the merges deliberately.

Also FYI: I resolved the merge conflict with main directly on the branch (1dbabf73). #799 had added structured field details to `SubmitFlowStep`'s decode-error path; the resolution keeps `WithDetails(...)` on your new `return nil, domErr` convention. Verified: 400 preserved through `errorResponse`, details survive `domainErrorDetails`, full regenerate is byte-identical, full test suite green.

api/internal/erroranalysis/flow.go
wim07101993: Fix: New bindAsTarget/bindAsTargets: propagates codes into the pointee of call.Args[1], wired into assignment RHS, if-conditions, switch tags, and
expression statements. Dropped "As" from errorUnwrapSource, since its first result is a bool.

api/internal/erroranalysis/flow.go
wim07101993: Fix: funcErrors tracks which active functions a walk bottomed out on and only caches when that set is empty (after removing its own key). Direct self-recursion still caches — its result is the exact least fixpoint; mutual recursion doesn't.

api/cmd/gen_openapi_errors/main.go
wim07101993: Fix: Split securityErrors out of transportErrors and gated it on the operation's effective security requirement (operation-level security overriding the root spec's).

api/cmd/gen_openapi_errors/main.go
wim07101993: Fix: Now skips with ℹ️ … not analyzed, doc comment not checked.

api/internal/erroranalysis/erroranalysis_test.go
wim07101993: Fix: sync.OnceValues shares one analysis; TestDumpAnalysis skips unless -v.

internal/service/user.go
wim07101993: Fix: Comment on NewUserService explaining the reflected floor only matches interface-typed handler fields.

api/internal/erroranalysis/erroranalysis.go
wim07101993: Fix: Added a # Prerequisites section: the analysis type-checks the module, so generated files must exist first.

I also added api/internal/erroranalysis/testdata/fixture/ — a two-package fixture pinning both fixed behaviours. I verified neither test is vacuous: reverting bindAsTarget fails AsPointerTarget/AsInCondition, and reverting the cache guard fails MutualB: missing domain.ErrFixtureA.

livio-a (approved):
All seven comments addressed — verified the fixes in the diff, not just the replies. The `bindAsTargets` wiring (including if-conditions and switch tags for the unassigned `errors.As` case) and the cycle-aware cache guard are both the right algorithms, the security gating shows up in the regenerated specs, and the testdata fixture is a nice bonus. Title fixed, CI green.

The bootstrap-ordering collision stays on my plate in #780 (enumer-before-api phase reorder) — no reason to hold this up for it. LGTM 🚀

PR #780, opened 2026-08-07
refactor: faster go codegen

wim07101993:
## Summary

I noticed that the `go generate ./...` command was getting slow on my computer so I started researching what was up:

## Impact                                                                                                                                                                                                               
                                                                                                                                                                                                                          
  | Command | Before | After | |                                                                                                                                                                                          
  |---|---|---|---|                                                                                                                                                                                                       
  | `go generate ./...` | 13.9s | **5.1s** | 2.7x |                                                                                                                                                                       
  | `moon run server:generate` / `server:check-generate` | 13.9s | **3.3s** | 4.2x |                                                                                                                                      

  ## Where the time went                                                                                                                                                                                                  
                                                                                                                                                                                                                          
  | Package | Before | After |                                                                                                                                                                                            
  |---|---|---|                                                                                                                                                                                                           
  | `internal/domain` | 5398ms | 2134ms |                                                                                                                                                                                 
  | `internal/service` | 5181ms | 886ms |                                                                                                                                                                                 
  | `internal/crypto` | 1219ms | 612ms |                                                                                                                                                                                  
  | `api` | 432ms | 469ms |                                                                                                                                                                                               
  | everything else | ~1000ms | ~1300ms |                                                                                                                                                                                 
  | **serial total** | **~13.9s** | **~5.1s** |                                                                                                                                                                           
  | **wall clock, packages in parallel** | — | **~3.3s** |                                                                                                                                                                
                                                                                                                                                                                                                          
  The two numbers differ because mockgen's cost is per *invocation*, not per interface:                                                                                                                                   
  it type-loads the package before mocking anything, so a directive covering one                                                                                                                                          
  interface costs the same ~800ms as one covering twenty. `internal/service` had six                                                                                                                                      
  such directives and paid the load six times.                                                                                                                                                                            
                                                                                                                                                                                                                          
  After the change the run is no longer a sum — it's a critical path of                                                                                                                                                   
  `api (469ms) + internal/domain (2134ms)`, with every other package finishing                                                                                                                                            
  underneath it.                                                                                                                                                                                                          
                                                                                                                                                                                                                          
## Validation

- `go generate ./...`

## Release notes / changeset

- changeset: .changeset/parallel-go-codegen.md - Make server codegen parallel and able to cleanup

## Notes

- I also added a command to clean all generated code so that ci can detect stale generated files


scripts/go-generate.mjs
```diff
@@ -0,0 +1,137 @@
...
+
+import { isDirectRun, mapWithConcurrency, run, runCapture } from "./dev-process.mjs";
+
+const ROOT = fileURLToPath(new URL("..", import.meta.url));
+
+// Never hold Go sources, and walking them dwarfs the rest of discovery.
+const SKIP_DIRS = new Set(["node_modules", ".git", ".moon", "dist", "target", ".next", ".turbo"]);
+
+// Runs alone in phase 1 — see the module comment.
+const ORDERED_FIRST = "api";
```
livio-a: **Blocking — this is the current `full-pr` failure.** Since #777, `gen_openapi_errors` type-checks `./internal/...` via `erroranalysis` (`packages.Load` with `NeedDeps`, hard-failing on any load error in `erroranalysis.go:208`). Running `api` first from a pruned tree now fails on two distinct missing prerequisites:

1. `undefined: domain.FlowDefinitionTransitionActionString` (+6 siblings) — the pruned `*_enumer.go` files, because enumer runs after api.
2. `could not import github.com/zitadel/nextgen/api/generated` — the pruned ogen output. This one is a genuine cycle: `internal/service` imports `api/openapi/endpoints/flow_definitions`, whose `embed.go` imports `api/generated`, so the type-check needs the *previous run's* ogen output — but ogen is the last directive in api's own chain (it consumes the yaml this very generator emits). Narrowing erroranalysis's load patterns wouldn't help; the dependency is transitively real.

Suggested fix, two parts:

- Reorder phases to **enumer → api → mockgen**. Enumer needs nothing from api, and this dissolves failure 1.
- **Drop `api/generated/**/*.go` from `GENERATED_GLOBS`.** ogen runs with `--clean`, so it already self-prunes orphans — external pruning buys nothing there and is what creates cycle 2. A comment on the glob list ("self-cleaning generators need no external prune") would keep the invariant legible.

The yamls are safe to keep pruning — I checked that `flow_definitions/embed.go` does not `go:embed` the `*-error-response.yaml` files, so they aren't part of the load graph.

(The alternative — making `erroranalysis` tolerate load errors in packages it doesn't index — would weaken #777's hard-fail-on-drift design, so I'd only go there if the exemption turns out untenable.)

scripts/go-generate.mjs
```diff
@@ -0,0 +1,137 @@
...
+ * Runs the module's `//go:generate` directives, packages in parallel.
+ *
+ * `go generate ./...` is strictly serial, and the directives are almost pure
+ * per-invocation overhead — mockgen spends ~800ms type-loading a package
+ * whether it mocks one interface or twenty. Running packages concurrently is
+ * therefore nearly a free 4x.
+ *
+ * The ordering that matters is a phase boundary, not a package-internal one:
+ * `api/cmd/gen_error_schemas` and `api/cmd/gen_openapi_errors` parse
+ * `internal/domain` and `internal/service` *source* to discover error codes,
```
livio-a: This premise is stale as of #777: the api generators no longer parse *source* — `gen_openapi_errors` type-checks `./internal/...`. The constraint has reversed direction: api doesn't just need the trees quiescent, it needs the module to *type-check*, which is exactly why it must run after enumer (see the blocking comment below).

Related doc claims that need the same update once the ordering is fixed:
- line ~59 here: "Both paths bootstrap from a tree with no generated files in it"
- AGENTS.md: "Generation bootstraps from an empty tree, so cleaning is safe at any time"
- the changeset: "Bare `go generate ./...` still produces identical output" / bootstrap claims

Bare `go generate ./...` also visits `api` before `internal/...` (import-path order), so a *fully* empty tree can't bootstrap by either path post-#777. With `api/generated` exempted from pruning, the supported path (`clean-generated && generate`) never produces a fully empty tree — so the honest phrasing is "bootstraps from a *cleaned* tree".

scripts/go-generate.mjs
```diff
@@ -0,0 +1,137 @@
...
+  // imports internal/domain, so its mockgen needs *domain's* enumer output,
+  // not just its own package's.
+  for (const phase of GENERATE_PHASES) {
+    await runPhase(rest, phase, log);
+  }
+}
+
+// `go generate -run` filters directives by matching against the command text.
+// A generator added later that neither writes into nor loads a package can go
+// in either phase; one that does both needs its own entry here.
```
livio-a: This isn't quite right — a directive outside `api/` that matches neither `-run enumer` nor `-run mockgen` (say, `stringer`) doesn't go "in either phase", it goes in **neither**: the moon task silently never runs it, while bare `go generate ./...` would. And because `check-generate` uses this same runner, the drift is undetectable — the output simply never comes into existence.

Since `findGeneratePackages` already reads every directive-bearing file, consider collecting the directive command lines and failing loudly when one matches no phase filter (excluding `ORDERED_FIRST`). That turns a silent gap into a one-line error at the moment someone introduces the third generator.

scripts/generated-outputs.mjs
```diff
@@ -0,0 +1,54 @@
...
+ * "generated by …". Checking it is the safety net that makes deletion safe:
+ * a glob that ever matches a hand-written file is caught here instead of
+ * silently eating someone's work.
+ */
+const MARKER = /^(\/\/|#).*generated by .*DO NOT EDIT/im;
+
+export function hasGeneratedMarker(absPath) {
+  // The marker is always in the header; reading whole files to find it would
+  // mean slurping all of api/generated for nothing.
+  return MARKER.test(readFileSync(absPath, "utf8").slice(0, 512));
```
livio-a: Nit: the comment says reading whole files "would mean slurping all of api/generated for nothing", but `readFileSync(absPath, "utf8").slice(0, 512)` reads the whole file and slices in memory anyway. Either read 512 bytes from a file descriptor (`open` + `read` into a buffer) or drop the comment — as-is it claims an optimization the code doesn't perform.

.changeset/parallel-go-codegen.md
```diff
@@ -0,0 +1,41 @@
+---
+---
```
livio-a, line 2: Question: the frontmatter is empty, i.e. no package version bump. Intentional (server isn't independently versioned), or should this declare a bump?

livio-a (changes requested):
Really nice work overall — the consolidation is exactly right (I verified every removed directive against the consolidated lists: crypto 6, domain 7, service 32 interfaces, all carried over), the new globs close real coverage gaps the old hand-maintained `GENERATED_PATHS` had (`idgenmock/`, `user_examples/`), and the double-guarded prune (precise globs + DO NOT EDIT marker) is the right amount of paranoia.

Requesting changes for one reason: **#777 (merged today, and now merged into this branch) breaks the pruned bootstrap, and `full-pr` currently fails in `server:check-generate-pruned`** — see https://github.com/zitadel/nextgen/actions/runs/31484590117/job/93756970995. The root cause is a bit deeper than a phase reorder; details and a suggested fix in the inline comments on `scripts/go-generate.mjs`. The remaining comments are smaller (one silent-skip gap worth closing, two nits).

scripts/go-generate.mjs
wim07101993: Thanks — the diagnosis is exactly right, and I took the first half of the fix (phases are now `enumer → api → mockgen`). I went a different way on the second half, so flagging it explicitly rather than quietly diverging.

**Instead of exempting `api/generated` from the prune, `gen_openapi_errors` now breaks its own cycle.** When `api/generated` is absent it runs two prerequisites before analyzing:

1. `go generate -run "go tool enumer" ./internal/...` — dissolves failure 1 from inside the generator, so it holds for bare `go generate ./...` too, which reaches `internal/` only after `api`;
2. `go generate -run "go tool ogen" .` over a placeholder spec it writes first, giving every operation every known error code. Wrong but complete, and deliberately a superset — whatever types the honest pass makes ogen generate, the placeholder made it generate too. That closes cycle 2.

The analysis then runs against a module that type-checks, the real sets overwrite the placeholders, and the ogen directive that follows in `api/generate.go` regenerates the client from them. Both prerequisites shell out to `go generate -run` rather than restating the commands, so ogen's flags stay declared only in `api/generate.go`.

Why not the exemption: you're right that `--clean` self-prunes, so external pruning buys nothing *for ogen*. But it costs something to drop it — `check-generate-pruned` stops proving that generation works from a genuinely empty tree, and the runner keeps a special case for one generator's cleanup semantics. With the cycle broken at its source, `api/generated/**/*.go` stays in `GENERATED_GLOBS` and the honest phrasing stays the strong one: both paths bootstrap from a tree with **no** generated files in it.

Verified:

| | result |
|---|---|
| `moon run server:check-generate-pruned` | pass, 10.7s |
| prune (163 files) + bare `go generate ./...` | exit 0, **0 files unrestored or changed** vs. committed |
| `moon run server:check-generate` (unpruned) | pass, 7.4s |

Cost is ~2s on the pruned path only. `bootstrapGeneratedClient` runs enumer unconditionally instead of checking disk first — under the Moon runner that's one redundant enumer run, which I kept deliberately so both entrypoints walk identical code and `check-generate-pruned` actually covers the `./...` bootstrap rather than a branch CI never takes.

If you'd still rather have the exemption, say so and I'll swap it in — but I'd then want to keep the self-bootstrap anyway, since it's what makes bare `go generate ./...` work from empty.

scripts/go-generate.mjs
wim07101993: Fixed — the module comment now says `gen_openapi_errors` type-loads the whole module, and that this is why `api` runs after enumer.

On the related doc claims: they're updated, but in the *stronger* direction rather than the hedged one. Since the generator now bootstraps its own prerequisites (see the thread above), a fully empty tree does regenerate by both paths, so "bootstraps from a *cleaned* tree" would be understating it. Concretely:

- line ~59 here → "Both paths generate from a tree with no generated files in it at all; there is a standing check for that (`server:check-generate-pruned`)."
- `AGENTS.md` → the pruned-tree paragraph now explains where the cycle lives and that the generator resolves it, with a "do not reintroduce a scheduler-side bootstrap pass" note.
- the changeset → keeps "Bare `go generate ./...` still produces identical output", and adds a paragraph on why `./api` needed the same treatment as the mockgen directives.

Both claims are checked, not just asserted: prune + bare `go generate ./...` restores all 163 files byte-identically, and `check-generate-pruned` is green.

scripts/go-generate.mjs
wim07101993: Good catch — "goes in neither" was the actual behavior, and the comment said the opposite. Closed as you suggested, since the files were already being read to find the directories.

`findGeneratePackages` now returns a `Map<dir, directives[]>`, and `assertEveryDirectiveHasAPhase` runs before anything else:

```
No generation phase claims these //go:generate directives:
  internal/phasecheck_tmp: //go:generate go tool stringer -type=Fake

This runner only ever runs -run enumer, -run mockgen, so a directive matching none of them never
runs here — and never shows up as drift, because its output is never written.
Add a phase for it in scripts/go-generate.mjs, ordered against the existing ones.
```

`ORDERED_PACKAGE` is exempt — `api` runs unfiltered, as its whole chain in file order. Verified by planting exactly that `stringer` directive: exit 1 from both `go-generate.mjs` and `check-go-generate.mjs`, so the moon task fails at the moment someone introduces the third generator rather than silently skipping it. All 21 current directives are claimed.

scripts/generated-outputs.mjs
wim07101993: Right — the comment claimed an optimization the code didn't perform. Took the first option: it now reads 512 bytes from a descriptor rather than slurping and slicing.

```js
const fd = openSync(absPath, "r");
try {
  const header = Buffer.alloc(HEADER_BYTES);
  const read = readSync(fd, header, 0, HEADER_BYTES, 0);
  return MARKER.test(header.toString("utf8", 0, read));
} finally {
  closeSync(fd);
}
```

The comment now also notes the one thing a byte-window read has to answer for: a multi-byte rune split at the boundary decodes to a replacement char, which the marker never contains. Still finds all 163 generated files and still leaves unmarked strays alone.

.changeset/parallel-go-codegen.md
wim07101993: Intentional. It's the last row of the [decision table](https://github.com/zitadel/nextgen/blob/main/.changeset/README.md#decision-table) — "touches a publishable path but ships nothing (rare) → empty changeset".

The PR touches `internal/**` and `api/**`, but the generated output is byte-identical to what's on `main` (that's what `check-generate` asserts), so there's no server behavior to bump a version for. The file exists rather than being omitted because the changelog entry is still worth having for the next person who wonders why the mockgen directives moved. 24 other changesets in the tree are empty for the same reason — `moon-ci-release-rebuild`, `try-depot-runners`, `stabilize-spanner-tests`, and so on.

Title stays `refactor:` for the same reason.

api/cmd/gen_openapi_errors/main.go
```diff
@@ -104,48 +149,57 @@ func main() {
...
+
+	// Step 3: On a pruned tree the analysis below has nothing to analyze — the
+	// module does not type-check until ogen has run, and ogen cannot run until
+	// this generator has written a spec. Break the cycle with a placeholder
+	// pass; see the package comment.
+	if !generatedClientExists(moduleRoot) {
+		bootstrapGeneratedClient(moduleRoot, apiDir, func() {
+			allErrorFuncs := slices.Sorted(maps.Keys(errorMapping))
+			writeErrorResponses(endpoints, tmpl, errorMapping, openapiBasePath,
+				func(endpointOperation) ([]string, bool) { return allErrorFuncs, true })
```
IAM-marco: The bootstrap pass and the honest pass disagree about which operations get a file, so the output depends on whether the tree was pruned.

This closure returns `true` for every endpoint, so the placeholder pass writes an `<operationID>-error-response.yaml` for all of them, each with the full superset of error codes. The honest pass does not: an operation with neither a handler nor a reflected service mapping returns `false` and is skipped ([#L219-L222](https://github.com/zitadel/nextgen/blob/289a9af69707a368bdd0d8e0a7ff55744386b068/api/cmd/gen_openapi_errors/main.go#L219-L222)), and `writeErrorResponses` just `continue`s ([#L259-L262](https://github.com/zitadel/nextgen/blob/289a9af69707a368bdd0d8e0a7ff55744386b068/api/cmd/gen_openapi_errors/main.go#L259-L262)).

What follows: on a pruned tree such an operation keeps the placeholder file, listing every error code the module can raise. On a populated tree it keeps whatever was committed. Same generator, same inputs, different bytes depending on starting state, which is the one property `check-generate-pruned` is meant to guarantee. It surfaces as an unexplained diff in CI rather than as the `No reflected service mapping` warning that actually describes the cause.

This may be dormant right now, but the skip branch exists and warns, so it is expected to fire, and a newly specced endpoint whose handler is not written yet lands exactly on it.

Fix: make the skip authoritative. When `errorFuncsFor` returns `false`, remove any existing file for that operation instead of leaving it. That also matches what the rest of the PR is fixing, since a stale file left behind by a skip is the same orphan class as the nine deleted here.

Coverage: an operation with a default error response and no handler would then be caught by `server:check-generate-pruned` as a deletion, which is the readable failure.

apps/server/moon.yml
```diff
@@ -57,10 +65,41 @@ tasks:
...
+      cache: false
+      runInCI: false
+      runFromWorkspaceRoot: true
+
   check-generate:
     command: "node scripts/check-go-generate.mjs"
     inputs:
       - "@group(serverInputs)"
+      - "/scripts/go-generate.mjs"
+      - "/scripts/dev-process.mjs"
```
IAM-marco, line 86: `check-generate` reads two scripts it does not declare as inputs.

`check-go-generate.mjs` imports `clean-generated.mjs` and `generated-outputs.mjs` at module scope, on every run and not only under `--prune`: [scripts/check-go-generate.mjs#L15-L18](https://github.com/zitadel/nextgen/blob/289a9af69707a368bdd0d8e0a7ff55744386b068/scripts/check-go-generate.mjs#L15-L18). `PATHSPECS` is derived from `GENERATED_GLOBS` ([#L20-L23](https://github.com/zitadel/nextgen/blob/289a9af69707a368bdd0d8e0a7ff55744386b068/scripts/check-go-generate.mjs#L20-L23)) and is what `gitState()` compares, so that glob list decides what this check looks at.

What follows: editing `GENERATED_GLOBS` does not change this task's input hash, so `check-generate` cache-hits and reports a pass computed against the old pathspecs. `vet` and `test` both depend on it, so the stale result propagates.

`check-generate-pruned` right below lists all four scripts ([#L95-L102](https://github.com/zitadel/nextgen/blob/289a9af69707a368bdd0d8e0a7ff55744386b068/apps/server/moon.yml#L95-L102)); the same two lines are missing here. This is the same class as Copilot's `dev-process.mjs` note, which was fixed, but those two files were not added. Same discipline as the `openapi` task's "Without these a lint-config change still cache-hits".

Fix: add `/scripts/clean-generated.mjs` and `/scripts/generated-outputs.mjs` to these inputs.

No coverage change needed; this is Moon cache metadata, and CI exercises it on the next glob edit.

AGENTS.md
```diff
@@ -293,8 +293,42 @@ call `idgen` outside dialect packages. Full minting contract:
 ## Generated Files
 
 - Do not hand-edit `api/generated/**`; update `api/openapi/**` and run
-  `moon run server:generate` or `go generate ./...`. CI enforces committed
-  generated output through `server:check-generate` (via `server:test`).
+  `moon run server:generate`. CI enforces committed generated output through
+  `server:check-generate` (via `server:test`). Bare `go generate ./...` produces
+  the same output, but runs every directive serially, roughly 4x slower. Both go
```
IAM-marco: This says bare `go generate ./...` is "roughly 4x slower", but the module comment added in the same PR retracts that number: "~21% end to end (6.4s against 8.1s), not the 4x the arithmetic suggests" ([scripts/go-generate.mjs#L11-L16](https://github.com/zitadel/nextgen/blob/289a9af69707a368bdd0d8e0a7ff55744386b068/scripts/go-generate.mjs#L11-L16)), because `./api` is a serial chain and now the larger half of the run.

Worth correcting here rather than leaving the two in conflict: `AGENTS.md` is what agents read, and 4x is the number that makes someone reach for this task over the plain one.

Same bullet, [line 330](https://github.com/zitadel/nextgen/blob/289a9af69707a368bdd0d8e0a7ff55744386b068/AGENTS.md#L328-L331): "Generation restores everything the clean removed" reads as a flat contradiction of the paragraph four lines above it, which is entirely about output that does not come back. "Generation restores everything that is still generated" keeps the guarantee without undoing the point.

scripts/go-generate.mjs
livio-a: Accepted — keep the self-bootstrap, don't swap in the exemption. Your trade-off analysis is right: `--clean` only self-prunes ogen's own tree, while the exemption would have quietly weakened what `check-generate-pruned` proves (empty-tree bootstrap) and left the runner carrying a special case for one generator's cleanup semantics. Breaking the cycle inside `gen_openapi_errors` keeps the strong property and makes bare `go generate ./...` work from empty too, which the exemption never could.

Also noting the pruned check immediately paid for itself — the 9 orphaned `*-error-response.yaml` files it flushed out (endpoints whose `methods.yaml` is long gone) are exactly the class this PR exists to catch.

Resolving this thread. Approval is waiting on the three open comments from @IAM-marco — in particular the bootstrap/honest-pass divergence on skipped operations, which is the one substantive gap left in the new approach.

.changeset/parallel-go-codegen.md
livio-a: Confirmed against the decision table — last row applies, empty changeset is correct. Resolving.

api/cmd/gen_openapi_errors/main.go
wim07101993: Confirmed and fixed — the divergence was real. Two passes, one writing a file for every endpoint and one writing a file for a subset, means the skipped operation's bytes are whatever the starting state left there. Dormant today (the run emits zero `No reflected service mapping` warnings), but as you say, the branch exists to fire.

I took a different fix than deleting, because **delete isn't available for most of these operations**: 32 of the 49 discovered operations `$ref` their error-response file by name from `methods.yaml` (`default: {$ref: setUserPassword-error-response.yaml}`); the other 17 reach it through the `error-details.yaml` content shape. Deleting a file in the first group leaves an unresolvable spec, not a smaller one — and on a pruned tree it would fail *during generation*: bootstrap writes all 49 → ogen succeeds → honest pass deletes one → the ogen directive that follows in `api/generate.go` fails on the dangling `$ref`. That trades a silent divergence for a broken build whenever an endpoint is specced ahead of its handler, which is the exact case you called out.

So instead the skip is gone: `errorFuncsFor` returns `[]string` rather than `([]string, bool)`, and an operation with neither a handler nor a reflected service mapping gets `boundaryErrors` — what the boundary raises with no handler behind it — plus the transport and security floor every operation gets. Same warning, but the file is written deterministically instead of being left as found:

```
⚠️  No handler or reflected service mapping for getFoo (…/methods.yaml); writing boundary errors only
```

`writeErrorResponses` now documents the invariant as the reason it takes no skip signal at all: *every* operation, never a subset, because a subset makes output depend on what was already on disk.

I closed the sibling hole in the same loop while I was there: `len(errorCodes) == 0` also used to `continue`, leaving a stale file for the same reason. It is unreachable while `transportErrors` maps, so it now exits 1 naming the operation and the unresolved constructors rather than skipping quietly.

No output change — the branch is still dormant, and `check-generate-pruned` plus prune + bare `go generate ./...` both come back byte-identical.

apps/server/moon.yml
wim07101993: Right — both are module-scope imports, and `GENERATED_GLOBS` is what `gitState()` compares through `PATHSPECS`, so a glob edit could cache-hit a pass computed against the old list. Added both, with the reason inline so the next person doesn't trim them:

```yaml
      # Imported at module scope, not only under --prune: GENERATED_GLOBS is
      # what this task's git comparison looks at, so a glob edit that does not
      # invalidate here would cache-hit and pass against the old pathspecs.
      - "/scripts/clean-generated.mjs"
      - "/scripts/generated-outputs.mjs"
```

That gap turned out to be more than theoretical — I edited `generated-outputs.mjs` in this round (dot-directory matching, see the PR comment) and it's exactly the kind of change that would have gone unvalidated by a cached `check-generate`. `check-generate` and `check-generate-pruned` now declare the same four scripts.

AGENTS.md
wim07101993: Both fixed.

On "4x": rather than swapping in another number that will rot the same way, I dropped the multiplier and point at the one place that's measured:

> Bare `go generate ./...` produces the same output, but runs every directive serially — slower, though not by the multiple the parallelism suggests, because `./api` is itself a serial chain and the larger half of the run (`scripts/go-generate.mjs` carries the numbers).

That way `AGENTS.md` says which task to prefer and why, and the numbers live next to the code that produces them. (For what it's worth, on this machine today the two are within ~7% of each other unpruned — 6.9s vs 7.4s — so 4x was badly stale in the direction that matters.)

Second one taken almost verbatim, with the point made explicit since the contradiction was four lines away:

> Generation restores everything that is still generated — which is the point: what stays missing was the orphan — so cleaning is safe at any time.

wim07101993:
## The prune had a blind spot, and it was hiding a tenth orphan

Found while checking @IAM-marco's divergence report, so flagging it separately — it's a bug in this PR's own feature rather than in a review comment.

`collectGeneratedFiles` used `fs.globSync`, and **Node's glob silently skips dot-directories**. `api/openapi/endpoints/.well-known/openid-configuration/` is generated output, so everything under it was invisible to `clean-generated` — and therefore to `check-generate-pruned`, which can only report an orphan as "deleted and not restored".

```
$ node -e '...' # api/openapi/endpoints/**/*-error-response.yaml
default   : 49        # fs.globSync
dot:true  : 49        # not an option Node honors
$ git ls-files -- ':(glob)api/openapi/endpoints/**/*-error-response.yaml' | wc -l
50
```

That 50th file is `getOpenIDConfiguration-error-response.yaml`, sitting in a directory that holds nothing else — its `methods.yaml` went away with the OIDC surface cleanup in #778, and the generated file has been committed and unreferenced ever since. Nothing generates it, nothing `$ref`s it, and the prune could not see it. So @livio-a's "the pruned check immediately paid for itself" is right, and it was one short: 9 flushed, a 10th out of reach.

The mismatch mattered beyond this one file, because the two halves of the check disagreed about what "generated" means: `clean-generated` deletes what Node's glob matches, while `check-go-generate` compares what git's `:(glob)` pathspecs match — and git has no dotfile rule. Anything orphaned under a dot-directory falls in the gap between them, permanently.

**Fix:** `generated-outputs.mjs` now matches the globs itself against a `readdirSync` walk, with git's semantics — `**` spans directories, `*` stays in one segment, a leading dot is not special. The two matchers now return the same set (164 = 164), so the cleaner deletes exactly what the checker compares.

`check-generate-pruned` then did its job on the first run:

```
Generated files that nothing generates any more:
  api/openapi/endpoints/.well-known/openid-configuration/getOpenIDConfiguration-error-response.yaml

A directive stopped producing these — a renamed destination, or an
interface no longer mocked. Commit their deletion.
```

Deletion committed, the empty directories are gone, and both paths are green again — `check-generate-pruned` passes, and prune + bare `go generate ./...` restores all 163 remaining files byte-identically.

livio-a (approved):
All threads resolved - LGTM

PR #783, opened 2026-08-07
fix: reject user property name containing dot

### LINKED ISSUE #771: bug(flow-engine): nested user-schema fields are unusable
## Summary

## Repro

```json
{
  "type": "object",
  "properties": {
    "address": {
      "type": "object",
      "required": ["street"],
    }
  }
}
```

| `fields` | Result |
|---|---|

## Root cause

  `FlowFieldTypeText`.

three defects.

## Expected

  keywords and `x-unique`.

vitorbari:
## Summary

Update the user-meta-schema to prevent user property names containing dot `.`. 

Screenshot 2026-08-07 at 17 46 12

Motivation:

Nested properties will be referenced with a dot notation on the flow-definition, so a property name itself can not contain a dot.

Flow-definition example:

```json
{ "name": "address", "fields": ["address.street", "address.city"]}
```

## Notes

- Applies to the `properties` chain at every depth. A name hidden under `$defs`, `allOf`, or
  `items` is not caught. (to be fixed later)
- Only applies where `ValidateAgainstMetaSchema` runs. A schema the resolver fetches from a URL
  is persisted without it. Unchanged by this PR.


fforootd (approved):

api/openapi/endpoints/schemas/user-property.json
```diff
@@ -70,7 +70,13 @@
     "properties": {
       "type": "object",
       "description": "A map of additional properties for the user definition, where the key is the property name and the value is the property schema",
-      "additionalProperties": true
+      "propertyNames": {
+        "pattern": "^[^.]+$",
+        "description": "The name of a single user attribute."
+      },
+      "additionalProperties": {
+        "$ref": "user-property.json"
```
fforootd, line 77: The wire-contract mirrors don't carry the new rule: `user-property.yaml` still has `additionalProperties: true` in its nested `properties` map, and `user-schema.yaml` has no `propertyNames` — so API docs and generated clients won't surface a constraint the server now rejects. The yamls deliberately drop the `allOf` for codegen compat, but `propertyNames` + `pattern` is plain OAS 3.1 and likely survives generators. Worth mirroring there in a follow-up, or at minimum stating the dot rule in the yaml `description`.

fforootd (commented):
Non-blocking recommendations, details inline — approval separate.

Verified locally on the PR head: `TestTenantSchemaValidator_ValidateAgainstMetaSchema` passes (both new cases), all 122 `config` tests pass (drift audit, Ajv dialect compile, shipped defaults/presets). The new `user-property.json` self-reference makes the schema graph cyclic, which I checked in all three consumers: the Go compiler caches the schema before `Resolve` (`json_schema.go`), Ajv resolves via filename registration, and editors resolve relative to `.zitadel/meta/`. Also nice that this lands before dot-notation flow fields ship — no grandfathered names to migrate.

.changeset/reject-dotted-schema-property-names.md
```diff
@@ -0,0 +1,9 @@
+---
+"@zitadel/server": minor
+"@zitadel/config": minor
+---
+
+A user schema property name must be a single attribute name and may no longer
+contain a dot. The rule lives in the user-schema meta-schema, so an editor
+validating against the shipped dialect flags it while authoring, and the server
+rejects it on create.
```
fforootd: Recommend one more sentence here: the `additionalProperties: true` → `$ref` recursion is a broader tightening than the dot rule. Nested property values must now be objects valid against `UserProperty` (and draft 2020-12), so e.g. a boolean subschema `"foo": true` or `{"type": 123}` two levels down passed before and is rejected now. That's the part most likely to surprise someone resubmitting an existing schema, so the release notes should mention it.

packages/config/meta-schemas/user-schema.json
```diff
@@ -42,6 +42,10 @@
     "properties": {
       "type": "object",
       "description": "A map of additional properties for the user definition, where the key is the property name and the value is the property schema",
+      "propertyNames": {
+        "pattern": "^[^.]+$",
```
fforootd: The changeset's headline claim is that editors flag this while authoring, but `meta-schemas.test.ts` has no negative case for the new rule (the branding dialect tests do assert negatives, e.g. the `http://` logo URL). A two-line Ajv case — dotted name fails, clean name passes — would pin the editor-facing artifact independently of the Go suite.

api/openapi/endpoints/schemas/user-schema.json
```diff
@@ -42,6 +42,10 @@
     "properties": {
       "type": "object",
       "description": "A map of additional properties for the user definition, where the key is the property name and the value is the property schema",
+      "propertyNames": {
+        "pattern": "^[^.]+$",
```
fforootd: On the "to be fixed later" note: the escape-hatch list is a bit wider than `$defs`/`allOf`/`items` — since `UserProperty` keeps top-level `additionalProperties: true` and embeds the full 2020-12 meta-schema, `patternProperties`, `additionalProperties`-as-schema, and `oneOf` branches are the same class. When picking it up: the textbook mechanism is a `$dynamicAnchor: meta` dialect extension, which makes the constraint recursive across every subschema position in one move; an imperative walk at flow-definition validation time may end up simpler.

api/openapi/endpoints/schemas/user-property.json
vitorbari: I've updated the YAML files as well

.changeset/reject-dotted-schema-property-names.md
vitorbari: Updated

packages/config/meta-schemas/user-schema.json
vitorbari: Added to meta-schemas.test.ts

api/openapi/endpoints/schemas/user-schema.json
vitorbari: Documented the exceptions in the design document: docs/design/flowengine/user-schema.md

api/openapi/endpoints/schemas/user-property.yaml
```diff
@@ -46,6 +46,10 @@ properties:
   properties:
     type: object
     description: |
-      A map of additional properties for the user definition, where the key is 
+      A map of additional properties for the user definition, where the key is
       the property name and the value is the property schema.
-    additionalProperties: true
\ No newline at end of file
+    propertyNames:
+      pattern: "^[^.]+$"
```
wim07101993: maybe make property name a separate schema, that way we only need to reference it once

vitorbari: Good idea 👍

wim07101993 (approved):

PR #784, opened 2026-08-07
fix(config): strip schema property defaults at every level

### LINKED ISSUE #771: bug(flow-engine): nested user-schema fields are unusable

vitorbari:
## Summary

Part of supporting nested properties in a user schema. Without it, `zitadel plan` can report a change on every run once a schema has properties nested inside an object, even when nothing has changed.

Three `x-*` property settings have defaults (`x-editable: true`, `x-sensitive: false`, `x-mfa: false`), so a property that spells one out is semantically identical to one that omits it. To compare local config against the platform fairly, the CLI strips those three from both sides before diffing.

It only stripped them from the top layer. Spell out `x-editable: true` on a property nested inside an object, apply, then delete that line — nothing changed semantically, but the two sides no longer matched. `plan` reported a change, `apply` published a new schema revision, and the next `plan` reported it again.

## One-time state-hash churn

`state.json` hashes are computed over the normalized form, so widening the normalizer changes the hash for the bodies it newly touches. A schema that spells out a default on a nested property therefore reads as changed on the first `plan` after upgrading: a revision with an empty field diff, published once, after which the new hash is stored and every later run is a skip. Schemas without a spelled-out nested default are unaffected.

This was a deliberate call over retaining the previous normalizer as a fourth accepted hash format. The affected population is small — nested leaves are not collectable by a flow until #785, so there is little reason to have authored a nested object yet — and the alternative sets a precedent that every future normalizer change carries its predecessor forever. It is also a bounded version of the cost that exists today: on `main` these schemas republish on *every* run rather than once. The changeset states it, and `loop.ts` records the reasoning next to the legacy-hash acceptance list.

## Docs correction

`apps/docs/content/docs/cli/plan-apply.mdx` claimed the platform stores a canonicalized form with meta-schema defaults spelled out. It stores the uploaded bytes verbatim (`NewJSONSchema` in `internal/domain/json_schema.go` keeps the document as-is). The difference decides whether deleting a spelled-out `x-editable: true` is safe, so the sentence now describes the real model and notes the rule holds at any depth.

## Validation

Reproduced end to end against a local server (`go run . server`) with a nested `address.street`:

1. set `"x-editable": true` on `address.street`, `cli apply` — the platform now holds the spelled-out default
2. delete that line locally
3. `cli plan` twice, then `cli apply`

On this branch all three report no changes. On `main` the same state diffs every run, because `normalizeSchemaBody` never descends past the top level and so cannot see the two are equivalent.

The immutability test is mutation-checked: cloning only to depth one — the plausible optimisation, since the top level is all the old code touched — fails the new nested test and nothing else.

- `moon run config:test` (153 passed), `moon run config:typecheck`, `moon run cli:typecheck`

## Release notes / changeset

- Changeset: `.changeset/normalize-nested-schema-defaults.md` — patch for `@zitadel/config`.

## Notes

- Independent of the nested-schema-fields stack (#785, #786, #804); base is `main` and it can merge on its own.
- The recursion follows `properties` only, skipping `$defs`, `items`, `allOf`/`oneOf`/`anyOf`, `patternProperties` and `additionalProperties`-as-schema. That matches the documented contract in `docs/design/flowengine/user-schema.md` and the recursive `$ref` in `meta-schemas/user-property.json`, both of which thread only through `properties`.


packages/config/src/normalize.ts
```diff
@@ -84,10 +84,20 @@ export function normalizeFlowBody(body: object): object {
  */
 export function normalizeSchemaBody(body: object): object {
   const result = structuredClone(body) as Record<string, unknown>;
-  if (!isPlainObject(result.properties)) {
-    return result;
+  stripPropertyDefaults(result);
```
fforootd, line 88: Deepening the normalizer also changes the **state hash**, and I think that needs one more line in `loop.ts` before this lands.

`hashForState` (`apps/cli/src/lib/sync/loop.ts:571`) feeds `normalize`'s output into the digest, so every `normalize` generation produces a different hash for the same file. `loop.ts:160-163` accepts exactly three forms — the current normalized hash, `sha256(JSON.stringify(raw))`, and `sha256(JSON.stringify(sorted raw))` — and the *previous* normalized form is not one of them.

I ran the real pipeline (both normalizers, `stableStringify` semantics, the acceptance rules verbatim), starting from a `state.json` hash written by today's CLI and asking whether this branch still reads it as unchanged:

```
  OK   no defaults spelled out anywhere
  OK   top-level default only
 FAIL  nested default only
 FAIL  top-level + nested default
  OK   flat schema, no nesting at all
```

So it's **any schema that spells out a nested default**, with zero local edits. Worth noting the two legacy fallbacks can't rescue these: they hash compact `JSON.stringify`, while `hashForState` hashes `stableStringify`'s pretty-printed output, so they only ever match genuinely ancient raw-content hashes.

What the user sees on the first `plan` after upgrading: `will publish a new revision` with an essentially **empty field diff**, because `plan-renderer.ts:726,754-758` diffs both sides *normalized* and they come out identical. `apply` then mints an immutable revision and rewrites `user_schema` pins in the flow files. That's the case `loop.ts:157` warns about and `loop.test.ts:478` pins:

> A legacy-format state hash must never read as an edit — a spurious mismatch would publish an unintended update/revision.

It's one-time and self-healing, and still strictly better than main's diff-forever behaviour, so I don't think it blocks — but it's worth a deliberate call rather than a surprise. Two ways out:

**a) Accept the pre-PR form as a third legacy hash** (my preference — it keeps the tested invariant true):

```ts
// Three hash generations now: raw, sorted-raw, and top-level-only
// normalization (before defaults were stripped at every depth).
const unchanged =
  entry.hash === hash ||
  entry.hash === hashResourceContent(content) ||
  entry.hash === hashResourceContent(JSON.parse(stableStringify(content)) as object) ||
  entry.hash === hashForState({ normalize: normalizeSchemaBodyTopLevel }, content);
```

with a `loop.test.ts` twin of the existing legacy-hash tests. Costs one retained legacy normalizer.

**b) Accept the churn deliberately** and say so in the changeset — one no-op revision on the first apply after upgrade. Cheaper, but it does spend an immutable revision and touch flow files in someone's git tree.

Happy to push either if useful.

---

packages/config/src/normalize.ts
```diff
@@ -84,10 +84,20 @@ export function normalizeFlowBody(body: object): object {
  */
 export function normalizeSchemaBody(body: object): object {
```
fforootd, line 86: Small docs nit: the new every-level contract is documented only on the private `stripPropertyDefaults`, while this doc block still reads as top-level-only ("a property carrying `x-editable: true` … normalizes to one omitting it", "every other field pass[es] through untouched").

This is the published entry point — `@zitadel/config/normalize`, called from `syncers.ts:81`, `setup-resources.ts:111`, and `plan-renderer.ts:645` — so someone checking whether a nested `address.street` default survives normalization reads the pre-PR answer here and never sees the helper. Since the depth contract is the whole point of the change, it's worth one clause at the call site:

```suggestion
 * Nested `properties` are descended into, so a leaf of an object property
 * normalizes the same way a top-level one does.
 */
export function normalizeSchemaBody(body: object): object {
```

---

packages/config/src/normalize.test.ts
```diff
@@ -75,6 +75,34 @@ describe("normalizeSchemaBody", () => {
...
+        address: {
+          type: "object",
+          properties: {
+            street: { type: "string" },
+            city: { type: "string", "x-sensitive": true },
+          },
+        },
+      },
+    });
+  });
```
fforootd, line 104: This covers the stripping nicely. The gap is one file down: `does not mutate its input and is idempotent` (line 128) asserts on a **flat** body, so it never reaches `stripPropertyDefaults(property)` — the recursive descent has no immutability guard at all.

That's not a theoretical worry, because of what the caller does with the object afterwards. `loop.ts:122` hashes the parsed on-disk `content` and then hands *that same object* to `create`/`revise` as the upload payload. So a later refactor swapping `structuredClone` for a shallow `{ ...body }` — a natural-looking optimisation, since the top level is all the old code touched — would pass every test in this file while mutating the caller's nested property objects in place, stripping nested `x-editable: true` out of what actually gets POSTed. Exactly what the module header at `normalize.ts:11-16` forbids.

A nested twin of the existing test closes it:

```suggestion
  });

  it("does not mutate a nested input and is idempotent", () => {
    const body = {
      properties: {
        address: {
          type: "object",
          properties: { street: { type: "string", "x-editable": true } },
        },
      },
    };
    const once = normalizeSchemaBody(body);
    expect(body).toEqual({
      properties: {
        address: {
          type: "object",
          properties: { street: { type: "string", "x-editable": true } },
        },
      },
    });
    expect(normalizeSchemaBody(once)).toEqual(once);
  });
```

---

fforootd (commented):
Nice, well-scoped fix — recursing the `properties` chain is the right call, and the recursion boundary matches the spec rather than over-reaching. CI is green across all 10 checks. Three recommendations inline, one of which I'd like a decision on before merge.

**The one worth deciding:** deepening `normalize` also deepens `hashForState`, which silently invalidates `state.json` hashes for any schema that spells out a nested default. I verified it against the real pipeline rather than by reading — details and two concrete options on `normalize.ts:87`. One-time and self-healing, but it currently trips the invariant `loop.test.ts:478` pins, so better as a deliberate choice than a surprise.

The other two are small: a JSDoc clause on the exported entry point, and a nested twin for the immutability/idempotency test. Both have one-click suggestions.

### Adjacent, pre-existing — not this PR's doing

`apps/docs/content/docs/cli/plan-apply.mdx:76` tells users the opposite of what this PR establishes:

> The platform stores a canonicalized form of what you upload: meta-schema defaults are spelled out (e.g. `"x-editable": true`), so every immutable revision is self-contained.

`normalize.ts:11-16`, `syncers.ts:82-85`, and this PR's own Notes all say the server stores schema bytes **verbatim** and never materialises those defaults. The difference decides whether deleting a spelled-out `x-editable: true` is safe: under the documented model the platform re-adds it, under the real one the next published revision quietly loses it. Since this PR is the definitive statement on that behaviour, it might be the natural place to fix the sentence — or a follow-up, no strong feelings.

### Checked and deliberately not flagged

The recursion follows only `properties` and skips `$defs`, `items`, `allOf`/`oneOf`/`anyOf`, `patternProperties`, and `additionalProperties`-as-schema. That reads like a gap but is the documented contract — `docs/design/flowengine/user-schema.md` ("Property names") puts those routes explicitly out of scope, and `meta-schemas/user-property.json` threads its recursive `$ref` only through `properties`. I also confirmed there's no parent→child inheritance of `x-editable`, so stripping at depth is semantically safe.

---

packages/config/src/normalize.ts
vitorbari: Confirmed your table, then went with **(b)** — accept the churn, document it.

I reran your probe against the real `normalizeSchemaBody` on this branch and got the same five rows, so the effect is exactly as you described:

```
 OK   no defaults spelled out anywhere
 OK   top-level default only
FAIL  nested default only
FAIL  top-level + nested default
 OK   flat schema, no nesting at all
```

Also confirmed the two fallbacks cannot rescue it, for the reason you gave: `hashResourceContent` hashes compact `JSON.stringify` while `hashForState` hashes `stableStringify` output, so no normalized-hash generation can ever collide with them.

Why (b) rather than (a):

The affected population is a schema that has a nested object **and** spells out a default on one of its leaves **and** has been applied. Nested leaves are not collectable by a flow until #785 lands, so a nested object in a user schema today buys nothing and there is little reason to have authored one. Set against that, (a) costs a retained copy of the old normalizer that has to be kept correct forever — and would set the precedent that every future widening of a normalizer adds another. As the comment I added to `loop.ts` puts it, keeping a copy of every past normalizer would outlast the state files it serves.

Worth adding: the churn is not new cost so much as a bounded version of the current cost. On `main` these users republish a revision on *every* run. After this PR they republish once and then stop.

So: `.changeset/normalize-nested-schema-defaults.md` now spells out the one-time revision and the empty field diff, and `loop.ts:159` records why only the two pre-normalization formats are accepted and not earlier normalizer generations. The invariant `loop.test.ts:478` pins still holds literally — both legacy tests run through `normalizeFlowBody` and raw hashes, neither of which this change touches — but the comment now scopes what "legacy" covers so the next person widening a normalizer sees the tradeoff rather than assuming it is free.

Happy to switch to (a) if you would rather not spend the revision.

packages/config/src/normalize.ts
vitorbari: Applied in `089e53c60` — the clause is on `normalizeSchemaBody` now, phrased as you suggested. Agreed the export is where someone checks whether a nested `address.street` default survives; the private helper is not on that path.

packages/config/src/normalize.test.ts
vitorbari: Added as `does not mutate a nested input and is idempotent`.

One correction to the mechanism, in your favour. I mutation-tested it, and swapping `structuredClone` for `{ ...body }` fails the **flat** test too — the spread shares the `properties` object, so stripping a top-level default mutates the input and the existing test catches it.

The mutation that gets through is one level deeper: clone the body, clone the `properties` map, clone each top-level property object, stop there. That is arguably the more plausible refactor, since it looks like a correct hand-rolled deep copy of everything the old code touched. It fails only the new test:

```
× does not mutate a nested input and is idempotent
  Tests  1 failed | 12 passed (13)
```

So the gap you identified was real, just reachable by a slightly different wrong turn than the one in the comment. Left the test as you wrote it apart from building the expected value from a factory rather than repeating the literal, so the assertion cannot drift from the input.

PR #785, opened 2026-08-07
feat: collect nested user-schema fields by dotted path

### LINKED ISSUE #771: bug(flow-engine): nested user-schema fields are unusable

vitorbari:
## Summary

Flow steps can collect nested user-schema properties by dotted path:

```json
{ "name": "address", "fields": ["address.street", "address.city"], "on_success": "create_user" }
```

## Tests:

Flow API response rendering step fields:

Screenshot 2026-08-10 at 15 01 48

Data stored under the `user_attributes` table after user creation:

```
╭─────────────────────────────────┬─────────┬─────────────────────────────────┬────────────────┬──────────────────────────────────╮
│           project_id            │ team_id │             user_id             │      key       │              value               │
╞═════════════════════════════════╪═════════╪═════════════════════════════════╪════════════════╪══════════════════════════════════╡
│ proj_01KZNVD9C6HASG9VR8S7HR8GZC │         │ user_01KZNW44MK0T6JFTZVD2DB9SWA │ email          │ "example@zitadel.com"            │
│ proj_01KZNVD9C6HASG9VR8S7HR8GZC │         │ user_01KZNW44MK0T6JFTZVD2DB9SWA │ $schema        │ "sch_01KZNVE3DG5QMGAKGWHJ36PJ23" │
│ proj_01KZNVD9C6HASG9VR8S7HR8GZC │         │ user_01KZNW44MK0T6JFTZVD2DB9SWA │ address.street │ "Main Street"                    │
╰─────────────────────────────────┴─────────┴─────────────────────────────────┴────────────────┴──────────────────────────────────╯
```


internal/domain/flow_field_resolver_schema.go
```diff
@@ -26,7 +26,8 @@ type SchemaResolver interface {
 // Translation map (user meta-schema → [FlowField]):
 //
 //   - `type` + `format` → [FlowFieldType]
-//   - top-level `required` membership → [FlowField.Required]
+//   - `required` membership at every level of the field's path →
+//     [FlowField.Required]
 //   - `minLength`, `maxLength`, `format` → [FlowFieldValidation]
```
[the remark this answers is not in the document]
vitorbari: There will be a follow up PR to update the flow validator: https://github.com/zitadel/nextgen/pull/805

internal/maputil/map_write.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// SetNested writes value at the dotted path, creating the intermediate
+// maps it descends through. It is the counterpart to [GetNested], and
+// errors when the path descends through a value that is not a map.
+func SetNested(m map[string]any, path string, value any) error {
```
wim07101993: Why not make the path a slice of string? It would reduce the responsibility of this function. I was thinking of doing the same with the GetNested function. That way, the AttributeKey describes how a key is separated and the Get/Set function handles the iteration logic

wim07101993 (commented):
Currently a required `object` cannot be satisfied it seems. The `FlowDefinitionValidator` still uses flat maps, so it cannot validat whether a nested property is present

internal/maputil/map_write.go
vitorbari: Do you mean something like this?

```go
maputil.SetNested(m, []string{"address", "geo", "latitude"}, 45.0)
```

internal/domain/flow_field_resolver_schema.go
```diff
@@ -102,16 +103,39 @@ func resolveUserPropertyField(properties map[string]schemaReader, required map[s
...
+// [ErrFlowFieldUnknown] naming the whole path.
+func walkUserProperty(root schemaReader, field Field) (schemaReader, bool, error) {
+	segments := AttributeKey(field.String()).Nodes()
+	parent, required := root, true
+	for i, segment := range segments {
+		prop, ok := parent.Properties()[segment]
+		if !ok {
+			return schemaReader{}, false, fmt.Errorf("%w: %q", ErrFlowFieldUnknown, field.String())
+		}
+		if _, ok := parent.RequiredSet()[segment]; !ok {
```
fforootd: This ancestor-`required` rule is the right semantics (matches what document-level validation accepts for the whole document), but right now it's unreachable through any saveable definition, and the same conflict blocks the feature's headline use case:

`validateRequiredUserSchemaFields` runs before field resolution on every save and exact-matches top-level `required` names against field names. For a schema whose top-level `required` contains an object property — exactly `nestedSchemaContent`'s `"required": ["email", "address"]` — no definition can pass:

- collect `address.street` + `address.city` without a literal `address` field → `required fields [address] in user schema are missing in the flow definition steps`
- add the literal `address` field → `ErrFlowFieldNotScalar` from this resolver

Consequences: a *required* nested block can't be expressed at all, and `Required: true` for a nested leaf can never happen in a saved definition (it needs the top-level parent in root `required`, which is the unsaveable case above) — so the changeset's "only marked required when every object above it is required too" is dead code until this is resolved.

Suggested direction: make `validateRequiredUserSchemaFields` treat a required object property as covered when some field descends into it (prefix match on `name + "."`), or — closer to document semantics — recursively require its required leaves to be collected. Either way this wants a `flow_definition_validator` test with a required object property; the resolver tests structurally can't catch this interaction.

internal/domain/flow_field_resolver_schema.go
```diff
@@ -102,16 +103,39 @@ func resolveUserPropertyField(properties map[string]schemaReader, required map[s
...
+			required = false
+		}
+		if i < len(segments)-1 && prop.Properties() == nil {
+			return schemaReader{}, false, fmt.Errorf("%w: %q", ErrFlowFieldUnknown, field.String())
+		}
```
fforootd: Optional simplification: this early-exit is behavior-identical to letting the next iteration miss — for a scalar `prop`, `Properties()` returns nil, so `Properties()[segment]` on the next pass fails with the same wrapped `ErrFlowFieldUnknown` naming the same full path. The doc comment above stays accurate either way.

```suggestion
```

internal/domain/flow_field_resolver_schema.go
```diff
@@ -355,7 +385,7 @@ func (r schemaReader) Properties() map[string]schemaReader {
 	return out
 }
 
-// RequiredSet returns the names listed in the top-level `required`
+// RequiredSet returns the names listed in this level's `required`
```
fforootd, line 434: Nit, same treatment as this line: `Properties()` just above still says "each top-level property of an object schema" — it now reads whatever level `walkUserProperty` hands it.

internal/domain/flow_state_machine.go
```diff
@@ -1263,7 +1265,12 @@ func mergeCollected(state *FlowState, fields map[string]any) error {
 			return fmt.Errorf("unknown auth method %s", k)
 		}
 
-		state.CollectedData.UserData[k] = v
+		// Field names are dotted paths for nested properties, so the
+		// collected document keeps the shape the user schema validates
+		// and the attribute store flattens back out.
+		if err := maputil.SetNested(state.CollectedData.UserData, k, v); err != nil {
+			return err
```
fforootd: The caller wraps this as `flow state machine: validate fields: %w`, which mislabels a merge failure. It's near-unreachable today (`Validate` has already rejected unknown names, and the definition-time not-scalar check prevents scalar/path collisions within one schema) — which is exactly why, when it does fire (schema drift against an in-flight session), it should say which field:

```suggestion
			return fmt.Errorf("merge collected field %q: %w", k, err)
```

internal/domain/flow_field_resolver_schema_test.go
```diff
@@ -65,6 +65,27 @@ const defaultSchemaContent string = `{
 		}
 	}`
 
+// nestedSchemaContent carries an object property whose leaves cover the
+// nested cases: a required leaf, an optional leaf, and one keyed with
+// x-unique.
+const nestedSchemaContent string = `{
+		"$schema": "https://json-schema.org/draft/2020-12/schema",
+		"type": "object",
+		"required": ["email", "address"],
```
fforootd, line 74: This fixture is exactly the shape the definition validator can't accept today (see the `walkUserProperty` thread): with `address` in top-level `required`, neither collecting its leaves nor naming it literally saves. Once the coverage rule learns dotted paths, it'd be great to reuse this schema in a `flow_definition_validator_test` case so the save path is exercised too, not only the resolver.

internal/maputil/map_write.go
```diff
@@ -0,0 +1,30 @@
+package maputil
+
+import (
+	"fmt"
+	"strings"
+)
+
+// SetNested writes value at the dotted path, creating the intermediate
+// maps it descends through. It is the counterpart to [GetNested], and
+// errors when the path descends through a value that is not a map.
```
fforootd: Worth pinning the one asymmetry in the contract — descending *through* a non-map errors, but a final-segment write onto a key that *holds* a map replaces it silently. Unreachable via validated flow fields (an object can't be a field name), but the next caller of this helper won't know that:

```suggestion
// SetNested writes value at the dotted path, creating the intermediate
// maps it descends through. It is the counterpart to [GetNested], and
// errors when the path descends through a value that is not a map. The
// converse is not guarded: a final-segment write onto a key that holds
// a map replaces the map silently.
```

fforootd (commented):
Really clean design end-to-end: one canonical splitter, `SetNested`/`GetNested` as true counterparts, the collected document staying schema-shaped so `create_user` validates it as-is, and object/array field names rejected at save time instead of failing mid-flow. I verified the #783 dot ban holds recursively (`user-property.json` re-applies `propertyNames` per nesting level), so dotted addressing is unambiguous — and the touched packages pass locally.

One thing to settle before merge (inline at `walkUserProperty`): a schema whose top-level `required` contains an object property can't save **any** definition — `validateRequiredUserSchemaFields` demands the literal name, `ErrFlowFieldNotScalar` forbids it. That blocks required nested blocks entirely and makes nested `Required: true` unreachable outside unit tests.

Follow-ups, non-blocking:
- **Login labels**: `fieldLabelFallback` (`packages/components/src/orchestrator/liquid.ts`) bails on dotted remainders, so a nested field without a locale entry renders its raw key (`profile.field.address.street`) as the visible label instead of humanising like top-level fields do. There's also a latent key-shape collision: a leaf named e.g. `email.help` shares its lookup key with field `email`'s `.help` sub-key. Deserves a client-side follow-up in this train.
- **Contract wording**: `Step.fields` in `api/openapi/endpoints/schemas/flow-definition.json` still reads "schema property names … must resolve to a property", and `docs/design/flowengine/flow-engine.md` / `flow-definition-rules.md` likewise — one sentence each for dotted paths and the new not-scalar rule.

internal/maputil/map_write.go
wim07101993: yes

vitorbari:
> Currently a required `object` cannot be satisfied it seems. The `FlowDefinitionValidator` still uses flat maps, so it cannot validat whether a nested property is present

@wim07101993 

I'm planning to raise a separate PR for the flow-definition validator, this one was already too big and hard to review.

Draft: https://github.com/zitadel/nextgen/pull/805

I will probably fold it into a single PR to avoid confusion. The copilot review also flagged it.

internal/domain/flow_field_resolver_schema.go
vitorbari: Fixed 727c3e7ed8d03397a5e41af54f11cc56ef82b5f8

internal/domain/flow_field_resolver_schema.go
vitorbari: The flow definition validator changes were initially in a separate PR (#805); they are now part of this one so they land together.

internal/domain/flow_field_resolver_schema_test.go
vitorbari: Done in 6c3f3ebb — TestValidator_RequiredNestedUserSchemaFields now has a case that drives nestedSchemaContent through ValidateFlowDefinition, so this fixture is exercised on the save path and not only in the resolver.

vitorbari:
@fforootd ,

> One thing to settle before merge (inline at `walkUserProperty`): a schema whose top-level `required` contains an object property can't save **any** definition — `validateRequiredUserSchemaFields` demands the literal name, `ErrFlowFieldNotScalar` forbids it. That blocks required nested blocks entirely and makes nested `Required: true` unreachable outside unit tests.

The blocker is settled. The flow definition validator changes were in a separate PR (#805); they are now part of this one (b9f61ab4), so a required object is satisfied by collecting one of its leaves.


> Follow-ups, non-blocking:
> 
> * **Login labels**: `fieldLabelFallback` (`packages/components/src/orchestrator/liquid.ts`) bails on dotted remainders, so a nested field without a locale entry renders its raw key (`profile.field.address.street`) as the visible label instead of humanising like top-level fields do. There's also a latent key-shape collision: a leaf named e.g. `email.help` shares its lookup key with field `email`'s `.help` sub-key. Deserves a client-side follow-up in this train.

Agreed. Tracking it separately rather than folding it in here.

> * **Contract wording**: `Step.fields` in `api/openapi/endpoints/schemas/flow-definition.json` still reads "schema property names … must resolve to a property", and `docs/design/flowengine/flow-engine.md` / `flow-definition-rules.md` likewise — one sentence each for dotted paths and the new not-scalar rule.

Covered by https://github.com/zitadel/nextgen/pull/786

internal/domain/flow_field_resolver_schema.go
```diff
@@ -142,12 +163,18 @@ func resolveAuthMethodField(authMethods xAuthMethodsReader, field Field, stepNam
...
+func deriveUserPropertyType(prop schemaReader, field Field) (FlowFieldType, error) {
 	jsonType, err := prop.JSONType()
 	if err != nil {
 		return "", err
 	}
+	// A property carrying `properties` is an object even when it omits
+	// the `type` keyword, which would otherwise fall through to text.
+	if jsonType == "object" || jsonType == "array" || prop.Properties() != nil {
+		return "", fmt.Errorf("%w: %q", ErrFlowFieldNotScalar, field.String())
+	}
```
fforootd: 🔴 **Blocking: this new rule fires at runtime for flows applied before this PR**

`deriveUserPropertyType` runs in the shared resolver, which the state machine calls on every `Start`/`Process` via `resolveStepFields` — not just at definition time.

Pre-PR, a schema with a required *object* property (e.g. `required: ["address"]`) actually **forced** definitions to name `"address"` as a field (the old top-level required check demanded it literally), and the old resolver happily rendered it as a text input. Those definitions are applied and valid today. After this PR deploys, `Resolve` returns `ErrFlowFieldNotScalar` for them on every render — the whole login/register surface errors instead of degrading.

That also contradicts the contract documented on `validatePasskeyActionsEnabled` in this package: *definition time is the only enforcement point; flows applied before a rule surface the violation on their next plan/apply*.

Suggestion: enforce not-scalar only on the definition-time path (like the passkey-actions rule), and keep the runtime resolver lenient (legacy text rendering, or skip-with-log) so existing deployments keep working until the definition is next re-applied.

---

internal/domain/flow_definition_validator.go
```diff
@@ -115,7 +116,7 @@ func resolveAllStepFields(schema *jsonschema.Schema, steps []FlowDefinitionStep)
 	}
 
 	// validate that all required fields in the schema are present in the flow definition
-	if err := validateRequiredUserSchemaFields(sr.RequiredSet(), steps); err != nil {
+	if err := validateRequiredUserSchemaFields(sr.RequiredLeafPaths(), steps); err != nil {
```
fforootd: 🔴 **Blocking: the CLI's TS port of this validator wasn't updated, so the feature is unusable through `zitadel plan`/`apply`**

`packages/config/src/validate.ts` still does a flat `properties[field]` lookup in `resolveFieldChallenge` and checks top-level `required` only, so the two sides now disagree in both directions:

- `"fields": ["address.street"]` → CLI throws `E_VALIDATION` (`flow-validation.ts:88` aborts plan *and* apply) for a definition the server accepts. The only workaround is `ZITADEL_SKIP_FLOW_VALIDATION`.
- `"fields": ["address"]` (object) → CLI passes plan, server rejects on apply with `ErrFlowFieldNotScalar`.

The contract comment at the top of this file (lines 23–28) asks for rule changes to be kept in sync with the TS port, and the drift-audit test only compares function names/literals, so it passes silently.

The dotted-path walk, `RequiredLeafPaths`, and the ancestor-coverage logic below all need porting to `validate.ts` in this PR (or the feature gated until the CLI catches up).

---

internal/domain/flow_field_resolver_schema.go
```diff
@@ -371,6 +398,30 @@ func (r schemaReader) RequiredSet() map[string]struct{} {
...
+// it satisfies the coverage check.
+func (r schemaReader) RequiredLeafPaths() map[string]struct{} {
+	out := map[string]struct{}{}
+	r.collectRequiredLeafPaths("", out)
+	return out
+}
+
+func (r schemaReader) collectRequiredLeafPaths(prefix AttributeKey, out map[string]struct{}) {
+	properties := r.Properties()
+	for name := range r.RequiredSet() {
```
fforootd: 🔴 **Blocking: `required` inside *optional* objects is invisible, so a saved-as-valid definition can dead-end every registration**

The descent only recurses through properties that are themselves in `required`, so a `required` list inside an optional object is never collected. But collecting *any* child of that object materializes it, and then schema validation enforces its nested `required`.

Repro (I confirmed this with a probe test against this branch):

```jsonc
// schema: "address" is optional, but has required: ["street"]
{ "properties": { "email": {...}, "address": {
    "type": "object", "required": ["street"],
    "properties": { "street": {...}, "city": {...} } } },
  "required": ["email"] }
```

A flow collecting `["email", "address.city"]` passes `ValidateFlowDefinition` (`address.street` is never demanded). At runtime the orchestrator submits every declared field (blank text defaults to `""`), `mergeCollected` materializes `address: {city: ""}`, and `NewCreateUser`'s schema validation fails `missing required field "street"` — **on every single submission** through this definition.

The coverage check needs to be conditional: once any collected path enters an object, that object's nested required leaves must be covered too. That means the check has to look at the collected fields, not just a statically precomputed required set.

---

internal/domain/flow_field_resolver_schema.go
```diff
@@ -371,6 +398,30 @@ func (r schemaReader) RequiredSet() map[string]struct{} {
...
+		path := prefix.AppendNode(name)
+		prop, ok := properties[name]
+		if !ok || len(prop.RequiredSet()) == 0 {
```
fforootd, line 476: 🟡 **A required object with no nested `required` is covered by a leaf that resolves as optional — late `create_user` failure**

When a required object declares no `required` of its own, the descent stops at the object and *any* collected leaf beneath it counts as coverage. But `walkUserProperty` marks that covering leaf `Required=false` (it isn't in the object's required set, line 126), so `MissingRequired` never demands it and `Validate` accepts its absence.

Scenario: schema `required: ["billing"]`, `billing` has properties but no `required` list; the definition collects `billing.vat_id` (accepted — covers `billing`). A headless/API client that omits `billing.vat_id` from the submit body never materializes `billing`, and `NewCreateUser` fails `missing required field "billing"` at the *end* of the flow — exactly the late failure the definition-time check exists to prevent. No definition over this schema shape can guarantee completion.

Suggestion: when a leaf is the sole coverage for a required ancestor object, surface it as `Required=true` (or reject the definition/schema shape with a clear error).

---

internal/domain/flow_field_resolver_schema.go
```diff
@@ -142,12 +163,18 @@ func resolveAuthMethodField(authMethods xAuthMethodsReader, field Field, stepNam
...
+// union the resolver cannot reduce to a single kind, and
+// [ErrFlowFieldNotScalar] when the property has no field-shaped input.
+func deriveUserPropertyType(prop schemaReader, field Field) (FlowFieldType, error) {
 	jsonType, err := prop.JSONType()
 	if err != nil {
 		return "", err
 	}
+	// A property carrying `properties` is an object even when it omits
+	// the `type` keyword, which would otherwise fall through to text.
+	if jsonType == "object" || jsonType == "array" || prop.Properties() != nil {
```
fforootd: 🟡 **The guard anticipates `properties`-without-`type` but misses `items`-without-`type`**

Nice catch handling schemas that omit the `type` keyword on objects — but arrays have the same hole. I confirmed with a probe test: `"tags": {"items": {"type": "string"}}` (no `type` keyword) resolves to `FlowFieldTypeText` with no error. The definition saves, the UI renders a text box, and the submitted string even passes `create_user` validation (JSON Schema `items` is vacuous on non-arrays) — so the attribute stores a string where the author declared an array, and reads back as the wrong shape.

Mirroring the object check with an `items` lookup closes it:

```go
if _, hasItems := prop.s.LookupKeyword("items"); jsonType == "object" || jsonType == "array" ||
	prop.Properties() != nil || hasItems {
```

(or a small `prop.HasItems()` helper next to `Properties()`).

---

internal/domain/flow_field_resolver_schema.go
```diff
@@ -100,16 +101,36 @@ func resolveUserPropertyField(properties map[string]schemaReader, required map[s
...
+// [ErrFlowFieldUnknown] naming the whole path.
+func walkUserProperty(root schemaReader, field Field) (schemaReader, bool, error) {
+	segments := AttributeKey(field.String()).Nodes()
+	parent, required := root, true
+	for _, segment := range segments {
+		prop, ok := parent.Properties()[segment]
+		if !ok {
+			return schemaReader{}, false, fmt.Errorf("%w: %q", ErrFlowFieldUnknown, field.String())
+		}
+		if _, ok := parent.RequiredSet()[segment]; !ok {
```
fforootd: 🔵 **Perf: `Properties()`/`RequiredSet()` are re-materialized per segment, per field, per render**

`Resolve` runs on every flow HTTP render/`Process`. Before this change, `root.Properties()` and `root.RequiredSet()` were hoisted once per `Resolve` call; now every field re-builds the full root-level map (allocating a `schemaReader` per *sibling*) plus a map per extra path segment.

For a step with N fields that's N re-materializations of the same root keywords each render. Hoisting the root readers once per `Resolve` (as before) — or adding a single-key lookup on `schemaReader` so the walk doesn't build the whole sibling map per level — restores the old behavior.

---

internal/domain/flow_definition_validator.go
```diff
@@ -179,28 +180,43 @@ func validatePasskeyActionsEnabled(sr schemaReader, steps []FlowDefinitionStep)
...
+	cover := func(path string) {
+		for {
+			covered[path] = struct{}{}
+			dot := strings.LastIndex(path, ".")
+			if dot < 0 {
+				return
+			}
+			path = path[:dot]
+		}
+	}
```
fforootd: 🔵 **Nit: reuse `AttributeKey` instead of hand-rolling the dotted-path walk**

`AttributeKey` (`Nodes()`/`AppendNode()`) already owns the `.` separator convention used everywhere else in this feature — the resolver, the state machine, and `RequiredLeafPaths` all go through it. Slicing with `strings.LastIndex` here duplicates that knowledge, so if the key syntax ever grows (escaped dots, alternate separator — the maputil tests already note the caller owns the separator), this walk would silently diverge from resolution.

Building the ancestor prefixes from `AttributeKey(...).Nodes()` keeps one authority:

```go
cover := func(path Field) {
	var prefix AttributeKey
	for _, node := range AttributeKey(path.String()).Nodes() {
		prefix = prefix.AppendNode(node)
		covered[string(prefix)] = struct{}{}
	}
}
```

---

fforootd (commented):
Reviewed at head `79d7a5b` against a local checkout. The core of the feature is solid — dotted-path resolution, the `maputil` nested write/query work, and the happy path from the PR description all behave as advertised, and the existing tests in the touched packages pass. The test coverage added here is also genuinely good.

That said, I found **three issues I'd consider blocking** and two smaller correctness gaps (details inline; two of them I confirmed by running probe tests against this branch):

**🔴 Blocking**
1. `ErrFlowFieldNotScalar` is enforced in the shared resolver that also runs at flow *runtime*, so definitions applied before this PR that name an object property (which the old validator actually **forced** for schemas with a required object) hard-fail on every render after upgrade — login/register goes down for those projects.
2. The CLI's TS validator port (`packages/config/src/validate.ts`) wasn't updated: `plan`/`apply` reject dotted fields the server now accepts, and accept object fields the server now rejects — the feature is unusable through the config path.
3. `RequiredLeafPaths` never sees `required` declared inside *optional* objects, so a definition collecting a sibling leaf (`address.city`) saves as valid but fails schema validation at `create_user` on **every** submission (probe-tested).

**🟡 Edge cases**
4. A required object with no nested `required` is "covered" by a leaf that resolves `Required=false` — an API client omitting it hits a late `create_user` failure the definition-time check exists to prevent.
5. The not-scalar guard catches `properties`-without-`type` but not `items`-without-`type` — such an array property renders as a text input and silently stores a string (probe-tested).

**🔵 Cleanups**: hoist `Properties()`/`RequiredSet()` once per `Resolve` again (they're now rebuilt per segment per field per render), and reuse `AttributeKey` in `cover()` instead of `strings.LastIndex`.

Happy to discuss any of these — 1 and 3 are the ones I'd definitely want resolved before merge.

---

internal/domain/flow_definition_validator.go
vitorbari: CLI changes will be covered by https://github.com/zitadel/nextgen/pull/804. This PR was intentionally only Golang changes.

internal/domain/flow_field_resolver_schema.go
vitorbari: Fixed by e3ea6680699f45abdc8d8c28b3ab1a30aba52f7d

internal/domain/flow_definition_validator.go
vitorbari: Fixed by e3ea6680699f45abdc8d8c28b3ab1a30aba52f7d

internal/domain/flow_field_resolver_schema.go
vitorbari: Flow validation already covers this: a definition must collect at least one property of every required object, or it's rejected at save time. So `billing` can't be silently skipped.

What flow validation can't do is decide which property makes a billing valid, that's the user schema's job.

If the schema says billing is required but never says which of its properties are required, it hasn't defined that, and the engine shouldn't guess: marking the covering leaf required would force users to fill a property the schema calls optional, and auto-creating `billing: {}` would make `required: ["billing"]` meaningless.

So for the case you found, the fix is one line in the schema: add `"required": ["vat_id"]` to billing, and the flow becomes correct by construction.

internal/domain/flow_field_resolver_schema.go
vitorbari: Fixed: 46b7bd845516a8052cfb88c8a7d898e830be1784

internal/domain/flow_field_resolver_schema.go
vitorbari: Fixed: d39c0f08e669653ef859d2bc2b4814375feecbd3

internal/domain/flow_field_resolver_schema.go
vitorbari: > this new rule fires at runtime for flows applied before this PR
> Those definitions are applied and valid today. After this PR deploys, Resolve returns ErrFlowFieldNotScalar for them on every render 

Existing flows are not a concern, there nothing is in production yet.

Rendering an object as a text input is the failure this PR solves: it saves, it renders, and then it crashers at `create_user`. Failing at apply is the improvement.

PR #786, opened 2026-08-07
docs: describe dotted field paths (for nested user attribute) and add a flow example

### LINKED ISSUE #771: bug(flow-engine): nested user-schema fields are unusable

vitorbari:
Documents dotted field paths and the optional-object required rule, and adds the `07-nested-profile-fields` worked example. Description text only; the `oas_schemas_gen.go` diff is regenerated doc comments.


vitorbari:
Superseded by #917 — same diff, clean history.

#785 was squash-merged, so its 30 original commits never entered `main`. Merging that branch in here to resolve the conflicts pulled all of them into this PR permanently, which is why it showed 33 commits for an 11-file docs change. #917 is the same net diff applied as a single commit on top of current `main`.

Copilot's two comments here still apply and are carried over in substance — they identified that a nested leaf renders `required: false` while the materialized object makes it mandatory at `create_user`. Tracking that separately, since it is a resolver change rather than a docs one.

PR #794, opened 2026-08-07
fix(database): harden keyset cursors and add battle tests

adlerhurst:
## Summary

- Encode sort `Direction` on keyset page tokens and reject direction / empty-column mismatches (`db.cursor_order_mismatch`) in `compileRead` and `user.ApplyCursor`.
- Coerce credential resource IDs as strings so password/TOTP/passkey/recovery cursor paging survives JSON round-trips.
- Collapse List next-cursor emission into `pagination.MarshalNext` across dialects; slim `user.NextCursor` to that helper.
- Make `stmttest` battle coverage data-driven via `drainIncarnation` (ASC+DESC ordered drain + NextCursor emission for every `List*`), with credential battles split out; adversarial / destructive / nullable cases stay in the special suite.
- Extract dialect cursor compile suites into `cursor_compile_test.go` and standardize constructions on `pagination.New`.
- Document that clients must resend the same `sorting` with `page_token` (OpenAPI on projects/teams query); API itest covers mismatch vs matching sorting.
- Document the battle registration rule (seed + `drainIncarnation`) in `internal/storage/AGENTS.md`.

## Validation

- `go test ./internal/storage/dialect/pagination/ ./internal/storage/dialect/{postgres,spanner,sqlite}/ -count=1`
- `go test -tags=sqlite_integration ./internal/storage/stmttest/ -count=1`
- Postgres/Spanner / API integration not run locally (no Docker); CI dialect + `full-pr` should cover them.

## Release notes / changeset

Changeset: `.changeset/cursor-battle-direction-coerce.md` — reject mismatched / empty-column cursor order; coerce credential IDs as strings so keyset paging works past the first page (`@zitadel/server` patch). Structure-only follow-ups (`MarshalNext`, battle harness, OpenAPI note, itest) need no additional changeset.

## Notes

- Merged with `main` (includes #781 nullable keyset fix and `internal/storage/v2` → `internal/storage` rename).
- Pre-direction tokens omit `direction` and unmarshal as ASC (hard cutover); covered by unit test.
- Wim review: harness tightened (no unused filter, ordered ASC/DESC asserts, emission remainder length, seed 5); clients must resend sorting (documented + tested). Schema Accessor/Coerce table unit test deferred as a follow-up.

<div><a href="https://cursor.com/agents/bc-16e784f9-9a30-4344-a3e9-a1d2d4b29340?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-16e784f9-9a30-4344-a3e9-a1d2d4b29340&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


internal/storage/dialect/pagination/cursor.go
```diff
@@ -37,9 +63,17 @@ func (c *Cursor[F]) Marshal() []byte {
 	return []byte(base64.RawURLEncoding.EncodeToString(payload))
 }
 
-func (c *Cursor[F]) MatchesOrderBy(columns []database.Column[F]) bool {
-	if len(c.Columns) != len(columns) {
+// MatchesOrderBy reports whether the cursor was issued for the same columns and direction.
+// Empty column sets never match: a zero-term keyset would compile to invalid SQL.
+func (c *Cursor[F]) MatchesOrderBy(orderBy database.OrderBy[F]) bool {
```
wim07101993, line 68: Does this validation mean that all api client now need to send the order by on each request?

If a request is made declaring: return me all users and order them by birthdate descening. A cursor would be returned containing that info. When the client then requests for the next page, but does not send the sorting info, it would fail the request since both are provided to the service and storage layers.

Maybe add an integration test against the api to verify?

wim07101993 (commented):
Some things claude found:

1. drainIncarnation's filter parameter is never used in the body — 15 call sites pass it for nothing. Drop it.
2. assertDrainMatch uses ElementsMatch, so the desc subtest never asserts descending order. A bug where direction is dropped from ORDER BY (consistently with the keyset predicate) still drains every row exactly once and passes. For a PR whose entire point is direction correctness, that's the wrong assertion — compare against the ordered slice, reversed for desc.
3. assertCursorEmission has a hidden precondition. It asserts the short page is smaller than emissionLimit, which requires emissionLimit > len(want)/2; the doc comment only promises emissionLimit < len(want). Seed 10 rows with limit 2 and you get a baffling failure. Document it or derive the assertion from the actual remainder.
4. 3 rows / limit 2 is a two-page drain — barely exercises the keyset tail. Bumping to 5 is nearly free.
5. Missing the cheap root-cause guard. The coerce bug is an Accessor-type vs Coerce-type mismatch. A table-driven unit test over every registered schema would catch that whole class in milliseconds with no integration tags — the battle matrix only catches it for fields that happen to be sort columns, under build tags, per dialect. Worth a follow-up.

wim07101993 (approved):

Issue #793, opened 2026-08-07
Remove the real Spanner test-instance bring-up once the emulator has proven itself


IAM-marco:
Follow-up to #788.

## Goal

Delete the real-Spanner-test-instance bring-up once the emulator has proven itself over time. #788 unwired it but deliberately left the code in place; this issue removes it.

## Current state

#788 found that the CI flake blamed on the emulator was an application bug: `exchangeConflict` in `internal/storage/v2/session/run_exchange.go` flattened its cause with `fmt.Errorf("%w: %v", ...)`, which stripped the gRPC status off a Spanner `ABORTED` and stopped `ReadWriteTransaction` from recognising it as retryable. With that fixed, the suite runs green at full parallelism against the emulator, and `TestTransactionContention` proves aborts are being retried (8 concurrent writers to one row produced 22-26 callback invocations, all committing, no lost updates).

So CI is emulator-only now, on purpose: the emulator serves one transaction at a time, and that is precisely what forces the aborts that keep the retry path honest. A real instance would hide a missing retry rather than fix it.

What was unwired but kept:

- `.github/workflows/ci.yml`: the `google-github-actions/auth@v2` Workload Identity Federation step, the `id-token: write` permission, and the separate "spanner test instance" test step are gone. One emulator step remains.
- `apps/server/moon.yml` `test-spanner`: no longer pins `-parallel 1`, which had been keeping the emulator green by removing contention instead of retrying it. `-p 1` stays for cross-package test isolation.

What is still in the tree, live but unreferenced by any config:

- `internal/storage/v2/testdb/instance.go` in full: `instanceEnv` (`ZITADEL_TEST_SPANNER_INSTANCE`), `provision`, `dropDatabase`, `parseInstancePath`, `uniqueDatabaseID`, `randomToken`. `createDatabase` in the same file is **also used by the emulator path** and must not be removed with the rest.
- `internal/storage/v2/testdb/spanner.go`: the leading `if strings.TrimSpace(os.Getenv(instanceEnv)) != ""` branch in `SpannerDSN`.
- Documentation of the dormant path in `CONTRIBUTING.md` and `docs/adrs/028-storage-v2-statements-and-dialects.md`.

[...]

Issue #792, opened 2026-08-07
Spanner statement suites: 7 created_at timestamp failures on arm64 (CI green)


IAM-marco:
Found while verifying #788.

> **Corrected after filing.** This was originally reported as failures "on main", which implied CI was broken. That was wrong: CI is green. These failures happen only on arm64 (Apple Silicon) local machines. Details and evidence below.

## Goal

Make the Spanner statement suites pass on arm64 development machines. Seven subtests fail deterministically there while passing in CI, so contributors on Apple Silicon cannot get a clean local Spanner run.

## Current state

On macOS arm64, with the emulator testcontainer:

```sh
go test -tags spanner_integration -timeout=10m -p 1 -count=1 \
  ./internal/storage/v2/stmttest/ ./internal/storage/v2/dialect/spanner/
```

fails 7 subtests, every run:

- `TestProjectStatements_Update/spanner/updates_name_and_refreshes_updated_at`
- `TestProjectStatements_List/spanner/filters_by_created_at_equal`
- `TestProjectStatements_List/spanner/filters_by_created_at_greater_than`
- `TestProjectStatements_List/spanner/filters_by_a_created_at_range`
- `TestProjectStatements_List/spanner/paginates_under_a_created_at_filter`
- `TestTeamStatements_UpdateTeam/spanner/updates_team`
- `TestJSONSchemaStatements_CRUD` (`internal/storage/v2/dialect/spanner/json_schema_test.go:43`)

They share one shape: a `created_at` read back after a second statement differs from the value captured on create, by tens of microseconds. From `stmttest/project_test.go:77`, which asserts `createdAt` is unchanged across an update:

```
expected: time.Date(2026, time.August, 7, 16, 30, 26, 403397000, time.UTC)
actual  : time.Date(2026, time.August, 7, 16, 30, 26, 403463000, time.UTC)
```

66µs of drift on a column that must not move at all. `json_schema_test.go:43` shows the same at 33µs.

**These same subtests pass in CI.** Confirmed in the run for #795 (job `full-pr`, step "Run Go integration tests (spanner emulator)"): `TestProjectStatements_Update/spanner/updates_name_and_refreshes_updated_at`, `TestTeamStatements_UpdateTeam/spanner` and `TestJSONSchemaStatements_CRUD` all report PASS.

[...]

Issue #791, opened 2026-08-07
Postgres silently loses concurrent updates (READ COMMITTED, no serialization retry)


IAM-marco:
Found while verifying #788.

## Goal

Stop Postgres silently losing concurrent updates. A read-modify-write inside `pool.Transaction` can be overwritten by a concurrent transaction with no error raised on either side.

## Current state

`internal/storage/v2/dialect/postgres/tx.go` `executeTransaction` opens transactions with `begin.Begin(ctx)` and never sets an isolation level; nothing in `internal/storage/v2/dialect/postgres/` sets one. pgx therefore uses the server default, READ COMMITTED. Under READ COMMITTED two transactions may both read a row, both write it, and the second commit silently wins. No serialization failure is raised, so there is nothing to retry and nothing to log.

The contention canary added in #788 (`internal/api/integration_test/transaction_contention_test.go`) demonstrates it. Eight concurrent transactions each read one team row, append a character to `name`, and write it back through `service.DB.Transaction`. On Spanner all eight land. On Postgres the run reports:

```
Error: "team-egftjzzaxxx" should have 21 item(s), but has 16
Messages: lost update: a retried transaction replayed stale state
```

Three of eight writers survived; the other five were lost. That test is currently tagged `spanner_integration` only, precisely so this failure does not red the Postgres suite. Flipping its build tag to `postgres_integration` reproduces the bug on demand.

Spanner is unaffected: it aborts the losers and `ReadWriteTransaction` retries them. SQLite serializes writes and is likewise unaffected.

## Decisions that shape this work

- Two candidate fixes, and the choice belongs in this issue: raise the isolation level for transactions that do read-modify-write (`REPEATABLE READ` is enough to turn a lost update into a serialization error; `SERIALIZABLE` is stricter), or keep READ COMMITTED and take row locks explicitly (`SELECT ... FOR UPDATE`) in the statements that read-then-write.
- Raising the isolation level means Postgres starts returning `40001` serialization failures, which nothing retries today: `executeTransaction` is a plain Begin / callback / Commit with no retry loop. So a retry loop is part of this work, and the callback has to be replayable the same way Spanner already requires (see the transaction rules in `internal/storage/v2/AGENTS.md`).
- Whatever is chosen should hold for SQLite too, or SQLite should be explicitly ruled out of scope with a reason.

## Dependencies

- None. #788 is independent and already merged the canary; this issue only widens it to Postgres.

## Test first (TDD)

[...]

PR #795, opened 2026-08-07
fix(session): keep Spanner ABORTED retryable through exchange conflicts

### LINKED ISSUE #788: Make Spanner abort-retry provable: preserve the ABORTED status, bound it, replay-safe closures
Follow-up to #604 and PR #629.

## Goal

Prove that Spanner `ABORTED` is retried by the application, under the emulator at full test parallelism. #629 fixed the CI flake by moving the `spanner_integration` suite onto a real Spanner instance. That removes the canary rather than the cause: Spanner aborts read-write transactions on concurrent updates in production too, and the emulator's deliberately aggressive single global lock is what surfaces a missing retry early. This issue keeps the emulator as the parallel canary and closes the gaps that let an abort escape.

## Current state

The v2 Spanner layer already retries. `internal/storage/v2/dialect/spanner/client.go:46` `Client.Transaction` runs the callback inside `spanner.Client.ReadWriteTransaction`, which retries `ABORTED` automatically. `with_transaction.go` joins an existing transaction when already inside one (Spanner forbids nesting) and otherwise opens a `ReadWriteTransaction`. In `db.go`, `client.Write` and `client.Update` each open their own `ReadWriteTransaction`; `client.Query` and `client.ReadRow` are read-only `Single()` reads. `tx.go` never uses the manual `ReadWriteStmtBasedTransaction`. There is no `codes.Aborted`, backoff or `MaxAttempts` handling anywhere in application code: retry is entirely delegated to the client.

That delegation is currently sound across the whole reachable surface, not just the paths above. `Client` holds `client *spanner.Client` as a named unexported field (`client.go:12`) rather than an embedded one, so no `*spanner.Client` method is promoted and the raw client can only be reached from inside `internal/storage/v2/dialect/spanner/`. Of the methods reachable there (checked against `cloud.google.com/go/spanner@v1.91.0`):

- Retry `ABORTED` themselves: `ReadWriteTransaction[WithOptions]` (`retry.go:108`), `Apply` in both modes (`client.go:1310` delegates to `ReadWriteTransactionWithOptions`; `ApplyAtLeastOnce` has its own retryer at `transaction.go:2256`), and `PartitionedUpdate[WithOptions]` (`pdml.go:76-95`).
- Cannot abort: `Single`, `ReadOnlyTransaction`, `BatchReadOnlyTransaction[FromID]`. Snapshot reads take no locks, so there is no wound-wait abort to retry.
- **Do not retry `ABORTED`:** `BatchWrite[WithOptions]` (`client.go:1473-1529` builds the response iterator with no retryer; mutation groups commit independently and per-group status comes back on the stream), and the package function `NewReadWriteStmtBasedTransaction`, which is caller-retried by design.

The repo uses neither of the non-retrying two today (`BatchWrite`, `PartitionedUpdate`, `ReadWriteStmtBasedTransaction` and `spanner.Client.Apply` have zero non-test call sites), so that is a future hazard to fence off rather than a live bug.

Three things do make the delegation fragile, and none of them is currently tested:

1. **The retry only fires if the gRPC status survives error mapping.** `cloud.google.com/go/spanner@v1.91.0/retry.go:120-133` retries only when the error returned by the callback satisfies `errors.As(err, &*spanner.Error)` or `status.FromError(err)`; otherwise it returns immediately and the raw `ABORTED` reaches the caller. The common path holds today (`wrapError` produces a `database.UnknownError` that unwraps, `domain.ErrInternal` sets `Parent`, `WithMessage` copies the struct), but several statement paths return a fresh domain error with no parent, which is the shape that would swallow an abort: `internal/storage/v2/dialect/spanner/auth_attempt.go:258,350,373` and `internal/storage/v2/dialect/spanner/session.go:44,77,218`. Those lines are reached on not-found / zero-rows rather than on an abort, so they are not the live bug, but nothing defends the invariant.

2. **`ReadWriteTransaction` re-runs the whole callback, so the callback must be replayable.** `internal/service/user.go:100-115` (`ApplyActions`) runs `action.Prepare` outside the transaction while `action.Apply` mutates action state inside it. `internal/service/project.go:93+` (`CreateProject`) mutates `project` through `GenerateNewKeySet` and calls `keyset.Activate` inside the callback. A retry re-enters both against already-mutated state.

3. **Retry is unbounded without a `ctx` deadline**, as the doc comment at `internal/storage/v2/dialect/spanner/client.go:43-45` already says. `internal/api/integration_test/setup_spanner_test.go` builds one shared pool from `context.Background()`. Under the emulator's global lock, with roughly 134 `t.Parallel()` call sites sharing that pool, a contended transaction can retry until the test timeout, which presents as a hang rather than a clean abort.

- **Do not add an application-level retry loop.** The client already retries `ABORTED`, and a second loop wrapped around it would compound the backoff and hide which layer gave up. The work is to stop defeating the existing retry, bound it, and prove it.
- **Spanner only.** Postgres and SQLite have no serialization retry either (`internal/storage/v2/dialect/postgres/tx.go` `executeTransaction` is a plain Begin / callback / Commit, and sqlite matches), but that is a separate concern from this CI flake and belongs in its own issue.
- **Diagnose before fixing.** Capture the actual failing trace first. If it shows something other than the three gaps above, adjust the remaining tasks rather than building for the wrong cause.
- **The emulator stays the default and stays parallel.** It is the canary. If the fix holds under the emulator at full parallelism, the real-instance path added by #629 has no remaining justification and should be retired in this issue: the operational provisioning it needs (GCP project, service account, Workload Identity Federation, the `GCP_WORKLOAD_IDENTITY_PROVIDER` / `GCP_SERVICE_ACCOUNT` secrets and the `SPANNER_TEST_INSTANCE` variable) is only worth carrying if the emulator genuinely cannot serve. Retire it only once the contention test passes; do not let a real instance mask a missing retry.

## Dependencies

- None blocking. #604 is closed and PR #629 is merged (`5e2f1b73`, 2026-07-31).
- Related: the operational item from #629, whether the GCP provisioning ever landed and whether `main` CI is currently red on the spanner auth step (it runs on push and merge_group, not only on PRs). Retiring that path as part of this issue would close the item.

## Test first (TDD)

Two tests come before any production change, and both are the deliverable rather than scaffolding:

1. A unit test in `internal/storage/v2/dialect/spanner/` asserting the invariant from gap 1: for an `ABORTED` `*spanner.Error` fed through `wrapError`, and again through the `domain.ErrInternal(...).WithMessage(...)` mapping the services apply, `status.FromError` on the result still reports `codes.Aborted`. Extend `error_test.go`, which already table-tests `wrapError`.
2. A contention integration test that forces two overlapping read-write transactions on the same row through `service.DB.Transaction` and asserts both commit. This is the canary that proves the retry loop, so it must live in the parallel `spanner_integration` suite, not be skipped locally.

Then implement, then confirm the raw `ABORTED` no longer surfaces.

## Tasks

- Reproduce first: run the suite against the emulator at full parallelism (`go test -tags spanner_integration -timeout=10m -parallel <N> ./internal/api/integration_test/...`, needs Docker) and capture the failing trace. Record which of the three gaps it points at: a `ctx` deadline consumed by retries, an error mapping that dropped the gRPC status, or a non-replayable callback.
- Add the two tests above.
- Guarantee the status survives mapping. Audit the statement paths that return a bare `domain.Err*()` with no parent (`internal/storage/v2/dialect/spanner/auth_attempt.go` and `session.go` are the known ones) and make them wrap the underlying error, or make `wrapError` classify `codes.Aborted` explicitly so it can never be mapped to a parentless domain error.
- Make the transaction callbacks replayable. Audit `internal/service/user.go` `ApplyActions` and `internal/service/project.go` `CreateProject` for state mutated inside the callback that a retry would re-apply, and move the non-replayable parts out or make them idempotent.
- Fence off the two `*spanner.Client` entry points that do not retry `ABORTED`, `BatchWrite[WithOptions]` and `NewReadWriteStmtBasedTransaction`. Neither is used today, and neither should be adopted without a caller-side retry. A sentence in `internal/storage/v2/AGENTS.md` naming them is enough; do not build a linter for it.
- Bound the retry. Ensure transactions run under a sensible `ctx` deadline so a conflict loop fails fast with a clear deadline error instead of hanging until the test timeout. Cover the shared integration pool in `internal/api/integration_test/setup_spanner_test.go`, which uses `context.Background()` today.
- Once the contention test passes on the emulator at full parallelism, retire the real-instance path from #629: `internal/storage/v2/testdb/{spanner.go,instance.go}`, the `ZITADEL_TEST_SPANNER_*` and Workload Identity Federation steps in `.github/workflows/ci.yml`, and the corresponding `CONTRIBUTING.md` section. If it does not pass, leave that path in place and say so in the issue.

## Reference

`internal/storage/v2/dialect/spanner/client.go` (`Transaction`, and the retry doc comment at lines 43-45), `with_transaction.go`, `db.go`, `tx.go`, `error.go` (`wrapError`) and `error_test.go`; `internal/service/database.go` (`DB.Transaction`, `Transactioner`) with call sites in `internal/service/{user,project,schema_service}.go` and `internal/bootstrap/users/import.go`; `internal/api/integration_test/setup_spanner_test.go` and `helpers/harness.go`; `internal/storage/v2/dbtest/spanner.go` and `internal/storage/v2/testdb/{spanner.go,instance.go}`; `internal/storage/v2/AGENTS.md` for the multi-write rule. The upstream retry decision lives in `cloud.google.com/go/spanner@v1.91.0/retry.go:108-146` and `transaction.go:1965-2005` (`runInTransaction`).

IAM-marco:
Closes #788.

## Summary

The Spanner CI flakiness from #604 was an application bug, not an emulator limitation.

`exchangeConflict` in `internal/storage/v2/session/run_exchange.go` wrapped the conflict sentinel with `%w` but formatted the real cause with `%v`, which flattens it to a string and drops it from the error chain. Spanner's `ReadWriteTransaction` decides whether to retry by looking for a gRPC status in the error the callback returns (`errors.As` for `*spanner.Error`, else `status.FromError`); with the cause flattened it finds neither and returns immediately. Session exchange runs inside that transaction, so under contention a retryable `ABORTED` became a user-facing conflict.

Instrumenting the failing test caught it exactly:

```
cause="*fmt.wrapError failed to insert session: unknown database error:
  spanner: code = \"Aborted\", desc = \"Transaction: 1786118321035086 aborted due to
  another transaction getting priority. The emulator only supports one transaction at a time.\""
grpc_code=Aborted
```

That is the verbatim #604 error, still intact on arrival and destroyed one line later. The fix is `%v` → `%w`.

With that fixed, two workarounds that were suppressing the signal are removed: `moon test-spanner` no longer pins `-parallel 1`, and CI no longer prefers a real Spanner instance over the emulator. The emulator's one-transaction-at-a-time limit is what forces the aborts that keep the retry path honest, so it is now the canary rather than the thing being worked around.

Three findings corrected the original plan in #788, and are worth a reviewer's attention:

- **`wrapError` needed no change.** It already preserves the status; that is why the abort survived all the way *to* `exchangeConflict`.
- **The transaction callbacks were already replay-safe.** `mapStorageError` preserves `Parent`, `GenerateNewKeySet` returns a fresh keyset, `CreateUserAction.Apply` is idempotent. No code change; the requirement is documented instead.
- **The real workaround was `-parallel 1`,** not the real instance. It had been keeping the suite green by removing the contention entirely.

Retry is now also bounded. Nothing in the stack supplied a deadline (the HTTP server sets `ReadTimeout`/`WriteTimeout`, but neither cancels the request context, and there is no timeout middleware), so a hot row could spin indefinitely. A 30s default is applied at every site that opens a `ReadWriteTransaction`, not only `Client.Transaction` — `client.Write`, `client.Update` and `withTransaction` each open their own. A caller with its own deadline keeps it.

The real test-instance path is **kept but unwired**: the Go code in `internal/storage/v2/testdb` still works and the CI steps are commented out rather than deleted, so it can be restored quickly if the emulator turns out not to hold. Removal is tracked in #793.

## Validation

- `go build ./...`, `go vet -tags spanner_integration ./...`, `go test ./internal/...` — clean.
- Spanner integration suite at `-parallel 32`: **5 consecutive green runs**. The same command failed on the first run before the fix.
- `TestTransactionContention` on the emulator: 8 concurrent writers to one row produce **22-26 callback invocations**, all committing, final length exact — so the aborts and retries are real, not incidental.
- The new regression tests fail without the fix (`0x2` Unknown instead of `0xa` Aborted, chain flattened to `domain.Error`) and pass with it.
- Postgres and SQLite integration suites: green.
- Each of the six commits compiles independently (`go vet -tags spanner_integration ./...`), so the history bisects.
- CI on this branch: all checks green. The Spanner step ran at full parallelism and `TestTransactionContention` passed there (0.49s), so the canary holds on CI's amd64 runners and not just locally.

## Release notes / changeset

Changeset: `.changeset/spanner-abort-retry.md` — patch for `@zitadel/server`. Sign-in no longer intermittently fails with a session exchange conflict on Spanner, and read-write transactions now run under a default 30s deadline.

## Notes

One defect found while verifying this, outside its scope and filed separately:

- **#791 — Postgres silently loses concurrent updates.** The new contention canary caught it immediately: only 3 of 8 writers survived. No isolation level is set anywhere, so pgx uses the server default READ COMMITTED, which permits lost updates *and raises no serialization error*, so there is nothing to retry. This is a silent data-correctness bug, and the one I would look at next. The canary is Spanner-tagged for now precisely so it does not red the Postgres suite; re-tagging it is the first step of fixing #791.

Also worth knowing for anyone running these suites on Apple Silicon: 7 subtests in `stmttest` and `TestJSONSchemaStatements_CRUD` fail locally on arm64 and pass in CI, because the arm64 emulator build rounds commit timestamps differently from the amd64 build CI runs. Deterministic, unrelated to parallelism, and nothing broken in CI or shipped. Investigated and closed as not planned (#792); the caveat and its workaround are documented in `CONTRIBUTING.md` so the next arm64 contributor is not left guessing at 7 red tests.

`-p 1` is retained in the moon task deliberately. The suites share one database and packages like `stmttest` assert on global list results, so they still need one package at a time; that is test isolation, not contention avoidance. Removing it surfaces cross-package leakage (a `ListProjects` assertion seeing another package's rows).


internal/storage/session/run_exchange_test.go
```diff
@@ -175,6 +178,43 @@ func TestRunExchange_missingSessionConflict(t *testing.T) {
...
+
+			// The conflict code still reaches the API layer, which maps it to 400.
+			assert.ErrorIs(t, err, domain.ErrSessionExchangeConflict())
+
+			// ...and the abort still reaches Spanner's retry predicate.
+			assert.Equal(t, codes.Aborted, status.Code(err),
+				"ABORTED was stripped; ReadWriteTransaction will not retry")
+			var se *spanner.Error
+			assert.ErrorAs(t, err, &se)
+		})
```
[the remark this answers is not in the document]
IAM-marco: `status.Code` does unwrap. It delegates to `Convert`, which calls `FromError`, which uses `errors.As` (grpc v1.82.1: `status.go:131-134`, `status.go:139-146`, `status.go:112-113`). The doc comment on `Code` states it directly: "Code returns the Code of the error if it is a Status error **or if it wraps a Status error**."

Verified empirically rather than by reading alone. Reverting just this line to `%v`:

```
--- FAIL: TestRunExchange_conflictPreservesAbortedStatus/DeleteAuthAttempt
    expected: 0xa
    actual  : 0x2
```

`0xa` is `codes.Aborted`, `0x2` is `codes.Unknown`. Restoring `%w` makes it pass. So this assertion is precisely the guard that catches the bug: if it always reported `Unknown` as described, it would fail in both states rather than distinguishing them.

On the suggestion itself, the test already asserts on the unwrapped error two lines down, which is the exact predicate `ReadWriteTransaction` uses (`errors.As` for `*spanner.Error`, `retry.go:120-133`):

```go
var se *spanner.Error
assert.ErrorAs(t, err, &se)
```

Keeping both: `status.Code` covers the `status.FromError` branch of the predicate, `ErrorAs` covers the `*spanner.Error` branch. No change.

internal/storage/dialect/spanner/error_test.go
```diff
@@ -123,3 +125,48 @@ func TestWrapError(t *testing.T) {
...
+	t.Parallel()
+
+	got := wrapError(spanner.ToSpannerError(status.Error(codes.Aborted,
+		"Transaction: 1 aborted due to another transaction getting priority.")))
+
+	require.Error(t, got)
+	assert.Equal(t, codes.Aborted, status.Code(got),
+		"ABORTED was stripped; ReadWriteTransaction will not retry")
+	assert.ErrorAs(t, got, new(*spanner.Error))
+}
```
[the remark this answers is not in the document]
IAM-marco: Same as the sibling comment on `run_exchange_test.go`: `status.Code` unwraps. `Code` delegates to `Convert` to `FromError`, which uses `errors.As` (grpc v1.82.1 `status.go:112-113`), and its doc comment says "or if it wraps a Status error".

`database.UnknownError` is specifically in the chain rather than terminating it, because it implements `Unwrap`:

```go
func (e *UnknownError) Unwrap() error { return e.original }
```

So the `*spanner.Error` stays reachable through it, which is why `wrapError` was already correct and needed no change in this PR. This test is a regression guard to keep it that way: `wrapError` maps unrecognised codes to `UnknownError`, and dropping that `Unwrap` (or constructing the error without a parent) would silently disable the retry.

The test does also check the unwrapped error, on the next line:

```go
assert.ErrorAs(t, got, new(*spanner.Error))
```

No change.

internal/api/integration_test/transaction_contention_test.go
```diff
@@ -0,0 +1,80 @@
...
+			errs[i] = harness.DB.Transaction(t.Context(), func(ctx context.Context, tx service.Statementer[service.AllStatements]) error {
+				current, err := tx.Statements().GetTeamByID(ctx, project.ID, team.ID)
+				if err != nil {
+					return err
+				}
+				current.Name += "x"
+				return tx.Statements().UpdateTeam(ctx, current)
+			})
+		}()
+	}
```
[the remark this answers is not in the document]
IAM-marco: Both halves of this are Go 1.22 changes, and `go.mod` declares `go 1.26`.

`for i := range contendingWriters` is range-over-int ([spec: for statements with range clause](https://go.dev/ref/spec#For_range), permitted since 1.22), so it compiles. `go vet -tags spanner_integration ./internal/api/integration_test/` passes, and the repo uses the same form elsewhere (for example `for attempt := range 10` in `internal/storage/testdb/spanner.go`).

Loop variables have also been per-iteration since 1.22 ([wiki: loopvar](https://go.dev/wiki/LoopvarExperiment)), so each goroutine captures its own `i` and writes a distinct index of a pre-sized slice. Distinct elements, no reallocation, so no race.

Confirmed three ways rather than by reading the spec alone:

1. CI ran this test and passed it: `--- PASS: TestTransactionContention (0.49s)`.
2. A standalone `-race` check of the same pattern reports no race, with every goroutine writing its own index: `OK: each goroutine wrote its own index: [0 1 2 3 4 5 6 7]`.
3. The test asserts the exact final length (`initialLen + contendingWriters`), so a wrong-index write would fail it rather than pass silently.

No change.

CONTRIBUTING.md
```diff
@@ -341,20 +340,40 @@ Docker. Point them at a throwaway database — the suites run migrations that
...
+the emulator surfaces a missing retry immediately. The suite therefore runs at
+full parallelism against the emulator, and `TestTransactionContention` asserts
+that concurrent writers to one row all commit.
+
+If you see a raw `ABORTED` ("aborted due to another transaction getting
+priority"), do not serialize the tests and do not move them to a real instance —
+both hide the bug. It means something on that code path stripped the gRPC status
+off the error, so Spanner's `ReadWriteTransaction` stopped recognising it as
+retryable. See the error-wrapping rules in
+[internal/storage/v2/AGENTS.md](internal/storage/v2/AGENTS.md) and #788.
```
adlerhurst: outdated link

CONTRIBUTING.md
```diff
@@ -341,20 +340,40 @@ Docker. Point them at a throwaway database — the suites run migrations that
...
+
+**On Apple Silicon, expect 7 pre-existing failures.** `emulator:latest` has a
+native arm64 build, and it does not round commit timestamps the way the amd64
+build CI runs does. Six subtests in `internal/storage/v2/stmttest` and
+`TestJSONSchemaStatements_CRUD` compare a `created_at` read back after a second
+statement and see tens of microseconds of drift. They are deterministic, they
+are not caused by parallelism, and they pass in CI. Ignore them, or run the
+emulator under `linux/amd64` emulation if you want a clean local run. If you
+touch timestamp handling, verify on CI rather than trusting a local pass.
```
adlerhurst: I don't think we should go that direction and have failing tests on specific operating systems.

adlerhurst (changes requested):
The some links are outdated (`storage/v2/` got refactored to `storage/`)

CONTRIBUTING.md
IAM-marco: You are right, and I have fixed it properly instead: `853de9c2`.

I re-verified on Apple Silicon first, and the failures were still real (the same 7 subtests, deterministic). The cause is architecture: `emulator:latest` has a native arm64 build that returns commit timestamps at a different resolution than the amd64 build CI runs.

So the emulator container is now pinned:

```go
ImagePlatform: "linux/amd64",
```

That is a no-op on amd64 hosts including CI, and Rosetta elsewhere. All 7 failures are gone on arm64, verified with 3 consecutive green runs.

One thing the pin exposed that is worth flagging, because it would have hit every Apple Silicon contributor on their first run: with emulation the cold start (image pull plus emulated boot) exceeded the testcontainers default 60s wait, and the suite failed in setup with no test-level error at all, which is more confusing than the original problem. `postgres.go` already sets `WithStartupTimeout(120 * time.Second)` for the same reason, so the Spanner request now matches it. Verified by deleting the cached image: cold run passes in 28s where it previously failed at 88s.

The paragraph telling people to expect 7 failures is gone. What replaces it is one short note saying the platform is pinned and not to unpin it for speed.

CONTRIBUTING.md
IAM-marco: Fixed in `853de9c2`. The `v2` level was removed by #790, so the link target no longer existed:

```
[internal/storage/v2/AGENTS.md](internal/storage/v2/AGENTS.md)  ->  [internal/storage/AGENTS.md](internal/storage/AGENTS.md)
```

Two more of the same kind in this PR that I caught while checking: `internal/storage/v2/stmttest` in the arm64 paragraph (that paragraph is now gone entirely), and `internal/storage/v2/testdb` in a `ci.yml` comment.

Worth a separate look: the flattening left the same stale path in several files this PR does not touch, including root `AGENTS.md:291` and `CLAUDE.md:22`, which both link to `internal/storage/v2/AGENTS.md`, plus `docs/design/flowengine/architecture.md`, `docs/design/api/permission-storage.md`, `docs/operations/encryption-keys.md` and a number of ADRs. I left those alone to keep this PR scoped, but the two root files are the ones agents read first, so they are probably worth fixing soon. Happy to do it here or in a follow-up, whichever you prefer.

adlerhurst (approved):

Issue #585, opened 2026-07-22
API/domain: ADR 030 contract hardening (status map, details, schema sensitivity)


livio-a:
## Context

Remaining ADR 030 contract work after foundation and leak remediation.

**ADR:** [030 — Error model, mapping, and reporting](https://github.com/zitadel/nextgen/blob/main/docs/adrs/030-error-model-mapping-and-reporting.md)
- Decision 4 → HTTP status (dedup follow-up; **no** `ErrorClass` on `domain.Error`)
- Decision 4 → Response body / Decision 6 → `Details` redaction
- Follow-up table: selective API `details`, status-map dedup, schema-declared sensitivity + path/query audit

**Depends on (done):**
- [#521](https://github.com/zitadel/nextgen/pull/521) — `errreport` + `domain.Error` refinements
- Leak / flow / ogen follow-ups — [#580](https://github.com/zitadel/nextgen/pull/580), [#583](https://github.com/zitadel/nextgen/pull/583), [#584](https://github.com/zitadel/nextgen/pull/584)

**Sibling:** [#48](https://github.com/zitadel/nextgen/issues/48) — storage capture / log-safe `LogValue` (independent)

## Current state

1. **Status mapping** — each resource has a near-identical `*_errorResponse` switch (`user`, `project`, `team`, `session`, `schema`, `flow`, `auth_attempt`, …) keyed by `err.Code`. Prefix routing in `errorResponse` is fine; the per-code switches are repetitive.
2. **`details`** — OpenAPI already allows optional `details`. `domainErrorDetails` already marshals `Details` when a producer sets them. There is **no** per-code allow-list (ADR: setting `Details` *is* exposing it). Gaps are producer discipline, reviewed reference patterns, and cleaning accidental string/`WithDetails` payloads that are not structured client-actionable data.
3. **Schema sensitivity** — user-schema `x-sensitive` exists (audit redaction). Broader contract work: declare sensitivity consistently, and audit that secrets/PII never appear in path/query (safe-to-log by construction).

### A. Deduplicate `code → HTTP status` (API only)
- [ ] Introduce a shared helper and/or suffix conventions / small manifest in `internal/api` for common cases (`*.not_found` → 404, `*.already_exists` → 409, `*.permission_denied` → 403, `*.invalid*` → 400, …)
- [ ] Keep **explicit overrides** for auth/opaque cases (same logical failure ≠ always same status)
[...]

PR #799, opened 2026-08-10
fix(api): selective details producers for ADR 030

livio-a:
## Summary

Implements [#585](https://github.com/zitadel/nextgen/issues/585) **task B only** (selective / reviewed `details` producers), started fresh from current `main`.

- Extract `marshalErrorDetails` for the legacy `details.details` wire slot; keep `FullErrorInResponse` (`details.parent`) composing beside it
- Add `RequestInvalidFieldDetails` and attach `{field}` on flow `Fields` decode failures
- Move accidental string `WithDetails` blobs to `Message` (user create / schema / unsupported proof)
- Document producer contract + inventory in `docs/design/api/error-details-producers.md`
- Cross-link from ADR 030

**Not in this PR:** central `code → HTTP status` map (task A), path/query sensitivity audit (task C), `isoduration` / session TTL encoding.

Supersedes draft [#586](https://github.com/zitadel/nextgen/pull/586).

## Validation

- `go test ./internal/api/ ./internal/domain/ ./internal/service/`

## Release notes / changeset

- Changeset: `.changeset/issue-585-selective-details.md` — structured `details.field` on flow Fields decode errors; producer contract docs (`@zitadel/server` patch)

## Notes

- Authz opaque `WithDetails("project does not exist")`, flowdef string details, list/filter strings, and crypto/token details are inventoried as deferred
- API integration tests should keep asserting public `code`/`message`/`details` only — not `FullErrorInResponse` parent chains

<div><a href="https://cursor.com/agents/bc-acf6b6fd-2b35-4033-8ca0-1e395f67e055?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-acf6b6fd-2b35-4033-8ca0-1e395f67e055&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


docs/design/api/error-details-producers.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Reference producers
+
+### Request field validation — `RequestInvalidFieldDetails`
+
+Flow `Fields` value decode failures attach `{field}` only (no parser text, no payload fragments). See [`SubmitFlowStep`](../../../internal/api/flow.go) and [`flowFieldDecodeError`](../../../internal/api/flow.go).
+
+### Session TTL — `SessionInvalidTTLDetails`
+
+[`internal/domain/session.go`](../../../internal/domain/session.go) attaches `{ttl,max_ttl}` as structured duration values. Encoding stays with the shared details marshal path (ISO string form may be refined later).
```
IAM-marco: This sentence says two things about the session TTL error, and neither one matches the code in this branch.

**"Encoding stays with the shared details marshal path"**

It doesn't. The session error handler has a dedicated branch for the invalid-TTL error that builds the entire response body by hand with a JSON encoder, instead of going through the `domainErrorDetails` helper that the rest of this document describes: [`internal/api/session.go#L338-L366`](https://github.com/zitadel/nextgen/blob/24042a14314c3ccb36e9c00ba2576a6b1ba2324b/internal/api/session.go#L337-L367). Because of that, it never calls `marshalErrorDetails`, which is the helper this PR adds and which the "Wire shape" section above tells producers to use.

**"ISO string form may be refined later"**

That work is already done. The two durations are written through `ogenx.ISODuration`, whose `Encode` calls `e.Str(formatISODuration(...))`, so they go on the wire as ISO strings today: [`internal/api/session.go#L352-L355`](https://github.com/zitadel/nextgen/blob/24042a14314c3ccb36e9c00ba2576a6b1ba2324b/internal/api/session.go#L352-L355) and [`internal/api/ogenx/iso8601_duration.go#L15-L19`](https://github.com/zitadel/nextgen/blob/24042a14314c3ccb36e9c00ba2576a6b1ba2324b/internal/api/ogenx/iso8601_duration.go#L15-L19).

**Why this particular spot matters**

This file is being introduced as the contract that producers are supposed to follow, and ADR 030 now links to it as the authority. Putting session TTL under the heading "Reference producers" tells the next person to copy it. But it is the one producer in the codebase that does not follow the contract, so copying it would spread the exception rather than the rule.

**A consequence the doc currently hides**

Because that branch skips `domainErrorDetails` entirely, session TTL errors never get the `details.parent` diagnostic block, not even in integration tests where `FullErrorInResponse` is switched on ([`setup_pg_test.go#L23`](https://github.com/zitadel/nextgen/blob/24042a14314c3ccb36e9c00ba2576a6b1ba2324b/internal/api/integration_test/setup_pg_test.go#L20-L26)). Every other error in the API does get it. That is a real behavioural difference, and it is worth writing down rather than leaving for someone to rediscover while debugging a red test run.

**Suggested fix**

Move session TTL out of "Reference producers" and into the Inventory table with status **Deferred**, noting that it composes its own response envelope and should be folded into the shared path later. If you would rather it stay a reference, then the code is what needs to change: have the branch attach `SessionInvalidTTLDetails` and let `marshalErrorDetails` encode it, so that the ISO formatting lives in the shared path the doc claims it already lives in.

**Coverage**

Nothing in `internal/api` asserts the session TTL response shape today. The only test is [`internal/domain/session_ttl_test.go`](https://github.com/zitadel/nextgen/blob/24042a14314c3ccb36e9c00ba2576a6b1ba2324b/internal/domain/session_ttl_test.go#L40-L64), which checks the domain-side struct, not what reaches the client. So whichever direction you pick, this doc is currently the only place the decision is recorded, which is exactly why it is worth getting right here.

IAM-marco (commented):
Claude comments

IAM-marco (approved):

PR #804, opened 2026-08-10
feat(config): validate nested field paths in zitadel plan

vitorbari:
## Summary

Teaches `zitadel plan` about nested user-schema properties, so it reaches the same verdict the
server does. A step naming a leaf by its dotted path validates locally, an object- or
array-typed property is reported as not collectable, and a required object counts as covered
when a step collects one of its leaves.

Without this, `plan` accepts a step the server rejects, and rejects one the server accepts.

Mirrors server validations changes introduced on #785


packages/config/src/validate.ts
```diff
@@ -616,14 +658,42 @@ function resolveFieldChallenge(
...
+  // Mirrors deriveUserPropertyType: an object or array has no
+  // field-shaped input. A property carrying `properties` is an object,
+  // and one carrying `items` is an array, even when it omits the `type`
+  // keyword.
+  if (
+    property.type === "object" ||
+    property.type === "array" ||
+    schemaProperties(property) !== undefined ||
+    "items" in property
+  ) {
```
fforootd, line 710: **A nullable object/array slips through here as a scalar.**

This test reads the raw `type` keyword, but the nullable reduction that turns `["null", X]` into `X` doesn't run until line 699 — *after* this check. So a nullable composite matches none of the four clauses, falls through to the union reduction, and is accepted as a single non-null type.

I ran both cases through `validateFlowDefinition` on this branch:

| `address` in the user schema | `zitadel plan` (this code) | `zitadel apply` (Go) |
| --- | --- | --- |
| `{ "type": ["null", "object"] }` | *(no error at all)* | `not a scalar property, name a nested leaf instead` |
| `{ "type": ["string", "number"], "properties": {…} }` | `not a scalar property…` | `unsupported JSON type: [string number]` |

Row 1 is exactly the plan/apply split this PR exists to close. Row 2 is milder but still misleading — the CLI hands out advice that doesn't apply, and following it still fails on apply.

`deriveUserPropertyType` is reduce-then-test, which is what makes both rows come out right on the server:

```go
jsonType, err := prop.JSONType()   // union error returns here, first
if err != nil {
    return "", err
}
if jsonType == "object" || jsonType == "array" || prop.HasProperties() || prop.HasItems() {
```

Mirroring that ordering means hoisting the block currently at 697–705 above this one and testing the reduced value:

```ts
// Mirrors schemaReader.JSONType: the nullable idiom `["null", X]`
// reduces to X; any other multi-entry union is unsupported. Go runs
// this before the scalar test, so the union error wins and the
// object/array test sees the reduced type.
let jsonType = property.type;
if (Array.isArray(jsonType)) {
  const nonNull = jsonType.filter((t) => t !== "null");
  if (nonNull.length > 1) {
    return { message: `flow field: unsupported JSON type: [${jsonType.map(str).join(" ")}]` };
  }
  jsonType = nonNull[0]; // undefined when every entry was "null" — matches Go's "" case
}

// Mirrors deriveUserPropertyType: an object or array has no
// field-shaped input. …
if (
  jsonType === "object" ||
  jsonType === "array" ||
  schemaProperties(property) !== undefined ||
  "items" in property
) {
```

…and dropping the now-redundant reduction at 697–705. (Left as a code block rather than a suggestion since the edit spans past the end of this hunk.)

Worth fixing rather than deferring, I think, because the idiom is reachable in practice: `validate.test.ts:399` already treats `["null", "string"]` as a supported pattern, and `api/openapi/endpoints/schemas/user-property.yaml` is `additionalProperties: true` with no constraint on `type`, so nothing stops an author writing `["null", "object"]`.

---

packages/config/src/validate.ts
```diff
@@ -616,14 +658,42 @@ function resolveFieldChallenge(
...
-  const property = properties[field];
-  if (property === undefined) {
-    return { message: `flow field: not a property in the user schema: ${q(field)}` };
+  // Mirrors walkUserProperty: a nested property is addressed by its
+  // dotted path, descending one `properties` level per segment.
+  const segments = field.split(".");
+  let property: unknown = undefined;
+  let level: Record<string, unknown> | undefined = properties;
+  for (const [i, segment] of segments.entries()) {
+    property = level?.[segment];
```
fforootd: **`Object.prototype` members resolve as schema properties.**

`level?.[segment]` walks the prototype chain, so any inherited member is "found" and the `undefined` guard on the next line never fires. Confirmed on this branch — `fields: ["toString"]`, `["valueOf"]` and `["constructor"]` each produce zero issues, and the new walk extends that to nested forms like `address.toString`.

Go can't hit this because the lookup is a map index over `types.PartMapSchema`, where a missing key is unambiguously not-found:

```go
prop, ok := properties[name]
```

An own-property guard restores that:

```suggestion
    property = level !== undefined && Object.hasOwn(level, segment) ? level[segment] : undefined;
```

Fully your call whether this belongs in this PR — the single-segment case is pre-existing, so it's a latent bug you inherited rather than one you introduced. The reason I'd lean toward folding it in: this loop is where one lookup became N lookups per path, so the same guard now covers every level instead of just the top one. `requiredPaths`' `properties?.[name]` at line 611 has the same shape and would want the same treatment.

---

packages/config/src/validate.ts
```diff
@@ -586,6 +586,48 @@ function schemaRequired(schema: Record<string, unknown>): string[] {
...
+ * enforces the rest of its `required` list.
+ */
+function requiredPaths(
+  schema: Record<string, unknown>,
+  materialized: ReadonlySet<string>,
+  prefix = "",
+): string[] {
+  const properties = schemaProperties(schema);
+  const required = schemaRequired(schema);
+  const out: string[] = [];
```
fforootd: **Small one: a `Set` here would match the Go mirror's dedup.**

`RequiredPaths` accumulates into a set on both ends — the return type is `map[string]struct{}`, and even the input list is iterated as `RequiredSet()`:

```go
func (r schemaReader) RequiredPaths(materialized map[string]struct{}) map[string]struct{} {
	out := map[string]struct{}{}
```

An array can't collapse a repeat. With `"required": ["address", "address"]` and `address` declaring `required: ["street"]`, the first loop recurses twice and the CLI prints:

```
required fields [address.street address.street] in user schema are missing in the flow definition steps
```

where the server prints the path once. Verified on this branch.

Granted, a duplicated `required` entry is an odd schema to write — but since the whole point of this file is that the two messages agree, and the change is small:

```suggestion
  const out = new Set<string>();
```

with `out.push(path)` → `out.add(path)`, the two `out.push(...requiredPaths(…))` spreads → `for (const p of requiredPaths(…)) out.add(p)`, and `return [...out]` at the end. The `missing` filter/sort downstream is unaffected either way.

---

packages/config/src/validate.test.ts
```diff
@@ -396,6 +396,145 @@ describe("schema-dependent rules", () => {
...
+      };
+      return withNested;
+    };
+
+    it("accepts a nested leaf addressed by its dotted path", () => {
+      const def = flow();
+      step(def, "register").fields.push("address.street");
+      expect(messages(validateFlowDefinition(def, nested()))).not.toContain(
+        'step "register": flow field: not a property in the user schema: "address.street"',
+      );
```
fforootd: **The accept-case assertions pass for any wrong verdict, not just the right one.**

`.not.toContain(<one exact string>)` only rules out that single message. This test stays green if the walk starts returning `not a scalar property, name a nested leaf instead: "address.street"`, and also if an earlier phase begins erroring so `validateAgainstSchema` never runs at all — it's gated on `issues.length === 0`, so an upstream failure reads as success here.

Not hypothetical for this diff: the nullable-type issue I flagged in `validate.ts` is a mis-classification inside the exact function these tests cover, and none of the four accept-cases would go red on it.

The file already has the stronger idiom at line 269:

```ts
expect(errors(validateFlowDefinition(def, schema))).toEqual([]);
```

which pins the whole verdict and prints the unexpected issues when it fails:

```suggestion
      expect(errors(validateFlowDefinition(def, nested()))).toEqual([]);
```

Same shape would fit lines 470, 489 and 496. The reject-cases are fine as they are — `.toContain` is right when you're asserting one specific message appears.

---

fforootd (commented):
Nice port — I diffed the TypeScript against the Go it mirrors (`walkUserProperty` / `deriveUserPropertyType` / `schemaReader.RequiredPaths` from #785) clause by clause, since "the CLI reaches the same verdict as the server" is this PR's whole contract.

It holds up on everything the tests cover: the dotted-path walk, the intermediate-scalar rejection, required-object recursion, the optional-object materialization rule, and every error string down to the `%q`/`%v` formatting. The `covered` prefix expansion is byte-equivalent to `AttributeKey.Nodes()` + `AppendNode()`, including the degenerate-input quirks. Both production call sites (`apps/cli/src/lib/sync/flow-validation.ts:62`, `packages/api-mock/src/platform-handlers.ts:443`) are safe — the new `covered` set is a strict superset of the old one, so the change can only relax the required-fields rule, never tighten it spuriously.

Four divergences, left inline. The first is the one I'd want resolved before merge; the rest are your call on scope:

1. **`validate.ts:686`** — the not-scalar test reads the raw `type` before the nullable reduction runs, so `{ "type": ["null", "object"] }` validates clean in `plan` and then fails in `apply`. That's the exact split this PR closes, reopened for one schema shape.
2. **`validate.ts:667`** — path segments resolve through the prototype chain, so `fields: ["toString"]` passes. Pre-existing at one level; the new loop carries it to every level.
3. **`validate.ts:607`** — array where Go uses a set, so a repeated `required` name prints the path twice.
4. **`validate.test.ts:418`** — four accept-cases assert `.not.toContain(<one message>)`, which passes for any *other* wrong verdict. None of them would catch #1.

I verified 1–3 by running the schemas through `validateFlowDefinition` on this branch rather than reasoning from the source, so the CLI-side outputs quoted inline are observed, not predicted. The Go-side outputs are read off #785.

One caveat on method: the multi-agent fan-out and adversarial verify pass this review normally runs weren't available in my session, so this is a single-pass read with execution standing in for the verify stage. Worth a second pair of eyes on the parts I cleared, particularly the required-path recursion — I hand-traced it against `collectRequiredPaths` including the four-level `address.geo.lat` case and it matches, but that's the densest logic in the diff.

---

packages/config/src/validate.ts
vitorbari: Fixed in ee1df783b — reduce first, then test, matching `deriveUserPropertyType`.

Both rows in your table now agree with the server. Two regression pins:

- `rejects a nullable object` — `{ "type": ["null", "object"] }` reports `not a scalar property, name a nested leaf instead: "address"`.
- `reports an ambiguous union on a property that also declares properties` — the union error wins over the composite test, so the message is `unsupported JSON type: [string number]` rather than advice to name a nested leaf.

I mutation-checked the ordering rather than trusting the reordering by eye: reading the raw `type` in the composite test fails only the first pin, and re-inserting a composite test above the reduction fails only the second. Neither mutation disturbs the other 165 tests, so each pin covers exactly one direction of the ordering.

One deviation from your sketch: I kept `schemaProperties(property) !== undefined || "items" in property` reading `property` rather than the reduced type, since those two clauses are structural rather than type-derived — Go tests `prop.HasProperties() || prop.HasItems()` against the reader, not against `jsonType`. The `jsonType = nonNull[0]` assignment lands as you wrote it, including the `undefined`-when-all-null case matching Go returning `""` for `len(nonNull) == 0`.

packages/config/src/validate.ts
vitorbari: Folded in — agreed on your reasoning that the loop turning one lookup into N is what makes it this PR's business.

`ee1df783b` uses the own-property guard you suggested, and it applies at every segment since `level` is reassigned inside the loop. Two pins: `rejects Object.prototype member names as fields` covers `toString`, `valueOf` and `constructor` at the top level, and `rejects an inherited member reached through a nested path` covers `address.toString`. Reverting to `level?.[segment]` fails exactly those two.

On `requiredPaths` line 611 — I applied the same guard there for symmetry, but I want to be straight that it fixes nothing observable. Every prototype member that lookup can reach falls out the same way with or without it: `properties["toString"]` is a function so `isPlainObject` is false, `properties["__proto__"]` is `Object.prototype` whose `schemaRequired` is empty, and both arms push the bare path — which is also what Go does when the map index misses. So it is a consistency guard on a lookup that is semantically a map index, not a second bug. Happy to drop it if you would rather not carry a line that no test can hold down.

packages/config/src/validate.ts
vitorbari: Done in `ee1df783b` — `new Set<string>()`, `out.add(path)`, the two spreads became `for (const nested of …) out.add(nested)`, and `return [...out]`.

Pinned by `reports a repeated required name once`: `required: ["address", "address"]` with `address` declaring `required: ["street"]` now prints `required fields [address.street] …` once. Swapping the set back for an array fails only that test.

Agreed the schema is odd to write by hand, but it is cheap to generate one — and the whole contract of this file is that the two messages agree, so a divergence that only shows up on a weird input is still a divergence.

packages/config/src/validate.test.ts
vitorbari: Good catch, and the gating point is the sharper half of it — `validateAgainstSchema` only runs when `issues.length === 0`, so an upstream failure would have read as a pass in all four.

All four accept-cases now use `expect(errors(validateFlowDefinition(def, …))).toEqual([])`. That is the four you listed: the dotted-leaf accept, the required-object-covered-by-any-leaf case, the optional-object-satisfied case, and the untouched-optional-object case. Left the reject-cases on `.toContain`, agreed that is the right assertion when you are pinning one specific message.

Your prediction held: with the nullable fix reverted, the strengthened accept-cases still pass — the mis-classification needs its own pin, so it got one rather than being caught here. What the change does buy is that all four now go red if any earlier phase starts erroring, which was the failure mode that would have been silent.

PR #805, opened 2026-08-10
feat: satisfy a required object by collecting one of its leaves

vitorbari:
## Summary

Stacked on #785 — review that first.

A user schema declaring `required: ["address"]` has no valid flow definition: the object itself
is not collectable, and naming `address.street` does not count as covering it. Coverage compared
the schema root's `required` names against whole field names.

The walk now descends nested `required` arrays, and a required path is covered by itself or by
anything beneath it — an object materializes once one of its children is collected.

## Validation

- `moon run server:test`
- `moon run config:test` (the CLI mirror and its drift audit still pass unchanged)

## Release notes / changeset

- Changeset: `.changeset/nested-required-coverage.md` — minor for `@zitadel/server`.

## Notes

- `RequiredLeafPaths` travels with this PR rather than #785, since nothing else calls it.
- The matching `zitadel plan` rule is in #804.

### DIFF
--- a/.changeset/nested-required-coverage.md
+++ b/.changeset/nested-required-coverage.md
@@ -1,0 +1,9 @@
+---
+"@zitadel/server": minor
+---
+
+A flow definition satisfies a required object in the user schema by collecting
+one of its leaves. Required coverage walks the nested `required` arrays rather
+than only the schema root, so a schema declaring `required: ["address"]` with
+`address.required: ["street"]` is satisfied by a step collecting
+`address.street`.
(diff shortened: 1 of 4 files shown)


vitorbari:
Folded into #785 (commit b9f61ab4) and closing this.

Three reviewers independently flagged the same gap on #785 — [Copilot](https://github.com/zitadel/nextgen/pull/785#discussion_r3751291585), @wim07101993, and @fforootd — that a schema whose top-level `required` contains an object property could not save any definition: `validateRequiredUserSchemaFields` demanded the literal name while `ErrFlowFieldNotScalar` forbade it.

That is a regression #785 introduced and this PR removed, so splitting them left a reviewable-but-broken intermediate. Keeping them together makes the required-coverage rule verifiable against the resolver change it depends on.

The two changesets are merged into one release note, since it is now a single change.

PR #806, opened 2026-08-10
feat: session list with state, user and createdAt filters

### LINKED ISSUE #747: List sessions - Implement handler and service
**Context:** Storage already has a capable `ListSessions` (filter + cursor + ordering), but `sessionService.List` [is stubbed](https://github.com/zitadel/nextgen/blob/main/internal/service/session.go#L118).

**Specifications:**
- Replace the service stub so the handler calls the existing storage `ListSessions`
- Pass the request through: filters, sorting, and pagination from the body → storage; return the sessions and `next_page_token`
- Default sort to newest-first (`createdAt DESC`) when none is given
- Working filters this ticket: `user_id` (scoped by `project_id`). `state` is wired but only meaningful after https://github.com/zitadel/nextgen/issues/748

**Acceptance criteria:**
- [ ] `POST /sessions/query` returns real, project-scoped results
- [ ] Filter by `user_id` works
- [ ] Results are newest-first by default; `sorting` direction respected
- [ ] Pagination works; `next_page_token` round-trips
- [ ] Tests cover query, user filter, pagination, ordering

grvijayan:
Closes #747

## Summary

- `POST /sessions/query` supports filters on `user_id`, `created_at`, and `state`, sorting with the id tiebreaker, and cursor pagination with the page token round-tripped through the response.
- State is computed rather than stored, so the filter expresses it over stored data:
  - `expired`: `expires_at < now()`
  - `building`: `expires_at >= now()` and no verified factors
  - `active`: `expires_at >= now()` and at least one verified factor
- The factor test compiles to a correlated `EXISTS` over the session's verified checks, matching `State()`'s `len(Factors)` test directly:
https://github.com/zitadel/nextgen/blob/310014f1ec8df441b161d12bb01658d27aa1f478/internal/domain/session.go#L121-L129

## Validation

- `go test ./internal/service/ ./internal/api/ ./internal/storage/dialect/... -count=1`
- `go test -tags sqlite_integration ./internal/storage/stmttest/ -run TestSessionStatements -count=1`
- `go test -tags postgres_integration ./internal/api/integration_test/ -run TestQuerySessions -count=1`
- `go test -tags spanner_integration ./internal/api/integration_test/ -run TestQuerySessions -count=1`

## Notes

- An earlier revision used `user_id IS NULL` as a proxy, which misclassified password-only sessions (factors promoted, no user bound) as building.
- The equality compiles as bare `EXISTS` / `NOT EXISTS` (new `compare.CompileBoolEqual`, shared by all three dialects), so planners convert it to a semi/anti-join instead of evaluating a subquery per row. The probe is covered by the existing `checks (project_id, session_id)` index on every dialect.
- A session's state can drift across pages of a paginated run, but not within a request, because `now` is bound once per request.
- New harness helpers: `CreateSession`, `CreateActiveSession`, `CreateUserWithTeam`, and `SetScopedTokenOnApiClient`. The last one mints a bearer with explicit scopes because no real credential carries `session.*` until ADR 036's credential planes land.


internal/api/integration_test/session_test.go
```diff
@@ -0,0 +1,154 @@
...
+	require.NoError(t, err)
+	harness.SetScopedTokenOnApiClient(t, client, project, "session.read")
+
+	querySessions := func(t *testing.T, req *api.QuerySessionsRequest) *api.QuerySessionsResponse {
+		t.Helper()
+		res, err := client.QuerySessions(t.Context(), req, api.QuerySessionsParams{ProjectID: api.ProjectID(project.ID)})
+		require.NoError(t, err)
+		require.IsType(t, &api.QuerySessionsResponse{}, res, helpers.MustMarshal(t, res))
+		return res.(*api.QuerySessionsResponse)
+	}
```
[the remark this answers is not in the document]
grvijayan: fixed in https://github.com/zitadel/nextgen/pull/806/commits/67955b4741c39aca025d8d7dc8a3189dffb1bc22

internal/service/session.go
```diff
@@ -115,9 +140,127 @@ func (s *sessionService) Get(ctx context.Context, input GetSessionInput) (*domai
...
+	expiresAt := database.Col(domain.SessionFieldExpiresAt)
+	userID := database.Col(domain.SessionFieldUserID)
+	// todo (grvijayan): replace with a greater-or-equal compare once the filter layer supports it
+	activeSessionsFilter := database.Or(database.GreaterThan(expiresAt, now), database.Equal(expiresAt, now))
+
+	switch state {
+	case sessionStateExpired:
+		return database.LessThan(expiresAt, now), nil
+	case sessionStateBuilding:
+		// todo (@grvijayan) review: user_id stands in for State()'s factor test.
```
grvijayan: @livio-a / @vitorbari - would like your opinion on translating `building` based on `user_id` being unset.

internal/api/list.go
```diff
@@ -0,0 +1,37 @@
...
+
+// list-request translation shared by the query endpoints.
+
+// sortingToService maps a request sorting object; F is the endpoint's sort-field enum.
+func sortingToService[F ~string](field F, direction api.SortDirection) *service.Sorting {
+	return &service.Sorting{Field: string(field), Direction: string(direction)}
+}
+
+// filterToService maps a request filter item; F is the endpoint's filter-field enum.
+func filterToService[F ~string](field F, operation api.FilterOperation, value api.OptFilterValue) service.Filter {
```
vitorbari, line 16: The loop over req.Filter is repeated across all endpoints, maybe we can move it here as well

> for _, filter := range req.Filter {
		input.Filters = append(input.Filters, filterToService(filter.Field, filter.Operation, filter.Value))
	}

internal/service/session.go
vitorbari: The domain definition of `building` is `len(s.Factors) == 0`: https://github.com/zitadel/nextgen/blob/f720dbb5b2a8ea974bad87263bd3e1e0fd377eca/internal/domain/session.go#L125

I think we should keep it, because there might be flows where we identify the user (`sessions.user_id` set) but the authentication factors aren't completed yet, for example, an identifier-first step, or a passkey prompt awaiting completion. In that window the session should still report `building`.

That is a future concern, today user_id and factors are both only during exchange.

grvijayan: Thanks. I'll update it to be based on the factors then (`last_verified_at IS NOT NULL`)

internal/api/list.go
grvijayan: Moving the loop to the shared helper would mean generics over the generated filter types and accessing `Field`, `Operation`, and `Value` via their getters, along the lines of:

```
 func filtersToService[F ~string, T any, PT interface {
     *T
     GetField() F
     GetOperation() api.FilterOperation
     GetValue() api.OptFilterValue
 }](items []T) []service.Filter

 // call site — F can't be inferred from a constraint's method signature:
 svcReq.Filters = filtersToService[api.TeamFilterField](req.Filter)
```

The getter signatures are ogen-generated, and relying on them could be tricky: if an endpoint defines its filter differently, its item type might not match these getters and that endpoint can't use the helper.

It feels simpler to just duplicate the loop in the list endpoints. What do you think?

Also, I realize this helper (and the sort one) are just one line changes that could be inlined in the loops if we don't go with the generic helper.

vitorbari: 👍 
Sorry, I did not spot the complexity

internal/service/session.go
grvijayan: fixed in https://github.com/zitadel/nextgen/pull/806/commits/a3348605077fff4fa7f06322112474334975e56c

vitorbari (approved):
We still have to clarify how the user data is exposed in the session list, but we can follow up on that in a separate PR.

PR #808, opened 2026-08-10
feat: implement wide-events audit log

adlerhurst:
## Summary

Completes Path B audit emitters and addresses thermo-nuclear review follow-ups on the events machinery:

- Envelope owns identity (no primary-id echo in payload)
- Creates emit full allowlisted snapshots; updates emit deltas
- Auth checks include `auth_attempt_id`; attempt/schema create payloads stay empty
- Password factor events use the persisted password row id (`RETURNING` / `THEN RETURN id`, no follow-up Get)
- Shipper: per-(sink, project) deadline, sequential deliver, one cursor upsert per page
- RequestBuffer: `BatchSize` takes a prefix; drop `HighWatermark` / `FlushInterval`
- Dialect-minted sink ids on natural key `(type, scope, url)`; `deploymentSinkID` removed
- Claimed-project export list via storage `ListClaimedProjectIDs`; Path A inserts through `EventExportAdapter`
- `GET /events` `order=asc|desc` (default `desc`); inclusive `created_after` via `GreaterThanOrEqual`
- Retention `Close` cancels in-flight purge (5s bound); sink `Enabled` is opt-out
- Path A `request.api` `occurred_at` is HTTP request **start** so a `request_id` group sorts the envelope first
- Login/flow HTTP now emits `request.api` when `project_id` is stamped (no token required); `healthz` / `readyz` / `livez` still skip
- Path B `FromContext` copies `request_id` from the actor slot and HTTP middleware, so in-TX events from public login/flow and `POST /projects` share the same `req_…` as the API call
- `POST /projects` stamps the minted `project_id` after create so Path A emits `request.api` for that request
- `createFlow` mints `flow_id` and stamps the actor slot **before** the state machine starts, so Path B `auth.attempt.created` shares `flow_id` with Path A `request.api`

## Validation

- `go test ./internal/audit/ ./internal/api/ ./internal/service/`
- `go test ./internal/service/ ./internal/domain/ -run 'TestFlowService_Start_|TestAuthAttemptService_Create|TestFlowStateMachine'`
- `moon run server:format`

## Release notes / changeset

- `.changeset/events-api-get.md` — `GET /events` / `GET /events/{id}`, `order=asc|desc`, login/flow and `POST /projects` `request.api` when `project_id` is known, Path B copies `request_id` from the HTTP request, `createFlow` stamps `flow_id` before `auth.attempt.created` (`@zitadel/server` minor)
- `.changeset/wide-events-audit-log.md` — Path B payload shape alignment (`@zitadel/server` minor)

## Notes

- Follow-up issues for deferred catalog types: #877–#884
- Public flow events leave actor/token empty (`Authenticated` stays false)
- Invalid flow cookies still have no Path A row (no project to scope)
- Failed `POST /projects` still skip Path A (no project id to tenant-scope)
- `auth.attempt.created` payload stays empty; `flow_id` / `session_id` are event columns, not payload fields

<div><a href="https://cursor.com/agents/bc-132f4b3f-9044-4d5e-b3d1-d63ece4aef9e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-132f4b3f-9044-4d5e-b3d1-d63ece4aef9e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


internal/service/event_claim.go
```diff
@@ -0,0 +1,32 @@
...
+
+// projectIsClaimed reports whether a project has completed claim (ADR 046 / 049):
+// the project exists and resource_scope_index.team_id is set.
+func projectIsClaimed(ctx context.Context, stmts interface {
+	GetProjectByID(ctx context.Context, id string) (*domain.Project, error)
+	GetResourceScope(ctx context.Context, resourceID string) (*domain.ResourceScope, error)
+}, projectID string) (bool, error) {
+	_, err := stmts.GetProjectByID(ctx, projectID)
+	if err != nil {
+		if errors.Is(err, new(database.NoRowFoundError)) {
```
fforootd: nit (consistency): the codebase idiom for this check is `errors.AsType[*database.NoRowFoundError](err)` (~17 existing sites); this PR introduces `errors.Is(err, new(database.NoRowFoundError))` in four places (here ×2, `event.go:98`, `shipper.go:134`). Both work — `NoRowFoundError.Is` matches by type — but converging on one form keeps grep-ability. Fine as a follow-up sweep too.

api/openapi/endpoints/events/methods.yaml
```diff
@@ -0,0 +1,130 @@
+get:
+  operationId: listEvents
+  summary: List events
+  description: |
+    Returns project-scoped audit events, newest-or-oldest by keyset on
```
fforootd: optional: the description says “newest-or-oldest by keyset”, but the implementation is ascending-only with no direction parameter — so reaching the latest event means paginating the full history. Since an audit UI's default view is latest-first, an `order=asc|desc` param (or a desc default) would be much cheaper to add now, while the surface is unreleased, than after. Happy for this to be a fast-follow if you'd rather keep this PR contained.

internal/audit/shipper.go
```diff
@@ -0,0 +1,230 @@
...
+				slog.String("event_id", ev.ID),
+				slog.String("error", err.Error()),
+			)
+			return
+		}
+	}
+}
+
+func (s *Shipper) deliver(ctx context.Context, sink *domain.EventSink, ev *domain.Event) error {
+	body, err := json.Marshal(ev)
```
fforootd: optional: `json.Marshal(ev)` on the untagged `domain.Event` exports PascalCase field names (`ProjectID`, `EventType`, …) plus the internal `OccurredAtWait` hint — a different shape than the snake_case `Event` schema the API documents. Reusing the API wire marshal (e.g. hoisting `marshalDomainEvent` from `internal/api` into the shared `events` package) would give SIEM/webhook consumers the documented shape and one canonical encoder.

internal/audit/shipper.go
```diff
@@ -0,0 +1,230 @@
...
+func (s *Shipper) Close() {
+	select {
+	case <-s.stop:
+	default:
+		close(s.stop)
+	}
+	<-s.done
+}
+
+func deploymentSinkID(sinkType domain.EventSinkType, url string) string {
```
fforootd: non-blocking observation: hashing `(type, url)` makes sink identity stable across config reordering — nice. Two side effects worth a line in the ADR caveats: changing a webhook URL re-keys the cursor, so the new sink re-ships all retained history from epoch; and sinks removed from config leave their `event_sinks` row enabled plus orphaned cursors behind. Both defensible for v1, just worth documenting so ops aren't surprised.

internal/service/event_export.go
```diff
@@ -0,0 +1,85 @@
...
+
+// EventExportAdapter adapts AllStatements to audit retention/shipper interfaces.
+type EventExportAdapter struct {
+	Pool *DB
+}
+
+func (a EventExportAdapter) ListProjectIDs(ctx context.Context) ([]string, error) {
+	res, err := a.Pool.Statements().ListProjects(ctx, &database.ListOptions[domain.ProjectField]{
+		Pagination: database.Page[domain.ProjectField]{
+			Limit: 10000,
```
fforootd: optional (scale): the flat `Limit: 10000` silently truncates — projects beyond the first 10k never get purged or shipped. A cursor loop here would remove the cliff. Relatedly, `ListClaimedProjectIDs` does an N+1 `GetResourceScope` per project on every shipper tick (default 5s); a single filtered list/join over `resource_scope_index` would collapse that. Neither bites at today's fleet sizes, so follow-up-sized.

internal/service/user.go
```diff
@@ -143,20 +125,47 @@ func (s *userService) ApplyActions(ctx context.Context, actions ...UserAction) (
 	return nil
 }
 
-func (s *userService) CreateUser(ctx context.Context, input CreateUserInput) (_ map[string]any, err error) {
+func (s *UserService) emitUserCreateFailedBestEffort(ctx context.Context, actions []UserAction, applyErr error) {
```
fforootd: optional: three small things in this helper — (1) the `UniqueError` you already matched carries the violated key, but `UserCreateFailedPayload.KeyName` is left empty even though the catalog advertises it; (2) `EntityID` records an ID minted inside the rolled-back tx, so it references a user row that never existed — arguably better left empty; (3) a full `pool.Transaction` for a single insert could be a plain statements insert. All cosmetic for a best-effort signal, but (1) is the one operators would actually miss.

api/openapi/endpoints/events/event-base.yaml
```diff
@@ -0,0 +1,113 @@
...
+      - auth
+      - session
+      - admin
+      - entity
+      - signal
+    example: entity
+  occurred_at:
+    type: string
+    format: date-time
+    description: When the action happened (DB clock).
```
fforootd: doc nit: “(DB clock)” is only true on Postgres — the Spanner and SQLite dialects stamp `occurred_at`/`created_at` from the app-server clock. Maybe soften to “server clock” so the contract doesn't overpromise monotonicity across nodes.

cmd/server/server.go
```diff
@@ -367,8 +413,10 @@ func loadConfig(configPath string) (Config, error) {
 	// (Console ADR 0004 §3); set NEXTGEN_PLATFORM_PROJECT_ID to pin an
 	// existing project instead. The server never creates a project itself,
 	// unless platform.bootstrap_project explicitly opts in (#605).
-	v.SetDefault("platform.project_id", "")
-	v.SetDefault("platform.bootstrap_project", false)
+	v.SetDefault("events.retention_days", 30)
```
fforootd: nit: this hunk replaced the `platform.project_id` / `platform.bootstrap_project` SetDefaults, so the platform comment above (ADR 0004 §3 / #605) now orphans onto the events defaults. I verified env pinning still resolves via the config-struct path, so it's not functional — but restoring the two lines (or moving the comment) would keep the explicit-defaults convention and the comment honest.

fforootd (commented):
A handful of **non-blocking** polish notes from a deeper read — none of these need to hold the PR, take or leave as you see fit. The Spanner `BufferWrite` insert path and the generator completeness gates are really nicely done. 👏

internal/storage/dialect/postgres/migration/sql/000016_events.sql
```diff
@@ -0,0 +1,33 @@
+-- +goose Up
+CREATE TABLE zitadel_nextgen.events (
+    project_id          TEXT        NOT NULL
+        REFERENCES zitadel_nextgen.projects (id) ON DELETE CASCADE
```
fforootd: Could we rework this project FK/cascade before merging? `projectService.Delete` deletes the project and then inserts `project.deleted` in the same transaction, so this FK rejects the event because its parent is already gone and the whole project deletion rolls back. Emitting first would still let `ON DELETE CASCADE` erase both the deletion event and all earlier audit history. I reproduced the rollback with the SQLite service path (`Delete` returned `failed to delete project`), and the same FK shape is present in all three dialects. A non-cascading scope strategy or a retained project tombstone seems necessary here.

internal/service/team.go
```diff
@@ -175,7 +208,22 @@ func (s *TeamService) Update(ctx context.Context, input UpdateTeamInput) (*domai
 // Delete is idempotent: DeactivateTeam only touches an active team, so an
 // unknown or already-deactivated team reports success without persisting any changes.
 func (s *TeamService) Delete(ctx context.Context, projectID, teamID string) error {
-	if err := s.v2Pool.Statements().DeactivateTeam(ctx, projectID, teamID); err != nil {
+	err := s.v2Pool.Transaction(ctx, func(ctx context.Context, tx Statementer[AllStatements]) error {
+		if err := tx.Statements().DeactivateTeam(ctx, projectID, teamID); err != nil {
+			return err
+		}
+		return audit.Emit(ctx, tx.Statements(), audit.EmitSpec{
```
fforootd, line 219: Could we emit this only when `DeactivateTeam` actually changed an active row? The method intentionally returns success for an unknown or already-deactivated team, so this unconditional emit records a mutation that did not happen. Calling `Delete` twice against SQLite produced two `team.deactivated` events. Perhaps `DeactivateTeam` could report whether it changed a row, and the event could be gated on that result.

internal/audit/shipper.go
```diff
@@ -0,0 +1,230 @@
...
+		if err != nil {
+			return err
+		}
+		defer resp.Body.Close()
+		_, _ = io.Copy(io.Discard, resp.Body)
+		if resp.StatusCode >= 300 {
+			return &httpError{status: resp.StatusCode}
+		}
+		return nil
+	default:
```
fforootd, line 253: Could unsupported sink types return an error (and ideally fail configuration validation) instead of succeeding here? `Start` accepts any non-empty type, so a typo such as `webhkok` reaches this branch; returning `nil` then advances the cursor as though the event was delivered, permanently skipping it. Keeping the cursor in place on an unknown type would make the configuration failure visible and recoverable.

internal/audit/request_buffer.go
```diff
@@ -0,0 +1,216 @@
...
+		events[i] = ev
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	if err := b.insert.InsertEvents(ctx, events); err != nil {
+		slog.Error("failed to flush request audit event batch",
+			slog.Int("batch_size", len(events)),
+			slog.String("error", err.Error()),
+		)
+		return
```
fforootd: Could we add the ADR 048 retry/backoff behavior here, or at least increment a drop metric and alert when retries are exhausted? The batch has already been removed from the buffer before this insert, so one transient storage error currently loses up to the full batch, while `Dropped()` remains unchanged. That makes the loss visible only in a log line.

internal/audit/shipper.go
```diff
@@ -0,0 +1,230 @@
...
+			)
+			return
+		}
+		cursor = &domain.EventSinkCursor{
+			SinkID:    sink.ID,
+			ProjectID: projectID,
+		}
+	}
+
+	batch, err := s.src.ListEventsAfterCursor(ctx, projectID, cursor.LastCreatedAt, cursor.LastEventID, 100)
```
fforootd: Could `shipProject` continue fetching pages until it sees a short page (with a context or work budget)? It currently fetches exactly one 100-event page per tick; with the default five-second interval that caps a sink/project at 20 events/sec even when delivery is instantaneous. A project sustaining more audited requests than that can never catch up, so retained events may eventually be purged before export.

docs/design/api/events-catalog.md
```diff
@@ -0,0 +1,78 @@
...
+| JSON schema create | `schema.created` | `admin` | `json_schema` | `schema_id` |
+| JSON schema delete | `schema.deleted` | `admin` | `json_schema` | `schema_id` |
+| Branding create | `branding.created` | `admin` | `branding` | `layout` |
+| Authz grant | `authz.granted` | `admin` | `authz_assignment` | `principal_type`, `principal_id`, `relation` |
+| Authz revoke | `authz.revoked` | `admin` | `authz_assignment` | `principal_type`, `principal_id`, `relation` |
+| Set password | `auth.factor.password.set` | `auth` | `user_password` | `user_id`, `factor_id` |
+| Delete password | `auth.factor.password.removed` | `auth` | `user_password` | `user_id`, `factor_id` |
+| Create TOTP | `auth.factor.totp.enrolled` | `auth` | `user_totp` | `user_id`, `factor_id` |
+| Delete TOTP | `auth.factor.totp.removed` | `auth` | `user_totp` | `user_id`, `factor_id` |
+| Create passkey | `auth.factor.passkey.enrolled` | `auth` | `user_passkey` | `user_id`, `factor_id` |
```
fforootd: Could we wire this event into the successful passkey-registration transaction, or narrow the catalog until it is supported? `FinishPasskeyRegistration` stores the credential with `CreateUserPasskey` and returns without `audit.Emit`, and the new `EventTypeAuthFactorPasskeyEnrolled` constant currently has no producer. As written, a security-relevant factor enrollment succeeds without the catalogued audit record.

fforootd (commented):
Thanks for taking on such a broad audit-log slice. I did a service-backed pass over the mutation and export paths and found a few cases that seem worth tightening before this lands.

fforootd (commented):
A few follow-up findings from another current-head pass:

internal/audit/retention.go
```diff
@@ -0,0 +1,115 @@
...
+		case <-ticker.C:
+			j.runOnce()
+		}
+	}
+}
+
+func (j *RetentionJob) runOnce() {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	defer cancel()
+	ids, err := j.list.ListProjectIDs(ctx)
```
fforootd: Could the retention sweep enumerate project IDs from `events` (or use an unscoped cutoff delete) rather than `ListProjects`? Events intentionally survive a project hard-delete, but once the project row is gone `ListProjectIDs` can no longer return that ID, so this loop never calls `DeleteEventsOlderThan` for its retained rows—including `project.deleted`. Those rows therefore accumulate indefinitely. The unit test injects IDs and the hard-delete stmttest only checks survival, so neither covers this lifecycle.

docs/design/api/events-catalog.md
```diff
@@ -0,0 +1,78 @@
...
+| `CreateUser` | `user.created` | `entity` | `user` | `user_id`, `schema_id` |
+| User attribute patch | `user.updated` | `entity` | `user` | `changed_keys[]` |
+| Unique violation on create | `user.create.failed` | `entity` | `user` | `key_name` |
+| `DeactivateUser` | `user.deactivated` | `entity` | `user` | `reason` |
+| `DeleteUserByID` | `user.deleted` | `entity` | `user` | _(empty)_ |
+| `CreateTeam` | `team.created` | `admin` | `team` | `name` |
+| `UpdateTeam` | `team.updated` | `admin` | `team` | `changed_keys[]` |
+| `DeactivateTeam` | `team.deactivated` | `admin` | `team` | _(empty)_ |
+| Membership create/status | `team.membership.updated` | `admin` | `team_membership` | `user_id`, `team_id`, `status` |
+| `CreateToken` | `auth.token.issued` | `auth` | `token` | `scopes[]` |
```
fforootd: Could we wire these token events into every token mutation, or narrow the catalog until that is true? `GenerateJWE` and initial `CreateSession` persist tokens without `auth.token.issued`, while `RevokeToken` deletes a token without `auth.token.revoked`; only `SessionService.Exchange` currently emits issuance. A current-head producer census also finds 16 other declared event types with no non-test producer, so a catalog-to-producer contract test would help prevent these silent Path B gaps.

internal/service/passkey_registration.go
```diff
@@ -151,13 +152,25 @@ func (s *PasskeyRegistrationService) Finish(ctx context.Context, in FinishRegist
...
+	err = s.v2Pool.Transaction(ctx, func(ctx context.Context, tx Statementer[AllStatements]) error {
+		if err := tx.Statements().CreateUserPasskey(ctx, newPasskey); err != nil {
+			return fmt.Errorf("passkey registration: store credential: %w", err)
+		}
+		if err := audit.Emit(ctx, tx.Statements(), audit.EmitSpec{
+			Type:       domain.EventTypeAuthFactorPasskeyEnrolled,
+			Category:   domain.EventCategoryAuth,
+			ProjectID:  in.ProjectID,
+			EntityType: "user_passkey",
+			EntityID:   reg.UserID,
```
fforootd: Could this event identify the passkey rather than the user? With `entity_type: "user_passkey"`, using `reg.UserID` as `entity_id` and omitting `factor_id` prevents correlating enrollment with the specific credential or its later removal. `CreateUserPasskey` mints `newPasskey.ID`, which should be used for both fields. SQLite currently mints into a local `id` without copying it back to `p.ID`, so that dialect needs the same backfill before this can be correct there. The service mock accepts any `InsertEvent`, which is why the identity/payload mismatch stays green.

internal/service/project.go
```diff
@@ -346,7 +405,25 @@ func (s *projectService) Delete(ctx context.Context, id string) error {
 	if id == "" {
 		return domain.ErrProjectMissingID()
 	}
-	if err := s.v2Pool.Statements().DeleteProjectByID(ctx, id); err != nil {
+	err := s.v2Pool.Transaction(ctx, func(ctx context.Context, tx Statementer[AllStatements]) error {
+		// Emit before hard-delete: events.project_id is not an FK (audit must
+		// outlive the project row). Order still matters if a future migration
+		// reintroduces a non-cascade FK.
+		if err := audit.Emit(ctx, tx.Statements(), audit.EmitSpec{
```
fforootd: Could we gate this emission on an actual project deletion, like the new `team.deactivated` handling? `ProjectService.Delete` documents a missing project as a no-op, and every dialect ignores the delete's affected-row count. Because the event is inserted first, deleting an unknown or already-deleted ID commits a `project.deleted` record even though no mutation occurred. Returning a `changed` flag from `DeleteProjectByID` (or otherwise proving existence inside this transaction) would keep the audit trail truthful.

internal/audit/request_buffer.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (b *RequestBuffer) flush(batch []bufferedEvent) {
+	if len(batch) == 0 {
+		return
+	}
+	events := make([]*domain.Event, len(batch))
+	for i, item := range batch {
+		ev := item.event
+		ev.OccurredAtWait = time.Since(item.enqueuedAt)
```
fforootd: Could `OccurredAtWait` be recomputed immediately before each `InsertEvents` attempt? ADR 048 requires the wait at the successful insert, including after retries. Computing it once before this loop means a slow failed attempt plus backoff is omitted; the later successful insert then derives an `occurred_at` shifted forward by that retry delay. The retry test currently exercises only permanent failure, so a fail-then-succeed timestamp assertion would catch this.

PR #807, opened 2026-08-10
feat(authz): wire permission resolver into in-project HTTP management gate

adlerhurst:
## Summary

Follow-up to #801 / #423: wire the library permission resolver into in-project management HTTP handlers.

- Replace token-scope `requireProjectAccess` with `resolver.Check` (coarse `project.{viewer,editor,admin}` until #420)
- Map **D10**: `NotFound` → 404 resource miss shapes, `Forbidden` → 403 `permission_denied`
- Seed `sk_proj` ↔ `project.viewer` on `CreateProject` so the returned project secret can pass the HTTP gate
- Keep preview/`project.read` on the browser plane (operator ceiling still requires `project.write`)
- Enrich `ScopeContext` with `PrincipalType` / `PrincipalID` for the gate
- Docs: reconcile ADR 033 403/404 wording in `authz.md` / `url-architecture.md` / `permission-storage.md`

Out of scope: fine-grained catalog (#420), SQL list-predicate injection, RSI dual-write for schema/branding/flow_definition/session, #333, Leopard, runtime/cookie/anonymous planes.

## Validation

```sh
go test ./internal/api/ ./internal/domain/ ./internal/service/ ./internal/authz/resolver/ -count=1
go test -tags sqlite_integration ./internal/storage/stmttest/ -run 'AuthzResolver|ResourceScope' -count=1
```

## Release notes / changeset

Changeset included: `.changeset/authz-http-middleware.md` (`@zitadel/server` minor).

## Notes

- Base is #801 (`cursor/authz-resolver-b0de`); retarget `main` after #801 merges.
- Seeded relation is `project.viewer` (not admin) because the current catalog closure treats assigned `viewer` as the relation that satisfies viewer/editor/admin checks.
- Operator session get/list/revoke intentionally share the MVP `project.write` + Check gate (replacing the old fail-closed `session.*` scope row until #420).
- Gate cleanup after merge of #801: honest `hasOperatorProjectWrite` ceiling, shared miss helper, single Check for `listUserTeams`, loud missing-deps failure.
- **No assignment backfill (D14):** nextgen is (pre)alpha; no deployment predates this stack. `sk_proj` ↔ `project.viewer` assignments are seeded only by `CreateProject` — no migration backfill for pre-gate projects.

[...]

PR #809, opened 2026-08-10
feat(authz): dual-write schema/branding/flow/session into RSI

adlerhurst:
## Summary

Dual-write `resource_scope_index` rows for schema, branding, flow_definition, and session on create/delete (same-tx as entity mutations). Fixes delete ordering so entity delete runs before RSI cleanup.

- Schema/branding/flow/session RSI upsert on create; explicit delete on entity delete (branding relies on project FK cascade)
- Entity delete runs first — wrong `projectID` no longer strips foreign RSI rows
- **Composite RSI PK** `(resource_kind, project_id, resource_id)` — fixes cross-project schema `$id` URL collision (tenant isolation)
- `GetResourceScopeInProject` for project-scoped path.id lookup

## Validation

```sh
go test -tags sqlite_integration ./internal/storage/stmttest/ -run 'AuthzDualWrite|ResourceScope' -count=1
```

## Release notes / changeset

Changeset included: `.changeset/authz-rsi-dual-write.md` (`@zitadel/server` minor).

## Notes

- Base is #807; retarget after the stack below merges.
- User multi-write helpers remain in `internal/storage/dialect/authz` (edges + RSI).
- **No RSI backfill for pre-existing rows (D14):** nextgen is (pre)alpha; RSI for schema/branding/flow/session exists only via dual-write from this PR onward.
- Schema `$id` collision fixed in this PR ([#835](https://github.com/zitadel/nextgen/issues/835)); Wave-1 migrations amended in place (no follow-up goose file).


livio-a:
Review notes. The tx-bound dual-write pattern is correctly applied (savepoint/join semantics check out per dialect, wrong-project deletes can't strip foreign rows, all three dialects tested). Two substantive issues:

1. **Cross-project RSI collision for URL-keyed schemas (high — tenant isolation).** `resource_scope_index`'s PK is `resource_id` alone, and the upsert is `ON CONFLICT (resource_id) DO UPDATE SET … project_id = EXCLUDED.project_id`. Schemas dual-write RSI keyed by the **user-supplied `$id` URL**, which is only unique per `(project_id, url)` — and cross-project reuse of a public schema URL is *designed* behavior (`JSONSchemaResolver` ingests remote schemas by URL per project). Scenario: project A ingests `$id=https://example.com/x.json`; project B does the same → the single RSI row's `project_id` flips to B, so scope resolution for A's schema now answers "project B" (the #810 gate 404s A's legitimate by-id request); if B then deletes, `DeleteResourceScope(url)` removes the row entirely, stranding A's still-existing schema with no scope. #811 then turns this into silent list omission — and its diff currently works around the symptom by removing `$id` from a test fixture rather than fixing the cause here. Fix options: key RSI per project (or `(resource_kind, resource_id)` with per-kind uniqueness), or restrict schema RSI to managed `sch_*` ids. The new dual-write test only exercises a unique URL, so this case is untested.

2. **No backfill for the four new kinds (high impact once the gate enforces).** `000015_authz_mvp.sql` backfilled RSI for project/team/user in the same change that introduced their dual-write; nothing equivalent ships here for `json_schemas`, `brandings`, `flow_definitions`, `sessions`. Every pre-existing row of these kinds has no RSI entry, so once #810/#811 enforce, by-id requests 404 and lists omit them — for fully authorized admins. A backfill migration (postgres + spanner) should land with or before the gate PR. The doc edit claiming RSI coverage for these kinds should carry the caveat until then.

3. Low: `InsertSession`'s RSI upsert isn't self-wrapped in a transaction — safe today because all callers (`CreateSession`, `ExchangeSession`/`RunExchange`) are tx-bound, but it's exported on `ExchangeStore`, so a future direct-pool caller would get three separate autocommits with partial-failure orphans. A `withTransaction` wrap (idempotent in all dialects) or an invariant comment would harden it.

livio-a:
Backfill for the four new RSI kinds (review item 2) is resolved by a recorded no-backfill decision — see https://github.com/zitadel/nextgen/pull/807#issuecomment-5264281432; please mirror it in this PR's `Notes`, and the D14 row covers it.

Item 1 (schema `$id` collision) is now tracked in #835; my recommendation stands that it's a tenant-isolation defect worth fixing in this PR rather than deferring, since #811 currently works around it in a test fixture.

livio-a (approved):
Review findings resolved: composite RSI PK (resource_kind, project_id, resource_id) fixes the cross-project schema $id collision (f9d378d5, #835 closed), backfill covered by D14. InsertSession tx wrap remains a low-severity follow-up for the consolidated issue. Approving.

PR #801, opened 2026-08-10
feat(authz): permission resolver Check and list oracle (#423)

adlerhurst:
## Summary

Implements issue #423 as a **library-only** permission resolver on top of the merged Wave 1 storage + authz testing stack:

- Shared resolver SQL in `internal/storage/dialect/authz` via `ArgWriter` + `Env`, emitted through each dialect’s `statementCompiler` (`WriteString` / `WriteArg`) — no `SQLStyle` / `fmt.Sprintf` placeholder templating
- `AuthzResolverStatements`: active system catalog, foothold helper, `CheckAuthz` (allowed + foothold in one round-trip), `ListAuthzObjectIDs` (L4/oracle helper; ADR 033 predicate injection stays for HTTP list wiring)
- `internal/authz/resolver`: `Decision` (`Allow` / `Forbidden` / `NotFound`), `sk_team_` allowlist, request-scoped memoization, `Check` / `ListObjects`
- Deterministic stmttest matrix covering closure, TTU membership + general arms, team/resource list scopes, foothold columns, home-project defaults; plus generative L4 oracle agreement

Out of scope (per plan): HTTP/middleware wiring, injectable list predicates at endpoints, #333 foreign team home project, Leopard, cross-request decision cache.

## Validation

```sh
go test ./internal/authz/resolver/ ./internal/storage/dialect/authz/ -count=1
go test ./internal/storage/dialect/postgres/ ./internal/storage/dialect/sqlite/ ./internal/storage/dialect/spanner/ -count=1
go test -tags sqlite_integration ./internal/storage/stmttest/ -run 'AuthzResolver' -count=1
```

Postgres/Spanner coverage via existing `forEachDialect` / CI integration tags.

## Release notes / changeset

Changeset included: `.changeset/authz-resolver-statements.md` (`@zitadel/server` minor).

## Notes

- Active catalog is resolver-owned (callers cannot pass `catalog_id`).
- MVP sets `$principal_home_project_id` to the protected project (#333 deferred).
- Check and List share full TTU (membership shortcut + scoped assignment implying source).
- HTTP 403/404 mapping remains a follow-up when endpoints wire (D10).
- `sk_team_` team-scope compensating requirement (resolver-enforced team bound + outside-team deny suite) is explicitly deferred and tracked in #831; this PR only ships the flat permission-name allowlist.

[...]

PR #810, opened 2026-08-10
feat(authz): resolve path-id scope from RSI before Check

adlerhurst:
## Summary

Path-id management handlers resolve scope from `resource_scope_index` before `resolver.Check` (flat-by-id; no required query `project_id`).

Review follow-ups addressed in this PR:
- RSI-miss delete 204 gated on `project.write` (preview → readMiss 404)
- Post-RSI `DecisionNotFound` always `readMiss`
- Expected `ResourceKind` check on path-id routes
- `requireTeamDelete` passes `ErrInternal` through
- MVP scoped-Allow ack → https://github.com/zitadel/nextgen/issues/833

## Validation

```sh
go test ./internal/api/ -count=1
```

## Release notes / changeset

Changeset included for server authz gate behavior.

## Notes

Base is #809; stack order #801 → #807 → #809 → #810 → #811.


livio-a:
Review notes. The mechanics are fail-closed and consistently wired (OpenAPI/client/CLI/mock all match, TOCTOU degrades gracefully, unauthenticated requests never reach the RSI lookup). The concerns are all about the decision→error-shape mapping:

1. **DELETE is now a cross-tenant existence oracle, and answers before authorization (high — needs an explicit decision).** On RSI miss with `op == opDelete`, `requireResourceAccess` returns the `errResourceGone` sentinel → 204 *before* `requireProjectAccess` consults the ceiling or runs Check. So any authenticated principal of any project — including a read-only preview secret — gets **204** for a nonexistent id, **404 `user.not_found`** for a real user in a foreign project, and **403 `team.permission_denied`** for a real foreign team; the updated integration tests encode exactly these pairs. That distinguishes "exists somewhere you can't see" from "doesn't exist" — the distinction the file's own anti-oracle comments say the gate must not leak, and which the pre-PR gate didn't. High-entropy ids make it a confirmation oracle (leaked logs, support tickets) rather than enumeration, but it inverts the documented contract, so it should be a deliberate choice, e.g. require a foothold before answering 204 (RSI miss + no foothold → the resource 404 shape, like read). Also: the "preview secret cannot delete" assertions were dropped — the code still denies via `res.denied()`, but coverage for deleting a *real* resource with a preview secret should come back.

2. **Second oracle on write ops via mismatched miss shapes (medium).** RSI miss on `opWrite` returns `res.readMiss()` (`user.not_found`), but RSI *hit* + `DecisionNotFound` returns `res.miss(opWrite)` → `writeMiss` (`user.invalid` "project does not exist", `team.project_not_found`, …). Distinguishable codes on PATCH/password routes = existence oracle, plus a factually false message. After an RSI hit on a by-id op, `DecisionNotFound` should force the read-miss shape (the `writeMiss` shapes were designed for create/list where `project_id` is explicit).

3. **`scope.ResourceKind` is returned but never checked against the route (low/medium).** `GET /users/{id}` with a schema id passes the gate whenever the id's project is accessible; wrong-kind deletes yield misleading 204 no-ops. No cross-project leak (the project Check still binds), but a one-line kind check would close it.

4. Expectation-setting (low): despite the branch framing, team-/resource-scoped grants still cannot authorize by-id requests — the gate always checks `ObjectType: "project"` and `resolver.Request` has no team/object field. Fail-closed and consistent with the MVP note, just making sure nobody reads scoped Allow into this layer.

(The `requireTeamDelete` internal-error→403 masking flagged on #807 gets wider here since the RSI lookup can also fail — same fix applies.)

livio-a:
`5981d528` looks good — the write-shape unification, kind check, `ErrInternal` passthrough, and restored preview-delete coverage all check out. One residual on the delete path:

Gating the 204 on `hasOperatorProjectWrite` removes the oracle for preview/anonymous callers, but a **foreign full project secret** still distinguishes: fabricated id → 204 (RSI miss + operator), real foreign resource → 404/403 (RSI hit + Check miss). Since anyone who can create a project holds an operator secret, this narrows the confirmation oracle rather than closing it — the commit message ("close authz by-id delete oracle") slightly overstates.

If keeping idempotent 204s for operators is the intended tradeoff, fine — but can we state it in the `errResourceGone` comment (and ideally the D10 docs) as a deliberate decision? The alternative (RSI miss on delete → readMiss unless the principal has a foothold somewhere) closes it fully at the cost of repeated-delete 204s.

livio-a:
`b823daff` + `0378c89b` check out — project-scoped resolution, the documented delete tradeoff, and the foreign-real-delete fix are all what the review asked for. One nit introduced along the way:

Both the wrong-kind fallback in `lookupResourceScope` and the delete branch in `requireResourceAccess` call the **global** `GetResourceScope(resourceID)` — whose doc comment (updated in `f9d378d5`) now restricts it to "when resource_id is globally unique (tests, oracle)". Post-composite-PK, schema `$id` URLs are legitimately non-unique across projects. When the same URL exists in **2+ foreign** projects, `CollectExactlyOneRow` returns a too-many-rows error that both call sites swallow as a miss — so a foreign schema delete answers **204** when the URL is shared by two projects but **404** when exactly one holds it. Not a leak (204 is the more masking answer), but it's nondeterministic against the tradeoff you just documented, and it works by accident of error-swallowing.

Suggest an explicit query for the purpose (e.g. `ExistsResourceScopeElsewhere(kind, resourceID, excludeProjectID)` or a `LIMIT 1` variant with defined semantics), and narrowing `GetResourceScope`'s doc to test/oracle use only.

livio-a (approved):
Review findings resolved: write-shape unification, ResourceKind route check, ErrInternal passthrough, restored preview-delete coverage (5981d528), project-scoped RSI lookup + documented delete-204 tradeoff (b823daff), deterministic ExistsResourceScopeElsewhere delete path (74e003b8). Scoped by-id Allow tracked in #833. Approving.

PR #811, opened 2026-08-10
feat(authz): inject EXISTS list predicates into management lists

adlerhurst:
## Summary

Inject authz EXISTS list predicates into management list endpoints after a successful **project-level** Check.

- `withAuthzListFilter` attaches `AuthzListFilter` to context; dialect compilers append correlated EXISTS (same branches as `ListObjects`)
- Wired: users, teams, schemas, branding, flow definitions
- Team-scoped HTTP list narrowing is **not** live yet (#834); predicate runs after project-level gate pass

## Validation

```sh
go test ./internal/api/ ./internal/storage/dialect/authz/ -count=1
go test -tags sqlite_integration ./internal/storage/stmttest/ -run AuthzList -count=1
```

## Release notes / changeset

Changeset included: `.changeset/authz-list-predicates.md` (`@zitadel/server` minor).

## Notes

- Base stack: #801 → #807 → #809 → #810 → #811.
- **No RSI backfill (D14):** pre-existing schema/branding/flow/session rows without dual-write RSI are out of scope for (pre)alpha.
- Schema `$id` RSI collision fixed in #809 (composite PK); integration tests still omit URL `$id` in fixtures because ogen leaf path params reject embedded `/`.


livio-a:
Review notes. The SQL side is clean: every runtime value goes through `WriteArg` (verified all 15 call sites — raw `WriteString` inputs are compile-time constants), the predicate lands before ORDER BY/LIMIT in all three dialects, cursor terms fold in ahead of it, and catalog-lookup failure is fail-closed. The issues are at the API layer:

1. **The team/resource-scoped predicate arms are dead code end-to-end (high).** Every wired list handler runs `requireProjectAccess` first, whose `Check` computes allowed = project-scoped closure OR full TTU. A principal whose only grant is `project.viewer` scoped to a team — the exact scenario in the new stmttest "team-scoped grant hides out-of-team rows" test — gets `DecisionForbidden` → 403 *before* `withAuthzListFilter` attaches. So for any principal that can pass the gate, the predicate degenerates to "an RSI row exists for this resource": the scoped narrowing the changeset advertises ("team-scoped grants filter rows in one SQL round-trip") never engages over HTTP. The stmttest tests pass only because they call storage directly, and there's no HTTP-level integration test for the filter — which is why this wasn't caught. Options: give list ops a `Forbidden → proceed-with-predicate` partial-view path in the gate, or scope the changeset/doc claims down and land the gate change as a follow-up. Either way an HTTP-level test should pin the intended behavior.

2. **Inherited from #809 (commented there): the schema `$id` RSI collision** now causes silent list omission — a colliding URL removes a tenant's schema from its *own* `ListSchemas`. The `schema_test.go` change that deletes `$id` from the fixture (with a comment acknowledging the parallel-create interference) is working around the defect in tests; the keying fix belongs in #809. Same for the missing RSI backfill: pre-#809 branding/schema/flow rows will list as empty for authorized admins.

3. **Performance (medium).** No project-wide fast path: since `requireProjectAccess` already ran Check, the filter could simply not attach when the decision was Allow via the row-independent arms — that skips the per-row correlated EXISTS for the common case (today: every principal that can pass the gate) and also softens the backfill symptom. Separately, `withAuthzListFilter` re-queries `ActiveSystemCatalogID` that the gate just fetched — one avoidable round-trip per list request.

4. **Fail-open-by-omission design (low).** Authz rides on a context value; a future list endpoint that forgets `withAuthzListFilter` silently returns unfiltered rows, and `compileRead` deliberately passes `context.Background()`. Today the coarse gate in front of every management list contains it, and the unwired lists (sessions, projects, passkeys, user-teams) each have a defensible reason — but nothing structural fails when one is missed. Worth an assertion or making the statement layer require the filter for management-kind lists.

Minor: `withAuthzListFilter` sets `PrincipalHomeProjectID` from `scope.ProjectID` while `resolver.Check` uses the target project id — equivalent today, diverges if #333 foreign footholds land; worth a comment.

livio-a (approved):
Review findings resolved: changeset/docs scoped to what ships, team-scoped 403 pinned by an HTTP integration test (57088660), partial-view list gate tracked in #834, inherited RSI issues fixed in #809/#810. Fast path + fail-open guard remain for the consolidated follow-ups issue. Approving.

PR #848, opened 2026-08-12
feat: filter and sort teams by name and status

### LINKED ISSUE #819: Teams: `name` and `status` filtering and sorting on `POST /teams/query`
## Context

The Teams directory (#797) has a search box and `Active` / `Inactive` tabs, but `POST /teams/query` only accepts `created_at` for filtering and sorting, so neither element can be built. The dev answers on that issue ([comment](https://github.com/zitadel/nextgen/issues/797#issuecomment-5251562323)) agreed to add both filters: search should cover all teams rather than the loaded page, matching partially and case-insensitively, and the search behaviour should be defined once so every resource can reuse it rather than solving it per screen.

## Scope

- Add `name` and `status` to the filter fields of `POST /teams/query`.
- `name` supports partial, case-insensitive matching through the existing `contains` operation. This defines `contains` on text fields as a case-insensitive substring match platform-wide, documented in the API contract.
- `status` supports `equals` with the contract's two values, `active` and `deactivated`.
- `name` and `status` also become sort fields. Name ordering follows the database's native collation.

Out of scope: a primary-domain field on `team-response` (answered in the same thread: needs its own model, not being added).

## Acceptance Criteria

- Filtering by `name` with `contains` matches substrings case-insensitively across all teams of the project, not just the fetched page.
- Filtering by `status` returns exactly the teams the API renders with that status.
- Sorting by `name` or `status` pages correctly through the cursor, with a stable order for ties.
- Invalid fields, operations, or values are rejected with a 400 and a message naming the offending input.
- The contract documents `contains` as a case-insensitive substring match to support partial, case-insensitive matching.

grvijayan:
Closes #819

## Summary

- `POST /teams/query` accepts `name` and `status` as filter and sort fields alongside `created_at`.
- `name` filters with `equals` and `contains`. The shared `stringFilter` now compiles `contains` to `LOWER(col) LIKE LOWER(?)` on all three dialects, so `contains` on text fields is case-insensitive platform-wide.
- `status` filters with `equals` on the contract's two values, `active` and `deactivated`. Unknown values and non-equality operations return 400; `not_equals` stays 501 like the other fields.
- Per the #622 decision (teams get no pending-purge lifecycle), reads now treat a `pending_purge` row as already gone: `List` ANDs an `Or(active, deactivated)` guard onto every query and `Get` returns not-found.
- Sorting by `name` follows the database's native collation and keeps the id tiebreaker for stable cursor pagination.

## Validation

- `go test ./... -count=1`
- `go test -tags postgres_integration -timeout 10m ./... -count=1`
- `go test -tags sqlite_integration -timeout 5m ./... -count=1`
- `go test -tags spanner_integration -timeout 9m -p 1 ./... -count=1`

## Changeset

- `@zitadel/server` minor: `name` and `status` become filter and sort fields on `POST /teams/query`; `contains` on text fields matches case-insensitively platform-wide.


api/openapi/endpoints/teams/query/methods.yaml
```diff
@@ -3,7 +3,9 @@ post:
   summary: Query teams
   description: |
     Returns the teams of a project, paginated with a cursor.
-    Teams of every lifecycle status are returned; each carries its `status`.
+    Teams of every lifecycle status are returned by default; each carries its
+    `status`. A `name` filter matches exactly with `equals` and
+    case-insensitively with `contains`.
```
IAM-marco: This documents case-insensitive `contains` on the teams endpoint, but the change is to the shared `stringFilter` at [list.go#L158-L165](https://github.com/zitadel/nextgen/blob/0654f5301de2857e3f2bceb1ec5d8f06645e493c/internal/service/list.go#L158-L165), so `POST /sessions/query`'s `user_id` filter changed with it.

[#819](https://github.com/zitadel/nextgen/issues/819) asks for that statement in the contract, not per endpoint: "This defines `contains` on text fields as a case-insensitive substring match platform-wide, documented in the API contract", repeated in the acceptance criteria. The shared home is [filter-operation.yaml](https://github.com/zitadel/nextgen/blob/0654f5301de2857e3f2bceb1ec5d8f06645e493c/api/openapi/components/schemas/filter-operation.yaml#L7-L9), which this PR touches only to strip trailing whitespace.

What follows is a sessions client seeing `contains` change behaviour with nothing in the contract saying so, and nothing in [the changeset](https://github.com/zitadel/nextgen/blob/0654f5301de2857e3f2bceb1ec5d8f06645e493c/.changeset/team-query-name-status.md) either. The PR description's changeset section says the note covers it; the file does not mention it.

Fix: describe the semantics on the `contains` entry in `filter-operation.yaml`, and add the platform-wide line to the changeset. Copilot raised the same gap on this file.

internal/service/list.go
```diff
@@ -145,13 +145,23 @@ func compareFilter[F ~uint8](op string, col database.Column[F], value any) (data
...
+		return "", domain.ErrRequestInvalid().WithDetails(fmt.Sprintf("%s filter value must be a string", f.Field))
+	}
+	return value, nil
+}
+
 // stringFilter maps an operation to a text filter.
 func stringFilter[F ~uint8](op string, col database.Column[F], value string) (database.Filter[F], error) {
 	switch op {
 	case filterOpEquals:
 		return database.StringEqual(col, value), nil
```
IAM-marco, line 162: `contains` on `name` now folds case, but `equals` on the same field stays byte-exact.

Team names are unique per project case-insensitively on every dialect: postgres indexes `(project_id, lower(name))` at [000002_teams.sql#L16-L18](https://github.com/zitadel/nextgen/blob/0654f5301de2857e3f2bceb1ec5d8f06645e493c/internal/storage/dialect/postgres/migration/sql/000002_teams.sql#L16-L18), and spanner and sqlite store a generated `name_lower`. So `Acme` and `acme` cannot both exist, yet `equals` treats them as different teams while `contains` treats them as the same one.

What follows: filtering `name equals "acme"` against a team stored as `Acme` returns an empty list, even though the schema guarantees that team is the only candidate.

[`database.StringEqualFold`](https://github.com/zitadel/nextgen/blob/0654f5301de2857e3f2bceb1ec5d8f06645e493c/internal/storage/database/filter_string.go#L47-L51) already exists. `stringFilter` is shared with sessions' `user_id`, so folding here would change that too; the choice likely belongs per field rather than in the shared helper.

Coverage: [team_test.go](https://github.com/zitadel/nextgen/blob/0654f5301de2857e3f2bceb1ec5d8f06645e493c/internal/service/team_test.go#L432-L446) pins the `StringEqual` shape, and the integration test does not exercise `equals` on `name`, so a casing mismatch would go unnoticed.

IAM-marco (changes requested):

internal/service/list.go
grvijayan: done in https://github.com/zitadel/nextgen/pull/848/commits/e488a44a1e4e5e51956fc604bc9dbe4987806a44

api/openapi/endpoints/teams/query/methods.yaml
grvijayan: done in https://github.com/zitadel/nextgen/pull/848/commits/e488a44a1e4e5e51956fc604bc9dbe4987806a44

IAM-marco (approved):

PR #855, opened 2026-08-12
fix: omit Secure on loopback HTTP cookies (Safari compatibility)

adlerhurst:
## Summary

Safari rejects `Secure` cookies on `http://localhost` (Chrome/Firefox accept them). That dropped `_zflow` after `POST /flow`, so the first register/login submit failed with `flow.cookie_invalid` and never reached the password step.

`_zflow` and `__nextgen_session` now set `Secure` only when appropriate for the request: HTTPS keeps it; **loopback HTTP only** (`localhost` / `127.0.0.0/8` / `::1`) omits it for Safari; non-loopback HTTP keeps `Secure` (fail closed on mis-set `X-Forwarded-Proto`).

## Validation

- `go test ./internal/api/ -count=1`
- Unit coverage for scheme/host-aware Secure on flow/session cookies and `WithRequestHostMiddleware`
- Review follow-up: narrowed Secure exception to loopback; cookie tests parse via `http.ParseSetCookie`

## Release notes / changeset

`.changeset/safari-http-cookie-secure.md` — patch on `@zitadel/server` (fixed group).

## Notes

- Still open apps at exactly `http://localhost:<port>` (not `127.0.0.1`); that remains an origin/WebAuthn constraint, documented in the quick start.
- To verify in Safari: rebuild/run a server with this change, then register at `http://localhost:3000/login` — DevTools should show `_zflow` stored without `Secure`, and email submit should advance to the password step.

<div><a href="https://cursor.com/agents/bc-b7f7babf-a629-4f84-b036-df2dfe56161c?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b7f7babf-a629-4f84-b036-df2dfe56161c&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


livio-a (approved):
Fix looks good — the loopback-only exception is the right scope, and I like that ambiguity fails closed: non-loopback HTTP, a mis-set `X-Forwarded-Proto`, or a missing/unparseable origin all keep `Secure`.

I also checked this against the header-injection class from GHSA-93m4-mfpg-c3xf: not applicable here. The forwarded headers only influence the `Set-Cookie` on the requester's own response, so forging them downgrades nothing for anyone else, and the https→http downgrade path requires a loopback host.

Note the behavior isn't Safari-gated — `Secure` is omitted for all clients on loopback HTTP, which is the right call (no UA sniffing, and the flag wasn't doing real work on loopback anyway). Maybe reword the merge title to "fix: omit Secure on loopback HTTP cookies (Safari compatibility)" so release notes don't imply browser detection.

PR #864, opened 2026-08-12
docs: propose anonymous session-probe contract (ADR 051)

fforootd:
## Summary

The written proposal the #830 review round asked to split out before implementation: `GET /sessions/me` answers cookieless visitors with `401 auth.unauthorized`, which is wrong-shaped twice — a probe whose purpose is distinguishing signed-in from signed-out treats its honest "nobody" answer as an error, and every cookieless page view logs a red 4xx for an app embedding session-aware chrome.

The platform has three visitor states. A valid session without a resolved user — either the pure shell `POST /sessions` mints or a factor-verified password-only exchange — already returns `200` with the full session body. The current server omits `user_id`; the OpenAPI property remains optional and nullable, so clients treat omission and explicit `null` as no bound user. `getSession()` already reads that response as signed out. Only the cookieless state is broken. ADR 051 weighs three contracts:

- **A (recommended):** a `session_state`-discriminated 200 on the same endpoint — `none` (no cookie; new, replaces the 401), `anonymous` (any resolved session without a bound user; body and omitted `user_id` preserved), and `authenticated` (requires non-null `user_id`). The mapping depends only on the resolved user binding, never on factors or lifecycle state. Invalid credentials stay `401`; `404 sess.not_found` is untouched. A minimal A′ adds only `none` and keeps inspecting optional `user_id`, but regeneration happens either way, so the explicit discriminator is recommended.
- **B:** a status envelope — the same blast radius as A plus a wrapper migration, so it is strictly dominated.
- **C:** a separate optional-auth probe endpoint — the smallest diff, but a permanent second endpoint describing the same resource to avoid a one-time alpha migration.

The implementation sequence covers spec → server → regenerated clients → **api-mock parity with its own conformance specs** → a representative browser fixture that actually calls `getSession()` through the framework proxy and asserts the cookieless `200 none` response → service-backed contract tests on PostgreSQL, Spanner, and SQLite. SQLite coverage explicitly includes the shared harness build tag and a `TestMain` backed by `dbtest.SQLite`; adding the tag only to `session_me_test.go` does not compile. The matrix spans absent, shell, password-only, authenticated, garbage, tampered, expired, revoked, and pruned-session cookies.

## Validation

- `moon run workspace:check-adrs` — passes (`adr index: ok (51 records)`)
- `node scripts/check-pr-title.mjs --title 'docs: propose anonymous session-probe contract (ADR 051)'` — passes
- `git diff --check` — passes

Docs only — no runtime change.

## Release notes / changeset

None — repository documentation only; nothing ships to a user.

## Notes

Decision requested before implementation; implementation would follow in a dedicated PR. Revised three times after review: round one restored the valid anonymous-session state, round two made the mapping exhaustive and added mock/browser/all-dialect coverage, and round three preserved the actual omitted-`user_id` wire shape while making the browser and SQLite verification mechanisms runnable rather than nominal.


fforootd:
Review round 1 (relayed) addressed in 8ef0c5b2 — both findings verified against the spec and accepted:

1. **P1 — collapsed states**: confirmed. `POST /sessions` mints valid anonymous session shells (pre-auth correlation, 10-minute TTL), their cookie authorizes `/sessions/me` with a full body and `user_id: null`, and `getSession()` already reads that 200 correctly — so my "aspirational contract" framing was wrong and the two-state union would have mislabeled or discarded the shell's `session_id`/`state`/`factors`/`expires_at`. The ADR now carries three states: `none` (absent cookie — the new variant, replacing the 401), `anonymous` (today's full body, unchanged apart from the discriminator), `authenticated`. The shell's body and TTL semantics are explicitly listed under "what must NOT change", and a minimal sub-option A′ (add only the body-less `none` variant, keep null-sniffing) is recorded with the discriminator recommended over it.
2. **P2 — test matrix**: extended to absent / valid anonymous shell / authenticated / garbage / tampered / expired / revoked / pruned-session (`404 sess.not_found`, untouched), service-backed on both dialects.

Changesets note agreed — docs-only, no changeset. `check-adrs-index` still ok (51 records).

fforootd:
Round-two review addressed in 158735ea — all four findings verified in-tree and accepted.

**1 (P1, non-exhaustive discriminator) — accepted.** Confirmed against `sessionStateFilter` in `internal/service/session.go` (its comment states the case outright: "An unexpired session with factors but no user (password-only exchange) is active, not building") and the stmttest fixture that builds exactly that session (`ExchangeSession` with a nil user → `Factors` non-empty, `UserID` nil). The ADR now defines the mapping as exhaustive and keyed on `user_id` alone: `anonymous` is *any* resolved session with `user_id: null` — the `building` shell and the `active` password-only session alike — never a function of factors or lifecycle state. The problem statement names both shapes with the code references, and the test matrix gains the password-only-exchange cookie → `200 anonymous` with factors intact.

**2 (P1, mock/browser path omitted) — accepted.** Confirmed the standalone mock's `GET /sessions/me` 401s without a cookie and returns an undiscriminated body with one, and that the framework demo e2e lanes are CI-excluded — so the Go contract tests alone could go green while the exact chrome path motivating the ADR stays broken. The sequence gains two steps: api-mock parity with direct conformance specs in the mock's own suite (named as load-bearing, not a courtesy — it's the always-on gate for the browser path), and a representative cookieless demo journey, honestly scoped as on-demand until those lanes join CI.

**3 (P2, dialects) — accepted.** SQLite is the zero-config default with its own CI lane, and nothing under `internal/api/integration_test/` carries the `sqlite_integration` tag today (SQLite coverage lives at the stmttest/bootstrap layers). The ADR now requires all three dialects explicitly and names the concrete mechanism: the session-me suite gains the `sqlite_integration` tag as part of the implementation.

**4 (P2, stale PR body) — accepted**, rewritten to match the revised ADR: three states with the `user_id`-keyed mapping, the shipped-behavior framing (no more "aspirational"), the mock/journey/three-dialect sequence, the full test matrix, and a revision log in the notes.

`check-adrs-index` still passes (51 records).

PR #865, opened 2026-08-12
docs: align the design docs with the shipped contracts

fforootd:
## Summary

Part 1 of the five-slice docs refresh (slice 5 is #856; merge order 1→5). This slice makes the hand-written design docs stop contradicting the shipped implementation — every corrected claim was verified against code or the OpenAPI source before editing:

- **`zitadel claim` is shipped** — five docs (README, VISION, platform design notes) still said it didn't exist; the guarded "MVP vs target design" framing is kept, only the false sentences moved to the shipped reality (ADR 046). VISION's shipped-list also now cites ADR 035 instead of the part-superseded ADR 007.
- **`docs/design/api/conventions.md` rewritten against the shipped contract**: real error envelope (`{code, message, details?}`), real list shape (`{<resource>: [], next_page_token}`), limit default 20, structured POST-query filtering (ADR 031), shipped PUTs, and the shipped `Idempotency-Key` surface (the three one-time auth ops). Everything unshipped (capabilities endpoint, `Zitadel-Version`, rate-limit headers, expand, uniform create idempotency) moved to an explicit "Direction (not shipped)" section, and the docs that referenced the nonexistent `/capabilities` endpoint as fact (hierarchy, glossary, resource-map) now mark it as direction.
- **Session docs match #757/#768/#770**: `POST /sessions/query` instead of the removed `GET /sessions` list, the `/sessions/me` vs operator-plane split, idempotent delete, no `revoked` state, `factors` as an array, one AAL urn scheme.
- **Flow docs match the spec**: `/flow` (not `/flows`), removed operations dropped (`/flow/{id}/event`, activate/deactivate/archive), `/flow_definitions` with the shipped verb set, and the ADR 021 ordered-arrays shape swept through every step example (capabilities, templates, flow-engine-nodes — including its Liquid snippets, now using the shipped `where: "name"` idiom).
- **False security claim fixed**: `flow-engine-storage.md` said the `_zflow` cookie is path-scoped to `/flows` and "never sent to other endpoints" — the code deliberately sets `Path=/` (rationale in `internal/api/flow.go`); the doc now states the real model (encryption, not path scoping).
- **CLI design docs rewritten** (PLAN.md had ✅ marks on never-built idp/app/locale commands): PLAN.md is now an honest current-state plan; bdui-renderer.md and identity-surface.md keep their concept framing with shipped-deviation notes, real paths, and no `/v1/` fiction. `@zitadel/ui-lit` stays documented as the deliberately reserved direction.
- **Dead references cleaned**: the ten links to the `flowengine/api/*.yaml` sketches deleted in May (branding docs called one "source of truth" — repointed at the shipped spec), the retired `river-8421` dictionary slugs (→ `proj_<ulid>`, ADR 047), and assorted moved-file refs.
- **credentials.md** no longer states secret-rotation-at-claim as fact (ADR 046 lists it as an explicit non-goal).

## Validation

- Scoped gates all clean on the merged tree: `rg "storage/v2" docs/ --glob '!docs/adrs/**'` → 0; `rg "river-8421" docs/` → 0; deleted-sketch references → 0; `/flow-definitions` and `/flows/` as API-path claims in design docs → 0 (console's hyphenated UI route untouched); bare `GET /sessions` list claims → 0.
- Relative-link check over all 40 changed markdown files → 0 broken.
- Ground truth spot-checks recorded per finding (e.g. `flow_state_machine.go:461` for the cross-flow rejection, `flow.go:268` for the cookie path, `flow-step.yaml` for the array shape).

## Release notes / changeset

- No changeset required — repo-only design docs; no shipped behavior changed.

## Notes

- One audit finding was **reversed by ground truth** during execution: cross-flow `switch`/`pivot` transitions are *not* runtime-supported (the validator accepts them in definitions; the runtime rejects with `flow.unsupported`). `capabilities.md` had this right all along — the customer-shipped guide was fixed in #856, and this slice corrects the error-code spelling in capabilities.md.
- resource-map.md got false-entry corrections only. Follow-up proposal for a maintainer decision: shrink its endpoint inventory to a pointer at `api/openapi/openapi-spec.yaml` — four hand-kept endpoint listings are the root cause of this drift class.
- ADR-file edits (statuses, storage/v2 link repoints inside `docs/adrs/`) are deliberately **not** in this slice — they're slice 2, which also extends `check-adrs-index.mjs` to gate status drift.


fforootd:
All six findings addressed (c1af4a2c + the follow-up commit):

1. **Flow handles**: the `/flow/{id}` handle is now consistently the flow's own `id` from the latest response — worked examples gained `"id": "flow_N"` alongside the stable `session_id`, all `/flow/sess_N/...` paths became `/flow/flow_N/...`, and the remaining plural `POST /flows` calls (mermaid diagrams in authn-and-auth-flows, flow-engine-storage, capability-handshake, the guide) are singular. The `session_id`-as-handle sentence in authn-and-auth-flows.md now states the id/session split per `flow-response.yaml`.
2. **Ordered arrays, completed**: every remaining `fields` map in flow-engine.md and flow-engine-guide.md is now a `FlowField[]` array with `name`, every action entry carries its required `kind`, and the guide's prose/tables/Liquid-context block teach arrays — dict-style lookups (`fields.email.…`) became the shipped `where: "name"` idiom (no `*_by_name` maps exist in the context; the default template uses `where` exclusively).
3. **Session factors**: examples now match `completed-factor.yaml` — `method` + `verified_at` + method-specific `payload` (`identifier` carries `user_id`), and the exchange example no longer exposes `amr` (projected by the OIDC adapter only, as the doc itself says).
4. **List envelope**: the conventions row now scopes the `{ <resource>: [], next_page_token }` shape to cursor-paginated lists and names both shipped exceptions as normalization candidates: `GET /branding` (bare array) and `GET /schemas` (offset/limit).
5. **Default-flow example**: the examples README no longer claims the root copy "does not gate anything" — it states that `POST /flow_definitions` still publishes it as the `defaultLoginFlow` `externalValue`, that it differs from the embedded default and sits outside `TestExampleFlowDefinitions`, and flags repointing the spec (or testing the copy) as the follow-up.
6. **Anchors + versioning claims**: `#idempotency` → `#idempotency-shipped` in authn-and-auth-flows and credentials; glossary and the API README now state header versioning is target design (`#direction-not-shipped`), not shipped.

Link check over all touched files: 0 broken.

fforootd:
All four findings addressed in 39aa07f3, plus adjacent instances of the same classes found while verifying:

1. **P1 — legacy definition shape**: all five flow-engine.md definitions and the guide's examples now use the shipped shape — slug-shaped `name` (no `slug`/display name), required `status`, `purposes` as a purpose→entry-step object (no `initial_steps`), `user_schema` as a server-assigned `sch_` id, and a declared `actions` entry for every `transitions` key the validator checks. **Every full definition example in both docs now validates against `api/openapi/endpoints/schemas/flow-definition.json`** (checked with ajv 8 / draft 2020-12) — which caught a bug beyond the review: step names must match `^[a-z][a-z0-9-]*$`, so `set_password` → `set-password` throughout. Same pass fixed gate `type`→`kind` (also in flow-engine-nodes.md, which additionally dropped the nonexistent `required`/`satisfied` keys), added the required SSO `template`, removed the author-forbidden `transitions.sso`, dropped the nonexistent `audience.schema_ids`, and corrected the prose bullets (`on_success` not `action`; gates as a keyed object; resolution over the `purposes` map, app > team > project-wide).
2. **P1 — submit bodies**: all submit examples send inputs under `fields` with the required `action` (guide pseudocode included); SSO selection uses the reserved `action: "sso"` + `sso_provider_id`. Also in this class: `POST /flow` examples now carry the required `project_id`, and the hint row is `user_schema_id` per `flow-hint.yaml`.
3. **P2 — `kind` on fields**: stripped from all seven runtime field examples (`Field` has no `kind`).
4. **P2 — cross-flow presented as working**: both the flow-engine.md pivot section and the guide's cross-flow section now carry an explicit direction guard — schema/validator accept `switch`/`pivot`, but the runtime rejects every non-null transition `action` with `{ "code": "flow.unsupported" }` (`internal/domain/flow_state_machine.go`; stubbed per capabilities.md) — and point at the shipped alternative, the default flow's local re-purposing transitions (`{ "target": "register", "purpose": "register" }`). Example 2's switch exchange and the guide's closing switch diagram are annotated direction-only; Example 4 (SSO) is direction-marked with a schema-valid definition. The guide's Definition Lifecycle section was rewritten to the shipped model (`active`/`draft`, PUT full replacement; validate/simulate marked planned — `archived`/`activate`/`PATCH` never shipped).

PR #876, opened 2026-08-13
docs(authz): define cross-project access and bootstrap

### LINKED ISSUE #333: [ARCH-ADR] Define Architecture for Cross-Project Identity and Collaboration

fforootd:
## Summary

- propose ADR 052 for cross-project human and team principals, target-scoped authorization, first-party Console session credentials, CSRF, authorized project discovery, and audit behavior
- propose ADR 053 for separate membership and ownership, direct-user and team-derived project grants, additive roles, offboarding, and customer-team ownership
- revise Console ADRs 0002–0004 so embedded management uses the HttpOnly first-party session cookie and bootstrap provisions a reserved platform project through the full provisioner
- keep standalone optimized for one customer project without imposing a one-project limit or removing the current first-created-project fallback before a human-usable seed transport ships
- incorporate review feedback: contiguous ADR numbering, legal direct-user FGA tuples, write-time owner/membership invariants, explicit alpha database reset, CSRF rollout impact, target-catalog wording, and repaired section citations

## Validation

- `node scripts/check-adrs-index.mjs` — `adr index: ok (53 records)`
- `git diff --check`
- focused `corepack pnpm exec oxfmt --check` across changed ADR/docs and formatter-clean comment files
- relative Markdown links checked across all changed Markdown files
- `node scripts/check-pr-title.mjs --title "docs(authz): define cross-project access and bootstrap"`
- `corepack pnpm exec changeset status --since origin/main` — no package bump planned

## Release notes / changeset

## Notes

- Documentation and architecture decisions only; TypeScript and Go edits update comments/citations and do not change runtime behavior.
- ADR 052 calls out the atomic implementation impact of adding CSRF enforcement to the already-implemented ADR 046 flow.
- ADR 053 requires dropping and recreating alpha databases that applied the old authorization seed; editing an applied Goose migration would otherwise leave the permissive closure active.
- Issue metadata and the issue owned by Marco were not modified.


apps/console/src/runtime/runtime.ts
```diff
@@ -1,13 +1,13 @@
...
+ * Console runtime discovery (Console ADR 0004 §3).
  *
  * The embedded console is one build artifact serving every deployment, so
  * deployment facts are discovered at boot instead of baked in: the server
  * publishes a small public runtime-metadata document and the console fetches
  * it once before the router renders (`src/main.tsx`). Everything in the
  * document is public runtime metadata in the root ADR 005 sense — an enum
  * and project ids, never secrets or feature inventories (per-surface gating
- * rides effective permissions, ADR 0004 §4).
+ * rides effective permissions, ADR 0004 §5).
```
IAM-marco, line 10: This pointer correctly moves to §5, but the renumbering pass stops at `apps/console/src/**` and `internal/service/project.go`. Twelve citations in six files the PR does not touch still use the old numbering, and they were correct before this change.

The clearest case is [console_runtime.go#L25-L29](https://github.com/zitadel/nextgen/blob/b0c2a7fa1c9ac85f4e58ccb43f9c564aaab3a474/cmd/server/console_runtime.go#L25-L29), the server side of this same document, which carries this exact sentence and still says §4.

Old §2 becomes §3, old §3 becomes §2, old §4 becomes §5, old §5 becomes §3:

| Location | Says | Should say |
| --- | --- | --- |
| `cmd/server/console_runtime.go:16` | §2 | §3 |
| `cmd/server/console_runtime.go:28` | §4 | §5 |
| `cmd/server/console_runtime.go:33` | §3 | §2 |
| `cmd/server/console_runtime_test.go:52` | §3 | §2 |
| `cmd/server/console_runtime_test.go:127` | §3 | §2 |
| `cmd/server/server.go:219` | §3 | §2 |
| `cmd/server/server.go:368` | §3 | §2 |
| `cmd/server/server.go:461` | §2 | §3 |
| `apps/console/vite.config.mts:112` | §2 | §3 |
| `apps/console-e2e/src/console-shell.spec.ts:13` | §3 | §2 |
| `apps/login-ui/src/main.ts:10` | §2 | §3 |
| `apps/login-ui/src/main.ts:115` | §3 | §2 |

One of these has drifted in substance too, not just numbering: [server.go#L218-L222](https://github.com/zitadel/nextgen/blob/b0c2a7fa1c9ac85f4e58ccb43f9c564aaab3a474/cmd/server/server.go#L218-L222) says the deployment "tracks exactly one project", which the revised §1 contradicts ("The data model and APIs do not enforce a one-project ceiling").

Nothing validates section anchors, so these stay wrong until someone reads them.

apps/console/docs/adrs/0004-console-deployment-modes.md
```diff
@@ -1,330 +1,261 @@
...
-   configuration; license-key mechanics should be evaluated for self-hosted
-   portals.
+One Console build must serve two product postures:
+
+- **Standalone** is the default self-hosted posture. It should be useful with
+  one customer project and one operator team, without forbidding an operator
+  from creating more projects later.
+- **Platform** hosts many customer projects. Platform-project users can receive
+  direct or team-derived access to selected customer projects. A platform
+  role alone never means access to every project.
```
IAM-marco, line 32: Are you impliying that a platform-project user can be granted access to another project directly without a team as middleman? I always understood that we don't grant permissions to users for objects outside their project, but it's always the team the user belongs to having that grant.

apps/console/docs/adrs/0004-console-deployment-modes.md
```diff
@@ -1,330 +1,261 @@
...
+both standalone and platform modes. The Console always signs into that
+project. Customer projects remain ordinary protected resources selected after
+sign-in.
+
+The reserved project is a **control project**, not a super-project:
+
+- a role in it does not imply access to any customer project;
+- access to a customer project comes only from that project's owning-team,
+  direct-user, team, or confidential-project assignment under ADRs 052 and
+  053; and
```
IAM-marco: To clarify, as pointed out above

apps/console/docs/adrs/0004-console-deployment-modes.md
```diff
@@ -1,330 +1,261 @@
...
+- a future CLI command or Console wizard may collect the values and invoke the
+  same server contract rather than inventing another provisioning path.
+
+All transports must converge on the same persisted resources and invariants.
+The minimum useful seed provisions, in one operation:
+
+1. the reserved platform project through the normal full project provisioner,
+   including its keys, default user schema, and login flow;
+2. at least one initial platform user who can sign into the Console;
+3. a default team plus that user's `team_memberships` row;
```
IAM-marco, line 112: @livio-a is `team_memberships` still relevant after the introduction of the permission system?

docs/adrs/053-cross-project-principals.md
```diff
@@ -0,0 +1,399 @@
...
+
+The server-authoritative principal context is:
+
+```text
+principal_type
+principal_id
+principal_home_project_id
+target_project_id
+credential_class
+```
```
IAM-marco, line 85: Not sure if this is how you expect it to look like once implemented, but I'd double check with Livio and Silvan

docs/adrs/053-cross-project-principals.md
```diff
@@ -0,0 +1,399 @@
...
+### 5. First-party human sessions may call the operator plane
+
+The embedded Console authenticates a human with the existing HttpOnly
+`__nextgen_session` cookie. The browser sends that first-party cookie directly
+to the same-origin API. For management endpoints that accept a human principal,
+the session proves who the human is and the permission resolver decides which
+target projects they may act on. The browser sends neither a project secret nor
+a script-readable session bearer.
+
+This explicitly amends ADR 036's statement that confidential-plane operations
```
IAM-marco, line 194: If this ADR is changing the behaviour of a previous one, I would just amend it rather than mentioning it here.

apps/console/docs/adrs/0002-console-api-access.md
```diff
@@ -232,13 +243,13 @@ while guaranteeing the auth-later work is an extension, not a rewrite.
...
   open, not only a note.
-- **Management calls still need more than the cookie.** The dev proxy's
-  secret is what carries `user.read` and friends today; the publishable key
-  is deliberately refused for them (`internal/api/user.go`). Root ADRs
-  032/033's session-derived permissions are what make the embedded console's
-  list screens work without a proxy — tracked there, not here.
+- **Management calls are fail-closed until the cookie is authorized.** The
+  dev proxy's secret is what carries `user.read` and friends today; the
+  publishable key is deliberately refused for them (`internal/api/user.go`).
+  Root ADRs 032/033/054's session-derived target permissions make the embedded
```
livio-a: **Typo (blocking):** “Root ADRs 032/033/054” — there is no ADR 054; the series ends at 053. The parallel bullet in ADR 0003 (line 153) and AGENTS.md both say 052, which is clearly what’s meant. Worth fixing before merge since a future unrelated ADR 054 would silently make this sentence mean something wrong.

apps/console/docs/adrs/0002-console-api-access.md
```diff
@@ -186,28 +196,28 @@ in whether the browser holds a secret — it never does. The base resolves in
...
+- A script-readable per-user bearer is not part of the Console contract. If a
+  future non-browser client needs one, it uses the SDK's existing token
+  mechanism rather than changing the first-party browser path.
 - The `401`/`403` handling from §3 becomes the "redirect to console login"
   trigger.
 
-The point of recording this now is that #440 can ship with zero console auth
-while guaranteeing the auth-later work is an extension, not a rewrite.
+Identity is already carried by the cookie. Root ADR 052 supplies the missing
+target-project authorization contract, while ADR 046 supplies the CSRF
```
livio-a: **Wrong attribution (blocking):** “ADR 046 supplies the CSRF requirements” — 046 §2 as checked in states the opposite (“SameSite cookie semantics are sufficient against CSRF and no separate CSRF token is introduced”). The requirements live only in 052 §5, which supersedes 046 *if accepted*. Suggest citing 052 alone here (and likewise for the “specified by ADRs 046 and 052” phrasing in 0003 §4 and 0004 §4) — or add the forward-pointer amendment note to 046’s header in this PR so the citation becomes true. As written, an implementer following the Implemented 046 ships SameSite-only protection — exactly the design 052 rejects.

docs/adrs/053-cross-project-principals.md
```diff
@@ -0,0 +1,399 @@
...
+
+### 8. Audit events are written in the protected project
+
+Every allowed cross-project mutation and every security-relevant read produces
+the same protected-project event as a local actor. It uses ADR 048's existing
+top-level columns:
+
+```text
+project_id = proj_customer  # protected target, not the actor's home project
+team_id    = ...            # protected-resource scope only, when applicable
```
livio-a, line 289: **Undeclared amendment of ADR 048:** §8 says it uses “ADR 048’s existing top-level columns” but annotates `team_id` as “protected-resource scope only”. 048 defines `team_id` as team-scoped *credential* context captured from the resolved credential `ScopeContext` (DDL comment, emit-time rule). For a foreign actor authorized via a platform team, 048’s rule would store the actor’s home-project team — which this section forbids. That’s a real semantics change for the emitter; please add it to “Amends if accepted” and the documentation-impact list rather than only “extend ADR 048’s actor metadata”.

apps/console/docs/adrs/0004-console-deployment-modes.md
```diff
@@ -1,330 +1,261 @@
...
+- it is never used as the hosted-login project for a customer's application.
+
+Standalone remains the default product posture, not a different authorization
+model. Its bootstrap may create one initial customer project and one owner
+team, and the initial Console can optimize for that common path. The data model
+and APIs do not enforce a one-project ceiling.
+
+### 2. Bootstrap is explicit desired state
+
+The server may start with no projects. In that state the embedded Console is
```
livio-a, line 86: **Conflict with a LOCKED design doc:** “The server may start with no projects” plus the seed-gated platform project contradict `docs/design/api/hierarchy.md`’s **LOCKED** “Self-hosted returns a singleton platform project and a singleton default team… The SDK does not branch on deployment mode.” No documentation-impact list schedules a hierarchy.md correction for this (053 touches it only for owner-vs-contributor wording). Either the LOCKED invariant needs amending, or bootstrap must guarantee the singleton — please adjudicate explicitly.

apps/console/docs/adrs/0004-console-deployment-modes.md
```diff
@@ -1,330 +1,261 @@
...
+participant.
+
+The existing `platform.bootstrap_project` row-only behavior is insufficient
+and must be replaced or routed through this provisioner. The same is true for
+`--user-file` behavior that inserts a bare project row only to satisfy foreign
+keys. Because the product is alpha, the checked-in seed behavior and test
+fixtures may be corrected directly; no compatibility migration or backfill is
+required for pre-release development databases.
+
+**Cutover rule:** the implemented first-created-project fallback remains
```
livio-a: **Dropped semantics still cited by code:** the cutover rule mentions only “the implemented first-created-project fallback”, but `internal/service/project.go` and `login.tsx` (updated in this PR) attribute the *configured-pin* branch to this rule too: `platform.project_id` must exist, missing = configuration error, deterministic `created_at` ordering. Those semantics no longer appear anywhere in the Decision. One restoring sentence would keep the cutover executable from the ADR alone — is the pin retained or removed at cutover?

apps/console/docs/adrs/0004-console-deployment-modes.md
```diff
@@ -1,330 +1,261 @@
...
+  "publishable_key": "pk_proj_..."
+}
+```
 
-The console never evaluates the first question per surface. The server folds
-it into the second: **effective permissions = the user's granted permissions
-∩ what the deployment offers.** One vocabulary, one source of truth; the
-console renders from the effective set alone.
+`console_project_id` and `publishable_key` identify the reserved platform
+project. They are omitted while bootstrap has not completed. The response is
```
livio-a, line 182: **Target state stated as current:** under §2’s own cutover rule, today’s server sends a first-created *customer* project id with no bootstrap at all — so “identify the reserved platform project” and “omitted while bootstrap has not completed” are false for every current deployment. Since §3 is where all the runtime-discovery code comments now point, suggest a one-line transitional caveat (“until the §2 cutover completes, the id is the first-created/pinned fallback project”).

apps/console/AGENTS.md
```diff
@@ -17,27 +17,29 @@ direction for the console build-out (issue
...
-  one default project (the first project created by `zitadel setup` through
-  `POST /projects`; the server never creates it). A minimal per-request
-  runtime document carries `mode` + the sign-in project id, and portal surfaces
-  (billing,
-  multi-project, support) render from **effective permissions** (user
-  grants ∩ deployment profile, computed server-side) via
-  `staticData.permission` — never build-time flags, never a parallel
-  console-facing feature array.
+  — one build and one authorization model serve cloud and self-host. Every
+  deployment uses a reserved platform project for Console identities; an
```
livio-a: **Same issue as 0004 §3:** “Every deployment uses a reserved platform project for Console identities” is present-tense target state; what ships is the first-created fallback. AGENTS.md is the layer agents treat as authoritative, so this bullet needs the transitional hint — e.g. “target: reserved platform project; today: first-created/pinned fallback per ADR 0004 §2”.

apps/console/docs/adrs/0004-console-deployment-modes.md
```diff
@@ -1,330 +1,261 @@
...
+file—ships with the full provisioner and an end-to-end first-operator test. The
+testkit can prove the target shape before that point, but test infrastructure
+alone must not strand a self-hoster. Only after that replacement exists may the
+server stop resolving the first-created project and require the reserved
+platform-project seed.
+
+Bootstrap inputs are server-side configuration. Secrets from them never enter
+`runtime.json`, the Console bundle, or browser-readable storage.
+
+### 3. Runtime discovery carries only the Console sign-in target
```
livio-a, line 166: **Two deleted decisions lost their home:** the rewrite dropped (a) the runtime-fetch fail-safe (“falls back to `standalone` so a broken endpoint degrades to the smallest surface, never to an accidental portal” — `main.tsx` now cites §3 for it, which no longer says it) and (b) runtime.json’s deliberate placement outside the OpenAPI surface (`cmd/server/console_runtime.go` and `docs/design/api/conventions.md:51` still defer here for it). Both are still implemented. One sentence each in §3 would keep them defensible on paper.

docs/adrs/053-cross-project-principals.md
```diff
@@ -0,0 +1,399 @@
...
+  assignments, relation closure, membership edges, and resource scope.
+- [`permission-storage.md` D13](../design/api/permission-storage.md#locked-decisions)
+  stores a foreign user or team assignment on the **protected** project.
+- `authz_membership_edges` remains in the principal's **home** project and is
+  not copied into every protected project.
+- The resolver already distinguishes the target project from
+  `principal_home_project_id`, but the HTTP management gate still assumes they
+  are equal.
+
+This ADR ratifies the cross-project mechanism and its security boundary. It does
```
livio-a, line 51: **Cite issue #333:** ADR 033 (§Out of scope) and `permission-storage.md` (“Left open for #333”) both defer cross-project human identity, staff tiers, and break-glass to #333 — and 052/053 are essentially that architecture. A line here — “this resolves #333’s cross-project identity question; staff tiers, consent, and break-glass remain open (§9)” — would close the dangling pointers those docs left.

docs/adrs/053-cross-project-principals.md
```diff
@@ -0,0 +1,399 @@
...
+not a cross-project directory read.
+
+### 9. Deployment-wide operators are a separate authority
+
+No role, membership, or administrative permission on the platform project
+implicitly grants access to every project. The platform project is an identity
+and customer-collaboration home, not a wildcard authorization object.
+
+Cloud operations and self-hosted break-glass may use an explicit
+deployment-wide operator relation or credential. Its issuance, activation,
```
livio-a, line 362: **Acknowledge the platform secret’s transitive reach:** the reserved platform project’s own `sk_proj` (which 053 §3 makes `project.admin` there, and Console ADR 0004 §2 mandates minting) can mutate any platform-team roster; team-derived grants expand through those membership edges (§2), so the secret transitively reaches every customer project — while 053’s test obligation 11 (“secret confined to its own project”) still passes. Possession is the guardrail, and for self-host that’s fine (the holder is already root). But in cloud this makes the platform secret an *implicit* version of exactly the deployment-wide operator credential this section defers to future governance, with audit only in the platform project. Suggest: (a) a sentence here acknowledging it and stating possession-as-guardrail; (b) noting its deployment-wide blast radius vs ordinary project secrets (handling/rotation tier); (c) reconsidering whether the full provisioner should mint a general-purpose platform secret at all — post-bootstrap the Console uses the session cookie and login uses the publishable key.

docs/adrs/053-cross-project-principals.md
```diff
@@ -0,0 +1,399 @@
...
+- does not appear in the protected project's user directory;
+- holds no local password, factor, session, or lifecycle owner there;
+- is not counted as a protected-project end-user merely because they manage it;
+- retains one stable actor identity for audit across every authorized project.
+
+### 2. Grants remain on the protected project
+
+Cross-project authority uses ordinary `authz_assignments` rows. The row carries
+the protected `project_id` and a foreign `user` or `team` principal. There is no
+`principal_type = project`, no wildcard project object, and no composite foreign
```
livio-a, line 103: **State that grant principals are platform-homed:** the foreign-principal mechanism here is home-project-generic — nothing forbids naming a *customer-project* user or team as principal on another customer project. Today the platform-only constraint is only implicit via §5’s credential rules (platform-homed Console sessions + confidential automation), so an app-homed human grant would be unexercisable authority sitting in the table until someone widens the credential rules. Suggest stating it explicitly: cross-project grant principals must be homed in the reserved platform project (or: app-homed rows are inert by construction and rejected at write time). Cross-app *automation* belongs to the future PAT/service-user or token-exchange leg, not human grants. (Transitional wrinkle: under Console ADR 0004’s fallback, the Console sign-in project is itself a customer project.)

livio-a (dismissed):
Strong, carefully-argued pair of ADRs — the cross-project model itself holds up well (I verified the claims about the current seed closure, D13, and the resolver shape against the code). Requesting changes only for the citation-level defects; everything else is discussion input for the author.

**Blocking (mechanical):** the two ADR 0002 citation bugs (nonexistent ADR 054; CSRF credited to 046) and the undeclared ADR 048 `team_id` amendment — see inline.

**Discussion points:** hierarchy.md LOCKED-singleton conflict, the dropped cutover/pin semantics, target-state-as-present-tense wording (§3 + AGENTS.md), the deleted fail-safe/OpenAPI-placement decisions, the #333 citation, the platform secret’s transitive reach, and making platform-homed principals explicit — all inline.

**Follow-up (files outside this diff, fine as an immediate follow-up PR):**
- The §-renumbering strands ~11 stale citations: `cmd/server/console_runtime.go:16,28,33`, `cmd/server/server.go:219,368,461`, `cmd/server/console_runtime_test.go:52,127`, `apps/login-ui/src/main.ts:10,115`, `apps/console/vite.config.mts:112`, and the console-e2e specs. Worst is `server.go:219`, which now cites “§3” as authority for “the deployment tracks exactly one project” — a claim this rewrite explicitly abolishes.
- `apps/console/README.md` (~86–90) still teaches “tracks exactly one project, and the server never creates it” — now contradicting the rewritten ADR 0004 and updated AGENTS.md.

Nit: 053 §4 states the “validates the principal exists and is active without exposing a global directory” rule twice two paragraphs apart; test obligation 12 already pins the behavior.

docs/adrs/053-customer-collaboration-grants.md
```diff
@@ -0,0 +1,505 @@
...
+compatibility migration or assignment backfill is provided.
+
+That choice requires an explicit destructive reset: every database that has
+already applied the old Goose migration must be dropped and recreated before
+running the updated binary. Editing an already-applied migration source does
+not update its catalog, and reusing such a database would leave the permissive
+`viewer -> editor -> admin` closure active. CI lanes must use fresh databases,
+and the implementation handoff must give developers and alpha operators the
+exact reset procedure. Catalog rows remain immutable within a running
+database—never patch an active catalog partially.
```
IAM-marco: I don't think this is needed. For the moment, we start everytime from scratch anyway and CI uses a fresh DB every time

IAM-marco (changes requested):

fforootd:
> **Discussion points:** hierarchy.md LOCKED-singleton conflict, the dropped cutover/pin semantics, target-state-as-present-tense wording (§3 + AGENTS.md), the deleted fail-safe/OpenAPI-placement decisions, the #333 citation, the platform secret’s transitive reach, and making platform-homed principals explicit — all inline.

Let me try and explain my thoughts around the hierarchy topic.

Standalone needs the platform project too as I see it today. It keeps admin users separate from the users inside the actual projects. We'd originally scoped standalone as single-project and that doesn't hold. It can have multiple projects; the UX just won't push it, since most self-hosters run one, maybe two.

"Singleton" here means the platform project and the default team, not the deployment's project count. hierarchy.md already reads that way, line 30 ("When self-hosted grows to multiple projects …") and line 34 ("one user + one team inside the singleton platform project. One customer project …").

What needs changing is line 28's unconditional present tense: self-hosted returns a singleton platform project, while §2 says the server may start with no projects. That's bootstrap sequencing:

> LOCKED. A bootstrapped self-hosted deployment returns a singleton platform project and a singleton default team with the identical JSON schema the cloud version returns. The SDK does not branch on deployment mode — it blindly works against both. Before bootstrap completes there is no platform project to return; that is a deployment-lifecycle state, not a second API shape (Console ADR 0004 §2).

"Singleton" describes the platform project and the default team, not the deployment's project count. Customer projects are unbounded in both modes.

The second paragraph also closes off the "one project total" reading, `cmd/server/server.go:219` and apps/console/README.md had both drifted into "tracks exactly one project" (fixed in this PR).

Do you see this the same @livio-a

fforootd:
Opened #903 — a failing test for ADR 053 §3's `owner from owning_team`.

It doesn't resolve. `project.owning_team` sits on the protected project, which the resolver finds, but `team.owner` sits at the team's home scope, and `writeFullTTUExists` constrains the source assignment to the protected project. The `team#member` path is special-cased to look up membership edges in the home project; `owner` is not. Membership crosses projects, ownership doesn't.

Two controls in that PR isolate the hop: the same rule with both rows in one project passes, and a cross-project team grant (052 obligation 2) passes. Fails identically on SQLite and Postgres.

The PR is expected to be red — `owning_team` isn't in the catalog yet, so nothing is broken today. But 053 obligations 1 and 5 require the behavior, and 052 obligation 6 needs the same edge for authorized-project listing, so I'd rather settle it before we accept 053 than after people design against it.

The suggested fix is in the PR. It depends on cross-project grant principals being platform-homed — which is what @livio-a asked for above, so the two want to land together.

Copilot found this, but it landed in a collapsed suppressed-comments block, which is why none of us saw it.

apps/console/docs/adrs/0004-console-deployment-modes.md
fforootd: Yes, that's intended — for **access**, never for ownership. Claim and ownership stay team-mediated (046 §1, restated in 052's header): a direct grant can't create or replace the owning team, complete claim, or make the user the lifecycle owner. It's ordinary revocable access to an already-owned project, same row shape as a team grant, just with a user principal. The reasoning lives in root ADR 053 §4 (Vercel individual project roles, GitLab direct members; agencies and one-off collaborators) — this bullet is only the console-side echo of it.

That said, we don't need them day one. I've updated 053 §4 to stage it: the model and testing obligations commit to both principal types from the start, but the collaboration API can ship team grants first, with direct grants following once §6's per-project grant listing and §9's source-annotated listing are implemented — a direct grant is only operable once an owner can enumerate and revoke it as easily as a membership. Your offboarding instinct is exactly what that gate encodes; suspension denies both paths immediately either way (§8).

apps/console/docs/adrs/0004-console-deployment-modes.md
fforootd: Yes — they're two different facts, and 053 §1 defines the split in a table. `team_memberships` stays the product roster: invitation, provisioning, status, billing-seat context (membership-as-resource is LOCKED by ADR 024 and hierarchy.md's ownership-vs-membership section). `authz_membership_edges` is the permission system's projection of it — an active roster mutation dual-writes the edge (ADR 033; `internal/storage/dialect/authz/dual_write.go`), and the resolver expands team usersets through the edges.

The permission system consumes membership, it doesn't replace it. 053 §1 leans on the roster directly: bootstrap step 3 writes membership as roster state only, and "an owner must be an active participant" is checked against roster status, not against the projection.

docs/adrs/053-cross-project-principals.md
fforootd: Adopted — essentially your (c). The provisioner no longer mints a platform secret at all: bootstrap provisions the publishable key only, and §9 now says so explicitly.

When platform-homed automation is needed, it's a named, explicitly scoped `sk_plat_` key: an ordinary principal whose authority is exactly its assignment rows, no implicit admin closure, minted deliberately via a session-authorized ceremony on the claim transport (Console page or `zitadel key create`). The first intended use is customer-boundary automation — a team owner's key for creating and managing their own team's projects — so the blast radius is bounded by the account that minted it.

That answers your (a) and (b) as a side effect: there's no default secret left to tier, and a created key's reach is visible in its grant rows rather than implied by its existence. The transitive-roster path you traced still exists in principle, but only for a key someone explicitly created with roster scopes someone explicitly granted, and 048 audits the mutations.

§9 also gained the guard for the failure mode you were pointing at: the deployment-wide operator credential must not be *approximated* by an `sk_plat_` with broad roles — it stays a separate, governed credential precisely so issuance, customer visibility, and audit posture can attach to it.

docs/adrs/053-cross-project-principals.md
fforootd: Half of this already exists: the block names `domain.AuthzCheckParams` (`internal/domain/authz.go`) — catalog, target project, principal home/type/id — plus `credential_class` from credential resolution. Livio verified the resolver shape against the code in his review. The genuinely-new part is the HTTP management gate, which today assumes home == target (052 §3 calls that out) — that is the piece worth Silvan's eyes at implementation time.

apps/console/docs/adrs/0004-console-deployment-modes.md
fforootd: Half restored, half reversed — deliberately. (b) The OpenAPI placement is back in §3, grounded in `hierarchy.md`'s planned `/capabilities`. (a) went the other way: an unreachable endpoint is now an error, not a mode (§3) — an error state satisfies the accidental-portal worry equally, and the fallback was reporting outages as "no project yet". The shipped behavior is recorded as a pending change in Consequences; code lands separately.

apps/console/src/runtime/runtime.ts
fforootd: All 12 fixed in-PR in e35c62e8 (rather than the follow-up PR), plus `apps/console/README.md`'s "tracks exactly one project" and the `server.go:219` substance drift.

docs/adrs/053-customer-collaboration-grants.md
fforootd: Trimmed in 4855950b — kept one sentence because CI freshness does not cover persistent local databases or alpha operators.

apps/console/docs/adrs/0002-console-api-access.md
fforootd: Fixed in e35c62e8 — 032/033/052.

apps/console/docs/adrs/0002-console-api-access.md
fforootd: Fixed in e35c62e8 — all three sites cite 052 §5, and 046/036 now carry the amendment in their own headers so the citation is true either way.

docs/adrs/053-cross-project-principals.md
fforootd: Declared in e35c62e8 — added to 052's Amends header, spelled out in §8, corrected in the doc-impact list, and 048 carries the forward-pointer.

apps/console/docs/adrs/0004-console-deployment-modes.md
fforootd: Answered in the conversation (https://github.com/zitadel/nextgen/pull/876#issuecomment-5295903174): "singleton" was always the platform project + default team, not the project count — lines 30/34 already anticipate multi-project self-host. Only line 28's unconditional present tense conflicts; proposed scoped wording is in that comment, hierarchy.md untouched pending your take.

apps/console/docs/adrs/0004-console-deployment-modes.md
fforootd: Restored in a10ea6e0, and the question is decided: the pin retires with the fallback at cutover — post-cutover the reserved project is found by a persisted marker only bootstrap writes, so there is no guess left to override. No adoption path; pre-existing alpha deployments recreate (consistent with 053 §3).

apps/console/docs/adrs/0004-console-deployment-modes.md
fforootd: Fixed in a10ea6e0 — §3 carries the transitional caveat.

apps/console/AGENTS.md
fforootd: Fixed in a10ea6e0 — target/today split plus an explicit do-not-remove-the-fallback warning.

docs/adrs/053-cross-project-principals.md
fforootd: Added in 4855950b.

docs/adrs/053-cross-project-principals.md
fforootd: Added in 4855950b — write-time rejection, the cross-app-automation carve-out, and the transitional wrinkle. It is also the precondition for #903's resolver fix, which leans on principal and team sharing a home project.

docs/adrs/053-cross-project-principals.md
fforootd: Done in e35c62e8 — 036, 046, and 048 now carry the amendment in their own headers, matching their existing convention.

apps/console/docs/adrs/0004-console-deployment-modes.md
fforootd: Answered on the thread above — and 053 §4 now stages the API: team grants can ship first, direct grants follow once the listing surfaces exist (c53e9b04).

livio-a:
> Do you see this the same @livio-a

Yes, agreed. Line 34 already reads that way ("one user + one team inside the singleton platform project. One customer project"), so my comment really only applies to line 28, and your wording fixes that.

I'd make the hierarchy.md change in this PR though — it's one paragraph and this is the PR that introduces the pre-bootstrap state. If you'd rather not touch it here, put it on 052's documentation impact list so it doesn't get lost. Also add a short note that self-host today has no platform project at all until the seed transport ships, so nobody reads the new wording as shipped behavior.

apps/console/docs/adrs/0004-console-deployment-modes.md
livio-a: Yes, still relevant. `team_memberships` is the roster (invites, provisioning, status, billing seats — ADR 024). `authz_membership_edges` is just the projection the permission system reads, dual-written from roster changes (ADR 033). 053 §1 has a table with the split.

docs/adrs/053-cross-project-principals.md
livio-a: I checked this against the code during review — it matches `AuthzCheckParams` plus `credential_class`, nothing new on the resolver side. What's missing is the HTTP gate handling home != target, which is what #903 demonstrated. @adlerhurst can you take a look at the evaluation order in §3?

IAM-marco (approved):

PR #890, opened 2026-08-14
fix(authz): reject non-project tokens before minting sk_proj ScopeContext

adlerhurst:
## Summary

- `HandleOAuth2` minted `PrincipalType: sk_proj` for any introspectable bearer. Session, OIDC, and PAT tokens could claim the project's `sk_proj` identity; containment rested only on the `project.write` scope ceiling.
- Only `TokenTypeProjectToken` and `TokenTypeProjectPreview` now mint an `sk_proj` `ScopeContext` via `TokenType.IsProjectSecret()`. Other types, including `TokenTypeUnspecified` (pre-#760 typeless secrets), fail the security requirement the same way a missing credential does.
- Preview secrets still cannot call management APIs (`hasOperatorProjectWrite`). This change only stops the wrong token class from becoming the project principal.

Closes #841.

## Validation

- `go test ./internal/api/ ./internal/domain/ -count=1 -run 'TestHandleOAuth2|TestHandleNextgenSession|TestTokenType'`
- `SetScopedTokenOnApiClient` now reuses `project.Token()` (same as `SetPreviewSecretOnApiClient`) so `TestQuerySessions/session.read_is_denied` still reaches the authz ceiling (403) instead of failing OAuth2 (401).

## Release notes / changeset

Changeset: `.changeset/authz-oauth2-token-type.md` — Management APIs no longer treat session, OIDC, or PAT bearers as the project secret. Only project and preview tokens mint an `sk_proj` principal. Pre-#760 project secrets with no Type field also stop authenticating. (`@zitadel/server`)

## Notes

- First PR in the authz follow-up workstream (#831–#843). Does not add `sk_team` minting or change the resolver.
- Review follow-ups parked: OpenAPI `session_token` Bearer wording (wrong blast radius); type-aware introspection to skip `GetTokenByID` before the gate.
- Next: #831 (`sk_team` team scope). This PR stays a sibling of #891; it is not merged into that stack.

[...]

PR #891, opened 2026-08-14
fix(authz): constrain sk_team checks to the token team

### LINKED ISSUE #831: authz: enforce sk_team resolver team scope (compensating requirement)
## Context

Locked text in [`docs/design/api/credentials.md`](docs/design/api/credentials.md) requires two things for `sk_team_` credentials:

1. A flat permission-name allowlist (e.g. `user.read` MAY, most other permissions MUST NOT).
2. A **compensating** resolver rule: every such grant must be limited to that token’s team, and the deny suite must include “team token reads/writes a user outside its team.”

Related: #423 (permission resolver library + L4 oracle), PR #801.

## Gap today

#423 / PR #801 only enforces (1) — the flat MAY allowlist in `internal/authz/resolver/sk_team.go`.

If an `sk_team` principal somehow holds a **project-scoped** `user.read` assignment, `Check` can still **Allow** for the whole project. The allowlist alone does not add team scope.

`sk_team` is **not** production-safe for `user.*` / `team_membership.*` until this issue is closed.

## Required work

- Pass the token’s `team_id` into resolver `Check` / `List` (and list params as needed).
- Constrain successful checks to that team: team-scoped grants and/or RSI `team_id` match for the token’s team.
- Grant-minting validation that `sk_team` grants are team-scoped (not project-wide for team-bound permissions).
- Deny-suite fixture: team token reads/writes a user **outside** its team → deny / NotFound as designed.

## Acceptance

- [ ] Resolver enforces team scope for `sk_team_` on `user.*` / `team_membership.*` (and any other team-bound allowlisted permissions).
- [ ] Outside-team deny suite exists and passes.
- [ ] Docs in `credentials.md` / `permission-storage.md` no longer mark this as deferred.

## Out of scope notes

Do not half-fix this inside the #801 library PR without `Request.TeamID` + SQL/grant-minting. Track and implement as a dedicated follow-up (this issue).

adlerhurst:
## Summary

- `sk_team_` had a permission-name allowlist but no team-id constraint. A project-scoped grant could Allow `user.read` for the whole project.
- `resolver.Request.TeamID` is required for `sk_team_` and is copied to `AuthzCheckParams.ConstraintTeamID`. Check/List/EXISTS then require the object to belong to that team: users via `authz_membership_edges`, teams by id, other kinds via RSI `team_id` / Check `ResourceTeamID`.
- `CreateAuthzAssignment` rejects `sk_team` + project scope for team-bound object types (`user`, `team`, `team_membership`, `event`) using the shared `domain.TeamBoundObjectType` set.
- sk_team Checks of team-bound types require `ResourceID` so a forgotten by-id id cannot skip the compensating constraint (List still omits it).
- No HTTP `sk_team_` token type is added. Tests inject `Request` / statement params directly.

Closes #831.

## Validation

- `go test ./internal/domain/ ./internal/authz/resolver/ ./internal/storage/dialect/authz/ -count=1`
- `go test ./internal/storage/stmttest/ -count=1 -tags sqlite_integration -run 'TestAuthzAssignment_SKTeam|TestAuthzSKTeam_'`

## Release notes / changeset

Changeset: `.changeset/authz-sk-team-scope.md` — Team service principals (`sk_team_`) can no longer act on users or teams outside the token's team, even if a project-wide grant exists. Project-scoped team-bound grants are rejected at mint time. HTTP wiring lands in #893–#895. (`@zitadel/server`)

## Notes

- Second PR in the authz follow-up workstream. Independent of #841 (sibling of #890, not stacked under it).
- MVP catalog still uses `project.viewer`; the deny suite uses a project-scoped `project.viewer` grant as the "somehow holds a project-wide grant" case because `user.read` is not in `cat_sys_1` yet (#420). The suite now also Checks `editor`/`admin`.
- Next: #833 decision note, then by-id scoped Allow.

### DIFF
--- a/.changeset/authz-sk-team-scope.md
+++ b/.changeset/authz-sk-team-scope.md
@@ -1,0 +1,5 @@
+---
+"@zitadel/server": patch
+---
+
+Team service principals (`sk_team_`) can no longer act on users or teams outside the token's team, even if a project-wide grant exists. Project-scoped team-bound grants are rejected at mint time.
--- a/.changeset/authz-sk-team-scope.md
+++ b/.changeset/authz-sk-team-scope.md
@@ -2,4 +2,4 @@
 "@zitadel/server": patch
 ---
 
-Team service principals (`sk_team_`) can no longer act on users or teams outside the token's team, even if a project-wide grant exists. Project-scoped team-bound grants are rejected at mint time.
+Team service principals (`sk_team_`) can no longer act on users or teams outside the token's team, even if a project-wide grant exists. Project-scoped team-bound grants are rejected at mint time. HTTP wiring for the constraint lands in the stacked follow-ups (#893–#895).
--- a/docs/design/api/credentials.md
+++ b/docs/design/api/credentials.md
@@ -84,11 +84,13 @@ permission string (old `team.users.*` did). Compensating requirement: every
 `sk_team_` grant must carry a resolver-enforced team scope, and the deny-list
 suite must include "team token reads/writes a user outside its team".
 
-> **LOCKED / deferred:** Wave 3 (#423) ships the flat permission-name allowlist
-> only (`internal/authz/resolver/sk_team.go`). Resolver-enforced team scope and
-> the outside-team deny suite are **not** implemented yet — tracked in
-> [#831](https://github.com/zitadel/nextgen/issues/831). `sk_team_` is not
-> production-safe for `user.*` / `team_membership.*` until that lands.
+> **LOCKED:** Wave 3 (#423) shipped the flat permission-name allowlist
+> (`internal/authz/resolver/sk_team.go`). Resolver-enforced team scope
+> (`Request.TeamID` → `ConstraintTeamID`) and the outside-team deny suite
+> landed in [#831](https://github.com/zitadel/nextgen/issues/831). Grant
+> minting rejects project-scoped `sk_team_` grants for team-bound object
+> types. There is still no HTTP `sk_team_` token type — tests inject
+> `ScopeContext` / `resolver.Request` directly.
 
 SCIM sync (`scim.sync`) is not listed yet — SCIM is a hosted interop surface
 (`/scim/v2/Users`, `/scim/v2/Groups` in [`resource-map.md`](resource-map.md))
--- a/docs/design/api/permission-storage.md
+++ b/docs/design/api/permission-storage.md
@@ -80,7 +80,8 @@ Authz statement interfaces in `internal/service/statement.go` stay table-shaped
   materialization helper, foothold smoke helper, active system catalog) plus
   `internal/authz/resolver` orchestration (`sk_team_` permission-name
   allowlist, decision kinds). Resolver-enforced `sk_team_` **team scope**
-  (token `team_id` + outside-team deny suite) is deferred —
+  (`Request.TeamID` → `ConstraintTeamID`, outside-team deny suite, grant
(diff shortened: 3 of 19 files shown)


livio-a (commented):
Reviewed together with the rest of the stack (#893–#896). The SQL and mint-time validation are solid; the asks below are mostly about the constraint being consistent across Check/List/oracle and about the compensating test suite actually covering what the docs claim. Recommendation: address before merge.

**Blockers**

1. **Check and List disagree on what "in the token team" means.** `writeCheckedObjectInConstraintTeam` accepts only a membership-edge user or `ResourceID == ConstraintTeamID`, while `writeListedObjectInConstraintTeam` additionally accepts rows with `resource_scope_index.team_id = ConstraintTeamID` — and the new doc comment on `domain.AuthzCheckParams` promises the RSI branch for "other kinds" on both paths. A team-owned non-user resource (the allowlist includes `event.read` and `team_membership.*`) is list-visible but check-denied on the same id. The oracle mirrors the same asymmetry (`constraintTeamAllows` vs `listedObjectInConstraintTeam`), so the parity suite can't catch it. Either add the RSI `team_id` branch to the Check side (#893's `ResourceTeamID` is the natural input) or narrow the doc comment to the implemented semantics.

2. **The deny suite covers reads only, while credentials.md (edited in this PR) claims the read/write suite landed.** `TestAuthzSKTeam_OutsideTeamDeny` only exercises `viewer`; credentials.md's compensating requirement is "team token reads/writes a user outside its team". Please add an `editor`/`admin` deny case — the suite is the compensating control the doc mandates, and a future relation-specific resolution path would regress unobserved.

**Should fix**

3. The third disjunct of `writeListedObjectInConstraintTeam` (`r.team_id IS NOT NULL AND r.team_id = ?`) is not restricted by resource kind. Nothing (no schema CHECK, no writer guard) prevents a user RSI row from carrying a `team_id`, and if one ever does, it passes the list filter without a membership edge — while the oracle accepts users only via membership. Gate the disjunct on non-user/non-team kinds.

4. Oracle/SQL parity gap in the other direction: the oracle's default branch treats `id == ConstraintTeamID` as in-team for *any* kind, while the SQL gates that equality on `resource_kind = 'team'`. Ids are unprefixed caller-supplied strings here, so the differential suite can diverge exactly where it's supposed to be authoritative.

5. `teamBoundObjectType` hard-codes `{user, team, team_membership, event}`, re-encoding the object-type surface of `skTeamAllowed` with nothing linking them. A new allowlist entry for a team-bound type silently reopens project-scoped minting. One shared exported set in `internal/domain` (both packages already import it, no cycle) fixes the drift risk.

6. `writeUserMembershipInTeam` is the third near-identical copy of the membership-edges EXISTS predicate in this file (`writeFullTTUExists`, `writePrincipalMatch` emit the same shape). Worth consolidating now: ADR 052 (#876, proposed) will move membership-edge lookup to the principal's **home** project, and one helper means one place to switch instead of three.

**Notes (stack context, no change needed here if the stack merges together)**

- The constraint is silently skipped whenever `ResourceID` is empty, and nothing in the resolver forces `ResourceID` for team-bound sk_team checks — the outside-team deny exists only if the caller remembers to pass it. #893 wires the by-id path, but the fail-open-on-omission shape survives for future callers; consider a resolver-side guard.
- The changeset reads as if enforcement is live end-to-end; the API wiring only lands in #893–#895 (and `ScopeContext.TeamID` is unread until then). Suggest rewording to point at the stack.

PR #893, opened 2026-08-14
feat(authz): allow by-id management ops with team- or resource-scoped grants

adlerhurst:
## Summary

- `WriteCheckAuthz` now ORs team- and resource-scoped grant arms when `ResourceTeamID` / `ResourceID` are present (same arms list SQL already uses).
- `requireResourceAccess` passes RSI `team_id` and path id into `resolver.Check`. Bare `requireProjectAccess` (create/list) still Checks without those ids, so a team-only grant cannot create project-wide resources.
- stmttest keeps “no object ids → deny” and adds matching team/resource id → Allow.
- HTTP: `GetTeam` with only a team-scoped `project.viewer` grant returns 200 for that team, 403 for another team, and `CreateTeam` stays 403.
- Absorbs #891 Check/List sk_team constraint parity and folds the #892 Scoped Allow spec into this branch (one `## Scoped Allow` section).

Implements #833. Base is `main` (squash of #890 and #891). Decision note is #892.

## Validation

- `go test ./internal/domain/ ./internal/authz/resolver/ ./internal/storage/dialect/authz/ ./internal/api/ -count=1`
- `go test ./internal/storage/stmttest/ -count=1 -tags sqlite_integration -run 'TestAuthzResolverStatements_Cases|TestAuthzSKTeam_'`
- `ci / full-pr` passed on the merge of `main` (`b008db14`).

## Release notes / changeset

Changeset: `.changeset/authz-scoped-by-id-allow.md` — A team- or resource-scoped grant can now authorize a by-id management read or write when the resource lives in that team (or is the granted resource). Creating project-wide resources still requires a project-scoped grant. (`@zitadel/server`)

## Notes

- Fourth PR in the authz follow-up workstream. #890 and #891 landed on `main`; this branch merges `main` and keeps combined Check SQL (`writeCheckAllowed` scoped arms **and** `writeCheckedObjectInConstraintTeam`).
- Next: #834 Forbidden → partial-view lists (#894). Do not GitHub-merge this PR from the agent; leave the stack for review.

[...]

PR #894, opened 2026-08-14
feat(authz): return filtered lists for team-scoped-only principals

### LINKED ISSUE #834: authz: Forbidden→partial-view list gate so team-scoped EXISTS arms engage over HTTP
## Summary

`withAuthzListFilter` team-/resource-scoped EXISTS arms are dead over HTTP: `requireProjectAccess` returns 403 for team-scoped-only principals before the list predicate attaches. stmttest covers SQL; HTTP never narrows.

Raised on #811 (Livio review notes, item 1). Depends on / related to scoped by-id Allow follow-up.

## Acceptance

- [ ] List gate has a Forbidden → proceed-with-predicate (or equivalent partial-view) path for management lists
- [ ] HTTP integration tests pin team-scoped narrowing (not only stmttest)
- [ ] Docs/changeset claim team-scoped HTTP filtering only when live

## Notes

Until then, docs should state team-scoped HTTP list narrowing is not live; SQL + stmttest are substrate.

adlerhurst:
## Summary

- Closes [#834](https://github.com/zitadel/nextgen/issues/834): management list handlers no longer 403 when Check is Forbidden (foothold, no project-wide Allow).
- New `requireProjectListAccess` proceeds on Allow or Forbidden and attaches `withAuthzListFilter`; NotFound still 404s.
- Wired into `ListUsers`, `QueryTeams`, `ListSchemas`, `ListBranding`, and `ListFlowDefinitions`. Create and by-id gates are unchanged.
- Carve-outs: `QueryProjects` (the project is the boundary), `ListEvents` (no RSI kind — do not migrate to EXISTS), `QuerySessions` (waits on `sessionService.List`). `ListUsers` is EXISTS-wired but user RSI `team_id` is NULL, so a team-scoped-only principal sees `200 []`.

## Validation

- `go test ./internal/api/ ./internal/authz/resolver/ ./internal/storage/dialect/authz/ -count=1`
- `go test -tags postgres_integration -c ./internal/api/integration_test/` (typecheck of `TestListAuthzTeamScopedOnlyPartialView`; HTTP run needs postgres/spanner)

## Release notes / changeset

- Changeset: `.changeset/authz-list-partial-view.md` — Management list endpoints now return a filtered view when the caller has a project foothold but only a team- or resource-scoped grant, instead of HTTP 403. (`@zitadel/server` minor)

## Notes

- Stacked on #893. Absorbs #891 Check/List parity and the folded Scoped Allow docs.
- Schema RSI rows are project-scoped (no `team_id`), so `ListSchemas` under a team-only grant is an empty 200. `QueryTeams` / stamped branding and flow_definition rows return only the granted team. `ListUsers` is empty for the same RSI-null reason; by-id user reads deny.
- `(proceed bool, error)` is replaced by `(context, error)` on #896; this PR does not land the bool shape alone.


livio-a (commented):
The Forbidden→partial-view direction is right and the gate refactor is clean. The asks are about the contract being narrower than the docs now claim, and about the partial view being asserted where it actually differs. Recommendation: address before merge.

**Blockers**

1. **The docs over-claim what's live.** permission-storage.md now states "Team-scoped HTTP list narrowing is live" with only `QuerySessions` carved out — but `ListEvents` (internal/api/event.go) and `QueryProjects` (internal/api/project.go) still hard-403 on Forbidden with no EXISTS predicate. Either migrate them or list them as explicit exceptions. Note ListEvents can't simply be flipped later: events have no RSI resource kind, so a naive migration would hand a foothold-only principal the whole audit stream.

2. **The user partial view is structurally empty and nothing pins that.** User RSI rows carry `NULL team_id` (`NewUserResourceScope`), and the membership-edge arm in the list predicate only fires for sk_team's `ConstraintTeamID` — so a principal whose only grant is team-scoped gets `200 []` from ListUsers for a team that has members, while by-id on those same users allows (the #833 team arm matches `RSI.team_id`… also NULL for users, so by-id user reads deny too, while team by-id allows). Whatever the intended semantics are, they're currently an accident of RSI shape: please either add a membership-edge arm for team-scoped grants on `ResourceKindUser`, or pin the empty-view behavior with a test and document it so console/client authors aren't surprised.

3. **Partial-view boundedness is only asserted for teams (and schemas-as-empty).** The Forbidden-proceed paths for branding, flow_definition, and user kinds ship with zero assertions that the EXISTS predicate actually bounds them. One integration case per kind (granted row visible, outsider row absent) would close it.

**Should fix**

4. authz.md still says "Bare `requireProjectAccess` (create / **list** / project-by-id …)" ten lines below the new list rule, and the comment block in internal/api/authz.go has the same staleness. The next list endpoint author will follow one of these and reintroduce the 403-before-predicate behavior.

5. The `(proceed bool, error)` gate contract: every caller does `if err != nil || !proceed { return nil, err }`, so a future `(false, nil)` return produces a `(nil, nil)` handler response that ogen turns into a panic/500. The later refactor in this stack moves to `(context, error)` which resolves it — fine if the stack merges together, but worth not letting this shape land alone.

livio-a (approved):
All review asks landed and verified (docs carve-outs incl. ListEvents/QueryProjects, pinned empty user partial view with the #876/ADR follow-up noted, boundedness tests for branding/flow, stale gate wording removed). Merge-conflict resolution and duplicate-test fix verified with go vet under the integration tag.

PR #895, opened 2026-08-14
fix(authz): fail closed when management lists omit the authz filter

### LINKED ISSUE #838: authz: structural fail-closed if management list omits withAuthzListFilter
## Summary

Authz list filtering rides on a context value. A future management list that forgets `withAuthzListFilter` silently returns unfiltered rows. `compileRead` deliberately uses `context.Background()`.

Raised on #811 (Livio, item 4).

## Acceptance

- [ ] Structural guarantee: management-kind list statements require an authz list filter on ctx (assertion, compile-time hook, or refuse `compileList` without filter for management tables)
- [ ] Document interaction with intentional carve-outs (see companion issue on unwired list endpoints)

## Related

#811 #423

adlerhurst:
## Summary

- Closes [#838](https://github.com/zitadel/nextgen/issues/838): `compileList` with a non-empty table + resource-id column fails closed unless `AuthzListFilter` (or a documented bypass) is on the context. Missing filter is a 500, not an unfiltered list.
- `ListUsers` is hand-built (not `compileList`); each dialect now calls `RequireManagementListFilter` so forgetting the filter still errors instead of returning every user. GetUser stays compileRead-shaped and bypasses that tripwire.
- Documents [#839](https://github.com/zitadel/nextgen/issues/839) carve-outs: `QuerySessions` (`compileRead`), `QueryProjects` (project is the boundary), `ListEvents` (no RSI kind), `ListUsers` (predicate helper, still fail-closed), `ListUserPasskeys` / `ListUserTeams` (nested under `requireResourceAccess`), nested `ListFlowDefinitions` (create uniqueness / login Resolve), branding `GetLatest`.
- Nested non-HTTP lists (`Create` uniqueness, `Resolve`, `GetLatest`) use `WithAuthzListFilterBypass` so fail-closed compile does not 500 those paths. Inherited-filter vs unrestricted is split on #896.
- HTTP list wrappers map `ErrListFilterRequired` to `"authz list filter missing"`.

## Validation

- `go test ./internal/service/ ./internal/storage/dialect/authz/ ./internal/storage/dialect/postgres/ ./internal/storage/dialect/sqlite/ ./internal/storage/dialect/spanner/ -count=1`
- `go test -tags sqlite_integration ./internal/storage/stmttest/ -count=1 -run 'TestAuthzListUsers_RequiresFilter|TestAuthzListPredicate_|TestUserStatements'`

## Release notes / changeset

- Changeset: `.changeset/authz-list-compile-fail-closed.md` — Management list queries now fail closed when the authz filter is missing from the request context, instead of returning an unfiltered page. (`@zitadel/server` patch)

## Notes

- Stacked on #894 (`cursor/authz-list-partial-view-0900`).
- Do **not** wire `QuerySessions` here; keep the #839 table accurate until `sessionService.List` exists.
- `#842` remains parked until someone compiles/flips `cat_sys_1`.
- Inherited filter beating nested unrestricted is fixed on #896, not here.


livio-a (commented):
Fail-closed at the compiler is the right structural move, and the carve-out table is exactly the kind of doc this needs. Two gaps, one of them on the most sensitive table. Recommendation: address before merge.

**Blockers**

1. **`ListUsers` is outside the gate in all three dialects.** postgres/spanner/sqlite `user.go` hand-build the users list SQL and call `maybeWriteAuthzListPredicate` directly — the query never passes through `compileList`, so the new `RequireManagementListFilter` check does not apply. A future caller (or a handler refactor that drops the filter attach) silently returns every user in the project instead of `ErrListFilterRequired` — the exact #838 bug class this PR closes, on the PII table. Either call `RequireManagementListFilter` inside the ListUsers implementations, or route them through `compileList`. Relatedly, the #839 carve-out table claims completeness but omits this row — a reviewer auditing coverage against the table would sign off on the gap.

2. **An `AuthzListFilter` already on the context beats a later unrestricted wrap.** `RequireManagementListFilter` checks the filter first, and `maybeWriteAuthzListPredicate` applies it regardless of the unrestricted marker — so a nested "unrestricted" list (flow resolve, uniqueness checks, branding GetLatest) running under a request that attached a filter gets the outer kind's EXISTS injected into the nested query. Wrong-kind EXISTS matches ~no RSI rows, so e.g. a uniqueness check sees zero rows and permits a duplicate. I believe this is unreachable today (no filtered request reaches a nested list yet), but it's a silent-corruption shape: the unrestricted wrapper should clear/neutralize an inherited filter, or the precedence should be explicit and tested.

**Minor**

3. `ErrListFilterRequired` surfaces as a bare generic 500 (e.g. `flowDefinitionService.List` returns the storage error raw, skipping the mapListError treatment teams get). Since this error specifically means "authz wiring bug", a distinct message/label would keep it from being triaged as a DB outage.

livio-a (approved):
Approving — both blockers resolved and verified: ListUsers is gated via RequireManagementListFilter in all three dialects (with the stmttest pin), the #839 carve-out table documents the hand-rolled ListUsers path, and ErrListFilterRequired maps to a distinct internal message. The filter/unrestricted precedence hazard is addressed at the tip of the stack (#896 marker split), which is fine since these merge in order.

PR #896, opened 2026-08-14
perf(authz): skip list EXISTS on project-wide Allow

### LINKED ISSUE #837: authz: skip EXISTS on project-wide Allow + request-scoped catalog for lists
## Summary

After `requireProjectAccess` Allow via row-independent arms, list handlers still attach a correlated EXISTS (`withAuthzListFilter`). Also re-queries `ActiveSystemCatalogID` that the gate just fetched.

Raised on #811 (Livio, item 3). Prefer after RSI backfill so skipping the filter does not hide missing RSI rows.

## Acceptance

- [ ] Do not attach `withAuthzListFilter` when Check already Allowed via row-independent arms (order relative to RSI backfill documented)
- [ ] Request-scoped catalog / `requireProjectListAccess` so catalog id is not queried twice per list

## Related

#811 #423

### LINKED ISSUE #843: authz: process-level ActiveSystemCatalogID cache
## Summary

`ActiveSystemCatalogID` is queried per gated request; the resolver instance memo never survives a single `requireProjectAccess` call. A process-level (or request-scoped shared) cache invalidated on catalog activation would cut a query from every management request.

Raised on #807 (Livio, item 5).

## Acceptance

- [ ] Process-level or shared request-scoped cache for active system catalog id
- [ ] Invalidation on catalog activation

## Related

#807 #423

adlerhurst:
## Summary

- Closes [#837](https://github.com/zitadel/nextgen/issues/837): project-wide Allow skips the list EXISTS predicate **once** (`WithAuthzListSkipOnce`, consumed by the next `compileList` / `ListUsers`). A forgotten nested management list on the same request still fails closed (#838). Forbidden still attaches EXISTS.
- Nested non-HTTP lists (create uniqueness, login Resolve, branding `GetLatest`, stmttest) use `WithAuthzListUnrestricted`, which ignores an inherited filter so uniqueness cannot see zero rows from the outer kind’s EXISTS.
- Skipping EXISTS on Allow means that SELECT does not re-check RSI dual-write, and a grant revoked between Check and SELECT can still appear. Docs and the changeset say that explicitly.
- `requireProjectListAccess` returns `(context, error)` and attaches the skip-once marker or the filter. Check and the list filter share one request-scoped `resolver.Resolver`.
- `CheckParams` keeps the home vs target seam (`HomeProjectID` from the credential; not defaulted to the target project). It does not apply the sk_team allowlist short-circuit; sk_team `project.viewer` Checks are NotFound so they never take the Forbidden/filter branch. `ConstraintTeamID` plumbing stays.
- Does **not** close [#843](https://github.com/zitadel/nextgen/issues/843): a process-global catalog-id cache is the wrong layer.

## Validation

- `go test ./internal/service/ ./internal/authz/resolver/ ./internal/api/ ./internal/storage/dialect/authz/ ./internal/storage/dialect/postgres/ ./internal/storage/dialect/sqlite/ ./internal/storage/dialect/spanner/ -count=1`
- `go test ./internal/service/ -count=1 -run 'TestBrandingServiceGetLatest|TestAuthzList|TestMapListError'`
- `go test -tags sqlite_integration ./internal/storage/stmttest/ -count=1 -run 'TestAuthzListUsers_RequiresFilter|TestAuthzListPredicate_|TestUserStatements|TestBrandingStatements'`

## Release notes / changeset

- Changeset: `.changeset/authz-list-catalog-cache.md` — Management lists skip the authz EXISTS filter for one compile when Check already returned project-wide Allow. That skip does not re-check RSI dual-write, and a grant revoked between Check and SELECT can still appear. Forbidden lists still get the EXISTS partial view. (`@zitadel/server` patch)

## Notes

- Stacked on #895 (`cursor/authz-list-compile-fail-closed-0900`).
- Livio asked to hold this PR; it stays open. The skip is kept and scoped to one compile.
- `#842` remains parked until someone compiles/flips `cat_sys_1`.
- `#843` remains open: cache belongs per pool/statements, not package `atomic.Value`.


livio-a (commented):
I'd hold this one — it partially undoes what #895 just built, and the perf claim hides a behavior change. Recommendation: request changes.

**Blockers**

1. **Allow stamps `WithAuthzListUnrestricted` on the whole request context.** Every later `compileList` in that request — any resource kind, any project, including a future nested list whose author forgot its own gate — silently passes the #838 tripwire and compiles with no EXISTS and no error. The old Allow path carried a kind-pinned filter; the new marker is an unscoped request-wide bypass with two merged meanings ("project-wide Allow" and "not an HTTP path"). Please scope the skip to the single gated query (e.g. a one-shot marker consumed by the next compile, or an explicit option on the statement call) instead of the context subtree.

2. **"Redundant" isn't accurate — this is a behavior change shipped as perf.** On Allow, the old EXISTS still enforced (a) an RSI row per listed resource and (b) query-time re-evaluation of the grant. After this PR, a row whose RSI dual-write drifted appears in Allow lists while by-id still 404s, and a grant revoked between Check and the SELECT returns a full page instead of an empty one — and Allow vs Forbidden callers now see different lists on the same endpoint. If the skip stays, the changeset/docs should own these consequences explicitly; permission-storage.md's wave paragraph still says "Allow **or** Forbidden both proceed; the EXISTS predicate returns the partial view", and the stmttest comment still claims production compileList always requires the filter.

**Should fix**

3. `CheckParams` hardcodes `PrincipalHomeProjectID: req.ProjectID` and deletes the old `home := scope.ProjectID` + "#333 foreign footholds" breadcrumb. ADR 052 (#876, proposed) explicitly forbids defaulting home to target once a credential is resolved — this moves in the opposite direction right as that rule is under review. Keep the home/target seam visible (thread it or reinstate the comment).

4. `CheckParams` skips the sk_team allowlist short-circuit that `Check` applies, so its own doc comment ("EXISTS SQL matches Check") is false for sk_team principals. Separately: sk_team can never reach the Forbidden/filter branch at all — the gate Checks `project.viewer`, which `skTeamAllowed` masks to NotFound — so the ConstraintTeamID list plumbing and `TestWithAuthzListFilterCopiesConstraintTeamID` pin a production-unreachable path. Worth a comment (or an explicit decision) on how sk_team is ever supposed to pass the list gate, so the dead path doesn't read as covered.

livio-a (approved):
Approving — both blockers from the original review are properly resolved: the request-wide unrestricted context is replaced by the one-shot WithAuthzListSkipOnce (consumed by exactly one compile, second unfiltered list still fails closed) with WithAuthzListUnrestricted masking inherited filters for nested lists, and the Allow-skip behavior change (no RSI re-check / revoked-in-flight grants on that SELECT) is now owned in the changeset and docs. Home/target seam kept per ADR 053 (cross-project principals), sk_team dead-path documented.

One non-blocking follow-up: ConsumeAuthzListSkipOnce flips an unsynchronized bool through a context pointer — if a handler ever runs two gated lists in parallel on the same request context, that's a data race. Harmless while gated lists are sequential; please switch to sync/atomic (or document the single-consumer assumption) in a follow-up.

PR #900, opened 2026-08-14
feat: nest user attributes under their own object

vitorbari:
A user's content is defined by a user schema created through `/schemas`. Today that content is spread across the top level of the user object, mixed in with the server's own fields. This moves it into `attributes`, which is then exactly the document that satisfies the user schema: its property names and types come from that schema, and it is what the schema validates.

## `GET /users`

Before:

```json
{
  "users": [
    {
      "id": "user_01KZZY8PX8K8ATNVDSMRCZY4N1",
      "metadata": {
        "created_at": "2026-08-14T10:49:43Z",
        "updated_at": "2026-08-14T10:49:43Z",
        "status": "active",
        "lifecycle_owner_team_id": null
      },
      "$schema": "sch_01KZZY705G04CPPG0QGZM9ERF2",
      "$schema": "sch_01KZZY705G04CPPG0QGZM9ERF2",
      "companyName": "",
      "dateOfBirth": "2026-08-11",
      "email": "alex@example.com",
      "familyName": "",
      "givenName": ""
    }
  ]
}
```

After:

```json
{
  "users": [
    {
      "id": "user_01KZZY8PX8K8ATNVDSMRCZY4N1",
      "schema": "sch_01KZZY705G04CPPG0QGZM9ERF2",
      "attributes": {
        "companyName": "",
        "dateOfBirth": "2026-08-11",
        "email": "alex@example.com",
        "familyName": "",
        "givenName": ""
      },
      "metadata": {
        "created_at": "2026-08-14T10:49:43Z",
        "updated_at": "2026-08-14T10:49:43Z",
        "status": "active",
        "lifecycle_owner_team_id": null
      }
    }
  ]
}
```

## `POST /users`

Before:

```json
{
  "$schema": "sch_01KZZY705G04CPPG0QGZM9ERF2",
  "email": "alex@example.com",
  "givenName": "Alex"
}
```

After:

```json
{
  "schema": "sch_01KZZY705G04CPPG0QGZM9ERF2",
  "attributes": {
    "email": "alex@example.com",
    "givenName": "Alex"
  }
}
```

## `POST /users` → `201`

Before:

```json
{
  "id": "user_01KZZY8PX8K8ATNVDSMRCZY4N1",
  "$schema": "sch_01KZZY705G04CPPG0QGZM9ERF2",
  "email": "alex@example.com",
  "givenName": "Alex"
}
```

After:

```json
{
  "id": "user_01KZZY8PX8K8ATNVDSMRCZY4N1",
  "schema": "sch_01KZZY705G04CPPG0QGZM9ERF2",
  "attributes": {
    "email": "alex@example.com",
    "givenName": "Alex"
  },
  "metadata": {
    "created_at": "2026-08-14T10:49:43Z",
    "updated_at": "2026-08-14T10:49:43Z",
    "status": "active",
    "lifecycle_owner_team_id": null
  }
}
```

`GET /users/{user_id}`, `GET /users/me` and the `201` all return the same `User` schema.

## Changes

- **Schema-defined content moves under `attributes`.** It becomes the document the user schema validates, so `additionalProperties: false` and friends work — today the schema is applied to the whole body, including `$schema`, and rejects every create.
- **`id` and `metadata` are freed up.** They are envelope names only, so a user schema can now declare properties called `id` or `metadata`. The reserved-names note in `user-property.yaml` goes.
- **`$schema` becomes `schema`.** No `$`-prefixed property survives the TypeScript codegen — it reaches no model type, zod schema or mock — and zod strips unknown keys on parse. `schema`, not `schema_id`/`schema_url`, because the value is both: a project's default schema is registered under a URL, a console-created one is an opaque `sch_…` id.
- **The duplicate `$schema` goes.** It is stored twice today, as the `users.schema_url` column and as an attribute row, and the encoder emits both.
- **`POST /users` gets its own request schema.** `id` and `metadata` are absent by construction rather than rejected at runtime.
- **Unknown top-level properties on create are rejected.** `additionalProperties: false` on the request object; `attributes` stays open. ogen emits `unexpected field`, so ADR 047 §4's forbid is contract-enforced instead of a handler check — which the old shape could not do, since the request object's top level held the schema-defined content.
- **`201` returns the `User` schema.** `CreateUserResponse` is gone: it described the same resource minus `metadata`, which the server already knows at create time. Create and read now return one representation, and ogen lets the handler build the 201 with the same conversion `GET /users/{user_id}` uses.
- **`id` and `metadata` are required on the response.** They were optional only because this schema doubled as the create body; the server has always sent both.
- **Field order is `id`, `schema`, `attributes`, `metadata`** — identity first, then what the user is, with server bookkeeping last.
- **The server validates `attributes` alone.** The schema no longer sees the envelope, so `additionalProperties: false` and friends behave, and a schema may declare a property named `id` or `metadata`. The three copies of that reservation — `NewCreateUser`, `audit.reservedUserMapKeys`, the console's `RESERVED_KEYS` — are gone rather than kept in step.
- **`NewCreateUser` takes the schema URL as an argument.** It no longer reads it back out of the user map, which removes `SchemaFromUserMap` and with it the attribute row that duplicated `users.schema_url`.
- **Create returns the stored user.** The insert does not report `created_at`, `updated_at` or `status`, so the service loads the row back and the handler runs the same conversion `GET /users/{user_id}` uses.
- **`@zitadel/testing`'s `seedUser` nests its body.** It posted the flat shape and would have broken against the new contract.
- **ADR 052 records the decision.** ADR 009 specifies `$schema` on the request body and ADR 020 says the user schema describes the whole `GET /users/{id}` response; both keep their text and gain an amendment pointer. ADR 020's claim was aspirational rather than reversed — the response always carried fields no user schema declares, so it never validated the whole body.

Left as `string`; `format: uri` would reject the `sch_…` form, so it waits on the identifier form being settled.


api/openapi/endpoints/users/create-user-request.yaml
```diff
@@ -0,0 +1,26 @@
+title: CreateUserRequest
+description: |
+  `attributes` is the document that must satisfy the user schema named by
+  `schema`.
+
+  The resource primary key is server-assigned and returned in the response, so
+  the request carries no `id`; `metadata` is server-owned for the same reason.
+  Neither is part of this schema, which is what frees a user schema to declare
+  properties of those names.
+type: object
```
fforootd, line 14: **[P1] Could we close the outer request object with `additionalProperties: false` before replacing the existing runtime guard? With the current shape, exact-head ogen generation produces a decoder whose default branch skips unknown fields. A request containing `id` or `metadata` therefore decodes successfully and the handler can no longer see or reject those fields. That conflicts with ADR 047 section 4, which requires create-user to reject client-selected resource IDs. I verified that adding `additionalProperties: false` makes ogen emit an `unexpected field` error instead.

api/openapi/endpoints/users/user.yaml
```diff
@@ -1,29 +1,36 @@
...
+  `schema`, `metadata` — is the server-owned envelope. Keeping the two apart
+  means a schema may name a property `id` or `metadata` without colliding with
+  the envelope, and closed-world keywords (`additionalProperties: false`,
+  `propertyNames`, `unevaluatedProperties`) behave as their author expects.
 type: object
-additionalProperties: true
 required:
-  - $schema
+  - schema
+  - attributes
```
fforootd, line 18: **[P2] Now that create has its own request schema, could we also require `id` and `metadata` on this response model? Exact-head generation currently makes them `OptUserID` and `OptUserMetadata`, although the GET/list handlers always populate both. Adding them to `required` while retaining `readOnly` preserves their response-only status and lets generated clients rely on the complete server-owned envelope.

api/openapi/endpoints/users/user.yaml
```diff
@@ -1,29 +1,36 @@
...
-  users, this is only a base schema.
+  A user represents an individual identity in the system. It can be used to
+  represent a human user, but also a service account or any other type of
+  identity.
 
-  The server assigns the resource primary key (`id`) on create. Clients must
-  not send `id` in the create body; it is returned in the create response.
-# ogen does not enforce JSON Schema `not`, and rejects `properties.id: false`.
-# Document the forbid here; the create handler rejects a client-supplied `id`.
+  `attributes` is the document that satisfies the user schema named by
```
fforootd, line 7: **[P2] Could this contract proposal amend the related ADRs as well? ADR 009 still explicitly specifies `$schema` on create/update, while ADR 020 says the user-schema document itself is a valid `GET /users/{id}` response. Renaming the pointer to `schema` and validating only the nested `attributes` document reverses both recorded shapes, so merging the OpenAPI alone would leave two incompatible canonical contracts in the repository.

fforootd (commented):
I found three contract gaps to address before implementation.

api/openapi/endpoints/users/user.yaml
vitorbari: I adopted the same response object from the `GET /users/{id}` endpoint

api/openapi/endpoints/users/user.yaml
vitorbari: Added a small ADR documenting the changes: https://github.com/zitadel/nextgen/pull/900/changes/27e4d1d5fae12bb058e846cd86923d7b650ec430

internal/domain/user.go
```diff
@@ -192,7 +200,7 @@ func NewCreateUser(projectID string, teamID *string, id string, schemabs []byte,
 		return nil, ErrInternal(err).WithMessage("failed to unmarshal schema map")
 	}
 
-	attrs, err := CreateAttributesFromMap(muser, mschema)
+	createAttrs, err := CreateAttributesFromMap(attrs, mschema)
```
fforootd: **`attributes: {}` now answers `500` where it used to be impossible.**

`create-user-request.yaml` puts no `minProperties` on `attributes`, so `{"schema": "sch_…", "attributes": {}}` is a contract-valid request. Against any schema with no `required` properties it passes `jschema.Validate`, and this line then returns zero attributes — I confirmed by calling `NewCreateUser` with an all-optional schema and `map[string]any{}`: nil error, `len(createAttrs) == 0`.

The dialects then refuse it:

```go
// internal/storage/dialect/postgres/user.go:124 (also sqlite:62, spanner:71)
if len(user.Attributes) == 0 {
    return fmt.Errorf("user create requires attributes")
}
```

`applyCreateUser` wraps that as `domain.ErrInternal` → `internalErrorResponse` → `500 internal`.

This is new. On `main` the guard was unreachable, because `SchemaFromUserMap` forced `$schema` into the map and `fromMap` iterates every key — I verified on `origin/main` that a create stores both an `email` row *and* a `$schema` row, so `len >= 1` always held. Removing that duplicate row is the right call, but it also removed the thing that was accidentally keeping this branch dead.

The same path is now reachable from the flow handlers, which pass `state.CollectedData.UserData` straight through — previously that map always carried the injected `$schema`.

**Recommendation:** decide what an empty attributes document *means* and say it in one place. Two clean options:

1. If it is legal (a schema with only optional properties genuinely describes an empty document), let the dialects write a header row with no attributes and drop the `len == 0` guard — it reads like a leftover invariant rather than a real constraint.
2. If it is not legal, reject it here as `user.invalid` so the caller gets a 400 that names the problem, rather than a 500 from three dialects:

```go
if len(createAttrs) == 0 {
    return nil, ErrUserInvalid().
        WithMessage("attributes is empty. A user must carry at least one schema-defined property.")
}
```

Either way a table case with `"attributes": {}` in `TestCreateUser` would pin it — the current suite can't reach it, since every case sends at least `email`.

---

internal/api/user.go
```diff
@@ -240,7 +238,7 @@ func domainUserToApiUser(user *domain.User) (*api.User, error) {
 		return nil, domain.ErrInternal(err).WithMessage("failed to parse user attributes")
 	}
 
-	props, err := convertUsingJson[api.UserAdditional](userData)
+	attributes, err := convertUsingJson[api.UserAttributes](userData)
```
fforootd, line 241: **Users written before this change still carry a `$schema` attribute row, and it will now surface *inside* `attributes`.**

This maps `user.Attributes.ToMap()` through wholesale, which is exactly right for the new contract — but it means whatever is in the EAV table lands in `attributes`. I checked `origin/main`: `NewCreateUser` with `{"$schema": …, "email": …}` stores attribute rows `email` **and** `$schema`, so every user created before this PR has one. The PR body notes the double-store; the diff stops writing it but doesn't clean up what's already there, and I don't see a migration in the changeset.

So after upgrading an existing instance, `GET /users`, `GET /users/{id}` and `GET /users/me` return:

```json
"attributes": { "$schema": "https://…/default-human-user.json", "email": "alex@example.com" }
```

That undercuts the headline claim of the PR — `attributes` is supposed to be *exactly* the document the schema validates. A client validating it with `additionalProperties: false` (the very thing this change enables) rejects its own users, and the console picks `$schema` up as a spurious column through the `columnsFor` fallback now that `RESERVED_KEYS` is gone.

**Recommendation:** add a migration alongside the existing `users` tables that drops the orphaned rows — it's a narrow, well-keyed delete:

```sql
DELETE FROM zitadel_nextgen.user_attributes WHERE key = '$schema';
```

(plus the `user_unique_attributes` twin if any schema ever marked it unique, and the sqlite/spanner equivalents).

If a data migration is out of scope for a preview, a one-line skip in `domainUserToApiUser` would at least keep the contract honest while the rows age out — but the migration is the version that lets you delete the workaround later. Worth a line in the changeset either way, since operators with existing data see the difference.

---

internal/service/user.go
```diff
@@ -161,13 +164,26 @@ func (s *userService) emitUserCreateFailedBestEffort(ctx context.Context, action
...
+	// TODO(vitorbari): have the insert return the row so create needs no read-back.
+	user, err := s.v2Pool.Statements().GetUser(ctx, database.And(
+		database.Equal(database.Col(domain.UserFieldProjectID), input.ProjectID),
+		database.Equal(database.Col(domain.UserFieldID), action.CreateUser.ID),
+	), UserQueryOptions{})
+	if err != nil {
+		return nil, domain.ErrInternal(err).
+			WithMessage("The user was created but could not be read back. Fetch it by id rather than retrying the create.").
+			WithDetails(domain.CreatedUserDetails{UserID: action.CreateUser.ID})
+	}
```
fforootd, line 184: **The read-back runs outside the write transaction, so a transient read failure turns a successful create into a `500` that `409`s on retry.**

`ApplyActions` commits via `s.v2Pool.Transaction(...)`, then this runs as a separate statement on `s.v2Pool.Statements()` — the non-transactional pool. If it fails for any reason (connection reset, statement timeout, the row not yet visible on a follower, a concurrent `DELETE /users/{id}`), `POST /users` answers 500 even though the user exists and `user.created` was already emitted. The natural client response — retry the create — then gets `409 user.already_exists` for a user it never saw.

Returning the stored row instead of echoing the request is a genuine improvement, and the `TODO` already names the real fix, so this is mostly about how the window is handled until then. Credit where due: the error message is unusually good, and `CreatedUserDetails` does reach the client — I checked that `domainErrorDetails` marshals `Details` on 500s too, so a caller can actually recover from it.

**Two suggestions to make that recovery real rather than theoretical:**

1. **Document the status.** `users/methods.yaml` lists no `500` for this operation, only `default`. A caller can't be expected to handle a shape the contract doesn't mention — especially one where the recovery is "read `details.user_id`, don't retry".

2. **Consider a distinct error code.** `internal` is what a caller sees for genuinely unknown failures, and the correct reaction to those *is* to retry. Something like `user.created_not_readable` would let a client branch on the code instead of sniffing for a `user_id` in the details.

Alternatively, folding the read into the same transaction closes the window entirely and needs no contract change — `ApplyActions` already hands the action a `Statementer[AllStatements]`, so `CreateUserAction` could hold the loaded user as an output field the way `Prepare` already holds `CreateUser`. That might be the smaller step toward the `TODO` than teaching every dialect's insert to return its row.

---

internal/domain/user.go
```diff
@@ -160,28 +171,25 @@ func (c *CreateUser) AttributeTeamScope() string {
...
-		return nil, err
+//
+// attrs is the instance the schema validates: envelope fields (id, schema,
+// metadata) are not part of it, so a schema may declare properties of those
+// names and closed-world keywords such as additionalProperties: false hold.
+func NewCreateUser(projectID string, teamID *string, id, schemaURL string, schemabs []byte, attrs map[string]any) (*CreateUser, error) {
+	if schemaURL == "" {
+		return nil, ErrUserInvalid().
+			WithMessage("No schema provided. A user must name the schema its attributes are validated against.")
 	}
```
fforootd: Two small things about the new signature.

**1. `id` and `schemaURL` are adjacent, same-typed, and transposable.**

```go
func NewCreateUser(projectID string, teamID *string, id, schemaURL string, schemabs []byte, attrs map[string]any)
```

The old signature derived the schema from the map, so this mistake couldn't be made. Now `NewCreateUser(projectID, teamID, schemaURL, id, …)` type-checks and produces a `CreateUser` whose `ID` is a schema URL — and `ensureManagedID` treats a non-empty `ID` as caller-supplied, so it goes straight into the row as the primary key. Nothing downstream would catch it.

Cheapest fix is ordering: put `schemaURL` next to `schemabs`, which is what it belongs with anyway.

```go
func NewCreateUser(projectID string, teamID *string, id string, schemaURL string, schemabs []byte, attrs map[string]any)
```
…reads better still as `(projectID, teamID, id string, schemaURL string, schemabs []byte, …)` with the schema pair adjacent — or as a small `CreateUserParams` struct, given this is up to six positional arguments now.

**2. The `schemaURL == ""` guard below can't fire from any real caller.**

`CreateUserAction.Prepare` resolves the schema first:

```go
schemaEntity, err := o.schemaStore.GetJSONSchemaByID(ctx, o.ProjectID, o.SchemaURL)
```

With an empty URL that returns `NoRowFoundError`, so the caller gets `user.invalid` / *"schema is not known to the system…"* with `details: {"schema": ""}` — not the precise message written here. The HTTP path can't reach it either, since `schema` is `required` with `minLength: 1` in `create-user-request.yaml`. Only `TestNewCreateUser_SchemaURLRequired` exercises the branch, by calling the domain function directly.

It's harmless, but it's the kind of guard that reads as load-bearing to the next person. Either drop it, or — better — have `Prepare` check `o.SchemaURL == ""` before the store lookup, so a flow definition whose `UserSchemaURL` resolves empty reports the real cause instead of "schema is not known" with an empty string echoed back.

---

internal/domain/user.go
```diff
@@ -40,6 +39,18 @@ func ErrUserPermissionDenied() Error {
...
+// project does not have.
+type UserSchemaUnknownDetails struct {
+	Schema string `json:"schema"`
+}
+
+// CreatedUserDetails names a user the server created, for the answer that
+// cannot carry its representation.
+type CreatedUserDetails struct {
+	UserID string `json:"user_id"`
+}
```
fforootd, line 52: Both payloads look fine against the contract in `docs/design/api/error-details-producers.md` — a schema id is a conflict key, a user id on a self-referential answer is the actionable part. Typed structs rather than the old bare strings are exactly what that doc asks for.

The catch is that the same doc's inventory currently says these two files no longer produce details at all:

```
| `domain/user.go`, `service/user.go` | was string `WithDetails` | **Fixed** — moved to `Message` |
```

That row is now wrong: `UserSchemaUnknownDetails` and `CreatedUserDetails` are attached at `service/user.go:374` and `:182`. The doc is marked **Status: Active** and is the thing #585 tracks, so leaving it stale means the next audit of what actually reaches clients skips exactly the producers this PR adds.

**Recommendation:** update the row in the same PR — something like:

```
| `domain/user.go`, `service/user.go` | typed `UserSchemaUnknownDetails`, `CreatedUserDetails` | **Reference** — schema id and created user id, both client-actionable |
```

Small, but it's the difference between the inventory being a record and being a trap.

---

packages/testing/src/seed.ts
```diff
@@ -38,15 +39,14 @@ export async function seedUser(
...
   const user = (await client.createUser(
     {
-      ...input.attributes,
-      $schema: context.schemaId,
-      email,
-    } as Parameters<ZitadelClient["createUser"]>[0],
+      schema: context.schemaId,
+      attributes: { ...input.attributes, email },
+    },
     { project_id: context.projectId },
```
fforootd, line 50: Nice catch migrating this — it would have broken silently against the new contract, and the reworked comment above (dropping the now-meaningless "`$schema` would yield credentials that cannot log in") is a good touch.

Two spots in the same package still describe the old body, though, and both ship to consumers:

- `packages/testing/README.md:240` — *"**Seeding** is `POST /users` (the body carries `$schema: <schema id>`)"*
- `packages/testing/src/types.ts:19` — *"User documents must reference it via their `$schema` field, so seeding needs it alongside the project credential."*

The README counts even though `package.json` has `files: ["dist"]` — npm always includes `README.md` in the tarball regardless of that field, so it's what a consumer reads on npmjs.com. And `types.ts:19` documents `SeedContext.schemaId`, which is the field they'd reach for.

Anyone following either writes `{ $schema, email }` and now gets `400 unexpected field "$schema"` from the closed `CreateUserRequest` — with the shipped docs telling them the opposite of the contract.

**Recommendation:** mirror the wording you already used here, e.g. for `types.ts`:

```ts
  /**
   * Server-assigned id of the seeded user schema. A user names it in `schema`
   * and puts the schema-defined content under `attributes`, so seeding needs
   * it alongside the project credential.
   */
  schemaId: string;
```

---

internal/api/integration_test/user_test.go
```diff
@@ -109,28 +119,59 @@ func TestCreateUser(t *testing.T) {
...
 		for _, tc := range tcs {
 			t.Run(tc.name, func(t *testing.T) {
 				t.Parallel()
 
-				user := &api.User{}
+				user := &api.CreateUserRequest{}
 				err := user.UnmarshalJSON([]byte(tc.userjson))
 				require.NoError(t, err)
 
 				resp, err := client.CreateUser(t.Context(), user, params)
```
fforootd: Not introduced here — but this PR rewrote all five `userjson` fields in this table, so it's a natural moment to notice it.

This passes `params`, the one declared at line 41 with `TeamID: {Set: true, Value: team.ID}` — not `tc.params`. The table's `params` field is never read by anything. Every case, including the ones that deliberately set `params: {ProjectID: project.ID}` with no team, therefore creates a user *with* an initial team membership.

So the no-team create path isn't actually covered. That path is not hypothetical: it's the `AttributeTeamScope()` empty-string branch, which decides the scope key for team-unique attributes.

**Recommendation:**

```suggestion
				resp, err := client.CreateUser(t.Context(), user, tc.params)
```

Worth running once before committing it — if a case was quietly relying on the team being set, that's itself worth knowing. If the field turns out to be genuinely unwanted, deleting it and the five literals is the other honest fix; what's not great is keeping a field that reads as coverage and isn't.

---

fforootd (commented):
Reviewed at `40fba46`. CI is green across all 10 checks; `go build ./...`, `go vet` and the Go unit tests pass locally against `origin/main` as the base.

**This is a good change and the reasoning in the description holds up.** Nesting schema content under `attributes` really does fix what it claims: the schema now validates a document it owns, so `additionalProperties: false` works instead of rejecting every create, and `id`/`metadata` become usable property names. Collapsing `CreateUserResponse` into `User` so create and read return one representation is the right call, and removing the three duplicated reserved-key lists (`NewCreateUser`, `audit.reservedUserMapKeys`, the console's `RESERVED_KEYS`) is the kind of deletion that makes the invariant real rather than replicated. The migration is thorough — console, `@zitadel/testing`, both flow handlers and the e2e specs all moved, `api-mock` genuinely needs no change since it returns an empty users list, and the new `TestNewCreateUser_EnvelopeNamesAreUsable` / `ClosedSchema` tests pin exactly the two behaviours the change exists to enable.

Seven comments inline. Ranked by what I'd act on:

**Worth fixing before merge**

1. **`attributes: {}` now returns 500.** `create-user-request.yaml` sets no `minProperties`, so an empty attributes object is a valid request; against an all-optional schema it produces zero EAV rows and trips the dialects' `user create requires attributes` guard, which surfaces as `500 internal`. It's a regression — on `main` the injected `$schema` row kept that branch unreachable. Verified by running `NewCreateUser` directly.
2. **Old users still carry a `$schema` attribute row, and it now lands inside `attributes`.** Verified on `origin/main` that every pre-existing user has one. There's no migration and no read-side filter, so after an upgrade `attributes` contains a key no schema declares — which is precisely the thing the new contract promises can't happen.

**Worth a look**

3. The post-create read-back runs outside the write transaction, so a transient read failure turns a successful create into a 500 that 409s on retry. The `TODO` already names the fix; the suggestion is mostly about documenting the 500 in `methods.yaml` and/or giving it a code a client can branch on.
4. `NewCreateUser`'s adjacent same-typed `id, schemaURL` params are silently transposable, and the new `schemaURL == ""` guard can't fire from any real caller — `Prepare` resolves the schema first and reports a different cause.

**Small**

5. `docs/design/api/error-details-producers.md` still lists these two files as "Fixed — moved to `Message`" while this PR reintroduces `WithDetails` in both. The payloads themselves look contract-legal.
6. `packages/testing`'s `README.md` and `types.ts` still document the `$schema` body that now 400s; `seed.ts` itself was migrated correctly.
7. Pre-existing, but adjacent: the dead `tc.params` in `TestCreateUser` means the no-team create path isn't actually covered.

Nothing here touches the shape of the design — (1) and (2) are both about the edges of a contract I think is the right one.

---

internal/api/user.go
vitorbari: Not relevant as there are no users in production.

internal/domain/user.go
vitorbari: Fixed.

internal/service/user.go
vitorbari: Documented the 500 case.

PR #909, opened 2026-08-15
refactor: add the claim service behind the 501 claim endpoints

### LINKED ISSUE #611: Claim E1: claim service (init/status/complete)

IAM-marco:
Closes #611. Part of #96.

## Summary

- add `ClaimService` (`internal/service/claim.go`) implementing the init/poll/complete dance of ADR 046: `Init` mints a handoff-style challenge (plaintext returned, hex SHA-256 stored, 10 minute TTL) and answers 409 with `{team_id, dashboard_url}` details when already claimed; `Status` verifies proof of possession before revealing any state, then reports pending, completed (team, claimed-at, dashboard URL reconstructed from the grant), or 410; `Complete` spends the challenge and writes the `project.team` grant, the resource-scope `team_id`, and the `authz.granted` audit event in one transaction
- add the domain primitives: challenge token mint/hash helpers, `ClaimChallengeTTL`, `ErrClaimNoPersonalTeam`, `ClaimConflictDetails`, and `NewClaimTeamAssignment`
- wire `claimService` into `Handler`/`NewHandler` and regenerate `MockClaimService`; the three ogen claim methods still fall through to `UnimplementedHandler`, so the endpoints keep answering 501 until #612 wires them
- unit tests first, on the existing gomock harness, covering the full matrix including the same-challenge race (`MarkChallengeCompleted` finds no pending row, 410) and the 409 details payload

## Validation

- `go test ./internal/service/ -run TestClaimService -v` (21 cases green)
- `go build ./...`, `go vet ./...`, `gofmt -l` clean on touched packages
- `node scripts/check-pr-title.mjs --title "refactor: add the claim service behind the 501 claim endpoints"`
- `corepack pnpm exec changeset status --since origin/main` reports no bumps

## Release notes / changeset

No changeset: nothing ships while the endpoints answer 501 (review feedback; the earlier empty changeset was dropped).

<details><summary><h2>Notes</h2></summary>

- Deliberate deviation from the ticket text: completing an already-completed challenge reports 409 already claimed with details, not 410. A completed challenge implies the grant exists (both are written in the same transaction), and the OpenAPI contract plus the api-mock (`platform-handlers.ts`) already describe 409 for a re-spend.
- Claimed-by provenance lives on the `authz.granted` event's actor: the grant's grantor columns are CHECK-reserved for delegations (migration 000015).
- Known accepted ceiling, commented at the claimed-check in `Complete`: two different pending challenges racing on one project can double-write the grant under read committed; the upgrade path is a conditional resource-scope update or a partial unique index.
- Claim and dashboard URLs derive from `builtin_public_base`'s origin plus the console path; a dedicated public-base setting should replace this when cloud deployment configuration lands.
- Handoffs recorded for #612: plumb the presented-bearer SHA-256 into `ScopeContext` in `HandleOAuth2`, add a claim branch to `errorResponse` (`ErrClaimChallengeNotFound` currently maps to 500, `ErrClaimNoPersonalTeam` to 400), and reconcile the handler's raw `cfg.Platform.ProjectID` pin with the service's `ResolvedProjectID()`.
- If PR #876 (ADRs 052/053) lands, `Complete` gains an explicit target `team_id` with an owner check and CSRF lands in the handler layer; both are contained changes.

</details>


internal/service/claim.go
```diff
@@ -0,0 +1,282 @@
...
+		if _, ok := errors.AsType[*database.NoRowFoundError](err); ok {
+			return nil, domain.ErrClaimChallengeNotFound()
+		}
+		return nil, domain.ErrInternal(mapStorageError(err)).WithMessage("failed to load claim challenge")
+	}
+	// Proof of possession before anything else: a caller without the
+	// initiating secret learns nothing about the challenge, not even expiry.
+	if challenge.InitiatingSecretHash != secretHash {
+		return nil, domain.ErrProjectPermissionDenied()
+	}
```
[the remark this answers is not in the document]
IAM-marco: Applied in 6f1db41d: the gate now uses crypto/subtle.ConstantTimeCompare on the two hex digests. Worth noting the practical exposure was limited even before, since both sides are SHA-256 digests of the bearer and recovering a digest prefix through timing does not yield a bearer that hashes to it without a preimage, but the constant-time form is one line and removes the question entirely.

internal/service/claim.go
```diff
@@ -0,0 +1,282 @@
...
+		ClaimURL:    s.claimURL(projectID, plain),
+		ChallengeID: plain,
+		ExpiresAt:   challenge.ExpiresAt,
+	}, nil
+}
+
+func (s *claimService) Status(ctx context.Context, projectID, challengeID, secretHash string) (*ClaimStatusResult, error) {
+	challenge, err := s.v2Pool.Statements().GetChallengeByID(ctx, projectID, domain.HashClaimChallengeToken(challengeID))
+	if err != nil {
+		if _, ok := errors.AsType[*database.NoRowFoundError](err); ok {
```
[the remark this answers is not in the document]
IAM-marco: errors.AsType is the generic added to the standard errors package in Go 1.26, and this module declares go 1.26. It is also the established convention across this package (internal/service/errors.go, event_claim.go, event.go, auth_attempt.go all use it). The full-pr gate compiles and tests this file, and it is green on this PR, so the code builds as written. Keeping AsType for consistency with the surrounding code.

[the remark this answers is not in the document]
IAM-marco: Same answer as the earlier thread on this line: errors.AsType is stdlib as of Go 1.26 (this module's declared version) and the repo-wide convention in this package, and the green full-pr gate on this PR compiles and tests this file. No change needed.

internal/service/claim.go
```diff
@@ -0,0 +1,282 @@
...
+		// completion wrote the grant in this same transaction shape, so a
+		// completed challenge always reports 409 with the owning team rather
+		// than 410 (matching the OpenAPI contract and the api-mock).
+		//
+		// Two different pending challenges racing on one project can still
+		// double-write the grant under read committed: this check is app-level
+		// and UpsertResourceScope overwrites team_id. Accepted for alpha; the
+		// upgrade path is a conditional update (team_id IS NULL) or a partial
+		// unique index on the grant.
+		teamID, err := claimedTeamID(ctx, stmts, projectID)
```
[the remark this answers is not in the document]
IAM-marco: This window is documented deliberately rather than left by accident. The ticket (#611, current-state update) chose UpsertResourceScope for the claim write, and the approved implementation plan recorded the residual cross-challenge race as accepted for alpha, with exactly the upgrade path this comment names (a conditional team_id IS NULL update or a partial unique index, one statement across the three dialects). The realistic race, two concurrent completes of the same challenge, is already serialized at the storage layer by MarkChallengeCompleted's pending-to-completed UPDATE, and the unit tests cover the loser getting 410. The residual case requires the same human to run claim init twice and complete both pending challenges within the 10 minute TTL against one project. Hardening it at the storage layer is a contained follow-up once #612 adds the end-to-end concurrency tests around this path.

cmd/server/server.go
```diff
@@ -183,6 +183,13 @@ func run(ctx context.Context, cfg Config, userFiles []string) error {
 		nil,
 	)
 	teamService := service.NewTeamService(serviceDBPool)
+	// The claim and dashboard URLs hang off the console. builtin_public_base is
+	// the only public-origin config today and carries the /api/schemas path, so
+	// strip it down to the origin before appending the console path; a
+	// dedicated server public-base setting should replace this when cloud
+	// deployment configuration lands.
+	consoleBase := (&url.URL{Scheme: builtinPublicBase.Scheme, Host: builtinPublicBase.Host}).String() + cfg.Server.ConsolePath
```
vitorbari, line 191: Just FYI, console base URL is also needed when we are generating schemas IDs

.changeset/claim-service-behind-501.md
```diff
@@ -0,0 +1,4 @@
+---
+---
+
+The claim service (#611) now backs claim init/status/complete inside the server, but the HTTP handlers are not wired until #612, so the endpoints keep answering 501 and no shipped behavior changes.
```
vitorbari: I think the changeset can just be skipped in this case.

internal/domain/claim.go
```diff
@@ -9,6 +13,44 @@ const (
...
+	return plain, HashClaimChallengeToken(plain), nil
+}
+
+// HashClaimChallengeToken derives the stored challenge id from a presented
+// plaintext token. The claim_challenges.id column is TEXT, so unlike the
+// handoff token's raw []byte hash this one is hex-encoded.
+func HashClaimChallengeToken(plain string) string {
+	sum := sha256.Sum256([]byte(plain))
+	return hex.EncodeToString(sum[:])
+}
```
vitorbari, line 38: Please add unit tests for these to make sure the behavior is deterministic.

internal/service/claim.go
```diff
@@ -0,0 +1,282 @@
...
+
+// claimedTeamID resolves the claim state all three legs branch on: a missing
+// project is ErrProjectNotFound, an unclaimed project (no scope row or no
+// team) is (nil, nil), a claimed project returns its owning team id.
+// projectIsClaimed (event_claim.go) is deliberately not reused: events
+// visibility treats a missing project as unclaimed, while claim needs the 404
+// and the team id for the 409 details.
+func claimedTeamID(ctx context.Context, stmts interface {
+	GetProjectByID(ctx context.Context, id string) (*domain.Project, error)
+	GetResourceScope(ctx context.Context, resourceID string) (*domain.ResourceScope, error)
```
vitorbari: You can create smaller interfaces here in the caller side. I think, in the future, the idea is to split the Statements per resource, so you could already do this split.

internal/service/claim.go
```diff
@@ -0,0 +1,282 @@
...
+// completedStatus reconstructs team and claim time from the grant written at
+// complete. Missing scope or grant on a completed challenge is corrupt state:
+// both are written in the same transaction that marks completion.
+func (s *claimService) completedStatus(ctx context.Context, projectID string) (*ClaimStatusResult, error) {
+	stmts := s.v2Pool.Statements()
+	scope, err := stmts.GetResourceScope(ctx, projectID)
+	if err != nil || scope.TeamID == nil || *scope.TeamID == "" {
+		return nil, domain.ErrInternal(err).WithMessage("completed claim challenge without a project-team scope")
+	}
+	assignments, err := stmts.ListAuthzAssignments(ctx, projectID, domain.AuthzPrincipalTypeTeam, *scope.TeamID, false)
```
vitorbari, line 164: Isn't it possible to introduce a `Get` instead of `List`?

internal/service/claim.go
```diff
@@ -0,0 +1,282 @@
...
+	var result *ClaimCompleteResult
+	err := s.v2Pool.Transaction(ctx, func(ctx context.Context, tx Statementer[AllStatements]) error {
+		stmts := tx.Statements()
+		challenge, err := stmts.GetChallengeByID(ctx, projectID, challengeIDHash)
+		if err != nil {
+			if _, ok := errors.AsType[*database.NoRowFoundError](err); ok {
+				return domain.ErrClaimChallengeNotFound()
+			}
+			return err
+		}
```
vitorbari, line 192: What happens when this gets called twice? Does the Challenge gets delete from DB after completion?

cmd/server/server.go
IAM-marco: Good point, and it is why this line cannot just redefine builtin_public_base: the schema $ids are minted from the full base including the /api/schemas path (api/openapi/endpoints/schemas/embed.go, DefaultHumanUserSchemaURL), so the base has to keep carrying it. This line strips to scheme+host only for the console-facing claim and dashboard URLs. A dedicated public-origin setting that both consumers derive from stays future work, as the inline comment records.

.changeset/claim-service-behind-501.md
IAM-marco: Done, the changeset is removed in de502d04.

internal/domain/claim.go
IAM-marco: Done in 8307682e: internal/domain/claim_test.go pins HashClaimChallengeToken against a fixed SHA-256 hex vector (so an algorithm or encoding change fails loudly instead of silently invalidating stored challenge ids) and covers the mint: prefix, id equals the hash of the plaintext, and two mints never collide.

internal/service/claim.go
IAM-marco: Done in 02c5537d: the anonymous interface is replaced by named caller-side interfaces. claimedProjectStatements covers the claimed-check reads, and claimStatements spells out the full slice of AllStatements the claim flow touches; Init/Status/Complete all bind their statements source to it, so the seam for the per-resource split is in place. The pool-level half (ClaimPool, Statementer[ClaimStatements]) is the part that still waits on go 1.27 generic methods, per the commented block in internal/service/database.go.

internal/service/claim.go
IAM-marco: I'll ask @livio-a and @adlerhurst

internal/service/claim.go
IAM-marco: The challenge row is deliberately never deleted: it flips pending to completed inside the same transaction that writes the grant, and it has to survive because the CLI keeps polling claim/status afterwards and needs the completed row to report success (team, claimed_at, dashboard_url). Cleanup is read-time by design; ADR 046 lists automated expiry/deletion as a non-goal since there is no scheduled-task infrastructure yet. Called twice sequentially, the second call hits the claimed-check and gets 409 already_claimed with the owning team in the details. Called twice concurrently, MarkChallengeCompleted's pending-to-completed UPDATE serializes them and the loser gets 410. Both paths are unit-tested in claim_test.go ("re-spent completed challenge reports already claimed" and "lost the same-challenge race").

internal/service/claim.go
livio-a: A generic `Get` keyed on `(project, principal_type, principal_id, object_type, relation)` isn't really possible here: that tuple isn't unique. The active-uniqueness index also spans `catalog_id`, `scope_kind`, and the scope ids — the same principal can hold the same relation project-wide, team-scoped, and resource-scoped as separate rows (that's exactly what the scoped-Allow work in #893 evaluates). A `Get` with those five params would need "which one?" semantics, and the full unique key is nine params — no longer a convenience. It would also duplicate the "active" filtering (revoked/expired/catalog) that today lives only in the resolver SQL, which is drift we've been actively hunting elsewhere.

What this call site actually wants is narrower: *the owning-team grant of this project*, which **is** unique per project (and ADR 053 §2 makes that explicit). So the options I'd consider:

1. Keep the `List` + scan — fine by me: the team has exactly one assignment on the project at claim time, and the scan doubles as the corrupt-state check.
2. If we want a `Get`, make it semantic, e.g. `GetProjectOwningTeamGrant(ctx, projectID, teamID)` — not the generic tuple getter. Small bonus: it would centralize the hardcoded `"project"`/`"team"` relation pair, which ADR 053 plans to rename to `project.owning_team`, so that migration becomes a one-line change instead of a grep-hunt through call sites.

Either 1 or 2 works for me; just not a generic `GetActiveAuthzAssignment`.

IAM-marco: On option 2: aren't we going to make a too specialized call? If so, what is stopping us to have one call per object-actor pair (e.g. `Get<ObjectPermission><Actor>`)?

vitorbari (approved):

PR #915, opened 2026-08-18
feat(login): offer a passkey after password registration

bastionstack:
## Summary

- The default login flow gains a `passkey-upsell` step. `register-password` submits into it instead of to `done`, and it offers `passkey_register` with `skip` beside it — both terminal. Registration no longer ends the moment a password is set.
- New projects get it at seed time (`internal/service/project.go` `createDefaultLoginFlowDefinitions`). Existing projects keep whatever flow definition is already in their database; there is no migration, and a tenant can edit theirs.
- The copy and the mock's state machine have carried this chain since #116 — `packages/components/src/orchestrator/locales/en.ts` documents `register → register-password → passkey-upsell → done` and defines all six strings, and `@zitadel/api-mock` already had the state. Only the flow definition never caught up.
- `@zitadel/testing`'s `registerWithPassword` now clears the offer, so callers land on their signed-in surface. The two suites that carried their own skip handling after it drop it — the two race each other.
- Split out of #913 (the login rebuild) so the behaviour change is reviewed on its own rather than inside a visual diff.

## Validation

- `moon run api-mock:typecheck api-mock:test config:test testing:test`
- `go test -tags postgres_integration -run 'TestCreateProjectProvisionsDefaultLoginFlow|TestPostCreateUserPasskeyUpsell' ./internal/api/integration_test/...`
- `node scripts/check-changesets-status.mjs`
- Not run: the upsell step has not been driven in a browser (see Notes).

## Release notes / changeset

- Changeset: `.changeset/passkey-upsell-step.md` — the default flow offers a passkey after password registration (`@zitadel/config`, `@zitadel/server` — the flow JSON is embedded into the server binary via `packages/config/defaults/defaults.go`).
- Changeset: `.changeset/testkit-absorbs-passkey-upsell.md` — `registerWithPassword` clears the offer (`@zitadel/testing`).

## Notes

- **Draft: the step has no browser-level coverage.** The server contract is proven end to end — `TestPostCreateUserPasskeyUpsell` drives a real WebAuthn ceremony with `virtualwebauthn` from this step and asserts a credential row lands, and the skip branch terminates cleanly. But no browser test takes the passkey *from the upsell screen*: `registerWithPasskey` enrols on the `register` step and never reaches it, and `registerWithPassword` is the helper that skips it. Worth adding before this leaves draft.
- Two design questions from the frame are unresolved: it shows no trustmark although attribution is injected on every step, and its copy ("Setup passkey", "With Face ID, Touch ID, or PIN.") differs from the locale strings ("Set up passkey", "Sign in with Face ID…"). The locale strings ship today.
- `packages/api-mock/package.json` declares `vitest` and `@vitest/browser`, without which `api-mock:typecheck` fails on any PR that makes the package affected. The same two lines are in #913; whichever merges second drops them on rebase.

### DIFF
--- a/.changeset/passkey-upsell-step.md
+++ b/.changeset/passkey-upsell-step.md
@@ -1,0 +1,6 @@
+---
+"@zitadel/config": minor
+"@zitadel/server": minor
+---
+
+The default login flow gains a `passkey-upsell` step between `register-password` and `done`. A user who has just set a password is offered a passkey (`passkey_register`) with a `skip` action beside it, so registration no longer ends on the password step. The copy already existed in the orchestrator's locales; only the flow definition was missing it. Tenants running an ejected flow definition are unaffected — the step is added to `defaults/default-login.json`, not injected at runtime — and `@zitadel/api-mock` mirrors it.
--- a/.changeset/testkit-absorbs-passkey-upsell.md
+++ b/.changeset/testkit-absorbs-passkey-upsell.md
@@ -1,0 +1,5 @@
+---
+"@zitadel/testing": minor
+---
+
+`registerWithPassword` now clears the passkey upsell the default flow shows after a password is set, so a caller still lands on their signed-in surface rather than on an intermediate step. Suites that added their own skip handling after `registerWithPassword` should drop it — the two race each other. A flow that routes `register-password` straight to `done` is unaffected: the helper only clicks a `skip` action that is actually rendered, and never waits for one.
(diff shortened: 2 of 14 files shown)


bastionstack:
Superseded by #1067, which carries the same change plus the scaffold update the social-login drift test now requires.

This branch had fallen behind both main and its own remote, so it was replayed onto current main rather than merged forward.

PR #920, opened 2026-08-19
docs: add decisions from the 12 + 19 Aug design reviews

jjjssswww:
## Summary

Logs the decisions from the two design reviews on 12 and 19 August in
`docs/design/decisions.md`. Docs only — no code, no schema, no API changes.

**12 Aug — user listing & session details**

- D29 user list ships as the raw merged schema table (all columns of all
  schemas, dashes where a property is missing) rather than pre-solving the
  display problem before release
- D30 display name: given + family name → email → user ID (already merged)
- D31 session list: name, email, user ID; company column removed
- D32 no search anywhere in the MVP — to be defined as a global concept first
- D33 revoked sessions leave the list; revoking needs a confirmation
- D34 session detail: minimal, one view, IP + user agent only
- D35 invitations hold an email only — no user and no schema behind them

**19 Aug — settings, members, flows, appearance**

- D18–D20 single project-admin role, member actions (remove / revoke /
  resend), toasts as the console-wide notification pattern
- D21–D23 flows list becomes a table, steps table shows actions, metadata
  block on every detail view
- D24 schema list stays as it is for now — no iterating on unreleased work
- D25–D28 appearance read-only for Alpha, five built-in layouts, JSON/YAML
  view, accessibility guaranteed for default templates only

**Note on numbering:** D16 (profile settings) and D17 (settings area) were
decided on 12 Aug and only re-confirmed on the 19th, so they are dated
12 Aug rather than duplicated. The new IDs therefore continue at D29 even
though those decisions are older than D18–D28.

Open questions and the follow-up blocks are updated for both reviews.

## Validation

Not run — documentation-only change. Rendering checked in the GitHub
markdown preview.

## Release notes / changeset

None — internal design documentation, no user-facing change.

### DIFF
--- a/docs/design/decisions.md
+++ b/docs/design/decisions.md
@@ -1,9 +1,16 @@
 # Design Decisions
-
 ---
-
 ## Open questions
-
+- [ ] **Last-admin protection** — you can currently remove every admin, leaving the account without one. Tim says enforcing "can't delete the final role" is hard because roles are flexible; asked whether the API can block it temporarily so we don't build on a broken default. Design shows remove/revoke as if it were unrestricted until confirmed. Spec issue incoming. *(owner: eng / Tim)*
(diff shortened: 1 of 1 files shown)


bastionstack:
D32 drops search console-wide and gives as part of its reasoning that the field is static. A working per-list filter has since shipped in the Teams directory, so the log and the console disagree as written.

Suggest amending D32 to record that the decorative field was removed, and that a per-list client-side filter is allowed where it genuinely filters. The open question about defining search as a console-wide concept is unaffected.

Issue #924, opened 2026-08-19
Wire up cursor pagination on listSchemas


vitorbari:
## Context

![Schema directory](https://github.com/user-attachments/assets/676d5095-5605-4e5e-9f5a-ca480e00e6b6)

`GET /schemas` declares both `page_token` and `offset`, but its response is a bare array — nothing can carry a next-page token back to the caller.

Neither parameter does anything: `ListSchemas` accepts both and reads neither, and the options it builds set no limit and no cursor, under a `// TODO: implement pagination` (`internal/service/schema_service.go:142`). Paging returns the whole list, with no error.

There is no cap either. Branding capped its list at 100 for exactly this reason (`internal/service/branding.go:107-113`); schemas accumulate the same way (#923) and have nothing. The console consuming this endpoint gets away with it by assuming the list stays small.

Storage already supports paging: `compileList` decodes an incoming cursor, validates it against the request's ordering, appends the keyset predicate and applies the limit (`internal/storage/dialect/postgres/compiler.go:38-56`, `:71`), and `ListJSONSchemas` already returns a `NextCursor`. Only the service and the API contract are missing. ADR 027 makes cursor pagination the rule for new endpoints and retires offset.

## Description

The endpoint advertises pagination it does not implement, and ignores `page_token` silently rather than rejecting it — so a client that believes it is paging reads page one forever, and drops schemas as soon as the list outgrows one response. The response is also unbounded, growing with every schema revision.

`GET /schemas` should page like the other list endpoints, return a bounded response, and stop advertising parameters it ignores.

One constraint: paging needs a stable total order, and `created_at` is not one — the table's key is `(project_id, url)` and timestamps can collide, so a page boundary can skip or repeat rows.

### Suggestion

`ListUsers` is a direct precedent (`internal/service/user.go:193-216`): normalised limit, `page_token` passed through as the cursor, ordering by `created_at, id` descending so the order is total, and `result.NextCursor` returned as `next_page_token`. The schema equivalent orders by `created_at, url`.

The response envelope this needs is the same change #921 makes to `list-schemas-response.yaml`, so the two want to land together or in that order. `offset` should go at the same time, per ADR 027.

## Acceptance Criteria

- [ ] A caller can page through every schema in a project using `page_token`, and each schema appears exactly once across the full set of pages.
- [ ] The response is bounded — a project with many schema revisions cannot produce an arbitrarily large response.
- [ ] `limit` is supported, with a default and maximum consistent with the other list endpoints.
[...]

Issue #923, opened 2026-08-19
Return the latest revision per object type from listSchemas


vitorbari:
## Context

![Schema directory](https://github.com/user-attachments/assets/676d5095-5605-4e5e-9f5a-ca480e00e6b6)

The screen shows one row per schema — `Minimal`, `Consumer`, `Business`.

Schemas are immutable and versioned by URL: editing one mints a new row sharing its `objectType`. Flows pin a revision by URL, and revision proliferation was accepted as the cost of that design (`internal/domain/flow_definition_validator.go:155-156`).

`GET /schemas` returns one row per revision, so the screen shows one row per *edit* — a project that has revised `Consumer` four times renders `Consumer` four times. The same endpoint is also the revision-history reader: `apps/cli/src/commands/schemas/list.ts` exists to page through revisions and depends on getting all of them.

`object_type` is also nullable, and schemas registered by URL currently persist without one (#812).

## Description

Two readers want different cardinality from one endpoint. The directory wants current state — the newest revision of each schema. Audit, export and the CLI want history. Only the second is expressible today.

Both should be, and which one is being asked for should be explicit in the request rather than inferred from whether another filter is set. Inference would make two legitimate queries impossible — "every revision of everything" and "the current revision of one schema" — and would leave a pagination cursor ambiguous, since the row set would change shape with an unrelated parameter.

"Current" also has to be deterministic: revisions can share a creation timestamp, so ordering by time alone picks no stable winner. Branding settled the same problem with an explicit tiebreak (`internal/storage/branding/list.go:8-18`).

### Suggestion

A boolean `latest_per_object_type` (default `false`), orthogonal to `object_type`. All four combinations stay meaningful:

| `object_type` | `latest_per_object_type` | result |
| --- | --- | --- |
| unset | `true` | the schema directory screen |
| set | `true` | the current revision of one schema |
| unset | `false` | full history — export, audit, GitOps diff |
| set | `false` | revision history of one schema (today's behaviour, what the CLI uses) |

Defaulting to `false` keeps the CLI working unchanged and means the endpoint never silently hides rows. The name is a suggestion — what matters is that the choice is a request parameter of its own.

Rows with no `object_type` have nothing to group by; #812 covers giving the remote path its metadata, and until then they need to pass through rather than be dropped.

## Acceptance Criteria

- [ ] A caller can list the current revision of every schema in a project in one request, getting one row per schema.
- [ ] A caller can still list full revision history, either for all schemas or for one.
[...]

Issue #922, opened 2026-08-19
Filter schemas by kind in listSchemas


vitorbari:
## Context

![Schema directory](https://github.com/user-attachments/assets/676d5095-5605-4e5e-9f5a-ca480e00e6b6)

The directory lists user schemas and only user schemas — it sits under `Users → User schemas`. `GET /schemas` filters on `object_type` and nothing else.

`kind` is not queryable. It is not a column (`json_schemas` is `project_id, url, object_type, created_at, payload`), there is no matching field in `internal/domain/json_schema.go`, and it exists only inside the stored payload.

The meta-schema already settles what a *stored* kind is: `api/openapi/endpoints/schemas/user-schema.json` defines `kind` as `const: "user-schema"` and requires it (`:5-9`, `:28-31`). So there is exactly one stored kind today, and **`schema-url` is not one of them** — it is a discriminator on the create request. `CreateSchemaByUrl` persists the fetched remote document, which carries `kind: user-schema` itself.

**Out of scope:** schemas fetched by URL, and `$ref` targets pulled in during resolution, are persisted without `kind` or `object_type` — that is #812, which also owns the root-versus-fragment question. Those rows simply carry a NULL kind and fall outside a `kind=user-schema` result, which is the behaviour the directory wants. Making them carry their metadata belongs to #812.

## Description

There is no way to ask for schemas of a particular kind. The directory gets away with it because only one kind is ever stored — the screen is scoped to user schemas by coincidence rather than by contract, and the moment a second kind exists the list is wrong.

Now that the meta-schema settles the enum, the work is to promote `kind` from something buried in a JSON payload to a stored, indexed, queryable property of a schema.

Keeping the column nullable means this lands independently of #812 and improves automatically when #812 populates metadata on the remote path.

## Acceptance Criteria

- [ ] A caller can request only the schemas of a given kind, and gets exactly those.
- [ ] The set of kinds a *stored* schema can have is documented, including explicitly that `schema-url` is not one.
- [ ] Filtering stays usable as a project's schema count grows; it does not degrade into scanning every stored payload.
- [ ] Rows with no recorded kind (see #812) are excluded from a kind-filtered result rather than erroring or leaking in.
- [ ] The console schema directory returns user schemas only, by contract rather than by coincidence.

Related: #921, #923. Remote-path metadata is #812.

Issue #921, opened 2026-08-19
listSchemas should return the full schema document


vitorbari:
## Context

![Schema directory](https://github.com/user-attachments/assets/676d5095-5605-4e5e-9f5a-ca480e00e6b6)

Each row renders the schema's name, sign-in methods, attribute names, creation date and id. Everything but the last two lives inside the schema document.

`GET /schemas` returns only `{ id, created_at }`, so both consumers fetch each document separately — one request per row, both commented as workarounds:

- `apps/console/src/routes/_authed/schemas/index.tsx:34-44`
- `apps/console/src/components/add-user-sheet.tsx:127-140`

The server already reads the full payload out of the database and discards it (`internal/service/schema_service.go:157`).

Users solved the analogous problem once: `user.yaml` is a single representation served by both `GET /users/{id}` and `GET /users`, keeping the customer-authored document nested in `attributes` so schema-defined keys cannot collide with the envelope (`api/openapi/endpoints/users/user.yaml:6-13`). Schemas does neither — get-by-id returns the bare document with no id or timestamp, list returns the envelope with no document.

## Description

Listing schemas cannot render a schema, so every screen that lists them fans out a request per row. The list response should carry what a row needs in one request, and the two read endpoints should converge on one description of a schema rather than the two disjoint halves they give today.

The document is customer-authored and open (`additionalProperties: true`), so whatever shape is chosen has to keep server-owned fields safe from collision and keep the resource id distinguishable from the document's own `$id`. Left open: what the nested field is called, and whether `object_type` and `kind` are promoted alongside it.

## Acceptance Criteria

- [ ] A single `GET /schemas` call returns everything the directory row renders — name, sign-in methods, attribute names, creation date, id.
- [ ] `GET /schemas/{id}` and `GET /schemas` describe a schema the same way, as one representation rather than two.
- [ ] Server-owned fields cannot be shadowed by a key in a customer-authored schema document, and the resource id stays distinguishable from the document's `$id`.
- [ ] The console renders the schema directory and the add-user schema picker with no per-row follow-up request; the N+1 loops in both files above are deleted.
- [ ] `apps/cli/src/commands/schemas/list.ts` still works against the new response.

Related: #922, #923, #924. Without #923 this list shows one row per revision rather than one per schema; #924 shares the response envelope change.

PR #930, opened 2026-08-20
feat: return the full schema document from the schema read endpoints

IAM-marco:
## Summary

Second PR of the three-PR stack for #921 (listSchemas should return the full schema document). Stacked on #929.

`GET /schemas/{id}` and `GET /schemas` described two disjoint halves of a schema: the bare customer document with no id or timestamp, and a bare `[{id, created_at}]` array with no document. Both now serve one representation, mirroring the user envelope (ADR 052):

```json
{
  "id": "sch_01K...",
  "schema": { "kind": "user-schema", "objectType": "human-user", "properties": { ... } },
  "metadata": { "created_at": "2026-08-14T10:49:43Z" }
}
```

- New `schema.yaml` / `schema-document.yaml` / `schema-metadata.yaml` specs; `GET /schemas` wraps rows in `{schemas: [...]}` like every other list endpoint. The envelope keeps server-owned fields collision-safe against schema-defined keys, and the resource `id` stays distinguishable from the document's own `$id`.
- `internal/api/schema.go` gains the single `domainSchemaToApiSchema` mapper (analogous to `domainUserToApiUser`); both read handlers go through it.
- api-mock serves the envelope from its stored records.
- Consumers adapt **mechanically only**: the CLI list command reads `{schemas}` and `metadata.created_at` (stdout contract unchanged), the CLI syncers and setup write-back unwrap `.schema` so `.zitadel/schemas/*.json` GitOps files stay bare documents, and console call sites unwrap `.schema`. The console's per-row N+1 loops intentionally survive this PR — deleting them is the third PR, keeping this diff reviewable as the contract flip.

Left open on purpose: pagination (#924), latest-revision cardinality (#923), kind filtering (#922), and the `POST /schemas` response (stays `{id}`).

## Validation

- `moon run server:openapi`, `moon run server:generate` (no drift), `moon run server:vet`, `moon run server:test`, and `moon run server:test-postgres` pass — `TestGetSchema` and `TestSchemaRevisions` now assert the envelope (id echoes/mints, `metadata.created_at` set, document present on every list row).
- `moon run api:generate`, then `api:test`, `api-mock:test` (spec-conformance gained a `GET /schemas` envelope case), `console:test` (118/118), `cli:test`, and package typechecks pass.
- `corepack pnpm exec changeset status --since origin/main` passes.
- `api-mock:typecheck` turned out to be a pre-existing landmine this PR stepped on: pnpm resolves two peer-hashed vitest instances, and tsc treats their identical types as unrelated as soon as `vitest.config.ts` is typechecked. It never surfaced before because moon only runs the task for PRs touching `packages/api-mock`. Fixed here by dropping the config from the typecheck include, matching `components` and `storybook` (second commit).

## Release notes / changeset

`.changeset/schema-read-envelope.md` — `@zitadel/server` minor, `@zitadel/cli` patch: the schema read endpoints now return the `{id, schema, metadata}` envelope and the list wraps rows in `{schemas: [...]}`; clients reading the bare document or bare array must unwrap.

## Notes

- Stack: #929 (service refactor) → **this PR** (contract flip) → console N+1 deletion (next).
- Nested field name `schema`, no promotion of `object_type`/`kind`, and the list-object flip were confirmed with Marco against the ticket's open questions.
- Also updates the stale bare-array note in `api/openapi/readme.md` and fixes the missed `listSchemas` stub in `auth-guard.spec.tsx`.

### DIFF
--- a/api/openapi/endpoints/schemas/schema.yaml
+++ b/api/openapi/endpoints/schemas/schema.yaml
@@ -1,0 +1,29 @@
+title: Schema
+description: |
+  A schema resource: the server-owned envelope around a customer-authored
+  JSON Schema document.
+
+  `schema` is the customer-authored document, served verbatim: its keys and
+  contents are defined entirely by its author. The rest of the object — `id`,
+  `metadata` — is the server-owned envelope. Keeping the two apart means the
+  document may declare any property (including `id` or `metadata`) without
+  colliding with the envelope, and the resource `id` stays distinguishable
+  from the document's own `$id`.
+type: object
+required:
+  - id
+  - schema
+  - metadata
+properties:
+  id:
+    readOnly: true
+    type: string
+    description: |
+      The resource id: the server-minted `sch_*` identifier, or the
+      customer-supplied `$id` URI when the document declared one at creation.
+    example: sch_01H455VB4213EF62K49QF6M73F
+  schema:
+    $ref: schema-document.yaml
+  metadata:
+    readOnly: true
+    $ref: schema-metadata.yaml
--- a/.changeset/schema-read-envelope.md
+++ b/.changeset/schema-read-envelope.md
@@ -1,0 +1,6 @@
+---
+"@zitadel/server": minor
+"@zitadel/cli": patch
+---
+
+`GET /schemas` and `GET /schemas/{id}` now return each schema as an `{id, schema, metadata}` envelope carrying the full customer-authored document, and `GET /schemas` wraps its rows in a `{schemas: [...]}` object. The two read endpoints share one representation; the resource `id` and `metadata.created_at` are server-owned and can no longer collide with keys in the document. Clients that read the bare document from `GET /schemas/{id}` or the bare `{id, created_at}` array from `GET /schemas` must unwrap the envelope.
--- a/api/generated/oas_json_gen.go
+++ b/api/generated/oas_json_gen.go
@@ -32321,118 +32321,6 @@ func (s *GetSchemaByIdNotFound) UnmarshalJSON(data []byte) error {
 	return s.Decode(d)
 }
 
-// Encode encodes GetSchemaByIdOK as json.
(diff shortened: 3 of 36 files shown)


api/openapi/endpoints/schemas/schema.yaml
```diff
@@ -0,0 +1,29 @@
...
+properties:
+  id:
+    readOnly: true
+    type: string
+    description: |
+      The resource id: the server-minted `sch_*` identifier, or the
+      customer-supplied `$id` URI when the document declared one at creation.
+    example: sch_01H455VB4213EF62K49QF6M73F
+  schema:
+    $ref: schema-document.yaml
```
vitorbari, line 26: This envelope pattern works well, but we should still allow the actual raw schema to be retrieved.

We can say the `GET /schemas` and `GET /schemas/{id}` return the "entities" (schema + metadata), but if you want to access the actual raw schema, we could offer another route e.g. `GET /schemas/{id}`. wdyt?

Accessing the raw schemas is useful when validating a json object against the schema.

We can tackle this in a different PR if you agree.

vitorbari (approved):

api/openapi/endpoints/schemas/schema.yaml
IAM-marco: Agreed on doing it separately, filed as #944.

One clarification: the document under `schema` is already the complete customer-authored schema, so validating a JSON object against it only requires unwrapping the field (that is what the CLI syncers do when writing `.zitadel/schemas/*.json`). And the pre-envelope `GET /schemas/{id}` was never byte-for-byte raw either: the handler already unmarshalled into the generated type and re-serialized, so key order was unstable before this PR.

What a dedicated raw route genuinely adds, and what #944 captures: standard JSON Schema tooling that dereferences a schema's URL/`$id` expects the bare document at that address (it would now get the envelope), and serving the stored bytes verbatim would also give byte fidelity by skipping the re-serialization.

Issue #812, opened 2026-08-11
[draft] bug(schema): a schema fetched by URL is stored without meta-schema


vitorbari:
## What happens

There are two ways a user schema gets persisted, and only one of them is gated.

`SchemaService.CreateSchema` (`internal/service/schema_service.go:52`) — the posted-body
path — runs both gates:

```go
model, err := domain.NewJSONSchema(input.ProjectID, input.Schema)   // :53
err = s.schemaValidator.ValidateAgainstMetaSchema(input.Schema)     // :58
```

`SchemaService.CreateSchemaByUrl` (`:94`) calls `schemaResolver.Resolve` and neither.

Resolution persists whatever it fetches. `JSONSchemaResolver.getFromDatabase`
(`internal/domain/json_schema.go:383`) writes a bare struct literal:

```go
dbSchema = &JSONSchema{
    ProjectID: projectID,
    URL:       schemaURL,
    Schema:    data,
}
if err := store.CreateJSONSchema(ctx, dbSchema); err != nil {
```

This is also reachable from the *validated* path: `CreateSchema` passes the root bytes to
`Resolve`, so the root is not refetched — but every `$ref` target is loaded through
`loadSchemaPayload` → `getFromDatabase` and persisted the same way. A validated root can
pull in unvalidated documents.

## What the unvalidated path skips

Against `NewJSONSchema` (`json_schema.go:67`):

- the reserved-name guards — `properties.id` and `properties.metadata` are rejected there (`:75-81`)
- `URL` is taken from the fetch URL, not the document's `$id`, so a document whose `$id`
  disagrees with where it was fetched is stored under the fetch URL
- `ObjectType` is left `nil`; `ListSchemas` filters on it (`schema_service.go`)
- `CreatedAt` is left zero

Against `ValidateAgainstMetaSchema` (`internal/domain/schema_validator.go:106`): the whole
dialect — required `x-auth-methods`, `kind` const, the `x-unique` enum, the property-name
rule from #783.

## The design question to settle first

`ValidateAgainstMetaSchema` dispatches on the document's `metaSchema` property and returns
`ErrMissingSchemaID` when it is absent (`schema_validator.go:112-115`). A `$ref` target —
a shared address fragment, say — legitimately has no `metaSchema`, so "validate everything
fetched" would reject valid sub-schemas.

So the rule needs to separate a *root* schema (carries `kind`/`metaSchema`) from a
*fragment* pulled in by `$ref`. Options:

[...]

PR #932, opened 2026-08-20
feat(api): filter schemas by kind on GET /schemas

### LINKED ISSUE #922: Filter schemas by kind in listSchemas

vitorbari:
Closes #922.

## Summary

Screenshot 2026-08-20 at 20 12 21

Lets a caller list only the user schemas in a project instead of every schema document stored, which is what the console's User schemas screen needs.

- Added `domain.JSONSchemaKind`, an enum mapped from the document's `kind` property.
- Declared it as a `NOT NULL` column on the `json_schemas` table in Postgres, Spanner and SQLite.
- Added a `kind` query parameter to `GET /schemas`.
- Pointed both console call sites — the schema directory and the add-user picker — at `kind=user-schema`.
- Gave the `schema.created` audit event a payload carrying the kind and object type, replacing an empty one.

## Notes

A schema persisted without its document being parsed — ingested by URL, or a `$ref` target pulled in during resolution — is recorded as `unknown`, which no filter matches. #812 covers giving that path its metadata; the remote path is otherwise untouched here.

Still out of scope, each tracked separately: #921 (the list returns only `{ id, created_at }`), #923 (one row per revision rather than per schema), #924 (`page_token` and `offset` declared but ignored).


internal/storage/dialect/spanner/migration/sql/000003_json_schemas.sql
```diff
@@ -5,6 +5,7 @@ CREATE TABLE json_schemas (
     project_id  STRING(MAX) NOT NULL,
     url         STRING(MAX) NOT NULL,
     object_type STRING(256),
+    kind        STRING(256) NOT NULL,
```
vitorbari, line 8: I'm editing the original migration because there is no production DB yet, otherwise spanner would reject the NOT NULL column.

internal/storage/dialect/postgres/migration/sql/000003_json_schemas.sql
```diff
@@ -4,6 +4,7 @@ CREATE TABLE zitadel_nextgen.json_schemas (
         REFERENCES zitadel_nextgen.projects (id) ON DELETE CASCADE
     , url         TEXT COLLATE "C" NOT NULL
     , object_type TEXT COLLATE "C"
+    , kind        TEXT COLLATE "C" NOT NULL
     , created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
     , payload    JSON        NOT NULL
```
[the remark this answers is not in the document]
vitorbari: I'm editing the original migration because there is no production DB yet, otherwise spanner would reject the NOT NULL column.

internal/storage/dialect/spanner/migration/sql/000003_json_schemas.sql
```diff
@@ -5,6 +5,7 @@ CREATE TABLE json_schemas (
...
     created_at  TIMESTAMP   NOT NULL DEFAULT (CURRENT_TIMESTAMP()),
     payload     JSON        NOT NULL,
```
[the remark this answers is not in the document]
vitorbari: I'm editing the original migration because there is no production DB yet, otherwise spanner would reject the NOT NULL column.

internal/domain/event_payload.go
```diff
@@ -105,6 +105,26 @@ type ClaimCompletedPayload struct {
 	TeamID string `json:"team_id,omitempty"`
 }
 
+// SchemaCreatedPayload is the allowlisted snapshot for schema.created. The
+// document itself is omitted — it is customer-authored and unbounded, and the
+// event's entity_id already identifies which schema this was.
+type SchemaCreatedPayload struct {
```
IAM-marco, line 111: This gives `schema.created` a real payload (`kind`, `object_type`) where it previously carried an empty one, but `docs/design/api/events-catalog.md` still documents the event as empty in both places: the Schema row of the payload-contents table ([events-catalog.md#L52](https://github.com/zitadel/nextgen/blob/69efe4521e0f5460810c623c9b77929a2e60d35e/docs/design/api/events-catalog.md#L51-L53)) and the Path B table row for `schema.created` ([events-catalog.md#L119](https://github.com/zitadel/nextgen/blob/69efe4521e0f5460810c623c9b77929a2e60d35e/docs/design/api/events-catalog.md#L118-L120)).

That file is the declared authority for payload shapes: [internal/service/AGENTS.md#L15-L18](https://github.com/zitadel/nextgen/blob/69efe4521e0f5460810c623c9b77929a2e60d35e/internal/service/AGENTS.md#L14-L20) says a change is incomplete if it "changes a payload so it no longer matches the catalog", and the comment at the top of this file (lines 8-9) says to keep it in sync with the catalog. Anyone reading the catalog to decide what audit data exists will now get a wrong answer for this event.

Fix: update both catalog rows to list `kind` and `object_type` (the document body staying excluded is worth keeping in the row). The OpenAPI payload schema, the event type YAML, and the `gen_event_schemas` mapping were all updated in this PR, so the catalog is the only piece left out of sync.

IAM-marco (changes requested):

internal/storage/dialect/sqlite/migration/sql/000001_init.sql
```diff
@@ -36,6 +36,7 @@ CREATE TABLE json_schemas (
     project_id  TEXT    NOT NULL,
     url         TEXT    NOT NULL,
     object_type TEXT,
+    kind        TEXT    NOT NULL,
```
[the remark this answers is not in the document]
vitorbari: This comment explains the reason: https://github.com/zitadel/nextgen/pull/932#discussion_r3824237312

internal/domain/event_payload.go
vitorbari: Fixed in b97cf0b1b — both rows now list `kind`, `object_type`, keeping the document body in the intentionally-omitted column.

vitorbari: Fixed: b97cf0b1b2ea7b373dba57d7b0216710896dbcd2

IAM-marco (approved):
Catalog rows updated in b97cf0b1 as requested; both tables now match the shipped schema.created payload.

PR #934, opened 2026-08-20
ci: wire the spanner real-instance lane for a one-off predicate bench…

IAM-marco:
**DIAGNOSTIC, DO NOT MERGE.** One-off benchmark: does `TestListAuthzTeamScopedOnlyPartialView` run fast on real Cloud Spanner? It costs **64.58s on the CI emulator lane vs 0.15s on postgres** in the same job, and ~19 minutes locally. This PR restores the pre-#795 real-instance wiring to get the real-Spanner number. The lane is **inert until an admin creates the `SPANNER_TEST_INSTANCE` Actions variable**; until then CI runs the emulator lane unchanged.

## Summary

The cost of this test scales super-linearly with the rows the rest of the package leaves in the shared database: 11s solo, 58s with 5 row-creating tests, 1127s (18m47s) with the full package at pre-#912 commit 401655b7. The suspect is the team-scoped EXISTS list predicate (`WriteListAuthzExistsPredicate`, `internal/storage/dialect/authz/resolver_sql.go`, from #811/#894/#896) under the emulator's naive planner. #912 did not introduce the pathology; its extra tests pushed the package past the timeout locally.

The real instance answers the one open question:

- **Fast (seconds):** the emulator planner is the multiplier. Test-environment concern; also an argument to keep #793 open as insurance.
- **Slow (tens of seconds or worse):** the predicate is intrinsically expensive. Production performance issue for Spanner-backed deployments; priority changes entirely.

<details>
<summary><h3>Activation prerequisites (needs an admin)</h3></summary>

1. Create the Actions variable `SPANNER_TEST_INSTANCE` = `projects/<project>/instances/<instance>` (repo or org level). The gate keys off it; nothing runs differently until it exists.
2. The auth step keeps the original names `secrets.GCP_WORKLOAD_IDENTITY_PROVIDER` / `secrets.GCP_SERVICE_ACCOUNT` (unchanged from the pre-#795 wiring). Heads-up: neither is visible today at repo or org level via the API, while `GCP_WIF_PROVIDER` / `GCP_WIF_SERVICE_ACCOUNT` exist as repo variables and look like renamed successors. If the secrets were rotated away, either recreate them or swap the auth step to the variables (one-line change on this branch).
3. Auth hard-fails when the variable is set but WIF is broken; no silent fallback (by design, from #795). The GCP side needs the service account to hold Spanner database create/drop on the instance.

</details>

<details>
<summary><h3>Reading the result</h3></summary>

In the `full-pr` run, open the step **Run Go integration tests (spanner test instance)** and find `--- PASS: TestListAuthzTeamScopedOnlyPartialView (Xs)`. Compare against 64.58s on the emulator lane in the same runner class (e.g. run 32401568967).

The run provisions a uniquely named `itest_*` database and drops it on teardown; a crashed run can leave one to clean up (traceable via run id, see `internal/storage/testdb/instance.go`).

</details>

## Validation

- Workflow YAML parses; the wiring is the pre-#795 shape restored verbatim (auth step, guarded emulator step, guarded real-instance step, `id-token: write`), plus an explanatory comment.
- With the variable unset, the auth step skips, the guard short-circuits, and the emulator lane runs exactly as on `main`.

## Release notes / changeset

None: CI-only, and this branch is never merged.

## Notes

Close without merging once the numbers are in, and delete the `SPANNER_TEST_INSTANCE` variable to return CI to the emulator-only state. Full local investigation write-up will land in a follow-up issue.


IAM-marco:
Baseline from this branch's first run ([32413497920](https://github.com/zitadel/nextgen/actions/runs/32413497920), variable unset, emulator lane): `TestListAuthzTeamScopedOnlyPartialView` **11.14s** (postgres in the same job: 0.20s). The emulator number swings between runs and branches: 64.58s on run 32401568967 (branch `worktree-schema-kind-filter`) earlier the same day, so when the real-instance arm runs, compare against the same-branch 11.14s rather than the body's 64.58s. The local macOS full-package number for the same test stays ~19 minutes, so the environment gap is large; the real-instance run remains the discriminator for whether the predicate itself is expensive.

IAM-marco:
**Update: the real-instance lane has never run, so activation is provisioning work, not a revert.** The gate `vars.SPANNER_TEST_INSTANCE != ''` was already false before #795 unwired the lane: sampled runs under the pre-#795 wiring on 2026-08-02 ([30755061992](https://github.com/zitadel/nextgen/actions/runs/30755061992)) and 2026-08-07 ([31190213827](https://github.com/zitadel/nextgen/actions/runs/31190213827)) both show `Authenticate to Google Cloud (Spanner test instance): skipped` and `Run Go integration tests (spanner test instance): skipped`, identical to this branch's run 32413497920. #629 merged the wiring with the GitHub/GCP setup explicitly out of scope ('One-time infra prerequisite (not in this PR)'), and that prerequisite was never completed, so there are no dormant credentials to revive and CI has been emulator-only in practice the whole time.

Activating this branch is therefore the #629 checklist:

1. A small Spanner instance in a GCP project (none was ever wired up; whether one exists is a GCP console question).
2. A service account with Spanner Database Admin plus Database User on it. Shortest path, if policy allows: reuse `github-deploy@zitadel-ops.iam.gserviceaccount.com` through the existing `github-actions` WIF pool (both already exist, per the repo's `GCP_WIF_*` Actions variables).
3. GitHub Actions config: the variable `SPANNER_TEST_INSTANCE = projects/<p>/instances/<i>`, plus either the secrets `GCP_WORKLOAD_IDENTITY_PROVIDER` / `GCP_SERVICE_ACCOUNT` (this branch's current wiring, same names as #629 specified) or a one-line swap of the auth step to the existing `GCP_WIF_PROVIDER` / `GCP_WIF_SERVICE_ACCOUNT` variables.

IAM-marco:
**Heads-up: run 32726396702's real-instance step did not actually run the tests.** Its log ends with `server:test-spanner (cached from remote, 141ms)`: `ZITADEL_TEST_SPANNER_INSTANCE` was not part of the moon task's input hash, so the real-instance step replayed the emulator lane's remote-cached result. The `TestListAuthzTeamScopedOnlyPartialView (32.47s)` line in that log is a replayed emulator measurement, not real Spanner; auth is proven working (thanks for the GCP_WIF_* switch and the instance), but the benchmark itself has not executed yet.

Fixed in 2081ce45 by declaring the variable as a `test-spanner` task input (same pattern as `ZITADEL_TELEMETRY_BUILD_CHANNEL` in apps/cli/moon.yml), which splits the hash in both directions, so a future real-instance result can also never poison the emulator lane. The push retriggers CI; the next 'spanner test instance' step should show a full `go test` run (minutes, not 141ms), and its `--- PASS: TestListAuthzTeamScopedOnlyPartialView (Xs)` line is the real number. Same-branch emulator baseline for comparison: 11.14s (run 32413497920, pre-merge-from-main) and main's current emulator lane around 32s.

IAM-marco:
Run [32743380406](https://github.com/zitadel/nextgen/actions/runs/32743380406) proves both cache fixes work: the suite now genuinely executes against GCP and fails on one precise, legible thing:

```
PermissionDenied: IAM permission 'spanner.databases.create' on resource
'projects/zitadel-dev/instances/adlerhurst'
```

Auth via WIF works; the service account just lacks the database-admin half of the #629 checklist. For whoever holds `zitadel-dev`: grant **roles/spanner.databaseAdmin** (create/drop databases, needed for the per-run `itest_*` databases) plus data access (**roles/spanner.databaseUser**) to the CI service account on the `adlerhurst` instance, noting the grant lands in `zitadel-dev` while the SA lives in `zitadel-ops`. After that, a re-run of this PR's jobs should finally produce the benchmark numbers.

IAM-marco:
Attempt 9 (thanks for the IAM grant and the re-run) produced the first genuine execution against real Spanner, and it surfaced a finding that stands on its own: **all three spanner packages died at the 10m go-test timeout inside `spanner/migration.Migrate`, before running a single test.** The goroutine dump pins it to migrations.go: goose runs with `WithIsolateDDL(true)`, so every DDL statement is its own `UpdateDatabaseDdl` long-running operation. Real Spanner takes seconds per LRO across the full migration stack; the emulator applies DDL instantly, which is why this was never visible. Beyond CI, this means a first boot of the server against real Cloud Spanner spends 10+ minutes in migrations as written; batching all statements into one `UpdateDatabaseDdl` call is the known fix and deserves its own ticket.

To still get the benchmark number, 3b7c279d scopes the real-instance step to the one package the question needs (`internal/api/integration_test`: the victim test plus the row-creating tests around it) with a 45m timeout: one migration pass instead of three, then the tests. The push retriggers CI.

IAM-marco:
**Benchmark complete: the predicate itself is the problem, not the emulator planner.** Run [32860917308](https://github.com/zitadel/nextgen/actions/runs/32860917308), scoped step, genuine execution (no cache lines, full migration pass, package total 2021.955s):

| Backend | TestListAuthzTeamScopedOnlyPartialView |
| --- | --- |
| Postgres (same job) | 0.14-0.20s |
| **Real Cloud Spanner (zitadel-dev/adlerhurst)** | **17.59s** |
| CI emulator (same branch baseline) | 11.14s |
| CI emulator (branch-dependent, main-era runs) | up to 64.58s |
| Local macOS emulator, full package | ~19 min |

Real Spanner lands squarely in the emulator's band, not in the low-seconds band that would have exonerated the query. The team-scoped EXISTS list predicate (`WriteListAuthzExistsPredicate`) costs ~100x postgres on identical, test-suite-sized data, and the local dose-response measurements (11s solo → 58s with 5 tests → 19 min with the full package) show it grows super-linearly with data volume, so production datasets make this worse, not better. Caveats: one data point, small test instance; but the shape of the result is structural, not capacity-related.

Three follow-ups fall out of this PR, each deserving a ticket:
1. **Authz list-predicate performance on Spanner** (this result): the EXISTS disjunction needs a Spanner-friendly rework for team-scoped-only principals.
2. **Migration DDL batching**: first boot against real Spanner spends 10+ min in `Migrate` because `WithIsolateDDL` issues one `UpdateDatabaseDdl` LRO per statement (previous comment).
3. **Go test-result caching in integration lanes**: setup-go's restored GOCACHE can replay integration results against a different backend; `-count=1` belongs on all three integration tasks (proven by run 32741899403).

This PR's purpose is fulfilled: it should be closed unmerged, and the `SPANNER_TEST_INSTANCE` variable deleted (or kept if the team wants the real-instance lane available while working on 1 and 2).

IAM-marco:
The three follow-ups are now tracked: #972 (Spanner-friendly rework of the team-scoped EXISTS list predicate, the headline result), #973 (batch migration DDL into one UpdateDatabaseDdl call), #974 (-count=1 for the integration lanes plus the task-input fix, both already proven on this branch as 9d0bd786 and 2081ce45). This PR can be closed unmerged whenever convenient; whether to delete `SPANNER_TEST_INSTANCE` or keep the lane for #972/#973 work is the team's call.

PR #398, opened 2026-06-24
refactor(components): extract field rendering into _fields.liquid partial and fix select enum mapping

peintnermax:
## Summary

Refactors the Liquid field rendering system into a dedicated partial and module, renames `mandatory-gates` → `field-patcher`, fixes `select` enum values being dropped by the Go API, and documents the real-server demo workflow.

### Field rendering refactor (`packages/components/src/orchestrator/`)
- **New `_fields.liquid` partial** — field rendering logic moved from `default.liquid` into a reusable partial with `{% case f.type %}` dispatch for `checkbox`, `select`, and default input types
- **New `liquid-fields.ts`** — all field-related filters (`fieldPlaceholder`, `fieldHelp`, `fieldError`, `selectOptions`, `formLevelError`) and the `{% fields %}` tag extracted from `liquid.ts`
- **`liquid.ts`** simplified to a thin wrapper that delegates field concerns to `liquid-fields.ts`
- **`mandatory-gates` → `field-patcher`** — renamed tag, file, and spec for clarity

### Bug fix: `select` enum values dropped by Go API (`internal/api/flow.go`)
- `toFlowFieldValidation` mapped `Format`, `MinLength`, `MaxLength` but silently dropped `Enum`. The OpenAPI spec already declared `validation.enum` — the mapper just never populated it.

### Default schema/flow (temporary, for testing)
- Added `gender` property with `enum` to `default-human-user-schema.json`
- Added `"gender"` to register step fields in `default-login-flow-definition.json`
- Added `register.field.gender` translations (en, de, it)

### Docs
- **AGENTS.md**: Documented real-server demo workflow and fixed mock server port (4000 → 8080)
- **`.env.example`**: Added `ZITADEL_PROJECT_SECRET` placeholder

## Validation

- 107 component tests passing (`moon run components:test`)
- Manual E2E with real Go server: register flow renders gender `<zl-select>` with enum options

## Release notes / changeset

Internal refactor + bug fix. Touches `@zitadel/components` — the `Enum` fix and field-patcher rename are patch-level.

PR #935, opened 2026-08-20
test(login): cover the select field branch in every branding design

fforootd:
## Summary

- The Liquid field-rendering loop (`{% for f in fields %}` with a `{% case f.type %}` dispatch over `checkbox` / `select` / default input) is duplicated byte-identically across six templates: the built-in `packages/components/src/orchestrator/templates/default.liquid` plus the five ejectable branding designs under `packages/config/defaults/branding/*/login.liquid`. The duplication is deliberate — each design is self-contained so `zitadel branding eject --design X` writes a standalone file (ADR 037, #563).
- The catalog parity test loops every design in `BRANDING_DESIGNS`, but its shared fixture declared only `email` (type `email`) and `remember` (type `checkbox`). The `select` arm — the one that renders `<zl-select options='{{ f.validation.enum | selectOptions | json }}'>` — was therefore never reached for the five branding designs; it was covered only through `default.liquid`'s own tests. A hand-edit that broke one design's select branch shipped green.
- Adds a required `select` field with a closed `validation.enum` to the fixture step and asserts, per design, that the rendered `<zl-select>` carries its localised label, its `required` flag, and `options` built from the enum. Existing email/checkbox assertions are unchanged.
- The `consumes its mandatory_gates marker` test gains a duplicate check for the new field: a *required* select must not trip the safety net into appending a generic `<zl-field>` alongside it. That regression was previously guarded only in isolation in `mandatory-gates.spec.ts`; it is now proven per design through the real pipeline.

No shared partial was extracted — that is a separate open design question, and the six copies stay as they are.

## Validation

- `moon run components:test` — 33 files, 425 tests passed (32 in `design-templates.spec.ts`, up from 27).
- `moon run components:typecheck components:lint` — clean; no new lint warnings attributable to the spec.
- `pnpm exec oxfmt --check packages/components/src/orchestrator/design-templates.spec.ts` — formatted.
- Mutation-tested to prove the assertion is not vacuous, restoring each template afterwards (`git status` clean, only the spec is modified):
  - dropped `options='…'` from `minimal/login.liquid` → exactly one failure, `minimal` only (`expected null to deeply equal [ Array(2) ]`);
  - deleted the whole `{% when 'select' %}` arm from `split/login.liquid` → exactly one failure, `split` only (`expected null not to be null` — the field falls through to the `<zl-field>` default arm).

## Release notes / changeset

- No changeset required — no shipped behavior changed. The PR is exclusively a `*.spec.ts` change (`.changeset/README.md#decision-table`, row 3), which is also why the title type is `test`.

## Notes

- The assertion parses the rendered HTML into a fragment and reads the decoded `options` attribute rather than string-matching it. The value passes through two escaping hops — Liquid runs with `outputEscape: "escape"`, so the `| json` output is HTML-encoded before the sanitiser's serialiser re-encodes it — and a `JSON.parse` round-trip is both stronger and immune to that churn.
- `options` is already on `zlSelectManifest.attrs`, so it survives the DOMPurify pass; the test would catch it if that ever changed.
- Fixture naming (`department`, enum `["Engineering", "Support"]`) matches the existing select fixture idiom in `embedding.browser.spec.ts`.
- Setup note for anyone reproducing locally in a fresh worktree: `moon run components:test` fails with `tsx/tsdown/orval: command not found` until `pnpm install --frozen-lockfile` has been run.

### DIFF
--- a/packages/components/src/orchestrator/design-templates.spec.ts
+++ b/packages/components/src/orchestrator/design-templates.spec.ts
@@ -17,6 +17,7 @@ const locale: Record<string, string> = {
   "identifier.title": "Sign in",
   "identifier.field.email": "Work email",
   "identifier.field.remember": "Remember me",
+  "identifier.field.department": "Department",
   "identifier.action.register.lead": "New here? ",
   "identifier.action.register.link": "Create an account",
   "submit.continue": "Continue",
@@ -29,6 +30,13 @@ const step: CreateFlow201Step = {
   fields: [
     { name: "email", type: "email", text_key: "identifier.field.email", required: true },
     { name: "remember", type: "checkbox", text_key: "identifier.field.remember" },
+    {
+      name: "department",
+      type: "select",
+      text_key: "identifier.field.department",
+      required: true,
+      validation: { enum: ["Engineering", "Support"] },
+    },
   ],
   actions: [
     { name: "submit", kind: "submit", text_key: "submit.continue", primary: true },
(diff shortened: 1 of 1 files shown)


fforootd:
relates to https://github.com/zitadel/nextgen/pull/398

PR #946, opened 2026-08-22
feat: wire up cursor pagination on listSchemas

### LINKED ISSUE #924: Wire up cursor pagination on listSchemas

IAM-marco:
Closes #924 (server half; the console/CLI consumers land in the stacked follow-up PR).

`GET /schemas` advertised `page_token` and `offset` but implemented neither: the service ignored both, applied no limit, and discarded the `NextCursor` the storage layer already returns. This PR wires the endpoint to the existing keyset machinery, so it pages like the other list endpoints, returns a bounded response, and stops advertising parameters it does not honour.

## What changed

<details open>
<summary><b>API contract</b></summary>

- `GET /schemas` takes `limit` (shared component: default 20, max 100) and `page_token`; the `offset` parameter is gone, and the now-unreferenced `components/parameters/offset.yaml` is deleted (ADR 027: no offset holdouts remain).
- `list-schemas-response.yaml` gains `next_page_token`, mirroring `list-users-response.yaml`.
- `req.invalid` was already in the endpoint's error catalog, so bad tokens needed no catalog change.
- ADR 027, `docs/design/api/conventions.md`, and `api/openapi/readme.md` no longer name `GET /schemas` as the remaining offset holdout.
</details>

<details open>
<summary><b>Service and handler</b></summary>

- `SchemaService.ListSchemas` now takes `ListSchemasInput{ProjectID, ObjectType, PageToken, Limit}` and returns `ListSchemasOutput{Items, NextPageToken}`, following `ListUsers`: `normalizeLimit`, cursor passthrough, `NextCursor` returned as the next page token.
- Ordering is `(created_at, url)` descending. `created_at` alone is not a total order (the table's key is `(project_id, url)` and timestamps can collide); with `project_id` fixed by the filter, `url` is a unique tiebreaker, so a page boundary can neither skip nor repeat rows.
- Malformed or order-mismatched tokens map to `req.invalid` (HTTP 400) through the existing `mapListError` path; nothing is silently ignored.
</details>

## Testing

- `internal/service/schema_service_test.go` (new): eight table cases with the full expected `ListOptions` passed to the mock expectation, so every field the service builds is validated per case (defaults, object-type filter, limit clamping, cursor passthrough, token emission, error mapping).
- `internal/storage/stmttest/cursor_battle_test.go`: `battleJSONSchemas` gains a `created_at_url_order` drain matching the service's real ordering, with the expected order derived from the timestamps the dialect scanned back at insert.
- `internal/api/integration_test/schema_list_test.go` (new): HTTP walk with `limit=1` (every full page carries a token, the page past the end is empty, pages cover the list in order exactly once) plus a 400 on a malformed token.
- Run locally: full Go unit suite, postgres integration (`stmttest` + `integration_test`), sqlite `stmttest`, `server:check-generate`, `workspace:check-openapi-rules`.

Changeset: `@zitadel/server` minor.


internal/api/schema.go
```diff
@@ -70,24 +70,27 @@ func (h *Handler) ListSchemas(ctx context.Context, params api.ListSchemasParams)
 	if err != nil {
 		return nil, err
 	}
-	schemas, err := h.schemaService.ListSchemas(ctx,
-		string(params.ProjectID),
-		params.ObjectType.Value,
-		params.Offset.Value,
-		string(params.PageToken.Value),
-	)
+	schemas, err := h.schemaService.ListSchemas(ctx, service.ListSchemasInput{
```
vitorbari, line 84: Can you wait for me to merge #932?
There will be a conflict in this file to solve: https://github.com/zitadel/nextgen/pull/932/changes#diff-53624ed2cc1585a8a8bac6914ec3a9dd36d2947468fbd5c14c2f78b41fc9094fR87

vitorbari: The service layer will also conflict

vitorbari (changes requested):

internal/api/schema.go
vitorbari: #932 was merged

vitorbari (approved):

PR #957, opened 2026-08-24
feat(api): return the latest revision per object type from listSchemas

### LINKED ISSUE #923: Return the latest revision per object type from listSchemas

vitorbari:
Closes #923.

## Summary

`GET /schemas` returned one row per revision. Schemas are immutable and versioned by URL, so editing one mints a new row sharing its `objectType` — the console's **User schemas** directory rendered one row per *edit*, while `zitadel schemas list` depends on getting the full history.

Adds a `revisions` parameter — `all` (the default, so the CLI is untouched) or `latest` — so which cardinality is being asked for is explicit in the request rather than inferred from another filter. The console directory and the add-user schema picker now pass `latest`.

## Decisions worth reviewing

- **A unique index on `(project_id, object_type, created_at)`.** Double duty: it makes `created_at` a total order *within* an object type, so "the latest" is decidable, and it is the seek the anti-join needs. Spanner needs `NULL_FILTERED`, or every row without an `objectType` collides on `(project_id, NULL, created_at)` — the trap `000016_checks_null_filtered_index.sql` exists to fix elsewhere.
- **`NOT EXISTS`, not `created_at = (SELECT MAX(…))`.** The `MAX` form reads tidier and uses the index just as well, but silently drops rows with a NULL `object_type`: the correlation is NULL, `MAX` yields NULL, `= NULL` is NULL. Those rows must pass through, and #812 keeps producing them.
- **No CTE or window function.** `compileList` grew an optional trailing conjunct, so the statement stays a plain single-table `SELECT` that the keyset cursor, the authz predicate, `ORDER BY` and `LIMIT` all still apply to. Existing callers untouched.
- **Page tokens carry their mode.** Both modes must sort by the same columns, or a page boundary would tie-break differently on either side of it — which means `MatchesOrderBy` can't tell the two apart. A cross-mode token is rejected with `req.invalid`.
- **New `sch.revision_conflict` (409).** A create that loses the `(object_type, created_at)` race answered `sch.already_exists` — wrong: the URL was free, the timestamp was not. Only Postgres reports which constraint fired, so the two are told apart by reading the URL back.
- `getSchemaById` no longer claims to return "the default revision of the schema" — it is a lookup by URL returning one row.

## Notes

- **Parameter naming.** The issue suggested a boolean `latest_per_object_type` and said the name was a suggestion. This ships `revisions=all|latest`: there is no boolean query parameter in the spec today (the house style is a string enum — `purpose`, `order`, `kind`), and it leaves room for a third mode. Easy to rename before merge.
- **The sub-query repeats none of the caller's filters.** So a `kind` filter that excludes the newest revision makes that schema disappear rather than falling back to its newest *matching* revision, and a newer revision the caller cannot see could suppress an older one they can. Both unreachable today, both one extra conjunct. Worth revisiting under #812.
- **The add-user picker is not named in the acceptance criteria.** Changed anyway — same reader, and a superseded revision is not something a new user should be created against. Flagging rather than folding it in silently.
- Migrations are edited in place rather than appended, matching how #932 added the `kind` column pre-production.
- The query-plan check ran on SQLite only (`SEARCH newer USING COVERING INDEX idx_json_schemas_object_type_revision (project_id=? AND object_type=? AND created_at>?)`); Postgres and Spanner need Docker, so CI covers them.


packages/api-mock/src/platform-handlers.ts
```diff
@@ -702,10 +728,16 @@ export function setupPlatformHandlers() {
...
+      const kindFilter = url.searchParams.get("kind") ?? undefined;
+      const matching = [...store.schemas.values()]
         .filter((r) => r.projectId === projectId)
         .filter((r) => !objectTypeFilter || r.objectType === objectTypeFilter)
+        .filter((r) => !kindFilter || schemaKind(r.body) === kindFilter)
         .sort((a, b) => (a.createdAt < b.createdAt ? 1 : -1));
+      const records =
+        url.searchParams.get("revisions") === "latest"
+          ? latestSchemaRevisions(matching)
+          : matching;
```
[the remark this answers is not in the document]
vitorbari: Both taken in 1286ef0.

Comparator: agreed, and it matters more than it looks — `createdAt` is millisecond-resolution here with no unique index behind it, so a tie is reachable in the mock in a way it is not on the server, and `latestSchemaRevisions` reads the first record of an `objectType`. Now ordered by `createdAt` then `id`, which is what the server lists by.

Filter order: also right, and the asymmetry is worth recording — only a row sharing an `object_type` can correlate in the anti-join, so narrowing by `object_type` before selecting the latest stays equivalent, while narrowing by `kind` does not. `kind` now filters after. Unreachable through the API today (the only kind the create path produces is `user-schema`, and unparsed rows carry a NULL `object_type`), but #812 makes it reachable, and the same gap is called out on the server side in the PR description.

internal/storage/dialect/postgres/json_schema.go
```diff
@@ -63,6 +63,19 @@ func (js jsonSchemaStatements) DeleteJSONSchemaByID(ctx context.Context, project
...
 
+// latestRevisionPerObjectType keeps only the newest revision of each
+// object_type. It is an anti-join rather than `created_at = (SELECT MAX(…))`
+// because the correlation on a NULL object_type is NULL either way: MAX then
+// yields NULL and the row is filtered out, while NOT EXISTS passes it through,
+// which is what a row that is a revision of nothing deserves.
+//
+// The uniqueness of (project_id, object_type, created_at) makes created_at a
+// total order within an object_type, so no tiebreak belongs in here.
+const latestRevisionPerObjectType = `NOT EXISTS (SELECT 1 FROM zitadel_nextgen.json_schemas newer` +
```
IAM-marco, line 82: This new anti-join decides "a newer revision exists" against the whole json_schemas table, before authorization is applied: `compileList` appends the per-resource authz filter after the conjuncts (https://github.com/zitadel/nextgen/blob/1286ef0adc7c5aa5490c53f6db0d31dba541a795/internal/storage/dialect/postgres/compiler.go#L75-L77), and that filter scopes rows by schema URL (https://github.com/zitadel/nextgen/blob/1286ef0adc7c5aa5490c53f6db0d31dba541a795/internal/storage/dialect/authz/resolver_sql.go#L99-L113), so two revisions of the same object type can differ in visibility for a team-scoped caller.

What follows: when such a caller may see an older revision but not the newer one, the subquery still finds the newer row and suppresses the older one, so `revisions=latest` returns nothing for that object type while `revisions=all` still returns the older revision. The same predicate is repeated in the spanner and sqlite dialects.

If "global latest regardless of visibility" is the intended semantics, this deserves a note on the const; otherwise the subquery needs the same authz scoping as the outer query (latest among visible rows).

Coverage note: the new integration test and the compile tests all run with unrestricted authz, so the partial-grant interaction is untested either way; a test with a team-scoped caller and a newer ungranted revision would pin the intended behavior.

internal/storage/stmttest/main_test.go
```diff
@@ -33,22 +34,28 @@ func registerDialect(
...
-	seedTiedProjects  func(ctx context.Context, ids []string, createdAt time.Time) error
-	hardDeleteTeam    func(ctx context.Context, projectID, teamID string) error
-	schemaNullability []schematest.ColumnNullability
-	liveNullability   func(ctx context.Context) (map[string]map[string]bool, error)
+	name             string
+	stmts            service.AllStatements
+	seedTiedProjects func(ctx context.Context, ids []string, createdAt time.Time) error
+	hardDeleteTeam   func(ctx context.Context, projectID, teamID string) error
+	// insertJSONSchemaAt writes a json_schemas row at an exact created_at, which
+	// CreateJSONSchema cannot: every dialect takes that timestamp from the
```
IAM-marco: This new comment says every dialect takes `created_at` from the database, but SQLite generates it in Go: `CreateJSONSchema` binds `nowUnixNano()` (https://github.com/zitadel/nextgen/blob/1286ef0adc7c5aa5490c53f6db0d31dba541a795/internal/storage/dialect/sqlite/json_schema.go#L45-L47, https://github.com/zitadel/nextgen/blob/1286ef0adc7c5aa5490c53f6db0d31dba541a795/internal/storage/dialect/sqlite/time.go#L18-L21), and the sqlite column has no DEFAULT. Only Postgres and Spanner default it in the database.

The helper is still needed for all three dialects (no `CreateJSONSchema` accepts a caller timestamp), so the code is fine; only the stated reason is wrong for SQLite. The same claim is repeated in https://github.com/zitadel/nextgen/blob/1286ef0adc7c5aa5490c53f6db0d31dba541a795/internal/storage/dialect/sqlite/seed_projects_integration.go#L48-L50, directly beside the contradicting implementation.

Since the tied-timestamp conflict semantics in this PR depend on whose clock stamps the row (database for Postgres/Spanner, application for SQLite), suggest rewording both places to: CreateJSONSchema never accepts a caller-supplied created_at (Postgres and Spanner default it in the database, SQLite stamps it in Go).

api/openapi/endpoints/schemas/methods.yaml
```diff
@@ -93,6 +93,27 @@ get:
...
+      description: |
+        Which revisions to return. Schemas are immutable and versioned by URL,
+        so editing one mints a new row sharing its `objectType`.
+
+        `all` (the default) returns every revision — the full history an
+        export, an audit or a GitOps diff needs. `latest` returns the newest
+        revision of each `objectType`, which is one row per schema.
+
+        A schema stored without an `objectType` is a revision of nothing, and
+        is returned by both.
```
IAM-marco, line 107: How is this possible and why this exist?

internal/storage/dialect/postgres/json_schema_compile_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/zitadel/nextgen/internal/domain"
+	"github.com/zitadel/nextgen/internal/service"
+	"github.com/zitadel/nextgen/internal/storage/database"
+)
+
+// Latest mode adds one WHERE conjunct and nothing else: the statement stays a
+// plain single-table SELECT, so the keyset predicate, the authz predicate,
+// ORDER BY and LIMIT all keep applying to it unchanged.
+func TestCompileListJSONSchemasLatestRevision(t *testing.T) {
```
IAM-marco: Why a separate test file if the tested function is in `compiler.go`?

IAM-marco: Same also for the other dialects

internal/storage/stmttest/cursor_battle_test.go
```diff
@@ -356,11 +357,55 @@ func battleJSONSchemas(t *testing.T, d dialect) {
...
+	orderAsc := database.OrderBy[domain.JSONSchemaField]{
+		Columns: []database.Column[domain.JSONSchemaField]{
+			database.Col(domain.JSONSchemaFieldCreatedAt),
+			database.Col(domain.JSONSchemaFieldURL),
+		},
+		Direction: database.OrderAsc,
+	}
+	// Seeded oldest first within an object type and object type by object type,
+	// so the surviving revisions are already in created_at order.
+	drainIncarnation(t, want, orderAsc, func(page database.Page[domain.JSONSchemaField]) (*database.ListResult[*domain.JSONSchema], error) {
```
IAM-marco, line 402: The name of these functions are the weirdest thing I've seen so far

IAM-marco (changes requested):

api/openapi/endpoints/schemas/methods.yaml
vitorbari: We added `json_schemas.object_type` as a `nullable` column.

The schemas API and storage is supposed to be generic and hold different types of schemas. 
Today, the only type of json-schema being stored is `user-schema`, which does not [enforce a object-type](https://github.com/zitadel/nextgen/blob/main/api/openapi/endpoints/schemas/user-schema.json#L5) (but it should).

We have the option to following options: 
- make object-type required for user-schemas (by updating the [meta-schema](https://github.com/zitadel/nextgen/blob/main/api/openapi/endpoints/schemas/user-schema.json#L5)), or 
- make object-type required for all types of schemas (DB constraint) - since we don't have other types of schemas stored today, that would be a bit risky

internal/storage/dialect/postgres/json_schema_compile_test.go
vitorbari: Fair — folded into `compiler_test.go` in all three dialects and deleted the extra file (ee8e7e3d6). It exercises `compileList`, so it sits next to `TestCompileListRequiresAuthzFilter` now; only the conjunct const it passes in comes from `json_schema.go`.

internal/storage/stmttest/main_test.go
vitorbari: Fixed: 8dce531bd2b20aaac22210beb88f52775c005a38

internal/storage/dialect/postgres/json_schema.go
vitorbari: No application flow can produce differing authorization between two schema revisions today, so I'm assuming this is fine.

IAM-marco (approved):

Issue #956, opened 2026-08-24
hard-coded user attributes on UI and API


vitorbari:
## Problem

User schemas are free-form, but the platform still assumes fixed property names whenever it needs to show *who* a user is.

- `GET /sessions/me` returns `name` and `email`, resolved by convention: `name` when present, otherwise `givenName`/`familyName` (with `given_name`/`family_name` also accepted) — see `domain.IdentityAttributeKeys` and `User.DisplayName()` in `internal/domain/user.go`.
- Console and the UI mocks render a user as "name + email", falling back name → email → user ID (`apps/console/src/components/app-shell/AppShell.tsx`).

A schema that spells things differently — `firstName`/`surname`, `preferred_username`, `employeeId`, a phone number as the primary identifier — produces no display name and no email, and the UI degrades to a raw ULID. Machine users, B2B and pseudonymous accounts may have neither a personal name nor an email at all.

The naming convention is invisible to the schema author: nothing tells them that calling a property `surname` instead of `familyName` costs them the display name everywhere in the product. Conversely, a schema that *does* carry `email` gets it surfaced in the session response whether or not that is the identifier the customer wants shown.

There is currently no way for a schema to declare which property (or combination of properties) represents the user's display identity, so the API cannot return it and the UI cannot render it.

## Where it shows up

- `GET /sessions/me` — `name` and `email` fields on the session response
- Console app shell / account menu
- User list and detail screens
- Login flow screens that greet or disambiguate the user

<img width="1597" alt="session identity in console" src="https://github.com/user-attachments/assets/4d59ef23-6bd6-46ba-9cd5-db573af51a64" />

<img width="1058" alt="user display in mock" src="https://github.com/user-attachments/assets/6231637b-35ce-411e-ab6f-ba47e7bccdf9" />

<img width="918" alt="user display in mock" src="https://github.com/user-attachments/assets/fe4fc081-bb35-4fae-ab50-b45642c48658" />

## Out of scope

Solution design. This issue only records the problem.

Issue #869, opened 2026-08-13
Return user attributes when listing sessions


grvijayan:
## Context

`POST /sessions/query` returns `user_id` and nothing else about the authenticated user. `GET /sessions/me` already resolves the user's display name and email from the conventional user-schema properties and returns them as `name` and `email`. The session response schema carries both fields and the API mapper populates them for any session that arrives with an identity attached, so the response shape does not change.

Operator surfaces are the consumer. A list keyed by opaque user IDs is not readable by a human, and the operator get-by-id has the same gap.

## Scope

- Listing sessions resolves the same identity attributes `GET /sessions/me` resolves, for every returned session that has an authenticated user.
- The operator get-by-id resolves them too, so both operator reads of a session answer the same shape.

## Out of scope

Filtering or sorting sessions by name or email. Session fields are columns and user attributes are not, so that is its own piece of work.

Requiring `user.read` in addition to `session.read`. Parked. `session.read` covers these reads for now.

## Acceptance criteria

- Listing sessions returns `name` and `email` for each session with an authenticated user, under the same resolution rules as `GET /sessions/me`, including schemas that spell the name parts either way.
- Reading a single session by ID returns the same fields for the same session.
- Anonymous sessions return neither field.
- A session whose user no longer exists is returned without identity, and the rest of the page is unaffected.
- A session whose user schema defines none of the conventional properties is returned without identity.
- An integration test covers a mixed page in one response: anonymous, authenticated, and deleted-user sessions.
- The `name` and `email` field descriptions no longer say the fields appear only on `GET /sessions/me`.

PR #959, opened 2026-08-24
docs: ADR 058 - user identity designation and references

livio-a:
## Summary

ADR 058 is the solution design for #956 (hard-coded user attributes on UI and API) and reframes #869 (session list identity):

- **`x-identifier`** (schema-root, singular) names the one project-unique leaf property that identifies a user — conditionally required when the schema enables an auth method needing identifier-first dispatch (password). **`x-display`** names the properties rendering the friendly display name. Reinstates the designation that #71/#142 removed, as a root-level keyword instead of the old unordered per-property boolean.
- A shared **`user-ref`** (`user_id`, `identifier`, `identifier_property`, `display`) replaces per-endpoint convention fields on every user-linked resource, resolved live via one batched lookup per page; the user envelope carries the same derived fields (amends ADR 052).
- **Resolution contract:** identifier lookups are scoped to designated, unique rows of in-scope schemas; a bare `login_name` on the direct API resolves across schemas by exactly-one-match, never precedence (classic's zitadel/zitadel#10782 lesson); designated values are unique across designations, contract from day one with staged enforcement. This also specifies the fix for the currently dead direct-API `IdentifierProof` path (empty attribute key — separate bug issue to follow).
- The `IdentityAttributeKeys` spelling convention and `User.DisplayName()` are retired; `/sessions/me` moves from flat `name`/`email` to the embedded ref (pre-release breaking change, in-repo consumers).
- Multi-identifier login, team-restricted login, the constraint-backed identifier namespace, and display templates are recorded as amendment paths with their triggers and constraints pre-agreed.

## Validation

Docs-only. `node scripts/check-pr-title.mjs` passes; README index row added in ADR-ID order.

## Release notes / changeset

No changeset — docs-only, ships nothing to any package (decision table: docs).

## Notes

- Numbering: main is at 055; 056 is claimed by #950. Heads-up: #937 currently claims 052/053, which already exist on main.
- Prior art context: the original `x-identifier` was removed by #71/#142 with no ADR; #820 swept the docs.


docs/adrs/057-user-identity-designation-and-references.md
```diff
@@ -0,0 +1,375 @@
...
+  identifiers in. Every resolution context today — auth attempts, refs, list
+  surfaces — is the project, so `x-unique: "project"` is required for now.
+  Team-scoped uniqueness is deliberate narrower semantics, not a defect: a
+  team-unique username anticipates login restricted to that team, and such a
+  value is intentionally unresolvable in a project-wide context. Once a
+  team-qualified resolution context exists (per-team login domains, a team
+  discovery step, a team-scoped auth attempt), `x-unique: "team"` properties
+  become admissible here; that relaxation amends this ADR without breaking
+  it.
+- The property must not be `x-sensitive`.
```
vitorbari: `x-sensitive` was removed, context: https://github.com/zitadel/nextgen/pull/901

docs/adrs/057-user-identity-designation-and-references.md
```diff
@@ -0,0 +1,375 @@
...
+- Meta-schema changes land in both copies (`packages/config/meta-schemas/` and
+  `api/openapi/endpoints/schemas/`), plus validator, flow resolver
+  (identifier fields restricted to designated properties), and the shipped
+  defaults.
+- `/sessions/me` is a breaking wire change for `<zitadel-session>` and the
+  console shell; `@zitadel/api-mock` session fixtures must be diffed and
+  updated in lockstep (its AGENTS.md parity rule).
+- Existing stored schemas are validated on write only; a schema predating the
+  keywords keeps working (no designation → `user_id`-only refs) until next
+  edited, when the conditional requirement in §1 applies.
```
vitorbari: not relevant - there are no schema stored in production yet.

docs/adrs/058-user-identity-designation-and-references.md
```diff
@@ -0,0 +1,375 @@
...
+- **A display template language instead of a property list.** Deferred;
+  `x-display` as an ordered list covers the known cases and a template can
+  amend this ADR later without breaking it.
+- **Requiring `x-identifier` unconditionally.** Machine and pseudonymous
+  schemas legitimately designate nothing, and schema-level requirement cannot
+  guarantee instance-level presence anyway (that would need the property in
+  `required`, which breaks imports and machine users). The conditional rule
+  in §1 catches the real misconfiguration class.
+
+## Consequences
```
vitorbari, line 360: We have to make it clear that these changes will force us to introduce golang schema validation, because probably not all of them can be enforced by the user meta json-schema: https://github.com/zitadel/nextgen/blob/main/api/openapi/endpoints/schemas/user-schema.json

(this same pattern is already used for the flow definition validation)

livio-a:
@fforootd i requested a review from you since I'm sure you have an opinion on the single identifier decision and your oxidel ADR and code allowed using multiple `x-identifier: true` annotations also on non-unique attributes.

IMO we don't need multiple identifiers for almost all cases. So the question is do we need / want to solve the other cases and do these even exist? 🤷 
Maybe also to @elinashoko. Do you know specific cases where a user needs to be identified by more than one attribute? So the classic either email or username, but which cannot already be solved by making the email the identifier / username.

docs/adrs/057-user-identity-designation-and-references.md
```diff
@@ -0,0 +1,375 @@
...
+
+## Decision
+
+User schemas declare which attributes identify and label a user. The platform
+resolves those declarations into one shared wire shape — the **user ref** —
+wherever a resource links a user. No fixed property names anywhere.
+
+### 1. `x-identifier` — the designated identifier
+
+A schema-root keyword naming one top-level property:
```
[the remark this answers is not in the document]
livio-a: Done in 84d4e17d.

docs/adrs/057-user-identity-designation-and-references.md
livio-a: Done in 84d4e17d.

docs/adrs/057-user-identity-designation-and-references.md
livio-a: Done in 84d4e17d.

docs/adrs/058-user-identity-designation-and-references.md
livio-a: Done in 84d4e17d.

elinashoko:
@livio-a I agree with you, there's no need for multiple identifiers, I think your proposed ordered list solves when we have username or email, employee ID or email etc etc., as long as the first one is used in user references and all of them can be used to sign in.

docs/adrs/058-user-identity-designation-and-references.md
```diff
@@ -0,0 +1,427 @@
...
+  legitimate identifier and break its login. This scoping is what makes the
+  flow path fully collision-proof: nothing outside the bound schema's
+  designated property can interfere.
+- **Designated identifier values are unique across designations.** Writing a
+  value into any designated identifier property requires that no other
+  designated identifier property in the same resolution scope holds it —
+  otherwise one user registering another user's identifier as their own
+  (e.g. a username equal to someone's email) locks the victim out of the
+  cross-schema resolution above. This rule is contract from day one;
+  enforcement is staged: a service-layer check at designated-identifier
```
wim07101993, line 236: can't we make this a storage layer issue as well? If we add an e.g. `__identifier` field to the attributes, let the service/storage layer map that and make it unique in the database. It isn't the cleanest solution but would require less database queries/transactions when creating/updating a user

docs/adrs/058-user-identity-designation-and-references.md
```diff
@@ -0,0 +1,427 @@
...
+
+- `user_id` is always present (the envelope ID, [ADR 052](052-user-envelope-and-attributes.md)).
+- `identifier` is the current value of the designated identifier. Absent when
+  the user is deleted, the schema designates nothing, or the user lacks the
+  value.
+- `identifier_property` names the schema property `identifier` came from, so
+  a client that wants semantics (mailto links, formatting, a field label) can
+  reach the property's schema instead of guessing from the value. Present
+  exactly when `identifier` is.
+- `display` is the `x-display` rendering. Absent under the same conditions.
```
elinashoko, line 120: How would the passkey-only use case be handled here? We’d still render the display name even if the user has no identifier, right?

livio-a:
@elinashoko I want to make sure I read you right, because the two halves of your answer point in different directions: "there's no need for multiple identifiers" — but then "username or email, employee ID or email … **all of them can be used to sign in**", which *is* multiple identifiers (a user reachable through more than one attribute at sign-in).

So which one is the requirement?

1. **Single**: each schema designates one identifier (e.g. email); it labels user references and is the one way to sign in. Username-or-email sign-in doesn't exist until we widen later.
2. **Ordered list**: first entry labels the references, all entries sign in — your examples work from day one, and we pull cross-property collision handling (someone registering a username equal to someone else's email) into the first iteration.

The ADR currently ships 1 with a non-breaking widening path to 2. If your examples are real near-term cases, we start with 2 directly — that choice also steers the storage discussion in the review thread below.

docs/adrs/058-user-identity-designation-and-references.md
livio-a: Exactly — `display` is resolved from `x-display` independently of the identifier. A passkey-only user's ref carries `display` (and `user_id`) with no `identifier`; only when the schema designates no display properties either — or the user lacks those values — does the UI fall back to `user_id`. I'll add a sentence to §3 making that independence explicit.

docs/adrs/058-user-identity-designation-and-references.md
livio-a: or what about an additional column on the `user_unique_attributes` with a partial unique index that deliberately excludes `key`?

elinashoko:
@livio-a Sorry, I didn’t capture the full context for both sides of that sentence :D I meant that we don’t need multiple identifiers in the first iteration. The proposed ordered list would cover future cases like username or email, or employee ID or email, as long as the first one is used in user references and all of them can be used to sign in.

Issue #962, opened 2026-08-24
Add user lifecycle endpoints (deactivate/reactivate)


IAM-marco:
**TL;DR:** #553 ships the enforcement half of user deactivation: a non-active user cannot log in, their sessions and tokens are revoked inside the deactivation transaction, and reads guard `pending_purge`. What still does not exist is an API to deactivate or reactivate a user. Today the only path to a deactivated user is deleting their lifecycle-owner team. The console's `Deactivate` row action (users list, #630) is blocked on these endpoints.

<details>
<summary><h3>What already exists (from #553)</h3></summary>

- `DeactivateUser` storage statements in all three dialects (postgres/sqlite/spanner), each revoking the user's sessions and tokens in the same transaction as the status flip, per ADR 024. The future endpoint inherits this for free.
- Login enforcement keyed on `status == active`: identifier lookup, discoverable passkey resolution, and a terminal gate at session exchange. A deactivated user is indistinguishable from an unknown login name.
- Management reads keep serving `deactivated` and `suspended` users with their status visible; only `pending_purge` reads as gone (parity with teams).
- Grants and resource-scope-index rows are deliberately left intact on deactivation so it stays reversible; access is cut at the authentication boundary and by membership-edge deletion.

</details>

<details>
<summary><h3>What the endpoints need</h3></summary>

- Deactivate: call the existing `DeactivateUser` statement (it already sweeps memberships, deletes authz membership edges, and revokes sessions/tokens). Decide the wire shape (e.g. `POST /users/{user_id}/deactivate` vs a status write); teams currently model deactivation as `DELETE /teams/{team_id}`.
- Reactivate: a `ReactivateUser` statement does not exist yet. Reactivation restores `active` and must re-sync authz membership edges from the roster (see `SyncUserTeamMembershipEdge`); direct grants were never removed, so nothing else needs rebuilding. Note the deactivation cascade sets the user's memberships to `removed`, so decide whether reactivation restores rosters or the user starts clean.
- Console: re-add the `Deactivate` row action in `apps/console/src/routes/_authed/users/index.tsx` (the row menu comment there points at this issue).

</details>

<details>
<summary><h3>Related note: sk_team_ keys and team status</h3></summary>

[...]

Issue #630, opened 2026-07-26
Frontend: Users list — schema-driven columns, sorting, and pagination


bastionstack:
## Context

Build out the console Users list.

**Design:** `User list` — internal Figma, node `277:288291` (`Filled`, `Minimal`, `ContextMenu`, `Dashboard xs`)

## Delivered

- **Schema-driven columns** (#701) — columns are the properties of the schemas the loaded users reference, unioned, with an em dash where a user's schema omits one. Search covers every rendered column.
- **Row menu** (#700) — `View details` and `Delete`, the latter opening the dialog from #632.
- **Horizontal scroll** for the wide column set.
- **Sorting is out of scope**, per decision D5: filtering and sorting are not in the MVP, and a client-side sort orders only the fetched page while implying a total order.

## Blocked

Struck-through rows are resolved by #721 (open) — `feat: user metadata` adds `metadata` (`createdAt`, `updatedAt`, `status`) to the user and moves `GET /users` to cursor pagination.

| Design element | Needs | Tracking |
|---|---|---|
| ~~**Status** column~~ | ~~`status` on the user read response~~ | ✅ #721 |
| ~~**`Load more`**~~ | ~~Cursor pagination~~ | ✅ #721 |
| **Team** column | A team reference on the user. Listing and naming teams is done — `POST /teams/query` landed with #620 — but nothing on the user read responses says which teams they belong to | #735 |
| ~~**Last login** column~~ | ~~The field itself — nothing carries it~~ | Not for MVP |
| ~~Row menu → **Edit user**~~ | ~~`PATCH /users/{user_id}`~~ | Not for MVP |

When #721 merges the list screen needs updating with it: `listUsers` returns `{ users, next_page_token }` rather than an array, and `metadata` becomes a reserved key alongside `id` and `$schema` that must stay out of the schema-driven columns.

## Notes

- A schema that defines its own `status` property already renders as a column, since columns are schema-driven. The blocked item is a status the console can rely on for every user.
- The sidebar `User schemas` sub-item belongs to #712, which builds the screen it points at. D8 settles the wording — "schema", always qualified — over the `Profile schema` label used in `520:77358`.

## Reference

- `apps/console/AGENTS.md`, console ADRs (routing `0001`, API access `0002`)
- `apps/console/docs/styling.md` — classify the component before building

bastionstack:
Updated the body to reflect what has landed and what has not.

[...]

PR #963, opened 2026-08-24
fix: block login and revoke sessions for non-active users

### LINKED ISSUE #553: Add status filters in service and api layer
One thing to track separately rather than in this PR: reads don't filter `users.status`/`teams.status` yet, so deactivation doesn't block anything (login included). That's consistent with the storage-only scope here — let's just make sure service-layer enforcement has its own issue.

_Originally posted by @fforootd in https://github.com/zitadel/nextgen/pull/287#pullrequestreview-4714672253_

IAM-marco:
## Summary

Fixes #553.

Deactivating a user now actually locks them out. Before this PR the status column was storage-only: a user deactivated by the team-delete cascade could still log in, and their live sessions and tokens survived until expiry.

<details>
<summary><h3>What is enforced, and where</h3></summary>

- **Login**: identifier resolution requires `status == active` (`UserStatementsLookup.GetByAttributes`), so a suspended, deactivated, or pending-purge user is indistinguishable from an unknown login name. The discoverable (usernameless) passkey path, which resolves the user cryptographically and never read the users row, gets the same guard with the same generic rejection. Session exchange carries a terminal active-status gate inside the mint transaction, so a deactivation cannot race an in-flight login.
- **Revocation**: `DeactivateUser` and the `DELETE /teams/{team_id}` owned-user cascade delete the affected users' sessions and tokens inside the same transaction that flips the status (ADR 024), in all three dialects. Hard user delete already cascades sessions and tokens via the FK and is unchanged.
- **Reads**: `GET /users/{user_id}` answers 404 for a `pending_purge` user and `GET /users` excludes them, matching the existing team guards. Deactivated and suspended users stay fully visible to the management plane so they can be audited and reactivated.

</details>

Scope decisions from grooming the issue:

- No API contract change: no `status` filter param on `GET /users` (the console MVP excludes list filtering, decision D5 in #630).
- Grants and resource-scope-index rows stay intact on deactivation, keeping it reversible; access is cut at the authentication boundary. Known residue: a deactivated user can still appear in authz-filtered lists.
- No user deactivate/reactivate endpoint here; that is #962, and the console row-menu TODO now points there instead of #553.
- `sk_team_` keys are unmintable groundwork today; the team-status check at authentication belongs to whoever ships minting (noted in #962).

## Validation

- New `stmttest` suites (`forEachDialect`): user deactivation revokes sessions and tokens (session token and PAT); the team cascade revokes lifecycle-owned users' credentials and spares other members'. Green on sqlite and postgres locally.
- New service unit tests: user service (`pending_purge` get/list guards, login lookup filter), auth attempt (discoverable passkey binds an active user, rejects a non-active one with the generic proof rejection), session (exchange gate rejects a user deactivated between verify and exchange).
- New integration test `TestTeamDelete_LocksOutDeactivatedUser`: the full wire journey. Login flow succeeds and mints a session; `DELETE /teams/{team_id}`; the old session cookie gets 401; a fresh login lands on the `not_found` terminal exactly like an unknown email; `GET /users/{user_id}` and `GET /users` still serve the user with `metadata.status = deactivated`. Green on postgres.
- `moon ci :lint :typecheck :build :test` green locally.

## Release notes / changeset

`.changeset/user-status-enforcement.md`, `@zitadel/server` patch: enforce user status at login, revoke sessions and tokens on deactivation, hide `pending_purge` users from reads.

## Notes

- The enforcement predicate is `status == active`, so the currently unwritten `suspended` status is enforced the day something starts writing it.
- Nothing in the API can set a user `deactivated` directly yet; the team-delete cascade is the only trigger. The lifecycle endpoints are tracked in #962 and will inherit the revocation for free, since it lives in the `DeactivateUser` statement.

### DIFF
--- a/.changeset/user-status-enforcement.md
+++ b/.changeset/user-status-enforcement.md
@@ -1,0 +1,5 @@
+---
+"@zitadel/server": patch
+---
+
+Enforce user status: a non-active (suspended, deactivated, or pending purge) user can no longer log in, deactivation revokes the user's sessions and tokens in the same transaction, and user reads hide `pending_purge` records while keeping deactivated users visible to management APIs.
--- a/apps/console/src/routes/_authed/users/index.tsx
+++ b/apps/console/src/routes/_authed/users/index.tsx
@@ -392,7 +392,8 @@ function userStatus(user: Record<string, unknown>): string | undefined {
  * as live controls that did nothing:
  *
  *   - `Edit user` — no `PATCH`/`PUT /users/{user_id}` endpoint (#693)
- *   - `Deactivate` — no `status` field and no lifecycle endpoint (#553)
+ *   - `Deactivate` — no lifecycle endpoint (#962; the `status` field landed
+ *     with #721, and #553 made deactivation actually lock the user out)
  *   - `Reset password` — `PUT /users/{user_id}/password` exists, but the screen
  *     that would collect the new password does not; the menu item alone had
  *     nothing behind it
--- a/internal/api/integration_test/user_status_test.go
+++ b/internal/api/integration_test/user_status_test.go
@@ -1,0 +1,198 @@
+//go:build postgres_integration || spanner_integration
+
+package integration_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"net/http"
+	"testing"
+
+	"github.com/go-faster/jx"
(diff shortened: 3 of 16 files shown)


IAM-marco:
Closing as not needed

Issue #975, opened 2026-08-25
Explicit session-to-team binding for team-scoped revocation


muhlemmer:
**Context:** Follow-up to #749, recorded in [ADR 056 — Sessions Join Teams Through User Membership](https://github.com/zitadel/nextgen/blob/main/docs/adrs/056-session-team-membership-join.md).

#749 shipped the team filter on `POST /sessions/query` using the **implicit** model: a session stays project-scoped and joins a team through its bound user's roster membership, derived at read time. That is enough to *list* a team's sessions, but it deliberately leaves one thing out.

**A session holds no per-team state, so revocation is project-level only.** `DELETE /sessions/{id}` ends the session outright — the user is signed out of the project and therefore out of every team. There is no way to express "end this session's access to team A while leaving B and C alone".

Two cases need that, and both need an explicit session→team binding:

1. **Privilege inversion.** A team admin who revokes a session seen in their team logs the user out of every other team too — including a project admin who happens to be a member of that team. A lower-privileged principal can evict a higher-privileged one. Authorization on the revoke action mitigates this (see ADR 056 § Revocation scope) but does not make revocation scopable.
2. **Scoped agent sessions.** An AI agent with permissions across many teams should be able to hold a session valid only for the team it was tasked against, so a mistake cannot reach the others. This is the OAuth2 `scope` shape, which the implicit model cannot express.

## Proposed table

A session can be in many teams, so this is a join table, not a column:

```sql
-- team_sessions is a many-to-many relation for sessions and teams.
CREATE TABLE zitadel_nextgen.team_sessions (
    project_id  TEXT COLLATE "C" NOT NULL,
    team_id     TEXT COLLATE "C" NOT NULL,
    user_id     TEXT COLLATE "C" NOT NULL,
    session_id  TEXT COLLATE "C" NOT NULL,

    FOREIGN KEY (project_id, session_id)
        REFERENCES zitadel_nextgen.sessions (project_id, id)
        ON DELETE CASCADE,
    FOREIGN KEY (project_id, team_id, user_id)
        REFERENCES zitadel_nextgen.team_memberships (project_id, team_id, user_id)
        ON DELETE CASCADE
);
```

The second foreign key is the load-bearing part: pointing at `team_memberships`' full primary key means a session can never be bound to a team its user is not a member of, and losing the membership cascades the binding away. Deleting one `team_sessions` row is then team-scoped revocation — it ends that session's access to that team and nothing else.

## Scope

[...]

Issue #745, opened 2026-08-04
[API] Session management


vitorbari:
**Context:** Operators need to manage end-user sessions from the console. Today `GET /sessions` (list) is unimplemented, revoke hard-deletes the row (so revoked sessions can't be shown), and there is no team linkage. Get-by-id, revoke-by-id, and the `me` endpoints already work.

**Specifications:**
- List via `POST /sessions/query` (ADR 031): filter by project/team/user/status,
  newest-first default, cursor pagination.
- Open a session for details (user, session ID, status, factors, created/expiry time).
- ~Revoke an active session; revoked sessions remain visible with `state: revoked`.~

**Acceptance criteria:**
- [ ] Operators can view active, expired, ~and revoked~ sessions
- [ ] Results filterable by project, team, user, and status
- [ ] A session can be opened to see its details
- [ ] ~An active session can be revoked and then shows as `revoked`~
- [ ] List is paginated and newest-first by default

Issue #288, opened 2026-06-15
AI agent identity


elinashoko:
# Problem

Our customers are increasingly exploring how AI agents can automate work across engineering, support, security, operations, and other business functions.

The real value isn't in agents answering questions or generating content. It's in agents taking actions: investigating incidents, updating systems, helping customers, managing infrastructure, or running business workflows.

Before a customer can trust an agent with meaningful work, they need answers to a few simple questions:

- Which agents exist?
- Who owns them?
- What can they do?
- Why can they do it?
- What have they done?
- How do I stop them?

Today, our customers typically solve this in one of two ways:

- Hacking custom agent management on top of Zitadel.
- Using token exchange or similar mechanisms to allow agents to act on behalf of users.

These approaches work for simple use cases, but start to break down as agents become more capable. Agents often end up acting as digital copies of users, inheriting broad permissions that were never intended for the task at hand.

As a result, it becomes difficult to understand what authority an agent actually has, why it was allowed to perform an action, or how to safely give it more responsibility over time.

Our customers need a better way to understand, control, and trust agents as they become increasingly autonomous.

# TL;DR

Introduce agents as first-class actors in Zitadel.

Customers should be able to:

- See which agents exist
- Understand who owns them
- Control what they are allowed to do
- Review what they have done
- Revoke access when needed

Authority is granted through delegation rather than broad permission inheritance, allowing agents to operate within clearly defined boundaries.

This proposal introduces the core concepts needed to model, manage, and audit agents. 

# Concepts

## Agent

An agent is an actor within the system.

Like users, agents have their own identity and can authenticate independently. Unlike traditional automation, agents can make decisions and act with varying levels of autonomy.

Agents represent specific autonomous actors operating on behalf of users, teams, or systems.

Examples include:

- Assistant agents that help humans complete tasks or make decisions
- Execution agents that perform actions on behalf of users or teams
- Workflow agents that coordinate multi-step processes across systems
- Autonomous agents that operate continuously towards goals
- Cross-organisational agents that interact with external systems, services, or agents

For MVP, the model should primarily support assistant agents and execution agents. The same model should be designed to evolve towards workflow, autonomous, and cross-organisational agents later.

---

## Delegation

Delegation grants authority to an agent.

A delegation answers:

- Who authorised the agent?
- What is the agent allowed to do?
- Which resources can it access?
- How long is that authority valid?

Delegation is separate from identity.

Identity answers:

> Who is this agent?

Delegation answers:

[...]

PR #976, opened 2026-08-25
feat(session): filter sessions by lifecycle-owner team on POST /sessions/query

### LINKED ISSUE #749: List sessions - Filter by team
**Context:** Sessions have no team column or linkage today, so team filtering needs a data-model decision before any implementation.

### Use case

As a Project Administrator, I want to filter the Sessions list by Team so that I can quickly find the relevant sessions during a security or support investigation.

### Requirements

When a Team is selected in the Sessions list:

* Show sessions belonging to users who are currently members of the selected Team.
* If a user belongs to multiple Teams, their sessions appear when filtering by any of those Teams.

**Specifications:**
- Decide how a session maps to a team (likely join session → user → team membership)
- Add `team` (or `team_id`) to `session-filter-field.yaml`
- Implement the join-based filter in the query

**Acceptance criteria:**
- [x] Session→team mapping decided and documented
- [x] `filter` by team returns only that team's sessions
- [x] Tests cover the team filter

muhlemmer:
## Summary

Adds a `lifecycle_owner_team_id` filter to `POST /sessions/query`. A session stays project-scoped and joins the team that **owns its bound user's lifecycle** (`users.lifecycle_owner_team_id`, ADR 024), resolved at read time — no new tables or columns. A user has one lifecycle owner, so a session surfaces under one team or none; a self-owned user's session and a session with no user belong to no team.

The lens changed during review ([#976 review](https://github.com/zitadel/nextgen/pull/976#pullrequestreview-5066105295)): this PR originally joined through **roster membership**, which #749 had guessed at. Ownership is the relation the caller can act on — every row a team sees is a session it may revoke — and it is the relation behind offboarding, a compromised account, compliance ("no session outlived deactivation") and team-owned agents. The membership cases have better surfaces (events for the past, the roster for "who can access us"), and a membership filter would have needed the `team_membership.read` gate that meanwhile landed on the users query (#1004), while the ownership filter needs no gate at all.

- The filter compiles to a correlated `EXISTS` over `users`. Expressing it needed a new `database.CorrelatedEqual` filter plus a `FieldBinding.SQLSuffix`, because every existing compiler branch writes the bound value at a fixed position relative to the column, and this predicate needs it *inside* the expression. One shared compile branch in `dialect/compare`, one behavior asserted in `stmttest`.
- The wire name matches the users query, which now carries both lenses side by side (#984, #1004, #1071): `lifecycle_owner_team_id` is ownership there too. `team_id` is left unclaimed on sessions — a session gets a team relation of its own if the explicit binding (#975) lands, and that has the better claim to the plain name.

## Schema

**This PR ships a migration** — `000018_sessions_sort_index` on PostgreSQL, `000021` on Spanner, `000005` on SQLite. It is index-only, adds no column or table, and reverses cleanly:

- `sessions (project_id, created_at, id)` — the default sort. Benchmarking the filter surfaced a pre-existing gap: without it, *every* page of `POST /sessions/query` scanned and sorted a project's whole session table (127 ms for an unfiltered first page at 2M sessions, against 0.44 ms after).
- `sessions (project_id, user_id)` — the drive side of a selective team, and it also serves the existing `user_id` sort. Both are load-bearing: they cover opposite team selectivities, and with the sort index alone a 3-user team is *worse* than no index at all (116 ms against 0.47 ms with both).
- `users (project_id, lifecycle_owner_team_id)` on SQLite only, the peer PostgreSQL and Spanner already had from migration 000011.

On PostgreSQL both session indexes are built `CONCURRENTLY` under `-- +goose NO TRANSACTION`, since `sessions` is a hot write path and a plain `CREATE INDEX` would block session creation and exchange for the whole build; each `CREATE` is preceded by a `DROP` so a rerun after a failed build converges instead of colliding with the `INVALID` index left behind. Spanner needs no equivalent (index backfill is a background operation with the table writable) and SQLite has none to offer.

Decision, data model, rejected alternatives (roster membership among them) and the measured plans: [ADR 060](docs/adrs/060-session-team-lifecycle-join.md) — renumbered from 059, which main took for "Expanding Embedded Objects".

Revocation stays project-level only. Team-scoped revocation needs an explicit session→team binding — deferred, tracked in #975. What the ownership join fixes is the mismatch the membership join would have shipped: the filter and the revoke authority now describe the same relation.

Closes #749.

## Validation

- `go build ./... && go vet ./... && gofmt -l internal/`
- `go test ./internal/...`
- `go test -tags sqlite_integration ./internal/...`
- `go test -tags postgres_integration ./internal/...`
- `go test -tags spanner_integration ./internal/...`
- `node scripts/go-generate.mjs` (generated output reproduces the merge byte for byte), redocly lint of `api/openapi/openapi-spec.yaml`
- Performance **re-measured for the new join**, not carried over: `EXPLAIN (ANALYZE, BUFFERS)` on PostgreSQL 18.3, 2M sessions + 200k users per project with an equal-sized decoy project, one verified check on every second session, ownership skewed 40% / 1% / 3 users, median of three warm runs over the statement `ListSessions` compiles.

| 2M sessions, first page of 20 | no index | sort index only | both (000018) |
|---|---:|---:|---:|
| no filter (baseline) | 127 ms | 0.57 ms | 0.44 ms |
| 40%-owning team | 215 ms | 0.74 ms | 0.81 ms |
| 1% team | 123 ms | 5.9 ms | 8.8 ms |
| 3-user team | 111 ms | 116 ms | 0.47 ms |
| team owning nobody | 15.7 ms | 15.6 ms | 0.41 ms |
| team + `state=active` | 76 ms | 77 ms | 38 ms |
| cursor at page 50 | 128 ms | 5.1 ms | 6.0 ms |

## Release notes / changeset

- Changeset: `.changeset/session-team-filter.md` — filter sessions by the team owning the user's lifecycle on `POST /sessions/query`, plus the index-only migration and its write-lock cost (`@zitadel/server`, minor). Decision table: user-facing API surface *and* schema, so a changeset is required.

## Notes

- **Test coverage of the split.** `stmttest` pins the ownership-vs-roster divergence on all three dialects: a collaborator on team B's roster owned by team A does not appear under B; a self-owned user on a roster appears nowhere; and a **deactivated** user's still-live session stays visible to its owning team — deactivation drops memberships but not ownership, so under the roster join the row would have vanished exactly when the admin needs it.
- **One open naming question, deliberately left to whoever hits it.** On users, `team_id` means membership; on sessions this PR leaves `team_id` unclaimed for #975's binding. A roster filter on sessions, if a use case ever justifies one, has to pick between `member_of_team_id` and `team_id` (parity with users). Nothing shipping here depends on it — both names are free.
- **The worst case to watch** is the team filter combined with `state`: two computed `EXISTS` on one list, 38 ms against 0.44 ms for the same page unfiltered. Bounded, and the 10.5 s pathological plan recorded under the membership shape does not reproduce here — but the shape can recur if a third computed predicate is added.
- The 40%-team regression recorded in the previous ADR revision does not reproduce under this join.


livio-a:
Question before this hardens into the API contract: I noted there were multiple discussions in #749, but I couldn't really figure out the now intended meaning of the `team_id` filter.

As far as I can trace it, we have three candidates for "this session belongs to team X":

1. the user is on the team's roster — what this PR ships
2. the session is explicitly bound to the team — the #975 plan, which even says the filter might later read the binding "instead of (or in addition to)" membership
3. the team owns the user's lifecycle — what ADR 056 itself says should decide who may *revoke*

The membership join seems to come from #749's "likely session → user → team membership" rather than from an actual use case — #745 only says "filterable by team". What makes me hesitate is the b2b / multi-tenant reading: per ADR 024 a team is a tenant grouping, and the tenant's own users are exactly the team-*owned* ones. So "tenant X's sessions" could just as well mean lifecycle ownership, not the roster. The two joins split exactly at the interesting cases: a rostered external collaborator (membership yes, lifecycle no) vs a managed user off the roster (lifecycle yes, membership no). With the current join, the team that *manages* a user might never see their sessions — while being the one that per ADR 056 may revoke them.

Related: I'd actually expect a tenant admin not to need this filter at all — their list should be scoped by permission ("listings are authorization queries", ADR 053), with the resolver injecting the team scope into the query the same way this filter does, and `team_id` only serving as drill-down for principals who see more than one team. That scoping doesn't exist yet (the endpoint is all-or-nothing on project-level session read), and it needs the exact same decision: whichever join defines a team's sessions has to be identical for the permission scope and the filter, or the two views diverge.

So my actual question: who is the team session view for, and what question should it answer — "who can access this team" or "the sessions of this tenant's users"? Once that's decided, the join, the filter semantics and the permission scoping all follow from it.

I'll review the implementation itself once we've settled this. Happy to open follow-up issues as needed. cc @vitorbari @elinashoko

livio-a:
Update from today's standup: @elinashoko will write down a use-case matrix for the team semantics (who views which sessions/users through which relationship, and with which actions). I'll continue the review here once that lands, so we can also settle whether the filter field name stays `team_id` before it enters the released contract.

livio-a (changes requested):
First the actual code review, which you were owed for a while — sorry for that.

The implementation is really solid: the shared compile branch in `dialect/compare` with the placeholder tests per dialect, the stmttest suite (multi-team, anonymous, status transitions) and the honest perf docs incl. the 40%-team regression. Two findings:

1. the roster status list `('pending','active','inactive')` is hardcoded in three dialect schemas, the ADR derives it from `domain.RosterMembershipStatuses` — a new status would silently diverge. a test asserting the SQL fragment against the constant would pin it. (moot if the lens changes, see below)
2. ADR 056 got taken on main in the meantime (056–058 all landed) — needs renumbering, 059?

Now my opinion on the lens, as agreed in the sync: I think the filter should be lifecycle-owner and the membership variant shouldn't ship.

One thing I noticed re-reading #749: we chose membership over the *explicit binding* — lifecycle was never on the table, so membership vs lifecycle was actually never decided. And the argument for membership back then (show sessions of users who currently have access) is the one Elina's matrix has since moved away from — both session rows there are lifecycle.

Why lifecycle:

- it matches the revoke authority we agreed on. every row you see is one you can act on. a membership row for a user you don't manage has no action — you can't revoke, and your actual remedy (remove the membership) doesn't need a session list.
- the real use cases are all lifecycle: offboarding (the matrix names it explicitly), a compromised employee account (owning team kills the sessions, other teams untouched), compliance (no session outlives deactivation — only answerable per owning team), and agents (#288): an agent owned by a team is exactly whose sessions that team watches and kills.
- even support: "I can't log in" is user-rooted → `user_id` filter, no team lens. and "team X says their people can't log in" — login trouble that correlates with a team correlates with their IdP/SSO/provisioning, and those are exactly the users they lifecycle-own. a guest on their roster logs in through their own context and isn't affected. so the team-rooted support case actually wants lifecycle too.
- the membership cases all have better surfaces: suspended member → suspension already cut access, permissions resolve per check; "who accessed us during the incident" → events, a session list can't see the past; "who can access us" → roster. the one real sweep ("sessions of everyone associated with X") composes from the users query membership filter + sessions `user_id` — two hops, each properly permissioned.
- and membership burns real principals: a help desk holds `user.read` without membership read on purpose, a SIEM key holds `session.read` alone on purpose. an ungated membership filter lets both enumerate every roster one bit per row; gating it means importing permission machinery this PR doesn't need. the lifecycle filter needs no gate — it filters along the relation that scopes the caller anyway.
- fwiw nobody ships roster→global-sessions: github shows org admins only the org-bound SSO sessions (basically #975), workspace/EMU show sessions to the account owner, auth0 has no org session list at all.

If we want membership additionally anyway: then gated by `team_membership.read` on the named team, named `member_of_team_id` (not `team_id`), with an actual use case in the matrix, and a note on the #975 interaction (its criteria already say the team filter might read the binding later). that's more work than the swap, which is part of why I wouldn't.

Good news is this PR already built everything the lifecycle filter needs: `CorrelatedEqual`/`SQLSuffix`, the compile branches, the tests, both indexes all carry over. the swap is the EXISTS reading `users` (`u.project_id = s.project_id AND u.id = s.user_id AND u.lifecycle_owner_team_id = ?`), the filter name, dropping the status set and the ADR text (+ renumber). finding 1 above disappears with it. happy to pair on it or push the swap as a commit on your branch — your foundation is what makes it cheap.

muhlemmer:
@livio-a Swapped to the lifecycle lens, as agreed. Full reasoning in the commit message: https://github.com/zitadel/nextgen/commit/09a2a1c09cd9e701574bd90b3d0d52c32d246857

**TL;DR**

- The `EXISTS` now reads `users.lifecycle_owner_team_id` instead of `team_memberships`. Finding 1 dissolves with it — the hardcoded `('pending','active','inactive')` set is gone, so there is nothing left to pin against the constant.
- Field is `lifecycle_owner_team_id`, which turned out to match the users query that landed meanwhile (#1071). `team_id` stays unclaimed for #975. One naming question I recorded in the ADR rather than deciding alone: on users `team_id` means membership, so a future roster filter on sessions has to pick between that parity and your `member_of_team_id`.
- ADR is **060** now — 059 went to Expanding Embedded Objects while this sat. Rewritten for the ownership join; your sync arguments (revoke authority, offboarding/compromise/compliance/agents, the help-desk and SIEM principals, the product survey) are recorded in it, and it now cites the `team_membership.read` gate from #1004 instead of predicting it.
- Perf re-measured for the new shape rather than carried over: 2M sessions + 200k users, postgres 18.3. Both indexes stay load-bearing — the 3-user team is 116 ms with the sort index alone against 0.47 ms with both. The 40%-team regression does not reproduce; worst case is now team + `state=active` at 38 ms, and the old 10.5 s plan is gone.
- `stmttest` pins the split on all three dialects, including the case that convinced me most: a **deactivated** user's live session stays visible to the owning team, where the roster join would have dropped it exactly when the admin needs it.
- `origin/main` merged in, all three dialect suites green.

docs/adrs/060-session-team-lifecycle-join.md
```diff
@@ -0,0 +1,354 @@
...
+
+The cost is the build, not the shape. PostgreSQL's plain `CREATE INDEX` takes a
+`SHARE` lock: session **writes** — creation, exchange, token issuance — block
+while each index builds, though reads are unaffected. On the 4M-row `sessions`
+table this ADR measured against, one index took ~4 s on container hardware. That
+is acceptable pre-release, where no deployment carries a session table worth
+protecting. If one ever does, the migration should be re-cut as
+`CREATE INDEX CONCURRENTLY` under `-- +goose NO TRANSACTION` (which trades the
+lock for a second table pass and a possible `INVALID` index to clean up on
+failure) — a deliberate decision, not something to discover during an upgrade.
```
muhlemmer: Don't care. This shouldn't be in an ADR.

livio-a (approved):
The swap resolves both findings. Verified against the new shape: the stmttest pins every boundary that matters — the roster collaborator owned elsewhere stays out, the self-owned user's NULL owner matches no team, anonymous sessions fall out, and the deactivated user's live session stays visible to the owning team, which is the case that separates the two lenses in practice. Finding 1 dissolves with the join (no status set left to drift), finding 2 is done (ADR 060, README row updated).

The migration got better than asked: `CONCURRENTLY` with the DROP-first convergence pattern on postgres, and SQLite picking up the missing `idx_users_lifecycle_owner_team_id`. Re-measuring the plans for the new shape instead of carrying the old numbers over is appreciated — same for recording the naming question in the ADR instead of deciding it here.

Approving.

livio-a:
Pushed the fix directly to keep things moving: the environments migration on main took the next slot on all three dialects, so the sort-index migrations collided everywhere — sqlite failed in `full-pr` (`TestZeroConfigStyleConnectMigrateCRUD`, duplicate version 000005), postgres/spanner were latent behind their integration tags. Renumbered to 000006/000019/000022 and updated the by-name references (postgres-peer pointers, ADR 060, ADR README). Dialect suites green locally; approval stands.

PR #979, opened 2026-08-26
feat: guarantee platform users their personal team at registration and sign-in

peintnermax:
## Summary

Part of #527; unblocks the browser leg of #615/#96. Platform-plane users are guaranteed a team — the one `claim/complete` attaches claimed projects to — and the platform project itself is bootstrapped into a state that can actually serve a registration.

## What it does

**Provisioning runs at session exchange.** `EnsurePersonalTeam` is invoked from `POST /sessions/exchange`, which is the one credential-agnostic point every registration and every sign-in passes through before any authenticated call. A registration that never exchanges its handoff has no session either, so it cannot claim; the team appears on the next exchange. Users created before this change converge the same way, so there is no migration.

**One team per user, enforced by the database.** The team name is a pure function of the user id. Team names are unique per project (case-insensitively), so that turns the existing index into the one-team-per-user constraint: two ensures racing compute the same name and one loses the insert instead of minting a second team. Deriving from the user id rather than the email keeps the name bounded and unguessable, so it cannot be invalidated by a long address or squatted by an unrelated team — which is what makes failing closed safe. The name is a renamable placeholder, not an identifier (ADR 046 §4, amended in #1069).

**Absent and deactivated are told apart.** `GetPersonalTeamForUser` answers "may this user claim", so it collapses "holds no membership" and "membership not active" into one not-found. Provisioning must distinguish them: the first needs a team, the second must be left alone, because claim resolves the earliest membership and would keep seeing the deactivated one. The new `GetEarliestTeamMembership` asks that question per dialect, and a non-active membership reports `ErrPersonalTeamNotActive` rather than succeeding silently.

**Gated on the explicit opt-in.** The effect is wired with `PlatformConfig.ProvisioningProjectID()`, which is non-empty only when `platform.bootstrap_project` is set. A standalone deployment that merely pins `platform.project_id` is naming its console default project, not opting into the platform plane, and its end-user registrations must not mint teams (#605, #736).

**The bootstrapped platform project is usable.** `platform.Ensure` now seeds through the project service: encryption and signing keys, the default user schema, and the default login flow definitions. A bare row satisfies the foreign keys and nothing else. It also refuses to start when a pre-existing platform project has no keys, because #736's bootstrap inserted exactly such a bare row and treating that collision as success would leave the deployment permanently unable to serve a registration.

## Validation

- `go build ./...`, `go vet ./...`, `gofmt` clean; `moon run server:generate` reproduces the committed artifacts.
- `go test ./internal/...` green.
- `go test -tags sqlite_integration ./internal/bootstrap/platform/ ./internal/storage/stmttest/` green. The stmttest cases pin the split across dialects, including that a deactivated earliest membership stays invisible to `GetPersonalTeamForUser` while `GetEarliestTeamMembership` returns it with its status.
- Handler tests cover the exchange wiring: the ensure fires for an exchanged user session, is skipped for an anonymous one, does not cost the sign-in when it fails, and is absent without the setter.

## Release notes / changeset

`.changeset/personal-team-ensure.md` (`@zitadel/server`, minor).

## Notes

- Stacked follow-up #1070 gives `claim/complete` a distinct 403 code for the not-active case so the console can react.
- Deliberately not done here: the `team.membership.updated` audit event, which the catalog lists as deferred under #879 and which the claim path is also missing.

https://claude.ai/code/session_01NYqGw6vooPadT98REz1pqh


vitorbari:
> The trade against the input-side design (an on_success: create_user_with_personal_team variant, considered and discarded): that shape bought strict same-transaction atomicity with user creation, but at the cost of a password-only effect (passkey runs a different provisioning path that on_success cannot express), a per-credential handler matrix, wire-format/schema churn, and an unsolved backfill. This design accepts one narrow eventual-consistency window — a crash between user commit and ensure — healed at next sign-in, in exchange for covering every credential, every pre-existing user, and touching no contracts.


I don't get this part, it says `on_success: create_user_with_personal_team` was discarded, but that is the change I see in the PR.

cmd/server/server.go
```diff
@@ -198,16 +198,39 @@ func run(ctx context.Context, cfg Config, userFiles []string) error {
...
 		userService,
 		schemaStore,
 		serviceDBPool,
 	)
 	createUserForPasskeyHandler := service.NewFlowCreateUserForPasskeyHandler(userService, schemaStore)
+	// The platform project's registration side effect (#527): same identity
+	// provisioning as create_user, plus the personal team the claim flow
+	// attaches projects to. Wired with the resolved platform project id so the
+	// handler's own gate matches claim/complete's session verdict.
+	createUserWithTeamHandler := service.NewFlowCreateUserWithPersonalTeamHandler(
```
IAM-marco: This branch ships two implementations of the same feature and wires both. `cd649c04` added the `create_user_with_personal_team` on_success design; `b996c911` added the `EnsurePersonalTeam` design but does not revert the first, so at HEAD `createUserWithTeamHandler` (here) and `.WithUserWithTeamCreater(...)` on line 245 are live alongside `personalTeams` on line 206.

The on_success half is unreachable: the stock flow still declares `create_user`, and nothing else in the repo names the new value.

https://github.com/zitadel/nextgen/blob/b996c911ff3477324a4cfa3d5427944e018d98f0/packages/config/defaults/default-login.json#L133-L135

It nonetheless carries the full wire surface for a value no flow uses: `internal/domain/flow_definition.go`, `flow_on_success.go`, `flowonsuccess_enumer.go`, `flow_state_machine.go`, `api/openapi/components/flows/flow-definition-step.yaml`, `api/openapi/endpoints/schemas/flow-definition.json`, `packages/config/meta-schemas/flow-definition.json`, `packages/config/src/validate.ts`, plus a second changeset.

The PR description says the opposite: that shape was "considered and discarded", and "No new `on_success` value, no flow-definition/schema/OpenAPI/wire changes at all". Dropping `cd649c04` removes ~400 lines and makes the description true again.

internal/service/personal_team.go
```diff
@@ -0,0 +1,130 @@
...
+		platformProjectID: platformProjectID,
+	}
+}
+
+func (s *personalTeamService) EnsurePersonalTeam(ctx context.Context, projectID, userID string) error {
+	// Personal teams are a platform-plane concept: every registration in every
+	// customer project passes through the same funnel, and none of those may
+	// mint teams. A silent no-op, not an error — off-platform callers are the
+	// normal case.
+	if s.platformProjectID == "" || projectID != s.platformProjectID {
```
IAM-marco, line 62: This gate re-derives "is this deployment the platform" from the resolved project id, but #736 already added the explicit signal for exactly that question: `platform.bootstrap_project`.

`ResolvedProjectID()` returns the built-in platform id only when `BootstrapProject` is true. Otherwise it returns the operator's `platform.project_id`, documented as a standalone deployment's pin to the project `zitadel setup` created:

https://github.com/zitadel/nextgen/blob/b996c911ff3477324a4cfa3d5427944e018d98f0/cmd/server/config.go#L38-L53

So a plain standalone deployment that pins its own project now mints a personal team for every registering user. #736's stated doctrine was that no environment gets platform provisioning silently, which is why the flag is off by default; the same boolean should gate this write. Passing `cfg.Platform.BootstrapProject` (or an empty id when it is false) keeps the effect off outside platform mode.

`internal/service/claim.go` does gate on `ResolvedProjectID()`, but that is a deny gate returning 403. This one writes rows.

internal/service/flow_create_user_with_personal_team_handler.go
```diff
@@ -0,0 +1,171 @@
...
+// registration identifier. The name is user-facing (team lists, the claim
+// 409's owning-team details) and unique per project by the teams unique-name
+// constraint — which is why the email is used verbatim: it is already unique
+// among registrations, so a name collision carries the same signal as
+// user_already_exists and is mapped to the same step error above.
+func personalTeamName(attributes map[string]any) string {
+	if email, ok := attributes["email"].(string); ok && strings.TrimSpace(email) != "" {
+		return email
+	}
+	return "Personal"
```
IAM-marco: If this file survives the duplicate-implementation question: the literal `Personal` fallback is not unique, so it is safe only for the first user.

Teams carry `UNIQUE (project_id, name_lower)`:

https://github.com/zitadel/nextgen/blob/b996c911ff3477324a4cfa3d5427944e018d98f0/internal/storage/dialect/postgres/migration/sql/000002_teams.sql#L16-L18

The second platform registration whose attributes carry no `email` collides on that index, and `Handle` maps the resulting `UniqueError` to a `user_already_exists` step error (line 114) — the registration is blocked and the visitor is told their account already exists.

`personal_team.go` derives a collision-free fallback for the same case:

https://github.com/zitadel/nextgen/blob/b996c911ff3477324a4cfa3d5427944e018d98f0/internal/service/personal_team.go#L119-L121

IAM-marco (changes requested):
Requesting changes. Three points, heaviest first.

1. The branch carries two implementations of the same feature and wires both. `cd649c04` (the `create_user_with_personal_team` on_success design) is never reverted by `b996c911`, and the on_success half is unreachable: `default-login.json` still declares `create_user`. It still drags the enum, the enumer output, OpenAPI, the meta-schema, the TS validator and a second changeset behind it. The description states there is no new on_success value and no schema/OpenAPI/wire change, which does not match the diff. Dropping `cd649c04` settles this and takes point 3 with it.

2. The platform gate re-derives "is this the platform deployment" from `ResolvedProjectID()`, but #736 already added the explicit signal for that question: `platform.bootstrap_project`. As written, a standalone deployment that pins `platform.project_id` mints a personal team for every registering user.

3. The `Personal` fallback team name is not unique, so the second registration without an email collides on the teams unique index and is rejected as `user_already_exists`.

Details inline.

cmd/server/server.go
peintnermax: this is an accident of my branch surgery.. 
I'll remove the on_success dead code and clean it up

internal/service/personal_team.go
peintnermax: Agreed, fixed. The gate now runs through a new PlatformConfig.ProvisioningProjectID(): the built-in platform id only under platform.bootstrap_project, empty otherwise — so a standalone pin never opts a deployment into team provisioning

internal/service/flow_create_user_with_personal_team_handler.go
peintnermax: That literal only existed in the discarded handler that shouldn't have been on the branch

internal/service/personal_team.go
```diff
@@ -0,0 +1,132 @@
...
+	return nil
+}
+
+// personalTeamName derives the team's display name from the user's email —
+// user-facing (team lists, the claim 409's owning-team details) and unique per
+// project alongside the identifier itself. Determinism is load-bearing: it is
+// what turns a concurrent double-ensure into a unique-name conflict instead of
+// two teams. A user whose email cannot be read falls back to a name derived
+// from the user id — equally deterministic and collision-free.
+func (s *personalTeamService) personalTeamName(ctx context.Context, projectID, userID string) string {
```
IAM-marco: The self-heal cannot converge when the derived name is permanently unacceptable, because the name is deterministic.

`personalTeamName` fails closed in two ways that no later sign-in can escape:

1. The name collides with an unrelated existing team. Teams are `UNIQUE (project_id, name_lower)`, so if any team on the platform project is already named with this user's email, `CreateTeam` returns a `UniqueError`, the recovery lookup on line 110 finds no membership, and the ensure errors out.
2. The name is rejected by `domain.NewTeam`. This is the case copilot flagged separately in this file: emails longer than `TeamNameMaxLength` (200) fail `ValidateTeamName` before the transaction is ever opened.

In both cases the call site logs a warn and moves on, which is correct. But the next sign-in recomputes the identical name, hits the identical failure, and logs the identical warn. `claim/complete` then stays 403 forever, which is the one outcome the PR sets out to make unreachable. The determinism that makes the concurrent double-ensure safe is the same property that makes these two failures permanent.

Both fall out of one decision: the team name doubles as the idempotency key. Worth separating those roles, or giving `personalTeamName` a terminal fallback that cannot fail. The `"Personal " + userID` branch on line 131 already is one, since it is bounded and collision-free by construction: reaching for it when the email is unusable, rather than only when it is unreadable, closes both cases.

https://github.com/zitadel/nextgen/blob/f655f7f43657919d51a6a2ff7e777915eb1ec66d/internal/service/personal_team.go#L124-L132

IAM-marco (commented):
The three points from the first round are addressed and resolved: the duplicate on_success implementation is gone, `ProvisioningProjectID()` gates the effect on `platform.bootstrap_project` with tests covering the exact doctrine, and the non-unique fallback name went with the deleted handler. The `ProvisioningProjectID` split reads well, and the comment on the asymmetry with `claimService` is the right place for it.

One remaining point inline, on the durability of the self-heal.

internal/service/personal_team.go
IAM-marco: TL;DR: I would make it so `personalTeamName()` returns an auto-generated ID. Can directly use the user ID since that should be unique per DB (or at least per project)

peintnermax: We only need a registered user to end up with an initial team — the one claim/complete attaches to. It's deliberately not bound to whoever triggered it: an ordinary team that can be renamed, shared, or carry other members later, with "personal" staying an inferred notion (earliest active membership) rather than a modeled one.

The name follows from that: "max@example.com personal team", a convenience only, changeable via PATCH /teams/{id}. Email is x-unique: "project" in the platform schema, so it's collision-free in practice (if we do not change the uniqueness of email in the platform); if the name is taken anyway it falls back once to Personal <user_id> — the requirement is that everyone gets a team, not a pretty name.

Is that state fine? or is there anything we should ensure here?

IAM-marco (approved):

PR #981, opened 2026-08-26
feat(api): add POST /users/query

vitorbari:
## Summary

This PR adds `POST /users/query`, following the structure defined on [ADR 031](https://github.com/zitadel/nextgen/blob/main/docs/adrs/031-openapi-querying.md), alongside the existing `GET /users` which will be removed on upcoming PR. 

Screenshot 2026-08-27 at 10 47 23

Filter and sort by `created_at`, `id`, `schema`, `status`, or `lifecycle_owner_team_id`, with the same cursor contract as `POST /sessions/query` and `POST /teams/query`.

### DIFF
--- a/api/openapi/endpoints/users/query/methods.yaml
+++ b/api/openapi/endpoints/users/query/methods.yaml
@@ -1,0 +1,48 @@
+post:
+  operationId: queryUsers
+  summary: Query users
+  description: |
+    Returns the users of a project, paginated with a cursor.
+
+    The project comes from the credential, not from a parameter: the operation
+    is bound to the token's own project by construction. This is why it takes
+    no `project_id`, unlike the other query endpoints.
+  tags:
+    - Users
+  security:
+    - oauth2:
+        - user.read
+  requestBody:
+    description: Request to query the users of a project.
+    required: true
+    content:
+      application/json:
+        schema:
+          $ref: query-users-request.yaml
+  responses:
+    '200':
+      description: The list of users was returned successfully
+      content:
+        application/json:
+          schema:
+            $ref: query-users-response.yaml
+    '400':
+      description: Invalid request (e.g., unknown filter field or unusable value).
+      content:
+        application/json:
+          schema:
+            $ref: ../../../components/error-details.yaml
+    '401':
+      description: Missing or invalid credentials.
+      content:
+        application/json:
+          schema:
+            $ref: ../../../components/error-details.yaml
+    '403':
+      description: Insufficient permissions.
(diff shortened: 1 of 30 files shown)


api/openapi/endpoints/users/query/user-filter-field.yaml
```diff
@@ -0,0 +1,19 @@
+type: string
+description: |
+  Field to filter users by:
+  - `created_at`: RFC3339 timestamp
+  - `id`: user id
+  - `schema`: id of the schema the user's attributes satisfy
```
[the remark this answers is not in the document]
vitorbari: We are using opaque IDs for now.

api/openapi/endpoints/users/query/methods.yaml
```diff
@@ -0,0 +1,48 @@
...
+      content:
+        application/json:
+          schema:
+            $ref: ../../../components/error-details.yaml
+    '403':
+      description: Insufficient permissions.
+      content:
+        application/json:
+          schema:
+            $ref: ../../../components/error-details.yaml
```
wim07101993, line 46: I think all these `error-details` responses can be replaced with `queryUsers-error-response`

api/openapi/endpoints/users/listUsers-error-response.yaml
```diff
@@ -28,6 +30,11 @@ content:
         value:
           code: internal
           message: 'An unexpected error occurred.'
+      not_implemented:
```
wim07101993, line 33: is this the latest state? because I don't see any not-implemented errors anywhere  in code.

vitorbari: This is just transient state.

This PR adds filter mapping to `service.ListUsers`, and both `Handler.ListUsers` and `Handler.QueryUsers` call that one method, so the generator attributes the filter errors to both.

#982 removes `GET /users` and deletes this file with it.

api/openapi/endpoints/users/query/methods.yaml
vitorbari: I have opened a follow up issue for this: https://github.com/zitadel/nextgen/issues/1076
It affects other existing endpoints.

livio-a (approved):
the query endpoint with the lifecycle filter and the separate filter/sort enums is the right base for the rest of the stack. one note, not blocking: null filter value is rejected, so "self-owned only" can't be selected. fine for now, just remember it when a "managed by: self" view comes up.

PR #986, opened 2026-08-26
feat(api): return the full flow definition from listFlowDefinitions

### LINKED ISSUE #939: listFlowDefinitions returns none of what the flows directory renders
## Context

Flows directory

The flows directory renders one row per flow definition: the flow's name, the purposes it serves (`Login + Register`), one tag per step (`identifier`, `password`, `register`, `passkey-upsell`), the user schema it operates on (`Minimal`), and a last-changed date.

`GET /flow-definitions` returns `id`, `name`, `project_id`, `status`, `created_at` and `updated_at` (`api/openapi/components/flows/flow-definition-response.yaml`). Purposes, steps and user schema are not among them, so nothing in the middle of the row can be rendered from a list call.

The server already holds all of it. `domain.FlowDefinition` carries `UserSchema`, `Purposes`, `Audience` and `Steps` (`internal/domain/flow_definition.go:115-130`), storage selects the whole `definition` column, and `flowDefinitionResponse` drops every one of them on the way out (`internal/api/flow_definition.go:281-290`).

Two further mismatches sit in the same mapper. `schema_uri` is declared on the response but never populated. And the two read endpoints describe a flow definition differently: `GET /flow-definitions/{id}` returns `flow-definition-detail-response.yaml`, which nests the whole document under `flow_definition`, while the list returns the flat `flow-definition-response.yaml`. Schemas has the identical split, tracked as #921.

## Description

Listing flow definitions cannot render a flow definition. The list response is an envelope with no document in it, and the fields the directory needs are read from the database and then discarded a layer later.

The list response should carry what a row needs, and the two read endpoints should converge on one description of a flow definition rather than the two disjoint halves they give today. `steps` is the open question: the full graph is large and the row only needs each step's name, so the list may want a reduced projection rather than the whole array.

## Acceptance Criteria

- [ ] A single `GET /flow-definitions` call returns everything a directory row renders — name, purposes, step names, user schema, last-changed date.
- [ ] `GET /flow-definitions` and `GET /flow-definitions/{id}` describe a flow definition the same way, as one representation rather than two.
- [ ] The response no longer advertises `schema_uri` while never setting it — it is either populated or removed.
- [ ] The console flow list renders the directory design with no per-row follow-up request, and shows last change rather than creation date (`apps/console/src/routes/_authed/flow-definitions/index.tsx`).

Related: #940, #942. The list/get split is the same one #921 describes for schemas.

grvijayan:
## Summary

`GET /flow_definitions` returned a flat summary (`id, name, project_id, schema_uri, status, timestamps`) while `GET /flow_definitions/{id}` nested the whole document under `flow_definition`. Purposes, steps and user schema were read from storage and dropped in the API mapper, so a directory row could not be rendered from a list call (#939).

- Both reads now serve the same representation, `flow-definition-response.yaml`. The flat summary is gone, and with it `schema_uri`, which the response declared and the server never populated.
- `FlowDefinitionDetailResponse` is renamed to `FlowDefinitionResponse`. Every read serves it now, so "detail" distinguished nothing. Own commit; the generated diff is a pure rename.
- `POST` and `PUT /flow_definitions` answered with zero `created_at`/`updated_at`: the writes ran `NOW()` without reading it back. All three dialects now `RETURNING` the columns into the entity. Own commit. Surfaced by the new list test, which compares each row to its create response, timestamps included.
- The api-mock's detail response emitted a top-level `status` and `schema_uri` that were never in the spec. Gone; the conformance test now asserts a list row `toEqual` the by-id document.
- The console list adapts to the nested shape. Rendering purposes, steps and last-changed (AC4) lands separately.

Breaking for clients reading `name` or `status` from the top level of a list row: they now live under `flow_definition`.

## Validation

- `go build ./... && go vet -tags postgres_integration ./...`
- `go test -tags postgres_integration ./internal/api/integration_test/ ./internal/storage/dialect/postgres/ ./internal/storage/stmttest/`
- `go test -tags sqlite_integration ./internal/storage/stmttest/`
- `go test -tags spanner_integration ./internal/storage/dialect/spanner/ ./internal/storage/stmttest/` (emulator)
- `moon run api:generate api:build`, then `pnpm test` in `packages/api-mock` (67 pass) and `apps/console` flow-definitions specs; `tsc --noEmit` in console and cli

## Release notes / changeset

- Changeset: `.changeset/flow-definition-list-representation.md` — `@zitadel/server` minor: list representation, create/update timestamps

## Notes

- Postgres needed `.UTC()` after the scan: pgx returns `timestamptz` in Local while the read path (`flowdefinition.ToDomain`) is UTC. Without it create and list disagree on location only.
- `steps` in the list is the full graph, not a reduced projection. The issue leaves that open; one representation was the priority, and a projection can be added if the payload turns out to matter.
- Refs #939, not closes: AC4 (console rendering) is outstanding.


internal/api/flow_definition.go
```diff
@@ -151,25 +151,21 @@ func mapFlowDefinitionRequestToService(projectID string, schemaURI api.OptSchema
...
 			}
-			s.Actions = actions
+			actions = append(actions, domain.FlowStepAction{
+				Name:    apiAction.GetName(),
+				Kind:    kind,
+				Primary: apiAction.GetPrimary().Value,
+				TextKey: apiAction.GetTextKey().Value,
+			})
 		}
+		s.Actions = actions
```
vitorbari, line 168: we lose nil-vs-empty with this

grvijayan: The [storage layer doesn't persist empty `[]`](https://github.com/zitadel/nextgen/blob/main/internal/storage/flowdefinition/content.go#L106). 
In these changes, we're only aligning the create/update response with get.

vitorbari (approved):

PR #1026, opened 2026-08-28
feat(api): add grant create get and revoke

adlerhurst:
## Summary

- Add `POST /grants`, `GET /grants/{id}`, and `DELETE /grants/{id}` so a project secret can bind a user or team to `project.viewer`, `project.editor`, or `project.admin`.
- Create resolves the principal through `resource_scope_index` (home must be the platform project when one is pinned), loads the user/team, and writes an `authz_assignments` row plus `authz.granted`.
- Get returns an unrevoked collaboration grant by `(project_id, id)`. Misses, revoked rows, project-secret setup (`sk_proj`), owning-team (`relation=team`), and non-`project` object types are `grant.not_found`.
- Delete soft-revokes a managed grant and emits `authz.revoked` in the same transaction. Already-revoked, missing, setup, owning-team, and non-`project` object-type grants are 404; rows are not un-revoked.
- Assignment IDs stay `asgn_`. Public error codes use the `grant.*` prefix.
- Grant create filters principals with `status = active` at storage (`GetUser` / `GetTeam(filter)`).
- Expired grants stay visible on GET until DELETE. The unique index is `revoked_at IS NULL` (cannot include `now()`), so re-granting the same binding requires DELETE first. Authorization still ignores expiry.
- Get/Revoke class is the Create triple: principal type user or team, `object_type` project, relation in `{viewer, editor, admin}`.

## Validation

- Merged `origin/main` into this branch. The only content conflict was `api/generated/oas_router_gen.go` (ogen route-node numbering vs `GET /users` → `POST /users/query`). Resolved by regenerating with `node scripts/go-generate.mjs` from the merged OpenAPI spec (both `/grants` and `/users/query` present).
- `go build ./...`
- `go test ./internal/service -run TestGrantService` — including sk_proj/owning-team Get/Revoke 404, expired GET still returns, inactive user, unique conflict, revoke emits `authz.revoked`, non-project `object_type` 404
- `go test ./internal/api`
- HTTP tests added for revoke-then-recreate, GET/DELETE 404 on seeded setup and owning-team ids. Postgres was not reachable in this environment (no Docker, no `ZITADEL_TEST_POSTGRES_URL`); those cases should be run with `go test -tags postgres_integration -timeout 10m -v ./internal/api/integration_test -run TestGrantCreateGetRevoke`
- `gofmt -l` on the touched Go files (clean)

## Release notes / changeset

Changeset: `.changeset/grant-api-create-get-revoke.md` — Add grant create, get, and revoke APIs for project viewer/editor/admin bindings. Lists `@zitadel/server` (`minor`).

No extra changeset for the review follow-up or this merge — same grant API slice.

## Notes

- Out of scope for this slice: `POST /grants/query` (#1020), `PATCH /grants/{id}` (#1021), human `__nextgen_session` / CSRF (#1022), foothold SQL home-project (#1023), `grant.*` catalog gate (#1024), Console UI (#1025).
- `DELETE /grants/{id}` is the live producer for previously deferred `authz.revoked` (#883).
- Callers are project secrets. Owning-team (`project.team`) grants stay on claim; Get/Revoke treat those rows as `grant.not_found`.
- OpenAPI oauth2 scopes stay `project.write` / `project.read`; unused `grant.*` scopes are not added.
- `grant.principal_not_found` stays a distinct code with message “the principal could not be resolved” (not collapsed into `grant.not_found`).
- Expired grants are not auto-revoked on create; replace/PATCH is #1021.

<div><a href="https://cursor.com/agents/bc-fbf162d0-605d-46c1-816e-f39a29e49cec?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-fbf162d0-605d-46c1-816e-f39a29e49cec&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


internal/service/grant.go
```diff
@@ -0,0 +1,223 @@
...
+	return scope.ProjectID, nil
+}
+
+func (s *GrantService) loadPrincipal(ctx context.Context, stmts AllStatements, homeProjectID string, principalType domain.AuthzPrincipalType, principalID string) error {
+	switch principalType {
+	case domain.AuthzPrincipalTypeUser:
+		user, err := stmts.GetUser(ctx, database.And(
+			database.Equal(database.Col(domain.UserFieldProjectID), homeProjectID),
+			database.Equal(database.Col(domain.UserFieldID), principalID),
+		), UserQueryOptions{})
```
adlerhurst: Is there no possibility to filter only for the active users? to get rid of lines 187-190?

internal/service/grant.go
```diff
@@ -0,0 +1,223 @@
...
+		if err != nil {
+			if _, ok := errors.AsType[*database.NoRowFoundError](err); ok {
+				return domain.ErrGrantPrincipalNotFound()
+			}
+			return err
+		}
+		switch team.Status {
+		case domain.TeamStatusDeactivated, domain.TeamStatusPendingPurge:
+			return domain.ErrGrantPrincipalNotFound()
+		}
```
adlerhurst: Maybe it makes sense to add a `statements.GetTeam`-method which allows to define the filters (including the status) like in users.

internal/service/grant.go
```diff
@@ -0,0 +1,220 @@
...
+		if de, ok := errors.AsType[domain.Error](err); ok {
+			return nil, de
+		}
+		return nil, domain.ErrInternal(err).WithMessage("failed to create grant")
+	}
+	return created, nil
+}
+
+func (s *GrantService) Get(ctx context.Context, projectID, id string) (*domain.AuthzAssignment, error) {
```
livio-a, line 107: **blocking** — `Get` and `Revoke` load by bare `(project_id, id)`, but `authz_assignments` also holds rows this API must not touch: the `sk_proj` setup assignment seeded on project create and the owning-team grant written by claim (`relation = "team"`).

Create correctly allowlists principal types and relations, but nothing symmetric guards this side, so with the project secret:

- `DELETE /grants/{id}` on the owning-team assignment revokes it — and since `GetActiveOwningTeamGrant` / `ListClaimedProjectIDs` key on `relation = 'team' AND revoked_at IS NULL`, the project becomes claimable by a *different* team. The ids are discoverable via the events API (`authz.granted` carries `entity_id`), so a leaked project secret becomes an ownership-transfer primitive. That contradicts "Owning-team grants stay on claim / rows are not un-revoked". Revoking the `sk_proj` seed row is the same bug as self-lockout.
- `GET /grants/{id}` on those rows emits `principal_type: "sk_proj"` / `relation: "team"`, which are outside the enums `grant.yaml` declares.

Suggestion: treat rows outside this API's class as `grant.not_found` — e.g. check `principal_type IN (user, team) && relation IN allowedGrantRelations` after the load (or push it into the query), in both `Get` and `Revoke`, and add an integration test that GET/DELETE on the seed and owning-team assignment ids 404.

internal/service/grant.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+		if err := emitAuthzGranted(ctx, tx.Statements(), asgn); err != nil {
+			return err
+		}
+		created = asgn
+		return nil
+	})
+	if err != nil {
+		if _, ok := errors.AsType[*database.UniqueError](err); ok {
+			return nil, domain.ErrGrantAlreadyExists().WithParent(err)
```
livio-a, line 94: **blocking (or at least decide + document)** — expired grants are invisible to this mapping but still occupy the partial unique index (`WHERE revoked_at IS NULL`, no expiry term). Sequence: grant with `expires_at`, let it lapse, POST the same binding again → 409 `grant.already_exists` with the message "an *active* grant … already exists", which at that point is untrue. The caller has to know to DELETE the expired grant first. `Get` (line 96) likewise 404s revoked rows but happily returns expired ones as if active, while the OpenAPI description says "Loads a single active grant".

Either account for `expires_at` in Get/Create (e.g. treat expired like revoked here, or auto-revoke on conflict with an expired row), or document explicitly that expired grants stay visible and must be deleted before re-granting. Current behavior is a support-ticket generator.

internal/domain/grant.go
```diff
@@ -0,0 +1,25 @@
...
+
+func ErrGrantNotFound() Error {
+	return newError(PrefixGrant.ErrorCodePrefix("not_found"), "grant not found", nil, nil)
+}
+
+func ErrGrantAlreadyExists() Error {
+	return newError(PrefixGrant.ErrorCodePrefix("already_exists"), "an active grant with this principal and relation already exists on the project", nil, nil)
+}
+
+func ErrGrantPrincipalNotFound() Error {
```
livio-a, line 19: `grant.principal_not_found` carries the message "grant not found". If this is meant as an anti-oracle it doesn't work — the distinct error code already discloses which case occurred (and any own-project secret can probe platform principal ids via create anyway: 201 vs `principal_not_found`). If we want the oracle closed, return `grant.not_found` itself; if we don't, give this an honest message like "the principal could not be resolved". The half-measure is just confusing, and it's baked into the generated `grant-principal_not_found.yaml` description too.

internal/service/grant.go
```diff
@@ -0,0 +1,220 @@
...
+		return "", err
+	}
+	wantKind := domain.ResourceKindUser
+	if principalType == domain.AuthzPrincipalTypeTeam {
+		wantKind = domain.ResourceKindTeam
+	}
+	if scope.ResourceKind != wantKind {
+		return "", domain.ErrGrantPrincipalNotFound()
+	}
+	if s.platformProjectID != "" && scope.ProjectID != s.platformProjectID {
```
livio-a, line 190: When `platformProjectID` is empty (bootstrap mode) the pin check is skipped entirely, so a principal homed in *any* project can be bound. That mirrors the claim service and is probably intended, but it's worth a sentence here — the PR body only states the pinned case ("home must be the platform project when one is pinned"), and the next reader will wonder whether the unpinned fall-through is a decision or an oversight.

internal/api/integration_test/grant_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	api "github.com/zitadel/nextgen/api/generated"
+	"github.com/zitadel/nextgen/internal/api/integration_test/helpers"
+	"github.com/zitadel/nextgen/internal/service"
+)
+
+func TestGrantCreateGetRevoke(t *testing.T) {
```
livio-a, line 17: Nice coverage overall. Two cases I'd still add:

- **revoke → re-create of the same binding.** That this works at all hangs on the unique index being partial on `revoked_at IS NULL`; a test pins that down against a future index change.
- Once the Get/Revoke scoping fix lands: GET/DELETE against the seeded `sk_proj` and owning-team assignment ids returning `grant.not_found`.

internal/service/team.go
```diff
@@ -66,22 +66,30 @@ func (s *TeamService) Create(ctx context.Context, input CreateTeamInput) (*domai
...
-	if team.Status == domain.TeamStatusPendingPurge {
-		return nil, domain.ErrTeamNotFound()
-	}
 	return team, nil
 }
 
+// visibleTeamStatusFilter is the read-side allowlist for HTTP get and list.
+// A pending_purge team is awaiting deletion by the cleanup job (#622 dropped
+// that status from the team lifecycle), so reads treat it as already gone.
+func visibleTeamStatusFilter() database.Filter[domain.TeamField] {
```
livio-a, line 86: Praise, no action — moving the visibility allowlist into SQL and sharing it between Get and List is a clear improvement, and leaving `GetTeamByID` unfiltered for bootstrap/stmttest with the callers audited in the PR notes is the right call.

livio-a (commented):
Reviewed the non-generated diff. Two blocking items (Get/Revoke aren't scoped to the grant class this API manages; expired-grant 409/Get semantics), plus a few smaller notes inline. Everything else looks solid — race handling, event bookkeeping, and the team status-filter refactor in particular.

internal/service/grant.go
```diff
@@ -0,0 +1,237 @@
...
+	"viewer": {},
+	"editor": {},
+	"admin":  {},
+}
+
+// isManagedGrant is the class this HTTP API may Get or Revoke. Create already
+// allowlists the same principal types and relations. Other authz_assignments
+// rows (project-secret setup, owning-team / claim) live in the same table and
+// must 404 here so a project secret cannot self-lockout or transfer ownership.
+func isManagedGrant(a *domain.AuthzAssignment) bool {
```
livio-a, line 25: One line missing to state the full invariant: the guard checks principal type and relation but not `ObjectType`, while `grantResponse` hardcodes `object_type: project` in the response. Today that's unreachable — Create only writes `ObjectType: "project"` and the other two producers are rejected by the principal/relation checks — but the catalog is designed to grow other object types, and a future producer writing e.g. a user→`(catalog_x, viewer)` assignment would be exposed, mislabeled as `project`, and revocable through this API. The relation namespace staying collision-free across future object types shouldn't be what this guard hangs on.

```go
case domain.AuthzPrincipalTypeUser, domain.AuthzPrincipalTypeTeam:
	if a.ObjectType != "project" {
		return false
	}
	_, ok := allowedGrantRelations[a.Relation]
	return ok
```

With that in, this matches everything Create can produce and I'm happy to approve.

livio-a (commented):
Both blocking items are addressed — thanks for the thorough turnaround. The `isManagedGrant` guard with unit + integration coverage against the real seeded rows is exactly right, and pinning the expired-grant decision in the docs, the 409 wording, and a unit test makes that behavior contractual instead of accidental. One hardening nit inline before I approve.

livio-a (approved):
All review points addressed across the two follow-up commits:

- `isManagedGrant` now pins the full triple Create writes — principal type, `object_type == "project"`, and the relation allowlist — so the setup (`sk_proj`), owning-team, and any future non-project rows 404 on GET/DELETE, with unit tests for each class and integration tests against the real seeded rows that also assert nothing got revoked.
- The expired-grant semantics are now contractual: "unrevoked" wording on the 409, `expires_at` docs spelling out that expiry stops authorization but not the unique binding, and a test pinning that GET still returns expired grants.
- `grant.principal_not_found` carries an honest message, the bootstrap-mode fall-through is documented at the pin check, and the revoke→re-create test pins the partial unique index.

Nice, focused turnarounds on all of it. LGTM.

PR #1036, opened 2026-08-28
fix: normalize page limit for list flow definitions

### LINKED ISSUE #942: listFlowDefinitions has no default limit, and the console never pages
## Context

Flows directory

`GET /flow-definitions` declares `limit` and `page_token`, and its response carries `next_page_token`. Storage supports paging: `ListFlowDefinitions` defaults the ordering to `created_at, id` descending (`internal/storage/flowdefinition/list.go`) and returns a `NextCursor`.

The limit is not normalised. The service passes it through as-is — `Limit: uint32(req.Limit)` (`internal/service/flow_definition.go:353-357`) — and `compileLimit` emits no `LIMIT` clause at all when it is zero (`internal/storage/dialect/postgres/compiler.go:248-253`). A caller that omits `limit`, which is what the console does, gets every flow definition in the project in one response. Every other list service normalises: events, projects, teams, sessions and users all call `normalizeLimit` (`internal/service/list.go:47-55`).

The console never pages either. It calls `api.listFlowDefinitions({ project_id })`, ignores `next_page_token`, and derives its three summary cards — `Flow definitions`, `Active`, `Inactive` — by counting and filtering the array it got back (`apps/console/src/routes/_authed/flow-definitions/index.tsx`).

## Description

The endpoint is unbounded by default, and the one screen that consumes it treats a single response as the whole set. That is invisible while projects have three flows and silently wrong once they do not: the list truncates at whatever a page turns out to be, and the summary cards report the page rather than the project.

## Acceptance Criteria

- [ ] `limit` has a default and a maximum consistent with the other list endpoints; omitting it does not return the whole table.
- [ ] A caller can page through every flow definition in a project using `page_token`, and each appears exactly once across the full set of pages.
- [ ] The console list pages rather than assuming one response is the whole set.
- [ ] The active/inactive summary counts describe the project, or are removed — they do not silently describe one page.

Related: #939 (shares the response shape). #924 is the same gap on `listSchemas`.

wim07101993:
## Summary

Normalize page limit for list flow definitions.

`GET /flow_definitions` now bounds its result set. A request that omits `limit` returned every flow definition in the project; it now returns the same default page of 20 as the other list endpoints, capped at 100, and hands back a `next_page_token` to walk the rest.

## Validation

- `go test ./...`
- `go test -tags postgres_integration`
- `go test -tags sqlite_integration`
- `go test -tags spanner_integration`

## Release notes / changeset

- Changeset: .changeset/flow-definitions-list-paging.md — Normalize page limit for list flow definitions.

## Notes

-


wim07101993 (commented):
(deleted)

grvijayan (approved):

PR #1052, opened 2026-08-28
docs: note stmttest is already cached per dialect

adlerhurst:
## Summary

- `stmttest` is one package, but it is **already** cached per adapter. Each `-tags *_integration` compile includes a different `open_*.go`, so `go test` hashes a different test binary. Moon's `server:test-postgres` / `server:test-spanner` / `server:test-sqlite` are already separate task hashes.
- An earlier revision of this PR added a sentinel `TestAdapterCacheKey` plus `ZITADEL_STMTTEST_ADAPTER` so `Getenv` would run inside a Test (golang/go#44625). That duplicated a key the compiler already has, split local `go test` and Moon GOCACHE slots, and leaked cache policy into the statement-contract suite.
- This lands the real answer in `internal/storage/AGENTS.md` and drops the probe.

## Validation

- `go list -test -tags {postgres,spanner,sqlite}_integration ./internal/storage/stmttest/` — each tag compiles only that adapter's `open_*.go`
- `moon run server:format` (clean)
- `node scripts/check-pr-title.mjs --title "docs: note stmttest is already cached per dialect"`

## Release notes / changeset

## Notes

- Thermo-nuclear review: do not approve the sentinel Test; delete it rather than move `Getenv` into `forEachDialect`.
- Local `go test -tags postgres_integration ./internal/storage/stmttest/` then the sqlite tag should not `(cached)`-hit the first run. The binaries differ.

<div><a href="https://cursor.com/agents/bc-01a048dc-eb27-7759-a364-b614fb9fc6db?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a048dc-eb27-7759-a364-b614fb9fc6db&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


wim07101993 (approved):

IAM-marco:
Correct, and worth having written down. Verified: each `-tags *_integration` compile pulls in a different `open_*.go`, so the three `stmttest` binaries hash differently and no dialect can replay another's result.

Merging this alongside #1007 rather than instead of it, because they answer different questions. This one answers *"does the postgres lane replay the sqlite lane's result?"* — no, and that is what the per-tag binary hash guarantees. #1007 is about the **same lane, same binary, different database**, which no amount of per-dialect separation covers:

- Run 32741899403: the real-instance spanner step replayed the **emulator** lane's result. Same `spanner_integration` tag, same binary, different backend. There is no second `open_*.go` to tell those apart.
- Run 32743380406: the postgres lane printing `ok … (cached)` on an unchanged tree against a freshly created container.

One heads-up on the wording once #1007 lands: the sentence says Moon hashes the three `server:test-*` tasks separately, which is true, but #1007 sets `options.cache: false` on all three, so Moon will not cache them at all. The Moon half of this note becomes moot then, while the `go test` binary-hash half stays accurate and useful. Might be worth a follow-up line at that point so the two documents do not appear to contradict each other.

Issue #940, opened 2026-08-21
Flow definitions name their user schema by id, but the directory shows its name


vitorbari:
## Context

![Flows directory](https://github.com/user-attachments/assets/158993ae-804e-42b2-b57c-659d895e2a02)

Each directory row labels its flow with a user schema by name — `Minimal`, `Consumer`, `Business`.

`user_schema` on a flow definition is the schema's server-assigned identifier, explicitly opaque to the client (`api/openapi/components/flows/flow-definition.yaml`). The name the design shows lives inside the customer-authored schema document, which only `GET /schemas/{id}` returns.

So even once the list carries `user_schema`, a client still needs one schema fetch per row to render the label — the same N+1 the schema directory has, and the same one #921 removes there.

## Description

The identifier the flow definition stores and the label the directory shows are not the same thing, and nothing in the flow-definition API bridges them. Either the flow-definition list resolves the schema's display name server-side, or the console reads it from a schema list response that already carries it (#921), and the flows directory joins the two client-side against a single schema call rather than one per row.

Which of the two is right depends on #921 landing, so this wants to be decided after it.

## Acceptance Criteria

- [ ] The flows directory shows each flow's user schema by its display name, not by its identifier.
- [ ] Rendering the directory costs a bounded number of requests — no per-row schema fetch.
- [ ] A flow definition whose user schema was deleted, or whose name cannot be resolved, renders without breaking the row.

Related: #939. Depends on #921 deciding where a schema's display name is readable from.

IAM-marco:
@vitorbari scoping question before this gets implemented.

#939 closed with PR #986 delivering the server half: `listFlowDefinitions` now returns the full flow definition. The console side of its acceptance criteria never landed, though: the directory still renders Name / Status / Created only (`apps/console/src/routes/_authed/flow-definitions/index.tsx`), with no purposes, no step tags, and creation date instead of last-changed.

This issue's acceptance criteria cover only the user schema label. The plan for it is a client-side join, no API change: the flows loader fetches `listSchemas` alongside `listFlowDefinitions` and resolves names with the existing `schemaDisplayName` helper (title, else objectType, else id), falling back to the raw id when a schema is deleted or unreadable.

The question: where does the rest of the directory row belong?

1. Fold purposes / step tags / last-changed into this issue's PR (pure rendering, the list response already carries the data), or
2. keep this one label-only and file a follow-up issue for the rest of the row, or
3. it is already parked somewhere we have not found.

[...]

PR #1053, opened 2026-08-28
feat(api): filter listSchemas by ids

IAM-marco:
Part of #940. Consumer PR stacked on top: #1054.

`GET /schemas` accepts a repeated `id` query parameter (up to 100 entries) that exact-matches schema ids. A consumer holding a set of known ids, such as the `user_schema` references on flow definitions, resolves all of them in one request instead of one `getSchemaById` per schema. Ids that match nothing are simply absent from the result, and the filter composes (AND) with `object_type`, `kind`, and `revisions`.

<details>
<summary>Contract and server details</summary>

- **Spec**: `api/openapi/endpoints/schemas/methods.yaml` declares the param as `form`/`explode` with `maxItems: 100`. The generated ogen decoder enforces the cap: exceeding it maps through the existing decode-error branch to the same `request.invalid` 400 a malformed page token gets, so no hand-written validation exists.
- **Server**: the handler passes `params.ID` into `ListSchemasInput.IDs`; the service appends an OR-of-equals filter on the `url` column (the primary key) to the existing filter slice. No new storage primitives, works unchanged in all three dialects.
- **Permissions**: the filter sits under the existing enforcement. Every query ANDs `project_id` from the authorized project, so a cross-project id probe matches nothing (same as an unknown id, no existence oracle), and partial access keeps its row-level EXISTS predicate, which the id filter ANDs under.
- **TS client**: `packages/api/src/generated` is gitignored and regenerates at build time; the regenerated client emits `id?: string[]` with correct explode handling, so there is no client diff to land.

</details>

<details>
<summary>Tests</summary>

- `TestListSchemasIDFilter` (integration): ids subset returns exactly those schemas, including a superseded revision id; unknown id yields no row and no error; composes with `kind` and `object_type`; an id from another project returns empty; 101 ids return 400 `request.invalid`.
- `TestListAuthzTeamScopedOnlyPartialView` gained an id-filtered assertion: a team-scoped-only principal asking for a specific schema id still gets an empty list, proving the filter ANDs under the authz row predicate rather than around it.
- Full postgres integration suite, `go test ./internal/api/... ./internal/service/...`, `server:check-generate`, `server:openapi`, and `workspace:check-openapi-rules` are all green.

</details>


wim07101993 (approved):

IAM-marco:
Closing, and thanks @wim07101993 for the review. @vitorbari's direction on #940 is to embed the user schema in the flow-definition response via the `expand` pattern (ADR 059, #984) rather than have clients join against a schema list, so the public `id` filter loses its consumer for now. The service-layer OR-of-equals filter on `url` built here returns as an internal detail of the expand hydrator in the follow-up PR. If a public by-ids filter finds a consumer later (the users directory fan-out remains a candidate), this branch is reusable as is.

PR #1054, opened 2026-08-28
feat(console): label flow definitions with their user schema

### LINKED ISSUE #940: Flow definitions name their user schema by id, but the directory shows its name

IAM-marco:
Closes #940. Stacked on #1053; merge that first.

The flows directory now labels each flow definition with its user schema's display name instead of nothing. The loader collects the distinct `user_schema` ids from the flow list and resolves them with a single `listSchemas` call using the new `id` filter from #1053, then derives names with the existing `schemaDisplayName` helper (`title`, else `objectType`, else id).

A row never breaks on an unresolvable schema: a deleted schema, an id the response did not contain, and a failed schemas fetch all fall back to the raw id rendered muted/mono. The fetch itself is catch-and-degraded, so the page renders even when `listSchemas` errors.

<details>
<summary>Details and tests</summary>

- One new "User schema" column in `apps/console/src/routes/_authed/flow-definitions/index.tsx`; a single code path where the map lookup falls back to the id.
- Fetching by explicit ids (rather than `revisions: latest`) matters because schema revisions are id-distinct and a flow pins the id it was created with, so flows referencing non-latest revisions are the common case over time.
- Spec cases added to `flow-definitions.spec.tsx`: resolved name shown; schema absent from the response leaves the raw id; `listSchemas` rejecting still renders the page with ids.
- `moon run console:test` (Node 24): 121 tests green; typecheck and lint clean.

</details>

Follow-ups deliberately out of scope: the rest of the directory row (purposes, step tags, last-changed, the orphaned console half of #939) and switching the users directory's per-schema `getSchemaById` fan-out to the new id filter.

### DIFF
--- a/.changeset/console-flow-schema-name.md
+++ b/.changeset/console-flow-schema-name.md
@@ -1,0 +1,5 @@
+---
+"@zitadel/server": patch
+---
+
+The console's login flows directory shows which user schema each flow definition uses, labelled with the schema's own name instead of leaving it out. A schema that no longer resolves is shown by id rather than blanking the row.
--- a/apps/console/src/routes/_authed/flow-definitions/flow-definitions.spec.tsx
+++ b/apps/console/src/routes/_authed/flow-definitions/flow-definitions.spec.tsx
@@ -15,6 +15,28 @@ vi.mock("@/auth/session", async (importOriginal) => {
 vi.stubEnv("VITE_CONSOLE_API_BASE", "http://localhost/api");
 
 const FLOWS_URL = "http://localhost/api/flow_definitions";
+const SCHEMAS_URL = "http://localhost/api/schemas";
+
+/** One populated flow definition, pinning schema `sch_1`. */
+function flowDefinitionsResponse() {
+  return {
+    flow_definitions: [
+      {
+        id: "flow_1",
+        project_id: "proj_1",
+        flow_definition: {
+          name: "Login",
+          status: "active",
+          user_schema: "sch_1",
+          purposes: { login: "identifier" },
+          steps: [{ name: "identifier" }],
+        },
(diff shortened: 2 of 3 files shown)


IAM-marco:
Closing: @vitorbari scoped #940 to the API only (the flows directory page is a preliminary #451 version awaiting new designs), and the backend direction changed to embedding the schema via `expand=user_schema` (ADR 059), which makes this PR's client-side join obsolete. The frontend work now lives in #1077; this branch stays as salvage for the label column and its tests.

Issue #898, opened 2026-08-14
Password policies 


vitorbari:
## Context

Password authentication is supported in the MVP and needs secure defaults. We’ll base these on the latest [NIST SP 800-63B password guidance](https://pages.nist.gov/800-63-4/sp800-63b.html#passwordver), which covers password length, composition, compromised-password checks and protection against repeated guessing attempts.

Projects also need limited control where requirements legitimately vary between customers, particularly minimum password length and password reuse. This spec defines that small configurable surface while keeping the remaining password security behaviour as secure ZITADEL defaults.

Whether password authentication is available to a user type continues to be defined through the user schema.

## Goal

Provide NIST-aligned password handling by default, while allowing Projects to configure minimum password length and password history for their use cases.

## Product model

For MVP, each Project has one active password policy that applies to all password credentials in the Project. Its configuration and lifecycle follow the simplest case of the general policy model in [#899](https://github.com/zitadel/nextgen/issues/899).

More complex policy combination, inheritance and conflict resolution can be added through #899 when additional scopes are introduced.

### Minimum password length

- The minimum password length defaults to 15 characters.
- Projects can configure a minimum from 8 to 64 characters.
- The authoring workflow warns during validation when the configured minimum is below 15 characters and explains that it does not meet NIST’s requirement for passwords used as a single authentication factor.

### Password history

- Password history is disabled by default.
- Projects can prevent reuse of between 1 and 4 passwords.
- When password history is enabled, the user’s current password is included immediately.
- History accumulates through future password changes until it reaches the configured depth.
- Changing the configured history depth does not invalidate the user’s current password.

This addresses the customer request in [zitadel#8034](https://github.com/zitadel/zitadel/issues/8034), which has also come up directly from other enterprise customers.

### Secure password handling

ZITADEL must:

- Accept Unicode passwords containing between the configured minimum and 64 Unicode code points after NFC normalization.
- Normalize passwords using NFC before applying length requirements, blocklist checks, hashing or verification.
- Reject longer passwords with a clear validation error and process accepted passwords in full without truncation.
- Apply no composition rules or periodic password changes.
- Compare the complete normalized password against a built-in offline blocklist of commonly used, expected and compromised passwords whenever a password is created, changed or reset.
- Reject blocklisted passwords and explain that another password must be chosen.
- Update the built-in blocklist through ZITADEL releases without calling an external service while a password is being set.

These are secure ZITADEL defaults and are not configurable by Projects.

[...]

PR #1068, opened 2026-08-31
docs(adrs): operation policies

vitorbari:
Adds operation policies ADR.

Note for reviewers, read only up to `Advanced topics`, the second half of the document is not ready for review yet.


docs/adrs/059-operation-policies.md
```diff
@@ -0,0 +1,630 @@
...
+  "$id": "${SERVER_URL}/policies/catalog-v1.json",
+  "title": "PolicyDocument",
+  "type": "object",
+  "required": ["kind", "metaSchema", "operation"],
+  "properties": {
+    "kind":       { "const": "policy" },
+    "metaSchema": { "type": "string", "format": "uri" },
+    "operation":  { "type": "string" }
+  },
+  "oneOf": [
```
wim07101993, line 139: i guess the oneof should be under metaschema instead of the root? Since the values in the given json documents won't match the schema described above.

docs/adrs/059-operation-policies.md
```diff
@@ -0,0 +1,630 @@
...
+
+10. **Return error.**
+
+11. **Proceed with operation.**
+
+### The rule engine: OPA/Rego or Go
+
+**This is an open decision.**
+
+**Choosing Rego/OPA** means a policy is two files. Zitadel ships a `.rego` next to the `.json`, and the server embeds an evaluator to run it. The rule is also part of the release, a developer can modify it.
```
wim07101993, line 256: In general I would go for the OPA because of the growing standard and flexibility. However, I do have concerns about performance. Would these policies also run in the hot path? `user.password.verify` might I think.

vitorbari: Most likely it will be in the hot path.

You can evaluate the policy locally via OPA SDK or remote if OPA is running as a Daemon / Sidecar.

Remote is definitely slower because of the network call.

These are some benchmark numbers I found online, can give us an idea, but we can always run our own benchmarks for our specific use-case:

Screenshot 2026-09-01 at 09 37 09

docs/adrs/059-operation-policies.md
```diff
@@ -0,0 +1,630 @@
...
+> defines the product model; [#898](https://github.com/zitadel/nextgen/issues/898)
+> is the first consumer.
+> **Related:** [ADR 020](020-credentials-out-of-user-schema.md),
+> [ADR 035](035-configuration-environments.md),
+> [ADR 042](042-scaffolded-file-ownership-and-drift-detection.md),
+> [ADR 048](048-wide-events-internal-audit-primitive.md)
+
+## Context
+
+This ADR proposes the data model and enforcement mechanism for policies in nextgen.
```
elinashoko, line 16: I understand this ADR is meant to define the broader policy model, with password policies as the first use case, rather than define the full implementation of #898. So I don’t expect it to repeat every password journey or requirement in detail.

From a product point of view, though, we should make sure the examples don’t accidentally define a different product from the one in #898.

A few things need aligning:

* Projects can set the minimum password length between **8 and 64**. The default is 15, with a warning when they choose anything below 15.
* Passwords support up to **64 Unicode characters**. The maximum isn’t something Projects configure.
* Password history is off by default and can cover up to the previous **four passwords**.
* When history is switched on, the current password is included immediately.
* The blocklist and rate limiting are built-in ZITADEL protections, not settings Projects can switch off or adjust in the MVP.
* There is one active password policy for the whole Project. Customers shouldn’t need to understand or manage separate policies for saving and verifying passwords.
* The same policy applies wherever a password is set—whether that happens through a ZITADEL login, the Console, an API or an SDK.
* Making a policy stricter doesn’t suddenly invalidate existing passwords or sessions. The new requirements apply the next time the password is changed or reset.

The main product question for me is around allowing developers to replace the policy rule completely. We’ve intentionally kept the configurable surface small: Projects can choose the minimum length and password history, while ZITADEL owns the secure baseline.

A custom rule shouldn’t be able to remove those built-in protections. If custom rules are part of the broader model, they should only be able to add or strengthen requirements.

Separately, we updated #898 after the security review, so some differences are expected. The main updates to the spec were:

* Changed the lowest configurable minimum from **4 to 8** and added a warning below 15.
* Set the supported maximum to **64 Unicode characters**.
* Finalised password history as off by default and configurable up to **four passwords**, including the current password as soon as it’s enabled.
* Clarified that the common and compromised-password blocklist is built into ZITADEL and updated through ZITADEL releases.
* Clarified that rate limiting works per user account, becomes more restrictive after repeated failures, recovers automatically and resets after a successful sign-in. The exact values still need Security/Engineering review.
* Added the forced-password-change journey: the user must change it before completing sign-in, and their existing sessions are revoked afterwards.
* Clarified that the policy applies consistently anywhere a password can be set.
* Added the standards coverage summary and clarified that PCI DSS is only partially covered.
* Clarified what happens to existing passwords when a policy changes.

Overall, I think the broader policy model can support #898. We just need to update the password examples and make the boundary between Project choices and ZITADEL’s built-in protections clear.

vitorbari: > The main product question for me is around allowing developers to replace the policy rule completely. We’ve intentionally kept the configurable surface small: Projects can choose the minimum length and password history, while ZITADEL owns the secure baseline.
> 
> A custom rule shouldn’t be able to remove those built-in protections. If custom rules are part of the broader model, they should only be able to add or strengthen requirements.

Ok, so this practically means we have the following options:
1. Ship only Golang policy evaluation rules (non-modifiable / built-in protections)
2. Ship Golang policy evaluation rules (non-modifiable / built-in protections) + Optional OPA/Rego policy rule as customizable extension (both would be evaluated sequentially)
3. Two layers of OPA/Rego: built-in and developer provided (both would be evaluated sequentially)

docs/adrs/059-operation-policies.md
```diff
@@ -0,0 +1,630 @@
...
+
+No defined settings / policy model. Some configuration is hard-coded, for example:
+- Password minimum length: [`MinLength: 8`](internal/domain/flow_field_resolver_schema.go)
+- Passkey user verification is a fixed [`"preferred"`](internal/domain/flow_state_machine.go)
+- Failed authentication attempts are recorded "for rate-limiting purposes" with no threshold reading them.
+
+The first priority is to introduce proper password policies ([#898](https://github.com/zitadel/nextgen/issues/898)):
+
+- Minimum password length
+- Password history
```
fforootd: I think also complexity and age should be worth testing the mechanism against

docs/adrs/059-operation-policies.md
```diff
@@ -0,0 +1,630 @@
...
+  "history_depth": 5,
+  "reject_current": true
+}
+```
+
+```json
+// .zitadel/policies/user.password.verify.json (illustrative)
+{
+  "kind": "policy",
+  "operation": "user.password.verify",
```
fforootd: I like this approach as the namespacing really helps.

One thing that I have not found yet in this ADR is how we define if a policy is applicable to a user, team, project,...

vitorbari:
Closing in favour of #1267

PR #1071, opened 2026-08-31
feat(api): embed a user's lifecycle owner team on the users query

vitorbari:
## Summary

`POST /users/query` accepts `expand: ["lifecycle_owner_team"]`, resolving each user's `metadata.lifecycle_owner_team_id` into the team itself at `metadata.lifecycle_owner_team`. Absent when not requested, `null` when the user is self-owned.

Screenshot 2026-08-31 at 15 06 10


livio-a (commented):
looked at both changes, this is exactly what we discussed, nice. the catalog rationale ("the same read, a page at a time") and "two relations, two gates" put it better than i did, and the lifecycle-filter exclusion test is a nice touch. two asks, both text only:

1. the `lifecycle_owner_team` description promises: requested ⇒ present, null only for self-owned. once granular scopes land (#420) the gates become per named team / per row, and a row whose owner the caller can't read has to show something else. null is taken (self-owned), so flag or omission - we can decide that with #420. but the doc shouldn't promise it away: one softening sentence here, plus one sentence in ADR 059 that records the per-target / per-row direction, so #420 doesn't just drop the fallback and keep the target-blind check.
2. one sentence connecting rule 8 with ADR 058: full embedded objects need the child's read (rule 8), 058 ref fields (id/identifier/display) ride the referencing resource. without that sentence the two ADRs contradict each other on first read.

and since it came up: null for self-owned is correct and stays - the personal team is not the lifecycle owner (ADR 024), there's nothing to resolve.

livio-a (approved):
both asks addressed - the "direction, not a rule yet" framing and the ref-field sentence are exactly what was missing, and the wire description now leaves room for the #420 answer. approving.

PR #1075, opened 2026-08-31
feat: added environments

### LINKED ISSUE #534: Service - Environments persistence & endpoints
**Context.** An environment is a runtime slot on a project. This ticket introduces the environment entity as a standalone concept — identity only, no reference to releases or deployments yet. The link between an env and the release it runs is added by #532 when deployments exist. Environment lifecycle (create / rename / retire) and everything else beyond "a slot" is #528.

**Description.**

Data model:

```json
{
  "id": "env_01KX...",
  "project_id": "proj_...",
  "name": "prod",
  "created_at": "2026-07-01T00:00:00Z"
}
```

| Field | Notes |
|---|---|
| `id` | Opaque, immutable. |
| `name` | Unique within a project. Immutable for now. |

Endpoints:

- `GET /environments` — list envs in the project.
- `GET /environments/{name}` — read one env by name (scoped to the caller's project).

**Temporary env seeding:**
Since this ticket adds no create endpoints, envs need to be created somehow. The short-term approach is to seed a fixed default set of envs at project creation (e.g. `dev`, `staging`, `prod`). 

Once #528 defines the proper lifecycle, the seed is replaced by the create/rename/retire flow.

**Deferred to #528 (ADR — Environments).**
- Lifecycle: create, rename, retire; per-project defaults on setup.
- Env classes (production-class vs. preview/ephemeral).
- Domain-based env selection.
- Cross-env data isolation (users, sessions, the user-schema exception).

**Blocks.** #532 depends on this landing first.

**Related ADR sections.** [Environments](https://github.com/zitadel/nextgen/blob/main/docs/adrs/035-configuration-environments.md#environments), [API — `environments` and `deployments`](https://github.com/zitadel/nextgen/blob/main/docs/adrs/035-configuration-environments.md#environments-and-deployments).

wim07101993:
## Summary

Projects now carry **environments** — the runtime slots a release is deployed onto (ADR 035). Two read endpoints ship: `GET /environments` lists a project's environments in pipeline order, and `GET /environments/{name}` reads one by its project-unique name, which is the handle a deploy target is already known by.

Environments are identity only for now: `{ id, project_id, name, created_at }`, with no link to a release yet. Because there is no create endpoint, every project — new or created through the CLI — is seeded with `dev`, `staging` and `prod` at creation, and the set cannot be changed until environment lifecycle lands. Existing projects are not backfilled.

Seeding emits one `environment.created` audit event per environment.

## Validation

- go test ./...
- go test -tags postgres_integration
- go test -tags sqlite_integration
- go test -tags spanner_integration

## Release notes / changeset

- Changeset: .changeset/environments-persistence-and-endpoints.md — environments


api/openapi/endpoints/environments/environment.yaml
```diff
@@ -0,0 +1,34 @@
+title: Environment
+description: |
+  A runtime slot on a project (ADR 035).
+
+  This is identity only. The release an environment currently runs, and its
+  deployment history, arrive with the deployments endpoints; environment
+  lifecycle (create, rename, retire) arrives with its own ADR. Until then
+  every project is seeded with a fixed set of environments at creation and
+  the set cannot be changed.
```
vitorbari, line 9: This is describing the implementation sequence. We should focus on what the Environment is. 

But I guess now, we don't have all the details yet. We will have more clarity once we conclude the pending research (data-isolation, lifecycle, etc)

vitorbari (approved):
I tested it locally. It is working fine 👍

PR #1085, opened 2026-09-01
feat(api)!: sessions return resolved user refs

### LINKED ISSUE #869: Return user attributes when listing sessions

livio-a:
## ADR 058 Phase D — sessions adopt user refs (closes #869)

Session responses identify their user through a resolved **user ref** derived from the user schema's own `x-identifier`/`x-display` designations, replacing the convention-resolved flat `name`/`email` fields. One rendering chain everywhere: `display` → `identifier` → `user_id`.

Fixes #869

### Structure

**Stacked on #1090** (Phase C — the resolution plumbing: `domain.UserRef`, designation readers, the §4 batch resolver, the `user-ref` wire component). This PR is Phase D, the adoption:

1. Wire + Go: `GET /sessions/me`, operator get, and **query sessions** (the #869 gap — listed sessions previously carried no identity at all) embed `user`; list hydration is one batch resolution per page. The convention resolver (`IdentityAttributeKeys`, `DisplayName`, `Email`, `UserIdentityReader`) is deleted, not kept as fallback. The example user schema gains `x-display`; the shipped default schema carries only `email` and correctly renders its identifier (its `name` was always empty before, too).
2. TS ripple: sdk-core session types → `{userId, identifier, identifierProperty, display}` (JWT claims map `name→display`, `email→identifier`); `zitadel-signout` detail → `{display, identifier}`; components render from the ref (logout templates keep `{{name}}`/`{{email}}` as filling aliases of the new `{{display}}`/`{{identifier}}` tokens); console shell, demos, CLI scaffolds, the Storybook session story, and the demo e2e suites updated; api-mock serves the ref and gains the previously missing `GET /sessions/me` zod conformance case.
3. Demo surfaces render the two identity forms deliberately: the badge is the compact chain (`display → identifier → user_id`), the signed-in-as line is the fuller form (`display (identifier)`); the api-mock ships two demo identities with a resolved display (`ada@example.com` → "Ada Lovelace", `grace@example.com` → "Grace Hopper") while every other email stays identifier-only, so both branches of the chain are exercisable against the mock (see the video below).

### Demo

Both branches of the rendering chain against the api-mock (`ada@example.com` is a fixture identity whose schema designates `x-display`; any other email resolves identifier-only):

https://github.com/user-attachments/assets/4e6aeaaa-ca40-40bd-a90c-ebbad957e062

The finale shows the two `GET /sessions/me` responses side by side — one `user` ref carrying `display`, one identifier-only, no flat `name`/`email` in either.

### Breaking

The flat `name`/`email` session fields and the old SDK session shape are gone. Unknown properties are dropped, not rejected — a stale client reads silently empty values, so server + SDKs + chrome update together (all in this PR; the changeset carries the warning).

### Verification

- Go: full `./internal/... ./api/...` sweep green; new unit tables for the readers, `ResolveUserRef`, the batch resolver, and the session service's ref hydration; integration tests assert the ref on `/sessions/me`, and the new query-sessions case pins refs on listed rows (the #869 acceptance) with anonymous rows carrying none.
- TS: sdk-next 141/141, sdk-nuxt 98/98, components 517 (the one failure is the pre-existing local-build `standalone-artifact` check, reproduced on a clean tree), all six SPA SDK suites, api-mock 70/70, console 122/122; typechecks clean for sdk-core/next/nuxt, components, api-mock, console, demo-next, demo-nuxt, cli; the mock-backed demo-next/demo-nuxt Playwright suites pass locally, the real-instance suites in CI.
- Spec: redocly valid, snake_case casing rules pass; ogen + orval regenerated (generated TS is gitignored and rebuilt by CI).

### Not in this PR

- User envelope derived fields (ADR 058 §3a) and console user list (#956) — next increment on the same primitives.
- `domain.DesignatedIdentifier` also exists on #1074 (Phase B, parallel track); whichever merges second rebases the duplicate away.


packages/components/src/orchestrator/zitadel-logout.ts
```diff
@@ -500,23 +502,27 @@ export class ZitadelLogout extends LitElement {
...
   while (node) {
     if (node.textContent) {
       node.textContent = node.textContent
-        .replace(/\{\{name\}\}/g, name)
-        .replace(/\{\{email\}\}/g, email)
+        .replace(/\{\{display\}\}/g, display)
+        .replace(/\{\{identifier\}\}/g, identifier)
+        .replace(/\{\{name\}\}/g, display)
+        .replace(/\{\{email\}\}/g, identifier)
         .replace(/\{\{initial\}\}/g, initial);
```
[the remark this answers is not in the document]
livio-a: Agreed — replaced the chained replaces with a single-pass regex + callback in 93805516, so substituted values are never rescanned (also avoids $&-style replacement expansion). Added a spec pinning that a display value containing a literal {{email}} renders literally.

internal/api/session.go
```diff
@@ -298,16 +299,26 @@ func sessionToAPI(session *domain.Session) *api.SessionResponse {
...
-		}
+		resp.User = api.NewOptUserRef(userRefToAPI(*session.User))
 	}
 	return resp
 }
 
+// userRefToAPI maps the resolved reference (ADR 058 §3): identifier and
+// identifier_property travel together, display independently; empty means
+// absent on the wire.
+func userRefToAPI(ref domain.UserRef) api.UserRef {
```
vitorbari: Better to keep this function closer to `user`, not in `internal/api/session.go`, since the same reference will be used across other endpoints as well.

internal/domain/user.go
```diff
@@ -125,41 +109,6 @@ func (u *User) StringAttribute(key AttributeKey) string {
 	return s
 }
 
-// DisplayName resolves the user's human-readable name from the conventional
-// identity attributes: `name` when present, otherwise the given and family
-// name parts joined — camelCase spelling first (the shipped presets'
-// convention), snake_case as fallback. Returns "" when the loaded
-// attributes carry none of them.
-func (u *User) DisplayName() string {
```
vitorbari, line 133: 😌

internal/api/session.go
livio-a: Moved it next to the user mapping in 6f13e15c.

vitorbari (approved):

Issue #545, opened 2026-07-15
Service - POST /configuration-releases (bundle constructor)


vitorbari:
**Context.** `POST /configuration-releases` is the CLI orchestrator entry point. It accepts a source-content bundle (the contents of `.zitadel/`), allocates a new revision for every changed resource, resolves handle references between resources, and constructs a release — all in one transaction. Backs `zitadel deploy` and `zitadel releases create`. The release entity itself is in #531; this ticket adds the bundle-in / release-out endpoint on top.

**Description.**

Bundle format (request payload):

```json
{
  "audit": {
    "message": "add phone_number to human-user",
    "git_sha": "4a5b6c7d8e9f...",
    "git_dirty": false
  },
  "schemas": [
    { "objectType": "human-user", "$schema": "...", "properties": { /* ... */ } }
  ],
  "flows": [
    { "name": "default-login", "user_schema": "human-user", "steps": [ /* ... */ ] }
  ]
  // additional per-kind arrays as more kinds land
}
```

| Field | Notes |
|---|---|
| `audit.message`, `audit.git_sha`, `audit.git_dirty` | Recorded on the resulting release (see #531). `created_at` and `created_by` are derived server-side, not sent in the payload. |
| Per-kind arrays (`schemas[]`, `flows[]`, ...) | Each entry is the resource's content as authored on disk. Cross-resource references use handles (e.g. flow's `user_schema: "human-user"`), not concrete revision ids. Kinds not present locally are sent as empty arrays or omitted. |

Endpoint:

- `POST /configuration-releases` — takes the bundle above; returns `{ release_id, revision_ids[] }`.

**Behaviors.**
- **Atomic construction.** Revision allocation for changed resources, handle-reference resolution, and release assembly happen in one transaction. Partial failure leaves no revisions and no release committed.
- **Revision reuse for unchanged resources.** If a bundled resource's content matches the current revision of that `(kind, handle)`, reuse the existing revision id — do not mint a new one.
- **Handle-reference resolution.** Cross-resource references inside the bundle (e.g. a flow's `user_schema: "human-user"`) are looked up against the bundle's own set of handles and rewritten to the target `revision_id` before persisting. Whether the resolved `revision_id` is baked into the stored resource revision or preserved as a handle for runtime resolution is decided here — see #536 for the runtime implication.
- **Empty bundle rejected.** A bundle with no resources across any kind returns an error; a release must contain at least one pointer.
[...]

PR #1088, opened 2026-09-01
feat(api): releases endpoints contract

### LINKED ISSUE #531: Service - Releases persistence & endpoints

**Description.**

Data model:

```json
{
  "id": "rel_01KX...",
  "project_id": "proj_...",
  "audit": {
    "git_sha": "4a5b6c7d8e9f...",
    "git_dirty": false,
    "created_by": "user_..."
  },
  "pointers": [
  ]
}
```

| Field | Notes |
|---|---|

Endpoints:

vitorbari:
This PR add the API contract for the `/releases` endpoints. There are the APIs that read and write the release data which is based on **pointers to resource revisions**.

- `POST /releases` — bundle from revision ids that already exist
- `GET /releases` — newest first, pinned set omitted (we can introduce `POST /releases/query` when search and sorting are needed)
- `GET /releases/{release_id}` — one release with its pointers

The endpoint `POST /configuration-releases` is out of the scope for now (https://github.com/zitadel/nextgen/issues/545). The difference, is that it takes the raw definition of the resources, instead of existing revisions.

image


api/openapi/endpoints/releases/release-pointer.yaml
```diff
@@ -0,0 +1,49 @@
+type: object
+description: |
+  One entry of the set a release pins: the revision of a single resource that
+  this release includes.
+required:
+  - kind
+  - handle
+  - revision_id
+properties:
+  kind:
```
grvijayan, line 10: as `kind` is also used in `create-release-pointer.yaml` above, we could extract this to be a common schema similar to `api/openapi/components/schemas/release-id.yaml`

api/openapi/endpoints/releases/methods.yaml
```diff
@@ -0,0 +1,241 @@
+post:
+  operationId: createRelease
+  summary: Create a release
+  description: |
+    Bundles a release from revisions that already exist, supplied as
+    `(kind, revision_id)` pairs. No new revisions are allocated — use the
+    per-kind create endpoints for that, then pin the ids here.
+
+    Every referenced `revision_id` must exist in the project. Each one's handle
```
grvijayan, line 9: should this also apply to the schemas/resources referenced within the resources (for example, the `user_schema` pinned in the flow definition revision should also exist in the project)? Or, is this implicit?

vitorbari: In that case, the `user_schema` should be part of the same release. There will be a validation step for the cross resource references.

grvijayan (approved):

PR #1090, opened 2026-09-02
feat: user-ref resolution plumbing (ADR 058 §3–§4)

livio-a:
## Phase C of ADR 058 — split out of #1085 for review focus

The resolution plumbing behind user refs, with **zero observable behavior change** — #1085 (sessions adoption, the #869 fix) stacks directly on this.

### What's in it

1. **Domain** — `domain.UserRef` + `ResolveUserRef` (ADR 058 §3: role-named fields, identifier/display resolved independently, clients render `display → identifier → user_id`), and the raw-document designation readers `DesignatedIdentifier`/`DesignatedDisplay`/`DesignatedAttributeKeys`. Designation paths are dot-joined leaf paths and address the EAV-flattened attribute keys verbatim.
2. **Service** — `StatementsUserRefResolver`, the §4 batch port: `(project_id, user_ids) → refs`, one by-ids user query per page hydrating only the union of designated attribute keys (§3a); every stored schema revision contributes its own designation; missing users are absent from the result. **Authz note:** resolution runs `WithAuthzListUnrestricted` — both queries are nested reads keyed on rows the caller already holds, not management lists (the `GetUser`/#839 pattern); without it the #838 list-filter tripwire 500s every resolving read.
3. **Wire** — the `user-ref` component (`{user_id, identifier, identifier_property, display}`), inert until a response references it (#1085 does).

### Notes for review

- `DesignatedIdentifier` is duplicated on #1074 (Phase B, parallel track); whichever merges second rebases the duplicate away.
- The `AttributeKeys: [""]` sentinel (hydrate nothing when no schema designates) behaves identically in all three dialects — postgres `key = ANY`, sqlite/spanner key-IN.
- Changeset is a `@zitadel/server` patch describing internal plumbing; the behavior-changing changeset rides #1085.

Part of #869 / ADR 058.

### DIFF
--- a/api/openapi/components/schemas/user-ref.yaml
+++ b/api/openapi/components/schemas/user-ref.yaml
@@ -1,0 +1,37 @@
+type: object
+title: UserRef
+description: |
+  A resolved reference to a user (ADR 058 §3). Fields are role-named, not
+  property-named, so responses can mix users from different schemas.
+  `identifier` and `display` are resolved independently from the user
+  schema's `x-identifier` and `x-display` designations, live at read time.
+  Clients render `display`, falling back to `identifier`, then `user_id`.
+required:
+  - user_id
+properties:
+  user_id:
+    $ref: ./user-id.yaml
+    description: The referenced user's id. Always present.
+  identifier:
+    type: string
+    description: |
+      The current value of the schema's designated identifier
+      (`x-identifier`). Absent when the schema designates no identifier or
+      the user carries no value for it.
+    example: ada@example.com
+  identifier_property:
+    type: string
+    description: |
+      The schema property `identifier` came from, so clients can reach the
+      property's schema for semantics (a mailto link, a field label)
+      instead of guessing from the value. Present exactly when
+      `identifier` is.
+    example: email
+  display:
+    type: string
+    description: |
+      The `x-display` rendering — the designated properties' values joined
+      in list order. Purely presentational, with no source attribution.
+      Absent when the schema designates no display properties or the user
+      carries no values for them.
+    example: Ada Lovelace
--- a/internal/service/user_ref.go
+++ b/internal/service/user_ref.go
@@ -38,6 +38,12 @@ func (r StatementsUserRefResolver) ResolveUserRefs(ctx context.Context, projectI
 		return map[string]domain.UserRef{}, nil
 	}
 
+	// Both queries below are nested reads keyed on rows the caller already
+	// holds (the page's user ids, the project's schemas) — not management
+	// lists — so the authz list-filter tripwire and any inherited filter are
+	// deliberately skipped, the same choice GetUser makes (#839).
+	ctx = WithAuthzListUnrestricted(ctx)
+
 	documents, attributeKeys, err := r.designatingSchemas(ctx, projectID)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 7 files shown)


internal/service/user_ref.go
```diff
@@ -0,0 +1,126 @@
...
+				attributeKeys = append(attributeKeys, key)
+			}
+		}
+	}
+	if attributeKeys == nil {
+		// AttributeKeys nil means "hydrate everything"; no designation
+		// anywhere means nothing to read — hydrate a key no property can
+		// use so the query returns no attribute rows.
+		attributeKeys = []string{""}
+	}
```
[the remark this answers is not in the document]
livio-a: Good catch — guarded on `len(attributeKeys) == 0` with the comment reworded to the UserQueryOptions contract (empty, nil or not, hydrates everything) in fbf4cc75.

api/openapi/components/schemas/user-ref.yaml
```diff
@@ -0,0 +1,37 @@
...
+title: UserRef
+description: |
+  A resolved reference to a user (ADR 058 §3). Fields are role-named, not
+  property-named, so responses can mix users from different schemas.
+  `identifier` and `display` are resolved independently from the user
+  schema's `x-identifier` and `x-display` designations, live at read time.
+  Clients render `display`, falling back to `identifier`, then `user_id`.
+required:
+  - user_id
+properties:
```
vitorbari, line 11: Would it be useful to return a reference to the user-schema, so the client knows the "user-type"?

internal/service/user_ref.go
```diff
@@ -0,0 +1,127 @@
...
+	// Both queries below are nested reads keyed on rows the caller already
+	// holds (the page's user ids, the project's schemas) — not management
+	// lists — so the authz list-filter tripwire and any inherited filter are
+	// deliberately skipped, the same choice GetUser makes (#839).
+	ctx = WithAuthzListUnrestricted(ctx)
+
+	documents, attributeKeys, err := r.designatingSchemas(ctx, projectID)
+	if err != nil {
+		return nil, err
+	}
```
vitorbari, line 50: This retrieves all user-schemas of a project. The order could be reversed: fetch users, then fetch only the schemas for those users.

api/openapi/components/schemas/user-ref.yaml
livio-a: IMO the ref should stay as small as possible — it's just for rendering and linking. If a client really needs to know what kind of user it is, it can fetch the user by id and gets `schema` there anyway. Putting it on every ref bloats every session row for something most UIs never look at. And since all ref fields are optional we can still add it later without breaking anything, so I'd wait until an actual screen needs it.

internal/service/user_ref.go
livio-a: The order actually matters here: we read the schemas first to know which attribute keys to hydrate on the user query — reversed we'd have to load all attributes for every user. For the case where the users are already loaded (the envelope resolution in #1092) we could filter the schema fetch down to their URLs, but it's one query either way and a project rarely has more than a handful of schemas, so I kept both paths on the same shape. Can change that one to a url filter if you'd rather.

vitorbari: Both are not ideal 😆 

But thinking that users has much higher cardinality than schemas, the current implementation makes sense.

api/openapi/components/schemas/user-ref.yaml
vitorbari: > Putting it on every ref bloats every session row for something most UIs never look at.

I will double check that. I assume it is important to distinguish , for example, session for human vs AI agents.

But for now, it implementation works, I'm approving the PR.

vitorbari (approved):

PR #1092, opened 2026-09-02
feat: user responses and console surfaces carry the derived identity

### LINKED ISSUE #956: hard-coded user attributes on UI and API

livio-a:
## ADR 058 §3a — user envelope identity + console user surfaces

**Third in the stack: #1090 (plumbing) ← #1085 (sessions) ← this PR.** Fixes #956; also completes the identity half of the users-list build-out tracked in #630.

### API

The user envelope gains read-only `identifier`, `identifier_property`, `display` — the same resolution that produces session user refs, inlined (`user_id` omitted; the envelope has `id`). Per ADR 059 rule 8, derived label fields ride the resource's own read permission: **no expand gate**, unconditional on every user response (query, get by id, `users/me`, create read-back).

Resolution cost: the read paths already hydrate the full attribute document, so the new `ResolveRefsForUsers` port method adds **one schema-list query per page and zero user queries** (§4: never per row). Same `WithAuthzListUnrestricted` posture as the batch resolver (#839 pattern).

### Console

- **Users list**: a leading **User** identity column (`display → identifier → id`) carrying the row link, plus a dedicated role-named **Identifier** column — platform-derived like Status/ID, deliberately outside the schema-driven column set (D4/#630), scannable down a mixed-schema list whether a row's identifier is an email or a loginname (cell tooltip names the source property) (the link used to hang on whatever schema property happened to sort first, falling back to the raw ULID: exactly #956's complaint). Schema-driven columns, Status, and ID follow unchanged; search covers the identity.
- **Detail**: heading follows the same chain, identifier as subtitle under a display.
- **Deleted**: `userDisplayName` — the client-side convention guesser (name → givenName/familyName, camel+snake), the console twin of the Go resolver #1085 removes. Identity now comes only from the server's resolution; the console carries zero designation logic.

### Screenshots

The list against a real seeded instance — `Grace Hopper` lives under a schema designating `loginname`/`first_name`/`last_name` (no property the platform has ever hardcoded), `Ada Lovelace` under one designating `email` + display, the seeded rest under the shipped default (identifier only). **Before**: see the screenshots in #956 — these same rows degraded to raw ULIDs or guessed attributes.

956userslist

956userdetail

### Out of scope

How **schema-defined properties** render across mixed schemas (the unioned schema-driven columns and their `—` placeholders, e.g. cross-schema property normalization) stays with #630 — this PR only adds the platform-derived identity fields, which sit outside that column set like Status and ID.

### Tests

- Unit: `ResolveRefsForUsers` (no user query asserted via mock, unstored-schema bare ref, empty short-circuit); console suites rewritten off the convention cases onto the chain (display / identifier-only / id fallback), 122/122.
- Integration: `TestUserResponses_CarryDerivedIdentity` — query rows carry each schema's own resolution (display-designating schema vs the shipped default), get-by-id carries it; create read-back asserted at the service level. Runs in CI (no local Docker).
- console-e2e real specs unchanged by design: seeded users live under the default schema, so their labels stay the email (identifier).

### Not in this PR

The api-mock serves no users ("a users list read from it is a fiction" — apps/console/AGENTS.md), so no mock change; verify visually with `moon run console:dev-real`.

### DIFF
--- a/apps/console/src/components/delete-user-dialog.spec.tsx
+++ b/apps/console/src/components/delete-user-dialog.spec.tsx
@@ -125,7 +125,7 @@ describe("delete user dialog", () => {
     // row is gone from the invalidated list.
     expect(await screen.findByText("Maya Patel deleted")).toBeInTheDocument();
     await waitFor(() =>
-      // The row's link is its first schema-driven column, not a combined name.
+      // The row's link lives on the User identity column.
       expect(screen.queryByRole("link", { name: "Maya Patel" })).not.toBeInTheDocument(),
     );
   });
--- a/apps/console/src/components/delete-user-dialog.spec.tsx
+++ b/apps/console/src/components/delete-user-dialog.spec.tsx
@@ -19,6 +19,11 @@ const USERS_QUERY_URL = `${USERS_URL}/query`;
 const USER = {
   id: "user_1",
   schema: "sch_business",
+  // The dialog's label comes from the envelope's resolved identity
+  // (ADR 058), not from attribute derivation.
+  display: "Maya Patel",
+  identifier: "maya@acme.com",
+  identifier_property: "email",
   attributes: { givenName: "Maya", familyName: "Patel", email: "maya@acme.com" },
 };
 const server = setupServer();
@@ -121,7 +126,7 @@ describe("delete user dialog", () => {
     expect(await screen.findByText("Maya Patel deleted")).toBeInTheDocument();
     await waitFor(() =>
       // The row's link is its first schema-driven column, not a combined name.
-      expect(screen.queryByRole("link", { name: "maya@acme.com" })).not.toBeInTheDocument(),
+      expect(screen.queryByRole("link", { name: "Maya Patel" })).not.toBeInTheDocument(),
     );
   });
 
@@ -150,7 +155,7 @@ describe("delete user dialog", () => {
     // found: opening the dialog from inside the row menu left the menu open
     // underneath, and its `aria-hidden` on the rest of the page outlived the
     // dialog — so the list was invisible to assistive tech after cancelling.
-    expect(screen.getByRole("link", { name: "maya@acme.com" })).toBeInTheDocument();
+    expect(screen.getByRole("link", { name: "Maya Patel" })).toBeInTheDocument();
   });
 
   it("keeps the dialog open and shows the server's message when the delete fails", async () => {
--- a/apps/console/src/lib/user.ts
+++ b/apps/console/src/lib/user.ts
@@ -15,26 +15,24 @@ export function userAttributes(user: Record<string, unknown>): Record<string, un
 }
 
 /**
- * The user's human-readable name, mirroring the platform's identity contract
- * (`User.DisplayName` in `internal/domain/user.go` — the same derivation behind
- * `GET /sessions/me`'s `name`): an explicit `name` wins, otherwise the given and
- * family parts joined. The shipped presets spell them camelCase
- * (`packages/config/defaults/*.json`); snake_case stays accepted for schemas
- * authored that way.
+ * The user's rendered identity per ADR 058: the envelope's derived `display`,
+ * falling back to `identifier`, then the id. The fields are resolved by the
+ * server from the schema's own `x-identifier`/`x-display` designations, so
+ * the console carries zero designation logic — the convention-guessing
+ * `userDisplayName` this replaces is gone with the platform's own resolver.
  *
- * User schemas are free-form and `queryUsers` returns the attribute tree verbatim,
- * so this reads only the conventional identity keys and returns `undefined` when
- * a schema names its properties differently. Callers fall back to the email and
- * then the id — inventing a name from an unrelated attribute is worse than
- * having none.
- *
- * Takes the attributes object, not the whole user: `name` is a schema property.
+ * Takes the whole user record: `display`/`identifier` are envelope fields,
+ * not schema properties.
  */
-export function userDisplayName(attributes: Record<string, unknown>): string | undefined {
-  const explicit = field(attributes, "name");
-  if (explicit) return explicit;
-  const given = field(attributes, "givenName") ?? field(attributes, "given_name");
-  const family = field(attributes, "familyName") ?? field(attributes, "family_name");
-  const joined = [given, family].filter(Boolean).join(" ");
-  return joined || undefined;
+export function userIdentity(user: Record<string, unknown>): string {
+  return field(user, "display") ?? field(user, "identifier") ?? (field(user, "id") ?? "");
+}
+
+/**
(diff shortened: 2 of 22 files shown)


internal/service/user.go
```diff
@@ -199,10 +202,28 @@ func (s *userService) CreateUser(ctx context.Context, input CreateUserInput) (_
 			WithMessage("The user was created but could not be read back. Fetch it by id rather than retrying the create.").
 			WithDetails(domain.CreatedUserDetails{UserID: action.CreateUser.ID})
 	}
+	if err := s.attachUserRefs(ctx, input.ProjectID, user); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
livio-a: Good catch — a decoration failure after the committed create would have invited a retry into the unique constraints. The create no longer fails on ref-resolution errors: the response simply carries no ref fields and clients fall back to the id (ADR 058 chain); the read-back failure path keeps its CreatedUserDetails hint unchanged. Pinned by TestCreateUser_SurvivesRefResolutionFailure in ed44c617.

apps/console/src/routes/_authed/users/index.tsx
```diff
@@ -277,7 +290,7 @@ function UsersScreen() {
             {rows.length === 0 ? (
               <TableRow className="border-0 hover:bg-transparent">
                 <TableCell
-                  colSpan={columns.length + 2}
+                  colSpan={columns.length + 4}
                   className="text-muted-foreground h-24 text-center"
                 >
```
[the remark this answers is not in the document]
livio-a: Fixed in ed44c617 — the empty-state cell now spans the full column count including the actions column (the previous count deliberately mirrored the file's original actions-column exclusion, but full-width is cleaner).

apps/console/src/routes/_authed/users/users.spec.tsx
```diff
@@ -61,10 +64,14 @@ describe("users screen", () => {
     );
     await renderUsers();
     expect(await screen.findByRole("heading", { name: "Users" })).toBeInTheDocument();
-    // Columns are the schema's properties (design `277:288291`, decisions log
-    // D4), so there is no combined "Name" column — the first property carries
-    // the link to the detail screen.
-    expect(await screen.findByRole("link", { name: "maya.patel@acme.com" })).toBeInTheDocument();
+    // The User column leads with the resolved display and carries the link;
+    // the identifier renders as its muted second line. Schema-driven
+    // attribute columns follow (design `277:288291`, decisions log D4).
```
peintnermax: it doesn't. the identifier is a separate column, and the assertion right below it checks the column header.

apps/console/src/components/delete-user-dialog.spec.tsx
```diff
@@ -121,7 +126,7 @@ describe("delete user dialog", () => {
     expect(await screen.findByText("Maya Patel deleted")).toBeInTheDocument();
     await waitFor(() =>
       // The row's link is its first schema-driven column, not a combined name.
```
peintnermax: describes the old behaviour. the link now lives on the User column.

apps/console/src/lib/user.ts
```diff
@@ -15,26 +15,29 @@ export function userAttributes(user: Record<string, unknown>): Record<string, un
...
- * a schema names its properties differently. Callers fall back to the email and
- * then the id — inventing a name from an unrelated attribute is worse than
- * having none.
- *
- * Takes the attributes object, not the whole user: `name` is a schema property.
+ * Takes the whole user record: `display`/`identifier` are envelope fields,
+ * not schema properties.
+ */
+export function userIdentity(user: Record<string, unknown>): string {
+  return field(user, "display") ?? field(user, "identifier") ?? (field(user, "id") ?? "");
```
peintnermax: userIdentity returns "" as its final fallback and both callers then do || id / || userId. Either let it return string | undefined like userIdentifier and userIdentitySecondary, and keep the caller fallback, or drop the caller fallback. Cosmetic, but the double fallback reads as uncertainty about the contract and inconsistent :D

peintnermax (commented):
Design question, non-blocking: with `x-identifier: email` this column now shows the same value as the schema-driven `email` column right next to it (also see screenshot in the PR description). Since `identifier_property` tells us exactly which property it is, could we drop that property from `columnsForUsers` when every loaded row shares one `identifier_property`, and name it in the header instead (e.g. "Identifier (email)" or the field's own label)? For mixed-schema lists we'd keep the plain "Identifier" header and the duplicate. Alternatively we could surface the property in the row itself (a small muted hint next to the value). also the title tooltip is easy to miss (no blocker for me). I am also fine to leave the duplication but wanted to leave it as follow up record

apps/console/src/routes/_authed/users/users.spec.tsx
livio-a: Right, that comment was left over from before the column existed — fixed in b4436f6c.

apps/console/src/components/delete-user-dialog.spec.tsx
livio-a: Fixed in b4436f6c.

apps/console/src/lib/user.ts
livio-a: Fair :D went with your first option — userIdentity returns string | undefined like the other readers and the callers keep the one explicit id fallback. b4436f6c

vitorbari (approved):

Issue #530, opened 2026-07-14
Service - Revisioning for flow definitions


vitorbari:
**Context.** User schemas already have revisions (#456 introduced `objectType` which works as a correlation ID between schema revisions). Flow definitions are the other configurable kind that exists today; releases can't work until flows are revisioned too. Other kinds (idps, brandings, apps, policies) get the same treatment when they land, but are not in scope here.

Flows: `POST`, `PATCH`, `PUT`, `DELETE` on `/flow_definitions` all exist today. This ticket adopts the schemas-style model for flows and specifies the effect on each existing endpoint.

**Endpoint effects.**

| Endpoint | Effect |
|---|---|
| `POST /flow_definitions` | **Kept, semantics change.** Every call allocates a new opaque `id` (`flowdef_*`). If a flow with the same `name` already exists, it is not a conflict. |
| `PATCH /flow_definitions/{id}` | **Removed.** Revisions are immutable; content changes go through a new POST. |
| `PUT /flow_definitions/{id}` | **Removed.** Same reason as PATCH. |
| `DELETE /flow_definitions/{id}` | **Removed (for now).** Deleting a revision that any release pins would break that release.  |
| `GET /flow_definitions/{id}` | **Kept.** Reads one specific revision by opaque id. |
| `GET /flow_definitions?name=<handle>` | **New.** Lists revisions of a specific flow (by handle), newest first. Used by `zitadel pull` (#541) and by runtime release resolution (#536) when handles are resolved at runtime. |

**Behaviors.**
- Revisions are immutable after write. Any content change allocates a new revision id.

**Related ADR sections.** [Releases](https://github.com/zitadel/nextgen/blob/main/docs/adrs/035-configuration-environments.md#releases).

PR #1093, opened 2026-09-02
feat(api): flow definitions are revisions

grvijayan:
## Summary

First step of #530. Flow definitions become revisions the way schemas
already are (#456): every `POST /flow_definitions` allocates a new
`flowdef_*` id, and publishing under an existing `name` adds a revision
to that flow instead of returning `flowdef.already_exists`.
`GET /flow_definitions?name=<handle>` lists one flow's revisions newest
by creation time first. `POST /flow` resolves by highest schema version,
then newest created; the two agree while every revision carries version
1.0.0.

- Spec: POST drops its 409, the list gains a `name` filter and documents
  newest-by-creation order.
- Service: Create no longer reads before insert; List filters on name.
- api-mock: list honors `name` and sorts newest by creation, with a
  store-held sequence as the tiebreak.
- Runtime: `POST /flow` by `flow_definition_name` resolves the newest
  active revision of that name (highest schema version, then newest
  created, then id). Before this PR at most one row per name and version
  existed, so the tie never arose.

Apart from that resolver rule this PR is additive. PUT and DELETE are
removed in a follow-up once the CLI syncer stops using them.

## Validation

- `go test ./internal/...` and the Postgres integration package
- redocly lint, openapi rules, api/cli/console typecheck
- api-mock typecheck, lint, tests

## Release notes / changeset

`@zitadel/server` and `@zitadel/api` minor.

## Notes

With the 409 gone, ogen returns the shared `ErrorDetails` type for the
POST 400 instead of a per-operation type. Wire format is unchanged.


internal/service/flow.go
```diff
@@ -306,12 +306,16 @@ func flowServesPurpose(def *domain.FlowDefinition, purpose domain.FlowDefinition
...
-// pickLatestFlowVersion is a lexicographic compare — sufficient while
-// versions stay zero-padded MAJOR.MINOR.PATCH. Caller ensures defs non-empty.
-func pickLatestFlowVersion(defs []*domain.FlowDefinition) *domain.FlowDefinition {
+// pickNewestFlowRevision prefers the highest schema version (a lexicographic
+// compare, sufficient while versions stay zero-padded MAJOR.MINOR.PATCH) and,
+// among revisions of one version, the newest created. Revisions of a name
+// share a version, so the pick must not depend on list order. Caller ensures
+// defs non-empty.
+func pickNewestFlowRevision(defs []*domain.FlowDefinition) *domain.FlowDefinition {
 	winner := defs[0]
```
IAM-marco, line 318: If `defs` is nil, this will panic. I would exit right away with a check on the length

internal/service/flow.go
```diff
@@ -306,12 +306,16 @@ func flowServesPurpose(def *domain.FlowDefinition, purpose domain.FlowDefinition
...
 }
 
-// pickLatestFlowVersion is a lexicographic compare — sufficient while
-// versions stay zero-padded MAJOR.MINOR.PATCH. Caller ensures defs non-empty.
-func pickLatestFlowVersion(defs []*domain.FlowDefinition) *domain.FlowDefinition {
+// pickNewestFlowRevision prefers the highest schema version (a lexicographic
+// compare, sufficient while versions stay zero-padded MAJOR.MINOR.PATCH) and,
+// among revisions of one version, the newest created. Revisions of a name
+// share a version, so the pick must not depend on list order. Caller ensures
+// defs non-empty.
```
IAM-marco: Yes, ensure defs non-empty, until it doesn't anymore

internal/service/flow_definition_test.go
```diff
@@ -393,7 +328,7 @@ func Test_flowDefinitionService_Create(t *testing.T) {
 					stmts := servicemocks.NewMockAllStatements(ctrl)
 					stmts.EXPECT().ListFlowDefinitions(gomock.Any(), gomock.Any()).DoAndReturn(func(context.Context, *database.ListOptions[domain.FlowDefinitionField]) (*database.ListResult[*domain.FlowDefinition], error) {
```
IAM-marco: Can you check if it's possible to enforce the second param instead of using `gomock.Any()`? It's not always possible, but if it is, it would make the mock stricker and verification more reliable

IAM-marco: Also for the other calls in this file

internal/service/flow_test.go
```diff
@@ -217,6 +217,59 @@ func TestResolve_ResolveByName_LatestActiveWhenVersionNil(t *testing.T) {
...
+	newer := newDef("login", "1.0.0", domain.FlowDefinitionAudience{}, domain.FlowDefinitionPurposeLogin)
+	newer.ID = "flowdef_newer"
+	newer.CreatedAt = time.Date(2026, 6, 1, 0, 0, 0, 0, time.UTC)
+
+	for _, defs := range [][]*domain.FlowDefinition{{older, newer}, {newer, older}} {
+		repo := stubListFlowDefinitions(t, defs)
+		got, err := service.NewFlowService(repo, nil).Resolve(t.Context(), service.ResolveFlowRequest{
+			ProjectID: "proj",
+			Purpose:   domain.FlowDefinitionPurposeLogin,
+			Name:      ptr("login"),
```
IAM-marco:
```suggestion
			Name:      new("login"),
```

IAM-marco: actually all ptr calls can be substituted by new()

internal/service/flow_test.go
IAM-marco, line 1: Not sure why we use t.Fatal here instead of the assert/require lib

IAM-marco (changes requested):

internal/service/flow_test.go
grvijayan: updated in https://github.com/zitadel/nextgen/pull/1093/commits/d90412185a6f0187516eca086a76278aadb57a56

internal/service/flow_test.go
grvijayan: updated in https://github.com/zitadel/nextgen/pull/1093/commits/d90412185a6f0187516eca086a76278aadb57a56

internal/service/flow_definition_test.go
grvijayan: done in https://github.com/zitadel/nextgen/pull/1093/commits/96d7ccb6efffd9a65de894fb02ec1af13f13978e

internal/service/flow.go
grvijayan: done in https://github.com/zitadel/nextgen/pull/1093/commits/96d7ccb6efffd9a65de894fb02ec1af13f13978e

internal/service/flow_definition.go
```diff
@@ -72,20 +72,6 @@ func NewFlowDefinitionService(
 }
 
 func (fd *flowDefinitionService) Create(ctx context.Context, req FlowDefinitionRequest) (*domain.FlowDefinition, error) {
```
IAM-marco, line 74: This PR removes the uniqueness check, but the design doc still teaches it. `docs/design/flowengine/flow-definition-rules.md` line 19 says a name is unique per project and schema version, line 68 even claims the database enforces it (it never did), and line 104 still asks which uniqueness key we should pick. This PR is the answer to that question, so the doc is now the only place the old rule survives, and the next person who reads it will build against behavior that no longer exists.

Please update the doc in this PR: drop the uniqueness wording on lines 19 and 68, and close the open question on line 104 with what this PR does (revisions coexist, no uniqueness key).

internal/service/flow_definition_test.go
```diff
@@ -62,6 +62,36 @@ func v2PoolFromStatements(t *testing.T, stmts service.AllStatements) *service.DB
 	return service.NewPool(pool)
 }
 
+// expectListFlowDefinitions stubs one ListFlowDefinitions call and asserts the
+// filter the service built, so a dropped or widened filter fails the test.
+func expectListFlowDefinitions(t *testing.T, stmts *servicemocks.MockAllStatements, wantFilter database.Filter[domain.FlowDefinitionField], items []*domain.FlowDefinition) {
```
IAM-marco, line 67: I like this helper, and the filter checks are exactly what I asked for. One gap: it can only return items, never an error. The old test "list flow definitions error bubbles up" was deleted together with the uniqueness check, so nothing in this file makes `ListFlowDefinitions` fail anymore. Create still calls it when a step pivots to another flow, and `validatePivotingTargets` passes that error straight through (flow_definition.go line 269), so the path is real but untested now.

Please add one Create case with a pivot transition where the list fails, and check the error comes back out. An error variant on this helper, or a direct stub for that one call, both work.

packages/api-mock/src/platform-handlers.ts
```diff
@@ -86,6 +86,15 @@ function nowIso(): string {
   return new Date().toISOString();
 }
 
+// Publication order. `nowIso()` is millisecond-resolution and ids are random,
+// so two records minted back to back would otherwise tie with no way to say
+// which came first; the server never ties because its ids are time-ordered.
+let lastSeq = 0;
```
IAM-marco: Two small things here.

First, this counter lives outside the store, but `resetPlatformStore()` says it resets all in-memory state, so the doc and the code disagree. It does no harm today because the counter only grows, but the next reader will believe the doc. If the counter moves into the object `makeStore()` builds, reset works by itself, and `nextSeq()` can go away too: `store.lastSeq += 1` at the four call sites is enough.

Second, the comment above says the real server never ties because its ids carry time. True on Postgres and SQLite, but Spanner hands out random UUIDs (`internal/storage/dialect/spanner/managed_id.go`), so there the id tiebreak is just as arbitrary as here. Worth rewording so it does not promise more than every database delivers.

internal/service/flow.go
```diff
@@ -306,12 +306,18 @@ func flowServesPurpose(def *domain.FlowDefinition, purpose domain.FlowDefinition
...
 }
 
-// pickLatestFlowVersion is a lexicographic compare — sufficient while
-// versions stay zero-padded MAJOR.MINOR.PATCH. Caller ensures defs non-empty.
-func pickLatestFlowVersion(defs []*domain.FlowDefinition) *domain.FlowDefinition {
+// pickNewestFlowRevision prefers the highest schema version (a lexicographic
+// compare, sufficient while versions stay zero-padded MAJOR.MINOR.PATCH) and,
+// among revisions of one version, the newest created. Revisions of a name
+// share a version, so the pick must not depend on list order.
+func pickNewestFlowRevision(defs []*domain.FlowDefinition) *domain.FlowDefinition {
```
IAM-marco, line 314: We now have two different answers to "which revision is the current one". This function prefers the highest schema version and only then the newest one. The list endpoint says newest first and sorts purely by creation time, and both the PR text and the api-mock say the first list entry is the current revision.

They agree today only because every definition gets schema version "1.0.0" (`mapFlowDefinitionRequestToService`, your TODO is on that line). Once versions are real, someone can publish an older version after a newer one, and then the list and `POST /flow` point at different revisions.

No code change needed now, just pick which answer wins and write it down: either the name-filtered list sorts by version first like this function, or the docs soften to "newest first by creation". One line in the OpenAPI description or the follow-up issue is enough.

IAM-marco (changes requested):

internal/service/flow.go
grvijayan: decided to update the docs, done in https://github.com/zitadel/nextgen/pull/1093/commits/773d9dad5b8e9fc14385c7f3644ede7d14bb00aa

packages/api-mock/src/platform-handlers.ts
grvijayan: done in https://github.com/zitadel/nextgen/commit/773d9dad5b8e9fc14385c7f3644ede7d14bb00aa

internal/service/flow_definition_test.go
grvijayan: done in https://github.com/zitadel/nextgen/commit/773d9dad5b8e9fc14385c7f3644ede7d14bb00aa

internal/service/flow_definition.go
grvijayan: done in https://github.com/zitadel/nextgen/commit/773d9dad5b8e9fc14385c7f3644ede7d14bb00aa

IAM-marco (approved):
All four points addressed cleanly: the design doc no longer teaches the deleted uniqueness rule, the pivot target lookup error is covered in Create, the mock seq counter lives in the store so reset clears it, and the list order is documented as newest by creation time with the resolver's version-first rule called out. Thanks!

PR #1118, opened 2026-09-03
feat(api): add POST /grants/query to list project grants

### LINKED ISSUE #1020: feat: add POST /grants/query to list project grants
Created on behalf of @adlerhurst.

## Context

The first grant API slice ships create, get, and revoke only. ADR 054 allows team grants without a listing path, so `POST /grants/query` was left out of that PR. There is no product API today to list grants on a project.

Storage has `AuthzAssignmentStatements` Create/Get/List-by-principal/Revoke. Listing by project needs a new list statement plus `stmttest` cursor-battle tests.

## Ask

Add `POST /grants/query` scoped to the caller’s project:

- New storage list statement over `authz_assignments` filtered by project (active grants).
- Cursor-battle coverage in `stmttest` for the new list statement.
- OpenAPI + GrantService query matching the existing collection/query pattern (copy teams/events).
- Keep IDs `asgn_<opaque>`. Error codes stay `grant.*`.

## Related

- ADR 054 (customer collaboration grants)
- First grant slice: create/get/revoke (no list)

adlerhurst:
## Summary

Adds `POST /grants/query` (#1020) for unrevoked user/team bindings to `viewer` / `editor` / `admin` (including expired). Project-secret setup (`sk_proj`) and owning-team (`relation=team`) rows are excluded in SQL so pages are not silently short. Grants stay out of RSI; the handler uses `requireProjectAccess(..., grantAccess, opRead)` with the existing `project.read` ceiling (operator `project.write` still required at the credential plane, same as Get).

Always-on **user-ref** / **team-ref** labels (ADR 058) on GET, create read-back, and query. Query-only `expand: ["principal"]` fills `principal` with the GET-user or GET-team body, or `null` if the principal cannot be loaded. Expand is gated **before** the list with both `user.read` and `team.read` (operator `project.write` fallback until #420). Hydration is a post-list batch (never joined into cursor SQL).

User refs now go through the designation `UserRefResolver` from #1090 (`x-identifier` / `x-display` live at read time). Missing users still degrade to `user_id` only.

`ListManagedGrants` takes an explicit `projectID` and always ANDs `project_id` in SQL, so a caller cannot list grants across projects even if `opts.Filter` omits it. The list conjunct also requires `catalog_id = cat_sys_1` and `scope_kind = project`, and Get/Revoke 404 the same rows.

## How to review

1. OpenAPI source under `api/openapi/endpoints/grants/` — wire contract, expand enum, `grant-expanded-principal.yaml`.
2. `internal/service/grant.go` — `List` + unexported `hydrate`; Create/Get return a hydrated `*Grant`; nested `GrantPrincipal` (nil omits, empty struct → wire null). Create treats post-commit hydrate as best-effort (id-only refs). Unpinned pages group principals by resource-scope home.
3. `internal/api/grant.go` — map only (`grantResponse`). Expand `user.read` / `team.read` constructors stay in `QueryGrants` so the generated error catalog includes those 403s (`requireUserRead` cannot be wired without dropping them from `queryGrants-error-response.yaml`).
4. Dialect `ListManagedGrants` + shared `authz.AuthzAssignmentSchema` + `internal/api/integration_test/grant_test.go` (`TestGrantQuery`, `TestGrantQueryExpand`).

ogen cannot infer `User|TeamResponse` from bare `$ref`s. `grant-expanded-principal.yaml` uses unique required fields on `allOf` wrappers. **Keep that filename** so the generated type is `GrantExpandedPrincipal` and does not collide with `GrantPrincipalType`.

## Out of scope

PATCH (#1021), human-session callers (#1022), foothold SQL (#1023), `grant.*` catalog (#1024), console (#1025), nested User expand.

## Validation

- `go test ./internal/service/ ./internal/api/ ./internal/storage/dialect/authz/ -count=1` — pass
- `go test -tags sqlite_integration ./internal/storage/stmttest/ -run 'TestAuthzAssignmentStatements_ListManagedGrants|TestCursorBattle_DrainAllListIncarnations|TestSchemaNullabilityMatchesDDL'` — pass (including `grants_expires_at` ASC/DESC drain)
- `go test -tags postgres_integration ./internal/api/integration_test/ ./internal/storage/stmttest/ -run 'TestGrant|TestAuthzAssignmentStatements_ListManagedGrants|TestSchemaNullabilityMatchesDDL'` — pass
- `moon run server:generate` — `queryGrants-error-response.yaml` still has nine errors, including `user.permission_denied` and `team.permission_denied`

## Release notes / changeset

Changeset: `.changeset/grant-query-list.md` — `@zitadel/server` **minor**.

Add POST /grants/query to list a project's collaboration grants, with always-on user-ref or team-ref labels and an opt-in `expand: ["principal"]` that embeds the GET user or GET team body.

No new changeset for these review fixes — unreleased corrections on the same minor.

## Notes

- Unrelated on this branch (do not squash/rewrite): `test(cli): retry doctor when ephemeral port collides`.
- Filter/sort stay on `principal_id` / `principal_type` / `relation` / `created_at` / `expires_at`, not display strings.
- QueryUsers `team.read` still uses `user.permission_denied`; grant-principal `team.read` uses `team.permission_denied`.
- Wiring `requireUserRead` / a grant-team helper from `QueryGrants` dropped both expand 403s from the generated error catalog (9 → 7). Constructors stay inline.

<div><a href="https://cursor.com/agents/bc-3438f64d-3961-4b3a-919d-883b0a07f3b5?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3438f64d-3961-4b3a-919d-883b0a07f3b5&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


internal/storage/dialect/postgres/authz_assignment.go
```diff
@@ -182,4 +220,43 @@ func scanAuthzAssignment(row pgx.CollectableRow) (*domain.AuthzAssignment, error
 	return a, nil
 }
 
+var authzAssignmentSchema = database.NewSchema(map[domain.AuthzAssignmentField]database.FieldBinding[domain.AuthzAssignment]{
```
IAM-marco: This same 39-line schema block is pasted three times, once per dialect (here, sqlite `authz_assignment.go:256`, spanner `authz_assignment.go:272`). On top of that, the check that verifies the schema's can-be-empty flags against the real DDL was never told this table exists, so it silently checks nothing: [internal/storage/AGENTS.md#L173-L176](https://github.com/zitadel/nextgen/blob/74eab988728c3bb47e7741d0c66ac5cd8c0a34a7/internal/storage/AGENTS.md#L173-L176) says a new schema must be registered or its columns go unchecked. Copilot's threads on these files point at the same gap.

Both problems disappear in one move: define the block once as `authz.AuthzAssignmentSchema`, next to [MembershipEdgeSchema](https://github.com/zitadel/nextgen/blob/74eab988728c3bb47e7741d0c66ac5cd8c0a34a7/internal/storage/dialect/authz/membership_edge_schema.go#L8-L10) which already does it this way, and register it with one `schematest.Columns("authz_assignments", authz.AuthzAssignmentSchema)` line in [sharedSchemaColumns](https://github.com/zitadel/nextgen/blob/74eab988728c3bb47e7741d0c66ac5cd8c0a34a7/internal/storage/stmttest/schema_nullability_test.go#L38-L40). That's roughly 78 duplicated lines gone and the coverage gap closed. Sharing is safe here: sqlite's unix-nano time handling lives in its compiler layer, not in these bindings.

internal/storage/stmttest/cursor_battle_test.go
```diff
@@ -134,6 +135,27 @@ func battleTeams(t *testing.T, d dialect) {
...
+		a := newTestAssignment(projectID, "", domain.AuthzPrincipalTypeUser, "user_battle_"+uniqueSuffix(t)+string(rune('a'+i)), "project", "viewer", domain.NewProjectAssignmentScope())
+		require.NoError(t, d.stmts.CreateAuthzAssignment(t.Context(), a))
+		want = append(want, a.ID)
+	}
+	orderAsc := database.OrderBy[domain.AuthzAssignmentField]{
+		Columns:   []database.Column[domain.AuthzAssignmentField]{database.Col(domain.AuthzAssignmentFieldID)},
+		Direction: database.OrderAsc,
+	}
+	slices.Sort(want)
+	drainIncarnation(t, want, orderAsc, func(page database.Page[domain.AuthzAssignmentField]) (*database.ListResult[*domain.AuthzAssignment], error) {
```
IAM-marco, line 154: Callers can sort grants by `expires_at` ([grant.go#L586-L589](https://github.com/zitadel/nextgen/blob/74eab988728c3bb47e7741d0c66ac5cd8c0a34a7/internal/service/grant.go#L586-L589) plus `grant-sorting-field.yaml`), and that column can be empty. Paging through a list sorted by a sometimes-empty column is exactly where cursors break, which is why [internal/storage/AGENTS.md#L165-L167](https://github.com/zitadel/nextgen/blob/74eab988728c3bb47e7741d0c66ac5cd8c0a34a7/internal/storage/AGENTS.md#L165-L167) asks for a dedicated drain case ("Nullable sort columns need an additional drain case").

This test never gets there: every seeded grant has no expiry, and the drain only sorts by `id`. The sessions tests show the shape to copy: [session_test.go#L188-L218](https://github.com/zitadel/nextgen/blob/74eab988728c3bb47e7741d0c66ac5cd8c0a34a7/internal/storage/stmttest/session_test.go#L188-L218) mixes empty and set values and drains in both directions. Copilot's comment here points the same way.

internal/api/authz.go
```diff
@@ -334,38 +334,34 @@ func hasOperatorProjectWrite(granted []string) bool {
...
-// is not minted yet (ADR 036). TODO(#420): drop it once granular scopes are.
+	return denied().WithMessage(msg)
+}
+
+func requireMembershipRead(ctx context.Context) error {
+	return requireExpandScope(ctx, "team_membership.read", domain.ErrUserPermissionDenied,
+		"reading a user's team memberships requires team_membership.read")
+}
+
+func requireUserRead(ctx context.Context) error {
```
IAM-marco: Nobody calls `requireUserRead`. `QueryGrants` does the same check by hand ([grant.go#L52-L63](https://github.com/zitadel/nextgen/blob/74eab988728c3bb47e7741d0c66ac5cd8c0a34a7/internal/api/grant.go#L52-L63)) because of the `gen_openapi_errors` concern noted there, so this helper only exists for its own unit test. And since the repo has no linter that flags unused functions, CI will never complain; it just sits here.

Either actually call it from `QueryGrants` (if `erroranalysis` can follow the constructor through `requireExpandScope`; the untouched `queryUsers-error-response.yaml` hints it can, but I could not confirm that for func parameters), or delete it together with its test case.

internal/api/grant.go
```diff
@@ -26,22 +26,79 @@ func (h *Handler) CreateGrant(ctx context.Context, req *api.CreateGrantRequest,
...
+		Limit:     int(req.Limit.Or(0)),
+		PageToken: string(req.PageToken.Or("")),
+	}
+	if sorting, ok := req.Sorting.Get(); ok {
+		svcReq.Sorting = sortingToService(sorting.Field, sorting.Direction)
+	}
+	for _, filter := range req.Filter {
+		svcReq.Filters = append(svcReq.Filters, filterToService(filter.Field, filter.Operation, filter.Value))
+	}
+	for _, expand := range req.Expand {
```
IAM-marco: A loop with a one-case switch, just to set one boolean. Unknown expand values are rejected by ogen validation before this code runs, so the switch's ignore-unknown behavior never triggers. One line does the same thing, the way [flow_definition.go#L60-L62](https://github.com/zitadel/nextgen/blob/74eab988728c3bb47e7741d0c66ac5cd8c0a34a7/internal/api/flow_definition.go#L60-L62) already does it:

```go
svcReq.IncludePrincipal = slices.Contains(req.Expand, api.GrantExpandPrincipal)
```

(`user.go`'s loop-switch is fine as is; it has two cases.)

IAM-marco (changes requested):

IAM-marco (approved):

PR #1119, opened 2026-09-03
docs(queue): propose SQL-backed background jobs

adlerhurst:
## Summary

Proposes **ADR 064** (`docs/adrs/064-background-jobs.md`) for a SQL-backed, in-process background job loop so periodic sweeps and queued work share one runtime across Postgres, Spanner, and SQLite.

The number is **064** because 061 (egress policy), 062 (per-environment variables), and 063 (resource revisions) landed on `main` while this PR was open. The pick-up step is **Take**, not Claim, so it does not collide with ADR 046 project claim.

Locked decisions (each has one home in the ADR):

- One jobs table; row model and states are specified up front (`pending` / `leased` / `done` / `dead`)
- `JobStatements` on `AllStatements`: `Enqueue`, `Take`, `Complete`, `Fail`, `UpsertPeriodic`, `DeleteCompleted`, `Heartbeat`. No memory/River/Pub/Sub backend in the first implementation
- Portable `Enqueue` is lookup-then-insert; Spanner live uniqueness is `active_unique_key` (NULL on `done`/`dead`); Postgres partial unique index on live rows
- Take owns expiry: pending or lease-expired rows, including `not_after <= now()` (exclusive); mark `dead` without Perform. GC deletes only terminal rows by `completed_at`
- `lease_token`: Heartbeat / Complete / Fail must match it so a late Complete cannot overwrite a new lease
- Queued Fail returns `pending` with backoff; `dead` at `max_attempts` or when retry would miss `not_after`. Periodic Fail reschedules `run_at = now() + period` and never reaches `dead`/`done`
- Take Perform only for names this binary registered; unknown names are left untouched
- Engine opens the tx for Take / Heartbeat / Complete / Fail / DeleteCompleted; Enqueue uses the caller’s product tx. Go 1.27 does not change that
- Lease 15m + heartbeat 5m; max_attempts 5; backoff cap 1h; retain 7d
- Metrics (§8) labeled by handler `name` only; `jobs_failed` is any Fail back to `pending`; `jobs_expired` is deadline skip (may have run once already)
- First implementation = table + engine + cut `RetentionJob` to `events.retention`; `sessions.gc` / #881 is the first follow-up

ADR 046’s Proposed amendment matches the current non-goal: the 14-day claim window stays; **deletion** of unclaimed projects can run on ADR 064 later and is still not added here.

## Validation

- `node scripts/check-pr-title.mjs --title "docs(queue): propose SQL-backed background jobs"` — ok
- Docs-only: second-pass review (Claim→Take, lease token, exclusive `not_after`, Spanner live key, `jobs_failed` / `jobs_expired`, token-row GC dropped, 046 amendment rewritten)

## Release notes / changeset

## Notes

- Second-pass review replies are on the threads. Threads left unresolved for Marco / Copilot.
- Implementation (migrations, `JobStatements`, cutting over `RetentionJob`, zotel wiring) is out of this PR.
- Follow-up after accept: `sessions.gc` / #881 as the first emit+delete job.

<div><a href="https://cursor.com/agents/bc-a858bb29-9627-486c-ac72-3e3030b1a679?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a858bb29-9627-486c-ac72-3e3030b1a679&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,237 @@
...
+### Row model
+
+One `jobs` table. Columns (logical; DDL is dialect-owned):
+
+| Column | Role |
+|--------|------|
+| `id` | dialect-minted `job_<opaque>` ([ADR 047](047-dialect-id-generation.md) §3). Not an HTTP resource. |
+| `name` | handler to run. Not unique. Many queued rows share one `name`. |
+| `payload` | opaque bytes; empty for sweeps. |
+| `unique_key` | uniqueness among `pending`/`claimed` only, nullable. Multiple nulls allowed. Periodic: required, equal to `name`. Queued: optional (idempotent enqueue among live rows). A `done`/`dead` row does not block a later Enqueue with the same key. |
```
IAM-marco: I don't understand what this sentence mean

docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,237 @@
...
+### 1. Two row kinds, one `ORDER BY run_at`
+
+Periodic sweeps and queued work are both rows.
+
+| Kind | Rows due at once | In flight cluster-wide |
+|------|------------------|------------------------|
+| Unique periodic (`jobs.gc`, later `sessions.gc`) | one row (`unique_key` = `name`) | at most one of that name |
+| Queued (`email.send`) | one per unit of work | up to remaining claimers |
+
+Parallelism is in-flight `Perform`s: `replicas × jobs.concurrency`. Named queues, per-name concurrency caps, and fair mixing are out of v1. A flood of queued rows can delay a due periodic row; that starvation is accepted.
```
IAM-marco: Please rephrase this so that it's understandable

docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,237 @@
...
+The dialect adapter is [`service.AllStatements`](../../internal/service/statement.go). `JobStatements` is tx-passive. Dialects hand-write the SQL.
+
+- `Enqueue` — product transaction; among live rows, a `unique_key` conflict keeps the existing row.
+- `Claim` — lease runnable rows; mark expired-unrun rows `dead` without Perform.
+- `Complete` / `Fail` — §3 (queued vs periodic).
+- `UpsertPeriodic` — boot; not `Enqueue`. Insert sets cursor fields. On conflict: update `period` from config only. Must not clobber `claimed_until`, `claimed_by`, or `run_at` of a live lease.
+- `DeleteCompleted` — `done`/`dead` where `completed_at < now() - retain`.
+- `Heartbeat` — `UPDATE claimed_until` where `id` and `claimed_by` are this replica.
+
+v1 does not add a `Backend` interface, a memory backend, River, Pub/Sub, or Cloud Tasks.
```
IAM-marco: What is v1??

docs/adrs/061-background-jobs.md
IAM-marco, line 1: In general, use the bro skill on this ADR because I'm doubting my understanding of the english language

docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,237 @@
...
+      Node->>Jobs: mark dead, no Perform
+    else runnable and registered
+      Node->>Handler: Perform
+      Node->>Jobs: Heartbeat during Perform
+      Node->>Jobs: Complete or Fail
+    end
+  end
+```
+
+1. **Claim** — select a due row (pending or lease-expired). Then:
```
IAM-marco: Why are you picking up `pending` ? And in any case, specifically for claiming, don't you just need to check on the creation date of the project (`WHERE project.created_at + '14 days' > now()' )?

IAM-marco (changes requested):
Made a first pass, got to paragraph 5.

Let's see how it gets on a second round

adlerhurst (commented):
Thanks for the hint to the bro skill @IAM-marco. The ADR is rewritten and IMO better to understand now.

docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,254 @@
...
+> **Amends if accepted:** [ADR 046](046-claim-lifecycle-v2.md) (the “no scheduled-task infrastructure” non-goal)
+> **Related:** [ADR 010](010-session-auth-attempt-check-model.md), [ADR 037](037-token-lifecycle.md), [ADR 039](039-signing-key-rotation-and-incident-response.md), [ADR 050](050-dev-inbox.md), [#881](https://github.com/zitadel/nextgen/issues/881)
+
+## Context
+
+Today each background sweeper has its own loop in the process:
+
+- Event retention ([`internal/audit/retention.go`](../../internal/audit/retention.go)) is started from [`cmd/server/server.go`](../../cmd/server/server.go).
+- The event shipper is a separate poller.
+- Path A request events sit in an in-process buffer; [ADR 048](048-wide-events-internal-audit-primitive.md) defers a durable outbox.
```
IAM-marco, line 16:
```suggestion
- Path A (request wide events) sit in an in-process buffer; [ADR 048](048-wide-events-internal-audit-primitive.md) defers a durable outbox.
```

docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,254 @@
...
+Storage is SQL on three dialects ([ADR 028](028-storage-v2-statements-and-dialects.md)): PostgreSQL, Spanner, and SQLite. A job system that exists only on Postgres (River) or only on GCP (Pub/Sub) does not cover that set.
+
+## Decision
+
+**First implementation** means what this ADR ships: the table, the in-process loop, and cutting event retention onto it. Later jobs (`sessions.gc`, mail) are called out as follow-ups, not as “v1.”
+
+How a job runs:
+
+1. Someone inserts a row (`Enqueue`) or, for a sweep, boot upserts one row (`UpsertPeriodic`).
+2. A running server **Claims** that row: it takes a time-limited lease so other servers leave it alone.
```
IAM-marco, line 31: Can we find an alternative word or a synonym for claims so that we don't get confused with actual claiming?

docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,254 @@
...
+| Unique periodic (`jobs.gc`, later `sessions.gc`) | one row (`unique_key` = `name`) | at most one of that name |
+| Queued (`email.send`) | one per unit of work | as many as there are free workers |
+
+How many jobs run at once: `number of processes × jobs.concurrency`. With three servers and `jobs.concurrency = 1`, three jobs run at a time.
+
+There is no separate mail queue versus GC queue. Workers take the next due row, whatever `name` it has. If 10,000 `email.send` rows are due, `jobs.gc` waits its turn. We accept that for the first implementation. Per-name caps and fair mixing are later work.
+
+### 2. Portability is `JobStatements`
+
+The dialect adapter is [`service.AllStatements`](../../internal/service/statement.go). Dialects write the SQL. The methods do not start their own transaction: they use the pool or the open transaction the caller already holds, same as `CreateSession`.
```
IAM-marco, line 76: Is this going to be true also for Golang 1.27?

docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,254 @@
...
+
+```mermaid
+sequenceDiagram
+  participant Node as Replica
+  participant Jobs as jobs_table
+  participant Handler as Perform
+
+  loop poll_interval
+    Node->>Jobs: Claim due rows
+    alt expired not_after
```
IAM-marco, line 99: What does Alt mean

docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,254 @@
...
+      Node->>Jobs: Heartbeat during Perform
+      Node->>Jobs: Complete or Fail
+    end
+  end
+```
+
+1. **Claim** — pick a due row (waiting, or lease expired). Then:
+   - If `not_after` is already in the past: mark `dead`, set `completed_at`, do not Perform. Any replica may do this. It does not need to know the handler. This is how a job that outlived its useful life is closed, including after a crash.
+   - Else if this process registered that `name`: write the lease (`claimed_until = now() + lease_duration`), commit, Perform.
+   - Else: leave the row. Do not Fail it. An old binary, or a process that did not register `events.retention`, must not burn retries on a job it cannot run.
```
IAM-marco, line 112: This part is not clear to me. By process do you mean the current container running? How does a process know that it wrote that name? How do we handle situations in which another job is processing the row so it's not dead but not_after is already expired?

docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,254 @@
...
+
+Operators watch the engine. There is no job HTTP API and no dead-letter table.
+
+Every series is labeled by handler `name` (`jobs.gc`, `email.send`, …). No other labels in the first implementation (`status`, replica, dialect). The outcome is the series name, not a `status` label. Increment one per row, not per batch.
+
+Counters:
+
+- `jobs_claimed` — Claim leased a row that will Perform.
+- `jobs_completed` — Complete succeeded (queued → `done`, or periodic reschedule).
+- `jobs_failed` — queued Fail with retry (`pending` + backoff).
```
IAM-marco, line 168: This line says `jobs_failed` only counts failed queued jobs. But line 172 says failed periodic jobs also count. Both can't be right as written, and someone setting up alerts will read this line and think periodic failures are counted nowhere. If it's both, just say so here, for example: "any Fail that puts the row back to `pending` (queued retry, or a periodic failure)".

https://github.com/zitadel/nextgen/blob/d755538c6d12fe0af603a86e8ce5767a8d7128f7/docs/adrs/061-background-jobs.md#L166-L173

docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,254 @@
...
+### Row model
+
+One `jobs` table. Column types are dialect-owned; this is the logical shape:
+
+| Column | Role |
+|--------|------|
+| `id` | dialect-minted `job_<opaque>` ([ADR 047](047-dialect-id-generation.md) §3). Not an HTTP resource. |
+| `name` | which handler to run (`jobs.gc`, `email.send`, …). Not unique. Many queued rows share one `name`. |
+| `payload` | opaque bytes. Empty for sweeps. |
+| `unique_key` | optional. Stops two **live** rows (`pending` or `claimed`) from being the same logical job. Many nulls are allowed. Details below. |
```
IAM-marco, line 44: The table says `unique_key` is optional, full stop. But the bullet at line 55 says periodic rows must always have it. So the table and the text below it disagree. An earlier commit (6208f2ff) had the table spell it out: "Periodic: required, equal to `name`. Queued: optional". The rewrite lost that. Let's just put it back in the cell.

https://github.com/zitadel/nextgen/blob/d755538c6d12fe0af603a86e8ce5767a8d7128f7/docs/adrs/061-background-jobs.md#L43-L57

docs/adrs/README.md
```diff
@@ -66,3 +66,4 @@ This directory contains architecture decision records (ADRs) for nextgen.
...
+| [061](061-background-jobs.md)                             | SQL-Backed Background Jobs | Proposed | One jobs table and in-process lease loop for periodic sweeps and queued work; `JobStatements` on `AllStatements`; no memory/River/Pub/Sub backend in v1. Cut event retention onto the loop; session reaper / #881 is the first follow-up. |
```
IAM-marco, line 69: We removed "v1" from the ADR itself after the "What is v1??" thread (the ADR now says "first implementation", line 26), but this index row still says "in v1". It's the last "v1" left in the whole index, and it's the first thing people read. Just change it to "in the first implementation".

https://github.com/zitadel/nextgen/blob/d755538c6d12fe0af603a86e8ce5767a8d7128f7/docs/adrs/061-background-jobs.md#L25-L27

docs/adrs/046-claim-lifecycle-v2.md
```diff
@@ -36,6 +36,13 @@
 > bounds *automatic provisioning* to one team per user; it does not change
 > §Context's rule that a user may belong to many teams.
 >
+> **Proposed amendment — [ADR 061](061-background-jobs.md):**
+> if ADR 061 is accepted, the [§Non-goals](#non-goals) statement that there is
+> no scheduled-task infrastructure (all TTLs are read-time filtering) no longer
```
IAM-marco, line 41: This block quotes the old wording of the Non-goals bullet. But main has already rewritten that bullet (e04e3ceb), and the new version already talks about ADR 061 and the reaper. Because the two spots are far apart in the file, the merge will go through without a conflict, and we'd end up with this block quoting text that no longer exists and repeating what the bullet already says. When rebasing, rewrite this against main's version of the bullet, or just edit that bullet directly instead.

Main's current bullet: https://github.com/zitadel/nextgen/blob/84d1c4b7dc636dc4439c0ee7a21eee8ecad44f36/docs/adrs/046-claim-lifecycle-v2.md#L260-L270

docs/adrs/061-background-jobs.md
```diff
@@ -0,0 +1,254 @@
...
+- Cut [`RetentionJob`](../../internal/audit/retention.go) over to unique periodic `events.retention` that still calls `DeleteEventsOlderThan` ([ADR 049](049-events-api-retention-export.md))
+- Unchanged: event shipper, Path A request buffer
+- `email.send` is an example producer, not a deliverable of this first implementation
+- Next job: `sessions.gc` / [#881](https://github.com/zitadel/nextgen/issues/881). Read-time session expiry stays; the reaper is the Path B producer (emit `session.expired` then delete or mark). One unique periodic row, not one job per session.
+
+## Non-goals
+
+- Unclaimed-project expiry ([ADR 046](046-claim-lifecycle-v2.md)). This ADR does not add `WHERE projects.created_at + interval` and does not delete unclaimed projects. A future sweeper would be its own periodic job on this loop.
+- [ADR 049](049-events-api-retention-export.md)’s dedicated `DELETE` role on the jobs table
+- Token-row GC, signing-key purge, and ADR 050 outbound as handlers in this first implementation
```
IAM-marco, line 226: This lists cleaning up token rows as future work for the job system. But ADR 037 (Accepted) already decided we never need that: a revoked token's row is deleted on the spot, and expired tokens never pile up because expiry is checked from the token itself, not the database. So there is nothing to collect. Drop "Token-row GC" here, or say what would actually need cleaning up. (The signing-key part of this line is fine, ADR 039 does plan a purge.)

https://github.com/zitadel/nextgen/blob/d755538c6d12fe0af603a86e8ce5767a8d7128f7/docs/adrs/037-token-lifecycle.md#L74-L80

IAM-marco (changes requested):
Second pass at d755538c. The rewrite reads much better; all threads from the first round are resolved. Ten inline comments this round: five questions, and five spots where the text contradicts itself or another ADR (the `jobs_failed` definition, the `unique_key` table cell, a leftover "v1" in the index row, the ADR 046 amendment vs main's rewritten bullet, and token-row GC vs ADR 037).

IAM-marco (approved):
Third pass at 0a869e50, all threads closed. Checked the new material against the code and the cited ADRs: the Take rename is swept clean through the row model, statements, metrics, and tests; the Spanner active_unique_key trick matches the authz migration comment it cites (000018_authz_mvp.sql lines 9-11); the Go 1.27 TODO in statement.go really is about generic pool types; and CreateSession does start its own withTransaction, so it is the right counter-example. The renumber to 064 is complete (no stale 061-background-jobs references, check-adrs-index passes).

One non-blocking note for a follow-up commit if you like: §3 step 1 says a late Complete after Take marked the row dead is a no-op because of lease_token, but the mark-dead path does not say it mints a new token or clears the lease, so as written the old token would still match. Saying that Complete/Fail also require status = leased (or that marking dead clears the lease) closes it. Both states are terminal, so nothing bad happens either way.

IAM-marco (changes requested):
Content is fine (see the approval just before this one), but the branch cannot merge

docs/adrs/065-background-jobs.md
```diff
@@ -0,0 +1,270 @@
...
+
+   If `not_after` passes **while this replica still holds a live lease**, it finishes Perform. The deadline is “do not **start** at or after.” Do not steal mid-run. After the lease expires, Take may see the row and mark `dead` without a second Perform. The old worker’s late Heartbeat, Complete, or Fail must send the old `lease_token` and is a no-op.
+
+2. **Perform** — do the handler work in its own transactions or I/O. Keep Heartbeat going so the lease does not expire under a long run.
+3. **Complete** or **Fail**. Both must match `lease_token` and only apply while the row is still `leased`.
+
+Complete:
+
+- Queued success → `done` + `completed_at`. Clear `active_unique_key` / the live unique projection. Reset `attempt`.
+- Periodic success → keep the same row. Clear the lease. Set `run_at = now() + period` (skip missed beats; do not add `period` onto the old `run_at`). Reset `attempt`. Never `done`. No history row.
```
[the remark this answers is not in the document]
IAM-marco: Fixed in 03f5148c: periodic Complete and Fail now say "Clear the lease and set status = pending", so Take's pending path picks the row up again (queued Fail already said it), and the stmttest bullet covers the reschedule landing back in pending.

docs/adrs/065-background-jobs.md
```diff
@@ -0,0 +1,270 @@
...
+3. **Perform** does the real work (delete old events, send mail, …) outside that Take transaction.
+4. **Complete** or **Fail** writes the outcome back.
+
+### Row model
+
+One `jobs` table. Column types are dialect-owned; this is the logical shape:
+
+| Column | Role |
+|--------|------|
+| `id` | dialect-minted `job_<opaque>` ([ADR 047](047-dialect-id-generation.md) §3). Not an HTTP resource. |
```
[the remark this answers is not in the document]
IAM-marco: Registered in 03f5148c: ADR 047 section 2 now has the "job row | job" entry with an "Amended by ADR 065" note, the 047 header records the amendment, and the ADR 065 header lists 047 section 2 under "Amends if accepted".

docs/adrs/065-background-jobs.md
```diff
@@ -0,0 +1,270 @@
...
+- `DeleteCompleted` — delete `done`/`dead` rows whose `completed_at` is older than the retain window.
+- `Heartbeat` — extend `lease_until` where `id` and `lease_token` match this Take.
+
+The first implementation does not add a `Backend` interface, a memory backend, River, Pub/Sub, or Cloud Tasks.
+
+`Enqueue` is lookup-then-insert inside the caller’s transaction. Spanner will not use `INSERT ... ON CONFLICT` on a `NULL_FILTERED` unique index ([`internal/storage/dialect/spanner/auth_attempt.go`](../../internal/storage/dialect/spanner/auth_attempt.go)). Lookup-then-insert is also not enough by itself: a unique index on `unique_key` would still see `done`/`dead` rows, so a resend would bounce until GC. Same trick as authz in [`internal/storage/dialect/spanner/migration/sql/000018_authz_mvp.sql`](../../internal/storage/dialect/spanner/migration/sql/000018_authz_mvp.sql) (lines 9–11): Spanner keeps `active_unique_key` equal to `unique_key` while the row is `pending`/`leased`, and **NULL** when `done`/`dead`. Postgres uses a partial unique index on live rows. When a queued row finishes or dies, that live key is cleared so the next Enqueue can insert.
```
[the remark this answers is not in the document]
IAM-marco: Reworked in 03f5148c: the portable contract is now "insert unless a live row exists; a concurrent loser keeps the winner's row and does not fail the product transaction". Postgres and SQLite implement it with INSERT ... ON CONFLICT DO NOTHING on the live-key partial index (plain lookup-then-insert is called out as insufficient there, citing executeTransaction's lack of retry); Spanner keeps lookup-then-insert because its read/write transaction serializes the read and the client retries aborted commits. Concurrent Enqueue is added to the contract-test list.

docs/adrs/065-background-jobs.md
```diff
@@ -0,0 +1,270 @@
...
+- SQLite: single writer
+
+Two clocks: `jobs.poll_interval` is how often a process asks the table for due work. `period` on a unique row is how often that sweep may run.
+
+Defaults for the first implementation (config key names are an implementation detail; the numbers are the contract):
+
+| Knob | Default | Role |
+|------|---------|------|
+| `jobs.concurrency` | `1` | How many Take/`Perform` loops this process runs in parallel |
+| `jobs.poll_interval` | ~1s | Idle wait between Takes |
```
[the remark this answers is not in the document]
IAM-marco: Pinned to 1s in 03f5148c. The jitter stays described as Spanner-specific poll behavior in section 9.

docs/adrs/065-background-jobs.md
```diff
@@ -0,0 +1,270 @@
+# ADR 065: SQL-Backed Background Jobs
```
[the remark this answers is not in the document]
IAM-marco: PR description updated: it now says ADR 065 and docs/adrs/065-background-jobs.md, and lists 064 (CLI resource commands) among the numbers that landed on main while this PR was open.

docs/adrs/065-background-jobs.md
```diff
@@ -0,0 +1,270 @@
...
+- Jobs table, `JobStatements`, in-process loop, unique periodic `jobs.gc`
+- Cut [`RetentionJob`](../../internal/audit/retention.go) over to unique periodic `events.retention` that still calls `DeleteEventsOlderThan` ([ADR 049](049-events-api-retention-export.md))
+- Unchanged: event shipper, Path A request buffer
+- `email.send` is an example producer, not a deliverable of this first implementation
+- Next job: `sessions.gc` / [#881](https://github.com/zitadel/nextgen/issues/881). Read-time session expiry stays; the reaper is the Path B producer (emit `session.expired` then delete or mark). One unique periodic row, not one job per session.
+
+## Non-goals
+
+- Unclaimed-project deletion ([ADR 046](046-claim-lifecycle-v2.md)). This ADR does not add `WHERE projects.created_at + interval` and does not delete unclaimed projects. A future sweeper would be its own periodic job on this loop.
+- [ADR 049](049-events-api-retention-export.md)’s dedicated `DELETE` role on the jobs table
```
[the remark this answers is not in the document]
IAM-marco: Rephrased in 03f5148c: the non-goal is now "a dedicated DELETE role on the jobs table, analogous to ADR 049's events-table role", so it no longer implies ADR 049's role covers the jobs table.

IAM-marco (approved):
Fourth pass at 03f5148c. The renumber to 065 is complete and the index gate passes with 65 records; the lease fixes close the request-changes items (mark-dead clears the lease, Complete/Fail need a live leased row, periodic Complete/Fail land back in pending); all six Copilot findings are fixed, replied to, and resolved; CI is green on both pushes. Good to squash.

PR #1123, opened 2026-09-03
fix(console): give CI test runs cold-cache headroom

livio-a:
`console:test` has failed on the first CI run after every large rebase (observed three times on the ADR 058 stack), always with the same signature: many unrelated `_authed` screens failing at ~5s `findBy` timeouts, while the identical head passes locally (132/132) and on every warm rerun.

The cause is the cold module-graph transform on a fresh runner: its startup cost lands inside the first specs' async clocks. This bumps two budgets **under CI only** — vitest `testTimeout` 5s → 20s and testing-library's `asyncUtilTimeout` 1s → 5s — keeping the tight local budgets so a genuinely hung test still fails fast at a desk.

No test changes; verified the full console suite green in both modes (default and `CI=1`).

### DIFF
--- a/apps/console/src/test-setup.ts
+++ b/apps/console/src/test-setup.ts
@@ -1,4 +1,12 @@
 import "@testing-library/jest-dom/vitest";
+import { configure } from "@testing-library/react";
+
+// Companion to the CI-only vitest testTimeout in vite.config.mts: the async
+// utils (findBy*, waitFor) default to 1s, which the cold-cache transform of
+// the first CI run eats before the screen under test ever renders.
+if (process.env.CI) {
+  configure({ asyncUtilTimeout: 5_000 });
+}
 
 // @ts-expect-error Needed for tests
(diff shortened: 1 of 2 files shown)


livio-a:
Closing — the console:test cold-start flake needs a proper look by the frontend team. Findings so far, for whoever picks it up: fails only on the first CI run after a large change (three times on the ADR 058 stack rebases), mass ~5s findBy timeouts across unrelated _authed specs, identical heads green locally (132/132) and on every warm rerun. Root cause is esbuild dep pre-bundling + cold transforms of the module graph landing inside the first specs' async budgets on a fresh runner. Options weighed: CI-only timeout bumps (blunt — slack applies to warm runs too), persisting node_modules/.vite via actions/cache split restore/save (fixes rebase-pushes outright, but a brand-new branch's first run stays cold since nothing saves caches on main), or a setup-time module-graph warmup (needs care: some specs rely on vi.stubEnv before dynamically importing the router).

PR #1129, opened 2026-09-03
feat(api): create releases

### LINKED ISSUE #531: Service - Releases persistence & endpoints

vitorbari:
Implements `POST /releases`.

Screenshot 2026-09-07 at 11 01 13

## What is in this PR

- `ReleaseService.Create`, and the service wired into the handler and the server
- Handle resolution per kind (schema `objectType`, flow definition `name`, branding `default`)
- Idempotency: content-hash lookup before insert, plus race resolution on the unique index

## Out of scope

- Cross-resource closure validation (reference between resources in the same release)
- Deploying a release to an environment


internal/service/release.go
```diff
@@ -0,0 +1,215 @@
...
+//
+// TODO(#1154): the pinned set is not checked for closure. A flow pinned here
+// may reference a user schema this release does not pin, and nothing objects.
+// The check belongs in this loop, which already holds every resolved handle and
+// has read each revision — but it is not expressible while a flow references
+// its schema by revision id instead of by handle (#1144).
+func (s *releaseService) resolvePointers(ctx context.Context, projectID string, requested []CreateReleasePointer) ([]domain.ReleasePointer, error) {
+	pointers := make([]domain.ReleasePointer, 0, len(requested))
+	for _, pointer := range requested {
+		handle, err := s.resolveHandle(ctx, projectID, pointer)
```
vitorbari, line 158: We can probably optimize this by retrieving resources in parallel. Not urgent, because creating a release is not in the hot-path.

api/openapi/endpoints/releases/create-release-request.yaml
```diff
@@ -11,6 +11,7 @@ properties:
   pointers:
     type: array
     minItems: 1
+    maxItems: 50
```
IAM-marco, line 14: Spec gap: this cap is new, and the commit message says a different number.

Issue #531 defines `pointers` as a plain list of (kind, handle, revision_id) tuples with no size limit. This line adds a hard cap: the decoder rejects any request with more than 50 pointers before the handler runs. The same cap sits on the response schema in `release.yaml` line 35.

The cap itself is reasonable: every pointer costs one revision lookup, so without a bound the request decides how much database work the server does. Two things are off around it though:

1. It quietly narrows the spec. A project with more than 50 configurable resources can never assemble a release that pins everything it owns. Today that is far above any real project, so all this needs is an explicit acknowledgement, in the PR body or on #531, so the limit reads as a decision rather than an accident.
2. The message of commit 2f9934de says "maxItems: 250 bounds it in the contract", but both contract files say 50. One of the two is not what was intended. Worth confirming 50 is the number you meant.

No test pins the boundary either way. If the cap is meant to be load-bearing, one request with 51 pointers in the integration test would keep it honest.

internal/service/release.go
```diff
@@ -0,0 +1,224 @@
...
+type CreateReleaseOutput struct {
+	Release *domain.Release
+	// Created distinguishes a new release from one that already pinned this
+	// exact set, which the caller answers 200 to rather than 201.
+	Created bool
+}
+
+type ReleaseService interface {
+	Create(ctx context.Context, input CreateReleaseInput) (*CreateReleaseOutput, error)
+}
```
IAM-marco, line 40: Style note: this interface has one implementation, no mock, and the test that receives it gets nil.

`ReleaseService` is declared as an interface, but only `releaseService` implements it, and it is not in the mockgen list in `internal/service/mocks/generate.go`. The package has two established shapes: plain concrete structs (`TeamService`, `BrandingService`, `GrantService`, `EventService`) and interfaces that earn their keep through a generated mock (`ProjectService`, `TokenService`). This one sits in between: the interface shape without the mock that is the reason to have one.

The visible consequence is `internal/api/flow_test.go` line 121, where the unit-test server is now built with `nil` for the release service. That holds only for as long as no test on that server touches a release endpoint; the first one that does will panic on the nil service instead of failing with a useful message.

Two honest ways out: make it a concrete struct like `TeamService`, or add it to the mockgen generate list and hand `flow_test.go` a mock instead of nil. Since the two read endpoints are about to land on top of this same service, the mock route probably pays for itself first.

internal/api/integration_test/release_test.go
```diff
@@ -0,0 +1,259 @@
...
+			Pointers: fixture.pointers(),
+			Message:  api.NewOptString("a different message entirely"),
+		})
+		require.IsType(t, &api.CreateReleaseOK{}, resp, "recreate release: %s", helpers.MustMarshal(t, resp))
+		reused := api.Release(*resp.(*api.CreateReleaseOK))
+
+		assert.Equal(t, created.ID, reused.ID)
+		assert.Equal(t, "initial import", reused.Metadata.Message.Value,
+			"the stored release keeps the message it was created with")
+	})
```
IAM-marco, line 150: Test note: the reuse path never proves the reused response carries the pointers.

This subtest re-submits the same set and checks the server answers 200 with the original release: same id, and the stored message rather than the new one. What it never looks at is `reused.Pointers`. The 200 body is built from the release the content-hash lookup found, which is a different code path from the 201 body built from the insert, so the one response shape this PR adds that goes completely unchecked is exactly the reused one.

The dialect round-trip tests already cover loading pointers back from storage, so the exposure is the handler mapping, not the storage. One extra assertion here, that `reused.Pointers` equals `created.Pointers`, closes it.

IAM-marco (changes requested):

api/openapi/endpoints/releases/create-release-request.yaml
vitorbari: Yes, 50 is the intended cap. It should be more than enough for now. 
The commits will be squashed, so that message does not really matter.

internal/service/release.go
vitorbari: Makes sense. I will add mockgen

api/openapi/endpoints/releases/create-release-request.yaml
vitorbari: Added a mention to the changeset and also tests for the max 50 pointers cap.

IAM-marco (approved):
All three points from the last review are addressed by the new commits: the 50-pointer cap is confirmed as intended, documented in the changeset, and pinned by a boundary test that posts 51 pointers through postRaw so the server-side rejection is what is actually exercised; ReleaseService now has a generated mock and flow_test.go no longer holds a nil service, with a good reason recorded for keeping the interface (gen_openapi_errors only walks interface-typed handler fields); and the reuse path now asserts the 200 body carries the same pointers as the 201. CI is green including full-pr. Nice work, ship it.

PR #1130, opened 2026-09-03
feat(api): read releases

### LINKED ISSUE #531: Service - Releases persistence & endpoints

vitorbari:
Implements `GET /releases` and `GET /releases/{release_id}`.

## `GET /releases`

Screenshot 2026-09-07 at 11 44 50

## `GET /releases/{release_id}`

Screenshot 2026-09-07 at 11 44 55

### DIFF
--- a/internal/api/release_internal_test.go
+++ b/internal/api/release_internal_test.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/stretchr/testify/require"
 	"github.com/zitadel/nextgen/internal/domain"
-	"github.com/zitadel/nextgen/internal/service"
 )
 
 // Pins the release resource-flavored answers through the resolver gate. The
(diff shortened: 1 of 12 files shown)


internal/service/release.go
```diff
@@ -35,8 +36,21 @@ type CreateReleaseOutput struct {
...
+
+type ListReleasesOutput struct {
+	Items         []*domain.Release
+	NextPageToken string
+}
+
 type ReleaseService interface {
 	Create(ctx context.Context, input CreateReleaseInput) (*CreateReleaseOutput, error)
+	Get(ctx context.Context, projectID, id string) (*domain.Release, error)
+	List(ctx context.Context, input ListReleasesInput) (*ListReleasesOutput, error)
```
IAM-marco: Past review note (confidence 100): this PR grows the `ReleaseService` interface from one method to three, and there is still an open review comment from PR #1129 asking for this type to get a proper shape first.

What changed: `Get` and `List` are added here, and both new endpoints call them through the handler:
https://github.com/zitadel/nextgen/blob/1b65bd691d683ea9178eb1a7b97cf972f7ebfead/internal/service/release.go#L50-L54

Why this spot matters: this is the only service in the package that is an interface with a single implementation and no generated mock. Every other service is either a plain struct (like `TeamService` or `BrandingService`) or an interface listed in the package's mock generation run, which this one is not:
https://github.com/zitadel/nextgen/blob/1b65bd691d683ea9178eb1a7b97cf972f7ebfead/internal/service/mocks/generate.go#L14-L16

What follows concretely: unit tests build the handler with nil in this slot today, for example here:
https://github.com/zitadel/nextgen/blob/1b65bd691d683ea9178eb1a7b97cf972f7ebfead/internal/api/flow_test.go#L120-L122
The first unit test that reaches a release path through the handler will crash with a nil panic instead of failing with a readable mock message. This PR doubles the surface where that can happen.

Separately: the PR #1129 comment explicitly predicted that these two read endpoints would land on top of the unresolved type, which is exactly what is happening here.

Fix direction: either make the service a plain struct like its siblings, or add `ReleaseService` to the mock generation list and hand the tests a mock instead of nil. With three methods now using the interface, the mock route likely pays for itself, as the #1129 comment suggested.

Coverage note: nothing fails today only because no unit test yet touches a release method through the handler, so the first one written is the one that hits the panic.

internal/api/release_internal_test.go
```diff
@@ -60,3 +61,61 @@ func TestReleaseAccessRow(t *testing.T) {
...
+	}
+	filter, ok := service.AuthzListFilterFromContext(ctx)
+	if !ok {
+		t.Fatal("Forbidden must attach the EXISTS list filter")
+	}
+	// The kind is what scopes the filter to releases; the wrong one here would
+	// silently filter the list against another resource's scope rows.
+	if filter.ResourceKind != domain.ResourceKindRelease {
+		t.Fatalf("filter kind = %q, want release", filter.ResourceKind)
+	}
```
IAM-marco: Simplification: this block re-tests shared machinery that already has its own test.

What changed: the new `TestReleaseListAccess` checks that an allowed call stamps the one-shot skip, that a narrowed call attaches the row filter, and that the filter carries the release kind.

Why this spot matters: the generic test already pins all of that machinery, for every resource at once:
https://github.com/zitadel/nextgen/blob/1b65bd691d683ea9178eb1a7b97cf972f7ebfead/internal/api/authz_internal_test.go#L370-L428
And the kind check on lines 105 to 109 cannot catch the one mistake it looks aimed at: this test passes `domain.ResourceKindRelease` into the call itself, so asserting the same value comes back out says nothing about what the real handler passes.

What follows concretely: about 27 lines of duplicated coverage that must be kept in sync with the shared test while adding no protection. No sibling resource repeats these machinery checks in its own test file.

Fix direction: drop lines 83 to 109 and keep only the release-flavored error checks below (lines 111 to 120), which are the one part the generic test cannot cover.

Coverage note: nothing is lost; the shared test already fails if the gate machinery breaks.

internal/api/release_internal_test.go
```diff
@@ -60,3 +61,61 @@ func TestReleaseAccessRow(t *testing.T) {
...
+		Scope:         []string{"project.write", "project.read"},
+		PrincipalType: domain.AuthzPrincipalTypeSKProj,
+		PrincipalID:   "proj_a",
+	})
+	preview := WithScopeContext(context.Background(), ScopeContext{
+		ProjectID:     "proj_a",
+		Scope:         []string{"project.read"},
+		PrincipalType: domain.AuthzPrincipalTypeSKProj,
+		PrincipalID:   "proj_a",
+	})
```
IAM-marco: Simplification: this setup is copied word for word from `TestReleaseAccessRow` a few lines up in the same file:
https://github.com/zitadel/nextgen/blob/1b65bd691d683ea9178eb1a7b97cf972f7ebfead/internal/api/release_internal_test.go#L15-L29

Why this spot matters: if the machinery checks in the middle of this test go (see the other comment on this file), only the three release-flavored error checks on lines 111 to 120 survive, and they need exactly the setup that `TestReleaseAccessRow` already builds.

What follows concretely: folding those three checks into `TestReleaseAccessRow` as a subtest removes this whole function, its duplicated setup, and the `service` import that only the machinery checks use.

Fix direction: move lines 111 to 120 into `TestReleaseAccessRow` as a subtest and delete `TestReleaseListAccess`.

Coverage note: none needed; this only moves existing checks, it does not change what is covered.

internal/api/release_internal_test.go
IAM-marco: Simplification: this block re-tests shared machinery that already has its own test.

What changed: the new `TestReleaseListAccess` checks that an allowed call stamps the one-shot skip, that a narrowed call attaches the row filter, and that the filter carries the release kind.

Why this spot matters: the generic test already pins all of that machinery, for every resource at once:
https://github.com/zitadel/nextgen/blob/d01e5655aed57ca0d6168c4e6fc67ece57361fea/internal/api/authz_internal_test.go#L370-L428
And the kind check on lines 105 to 109 cannot catch the one mistake it looks aimed at: this test passes `domain.ResourceKindRelease` into the call itself, so asserting the same value comes back out says nothing about what the real handler passes.

What follows concretely: about 27 lines of duplicated coverage that must be kept in sync with the shared test while adding no protection. No sibling resource repeats these machinery checks in its own test file.

Fix direction: drop lines 83 to 109 and keep only the release-flavored error checks below (lines 111 to 120), which are the one part the generic test cannot cover.

Coverage note: nothing is lost; the shared test already fails if the gate machinery breaks.

internal/api/release_internal_test.go
IAM-marco: Simplification: this setup is copied word for word from `TestReleaseAccessRow` a few lines up in the same file:
https://github.com/zitadel/nextgen/blob/d01e5655aed57ca0d6168c4e6fc67ece57361fea/internal/api/release_internal_test.go#L15-L29

Why this spot matters: if the machinery checks in the middle of this test go (see the other comment on this file), only the three release-flavored error checks on lines 111 to 120 survive, and they need exactly the setup that `TestReleaseAccessRow` already builds.

What follows concretely: folding those three checks into `TestReleaseAccessRow` as a subtest removes this whole function, its duplicated setup, and the `service` import that only the machinery checks use.

Fix direction: move lines 111 to 120 into `TestReleaseAccessRow` as a subtest and delete `TestReleaseListAccess`.

Coverage note: none needed; this only moves existing checks, it does not change what is covered.

internal/api/release_internal_test.go
IAM-marco: Please also use assert/require over `t.Fatal`

IAM-marco (changes requested):

IAM-marco (approved):
Two-axis review ran on this branch: one correctness finding (the unmocked ReleaseService interface from the #1129 thread) and two test-duplication findings. All three were addressed on the branch (mock generated and wired into the unit tests, list gate folded into the access row test), so nothing is left to flag. Approving.

Issue #1011, opened 2026-08-27
Weekly real-instance Spanner benchmark: we cannot tell a Spanner regression from an emulator artifact


IAM-marco:
**TL;DR:** We have no routine signal for Spanner performance, and the signal we do have is wrong in both directions: the emulator invented a production problem that does not exist (#972) and completely hid one that does (#973). A weekly scheduled run against the real instance fixes that. It needs its own workflow file, its own moon task, and #1008 landed first.

> [!NOTE]
> Rewritten after review. The original said this was "mostly a `schedule:` trigger"; it is closer to one PR plus a workflow restructure. What changed and why is in [the review](https://github.com/zitadel/nextgen/issues/1011#issuecomment-5443111533) and in *Decisions taken* below.

## Problem

Nothing tells us when a change makes Spanner slower, and the lane we do run misreports it.

| | emulator | real Cloud Spanner |
| --- | --- | --- |
| `TestListAuthzTeamScopedOnlyPartialView` | 2213s | **17.98s** |
| whole `integration_test` suite | 2245s | **51s** |

Those emulator figures are **pre-#1009**; after it the test is 786s, so the divergence narrows to about 44x but does not close. Re-measure against #1009 rather than against this table.

The emulator said "catastrophic production problem". The real instance said "fine". Resolving that took a day and a manually restored CI lane. It cuts the other way too: per-statement DDL cost is invisible on the emulator, which applies DDL instantly, so #973's 43 minutes of schema setup could only ever be seen on a real instance.

Without a routine real-instance signal we cannot tell an emulator artifact from a real regression, so we either chase phantoms or miss real ones. We have now done both.

## Design

**Its own workflow file, not a `schedule:` on `ci.yml`.** `scripts/ci-mode.mjs` falls back to `origin/main` when there is no PR base; on a scheduled run from `main` that diff is empty, and empty diffs fail open to `mode=full`. A bare schedule would pay `moon ci`, postgres, sqlite, the snapshot build, all four journeys, both console suites and browsers, on top of the spanner lane. A separate file with one job cannot regress PR CI and produces a summary worth reading.

[...]

Issue #734, opened 2026-08-03
Infrastructure setup


IAM-marco:
This ticket is to track progress on infrastructure deployment on GCP.

PR #1132, opened 2026-09-03
ci(deploy): add the OpenTofu configuration for the GCP dev environment

muhlemmer:
## Summary

Part of #734.

Recovers the infrastructure-as-code from the closed [#188](https://github.com/zitadel/nextgen/pull/188) (`feat/tofu-infra`), rebased onto current `main`.

The important thing about that PR: **the resources it declares are already applied** in `zitadel-dev-492704`. The CI plan comment on #188 shows every resource `Refreshing state...` against real IDs, and the final plan was `0 to add, 1 to change, 0 to destroy` (the one change being cosmetic drift — `cpu "1000m" → "1"`, short network names → fully qualified). The branch was abandoned before merge, so `main` has never carried the code that describes our live infrastructure. State lives in `gs://zitadel-ops-tofu-state` under prefix `infra/dev`.

Rather than rebase ten commits with merge commits in them, this takes the tree as a single commit off current `main` — the original PR is closed and its history is not worth preserving.

Changes on top of the abandoned branch:

- **New `secrets` module.** It owns the Secret Manager containers and the access bindings but never the values. A secret with no versions carries no credential material, so nothing sensitive reaches OpenTofu state, while who may read each secret stays reviewable in code. Values come from GitHub environment secrets (see the companion PR).
- **Least-privilege bindings.** The runtime SA gets `secretAccessor`; the `github-deploy` SA gets `secretVersionAdder` — version adder, not admin, so CI can write a new version but cannot read one back, delete the secret, or change who has access.
- **README correction.** The secrets section documented `NEXTGEN_SERVER_ENCRYPTION_KEY`, a variable that no longer exists — it was replaced by the rotatable master-key config, which is a map and therefore cannot be set from an environment variable at all. The rewritten section explains why the master key has to arrive as a mounted file.
- **Documents the unwired migrate job.** `zitadel-migrate-<env>` has no work to do: the released binary has no `migrate` subcommand, so migrations run at server startup.

Nothing here changes any GCP resource on its own — merging runs `tofu plan` on the mgmt tier and `tofu apply` on dev, which should be close to a no-op apart from the new (empty) secret containers and their bindings.

## Validation

- `tofu validate` — `Success! The configuration is valid.`
- `tofu fmt -check -recursive` — clean.
- `tofu init -backend=false` against the committed lock file.

Not run: `tofu plan` against the real project. This environment has no `gcloud` and no GCP credentials, so the plan will first appear on this PR from CI.

## Release notes / changeset

No changeset required — no shipped behavior changed. Deployment wiring only, which is why the title is `ci` rather than `feat`.

## Notes

- **Review the `secrets` module most closely.** It is the only genuinely new code; everything else is #188's, reviewed there.
- **Correction on the Spanner instance.** It is not managed here and never was — `spanner.googleapis.com` is not even in the enabled-APIs set — but it is *not* abandoned leftovers, as I first read it. It belongs to the benchmark work tracked in #1011, which plans a recurring real-instance run against it. Treat it as live infrastructure with an owner: **do not delete it.** Bringing it under OpenTofu is worth doing and belongs with #1011, not here. It does share the name `zitadel-dev` with the Cloud SQL instance, which makes console URLs easy to misread.
- `infra/environments/prod.tfvars` is inherited from #188 and still contains placeholders. Dev is the only wired environment.
- The companion PR adds the deploy workflow and depends on the secret containers this PR creates. Merge this one first.

### DIFF
--- a/.github/workflows/infra.yml
+++ b/.github/workflows/infra.yml
@@ -174,11 +174,20 @@ jobs:
     env:
       WORKING_DIR: infra/mgmt
     steps:
+      # Full history: the detection below diffs two arbitrary SHAs, and on the
+      # default depth-1 checkout neither is present. `git diff` then dies with
+      # "fatal: bad object" and — inside an `if` condition, where `set -e` does
+      # not fire — the failure was read as "nothing changed". This job reported
+      # success without running a plan on every PR it has ever seen, including
+      # the one that added infra/mgmt/ in the first place.
       - uses: actions/checkout@v6.0.2
+        with:
+          fetch-depth: 0
 
       - name: Detect mgmt changes
         id: changed
         run: |
+          set -euo pipefail
           if [ "${{ github.event_name }}" = "pull_request" ]; then
             base="${{ github.event.pull_request.base.sha }}"
             head="${{ github.event.pull_request.head.sha }}"
@@ -189,12 +198,18 @@ jobs:
           # An all-zero "before" sha means the branch was created by this
           # push; compare against origin/main instead.
           if [ -z "$base" ] || [ "$base" = "0000000000000000000000000000000000000000" ]; then
-            git fetch --depth=1 origin main || true
+            git fetch origin main
             base="$(git rev-parse origin/main^)"
           fi
-          if git diff --name-only "$base" "$head" -- infra/mgmt/ | grep -q .; then
+          # Run the diff on its own so a git failure fails the step, instead of
+          # being swallowed by the pipeline that reads it.
+          changes="$(git diff --name-only "$base" "$head" -- infra/mgmt/)"
+          if [ -n "$changes" ]; then
+            echo "mgmt files changed:"
+            echo "$changes"
             echo "changed=true" >> "$GITHUB_OUTPUT"
           else
+            echo "no mgmt changes between $base and $head"
             echo "changed=false" >> "$GITHUB_OUTPUT"
           fi
 
@@ -218,9 +233,17 @@ jobs:
         if: steps.changed.outputs.changed == 'true'
         working-directory: ${{ env.WORKING_DIR }}
         run: |
-          set -o pipefail
+          # -detailed-exitcode returns 2 for "there are changes", which is the
+          # case this job exists to report. Under the runner's `bash -e` that
+          # non-zero status killed the step before PLAN_EXIT was written, so the
+          # comment and summary steps were skipped exactly when they mattered.
+          # Capture the code, then fail only on a real error (1).
+          set +e
           tofu plan -var-file=mgmt.tfvars -no-color -detailed-exitcode 2>&1 | tee /tmp/mgmt-plan.txt
-          echo "PLAN_EXIT=${PIPESTATUS[0]}" >> "$GITHUB_ENV"
+          rc=${PIPESTATUS[0]}
+          set -e
+          echo "PLAN_EXIT=$rc" >> "$GITHUB_ENV"
+          [ "$rc" != 1 ]
 
       - name: Publish plan to step summary
         if: steps.changed.outputs.changed == 'true'
--- a/.github/workflows/infra.yml
+++ b/.github/workflows/infra.yml
@@ -1,0 +1,285 @@
+name: Infrastructure / Dev
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'infra/**'
+      - '!infra/README.md'
+  push:
+    branches: [main]
+    paths:
+      - 'infra/**'
+      - '!infra/README.md'
+
+permissions:
+  contents: read
+  id-token: write
+  pull-requests: write
+
+env:
+  TOFU_VERSION: '1.9.0'
+
+jobs:
+  # ─── Env tier (infra/) ──────────────────────────────────────────────────────
+  plan:
+    if: github.event_name == 'pull_request'
+    name: Plan / dev
+    runs-on: ubuntu-latest
+    env:
+      WORKING_DIR: infra
+      ENVIRONMENT: dev
+    steps:
+      - uses: actions/checkout@v6.0.2
+
+      - uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: ${{ vars.GCP_INFRA_WIF_PROVIDER }}
+          service_account: ${{ vars.GCP_INFRA_SERVICE_ACCOUNT }}
+
+      - uses: opentofu/setup-opentofu@v1
+        with:
+          tofu_version: ${{ env.TOFU_VERSION }}
+
+      - name: Init
+        working-directory: ${{ env.WORKING_DIR }}
+        run: |
+          tofu init -backend-config=environments/${{ env.ENVIRONMENT }}.tfbackend
+
+      - name: Plan
+        id: plan
+        working-directory: ${{ env.WORKING_DIR }}
+        run: |
+          tofu plan \
+            -var-file="environments/${{ env.ENVIRONMENT }}.tfvars" \
+            -no-color \
+            -out=tfplan \
+            2>&1 | tee /tmp/plan.txt
+
+      - name: Comment plan on PR
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const fs = require('fs');
+            const plan = fs.readFileSync('/tmp/plan.txt', 'utf8');
+            const maxLen = 60000;
+            const truncated = plan.length > maxLen
+              ? plan.substring(0, maxLen) + '\n\n... (truncated)'
+              : plan;
+
+            const body = `### OpenTofu Plan — dev
+
+            \`\`\`
+            ${truncated}
+            \`\`\`
+
+            *Pushed by @${context.actor}*`;
+
+            const { data: comments } = await github.rest.issues.listComments({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+            });
+            const existing = comments.find(c =>
+              c.user.type === 'Bot' && c.body.includes('### OpenTofu Plan — dev')
+            );
+
+            if (existing) {
+              await github.rest.issues.updateComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                comment_id: existing.id,
+                body,
+              });
+            } else {
+              await github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.issue.number,
+                body,
+              });
+            }
+
+  apply:
+    if: github.event_name == 'push'
+    name: Apply / dev
+    runs-on: ubuntu-latest
+    environment: dev
+    concurrency:
+      group: infra-dev
+      cancel-in-progress: false
+    env:
+      WORKING_DIR: infra
+      ENVIRONMENT: dev
+    steps:
+      - uses: actions/checkout@v6.0.2
+
+      - uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: ${{ vars.GCP_INFRA_WIF_PROVIDER }}
+          service_account: ${{ vars.GCP_INFRA_SERVICE_ACCOUNT }}
+
+      - uses: opentofu/setup-opentofu@v1
+        with:
+          tofu_version: ${{ env.TOFU_VERSION }}
+
+      - name: Init
+        working-directory: ${{ env.WORKING_DIR }}
+        run: |
+          tofu init -backend-config=environments/${{ env.ENVIRONMENT }}.tfbackend
+
+      - name: Plan
+        working-directory: ${{ env.WORKING_DIR }}
+        run: |
+          tofu plan \
+            -var-file="environments/${{ env.ENVIRONMENT }}.tfvars" \
+            -no-color \
+            -out=tfplan \
+            2>&1 | tee /tmp/plan.txt
+
+      - name: Publish plan to step summary
+        run: |
+          {
+            echo "### OpenTofu Plan — ${{ env.ENVIRONMENT }}"
+            echo
+            echo '```'
+            head -c 60000 /tmp/plan.txt
+            echo
+            echo '```'
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Apply
+        working-directory: ${{ env.WORKING_DIR }}
+        run: tofu apply tfplan
+
+  # ─── Mgmt tier (infra/mgmt/) — plan only, drift detection ──────────────────
+  # Mgmt applies are manual on purpose (cross-project IAM and WIF changes have
+  # a big blast radius). This job runs `tofu plan` on every PR touching
+  # mgmt files, posts the diff to the PR, and fails on push-to-main if the
+  # plan is non-empty — that indicates someone merged a mgmt change without
+  # applying it locally first.
+  mgmt-plan:
+    if: |
+      (github.event_name == 'pull_request' || github.event_name == 'push')
+    name: Mgmt plan
+    runs-on: ubuntu-latest
+    env:
+      WORKING_DIR: infra/mgmt
+    steps:
+      - uses: actions/checkout@v6.0.2
+
+      - name: Detect mgmt changes
+        id: changed
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            base="${{ github.event.pull_request.base.sha }}"
+            head="${{ github.event.pull_request.head.sha }}"
+          else
+            base="${{ github.event.before }}"
+            head="${{ github.sha }}"
+          fi
+          # An all-zero "before" sha means the branch was created by this
+          # push; compare against origin/main instead.
+          if [ -z "$base" ] || [ "$base" = "0000000000000000000000000000000000000000" ]; then
+            git fetch --depth=1 origin main || true
+            base="$(git rev-parse origin/main^)"
+          fi
+          if git diff --name-only "$base" "$head" -- infra/mgmt/ | grep -q .; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - uses: google-github-actions/auth@v2
+        if: steps.changed.outputs.changed == 'true'
+        with:
+          workload_identity_provider: ${{ vars.GCP_INFRA_WIF_PROVIDER }}
+          service_account: ${{ vars.GCP_INFRA_SERVICE_ACCOUNT }}
+
+      - uses: opentofu/setup-opentofu@v1
+        if: steps.changed.outputs.changed == 'true'
+        with:
+          tofu_version: ${{ env.TOFU_VERSION }}
+
+      - name: Init
+        if: steps.changed.outputs.changed == 'true'
+        working-directory: ${{ env.WORKING_DIR }}
+        run: tofu init -backend-config=mgmt.tfbackend
+
+      - name: Plan
+        if: steps.changed.outputs.changed == 'true'
+        working-directory: ${{ env.WORKING_DIR }}
+        run: |
+          set -o pipefail
+          tofu plan -var-file=mgmt.tfvars -no-color -detailed-exitcode 2>&1 | tee /tmp/mgmt-plan.txt
+          echo "PLAN_EXIT=${PIPESTATUS[0]}" >> "$GITHUB_ENV"
+
+      - name: Publish plan to step summary
+        if: steps.changed.outputs.changed == 'true'
+        run: |
+          {
+            echo "### OpenTofu Plan — mgmt"
+            echo
+            echo '```'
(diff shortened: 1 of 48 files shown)


infra/mgmt/mgmt.tfvars
```diff
@@ -0,0 +1,8 @@
+mgmt_project_id = "zitadel-ops"
+region          = "us-central1"
+github_repo     = "zitadel/nextgen"
+
+env_project_ids = {
+  dev = "zitadel-dev-492704"
+  # prod = "REPLACE_PROD_PROJECT"
```
muhlemmer, line 7: TBD: Shouldn't prod mgmt be created as this point as well?

infra/modules/cloud-run/main.tf
```diff
@@ -0,0 +1,117 @@
...
+    scaling {
+      min_instance_count = var.min_instances
+      max_instance_count = var.max_instances
+    }
+
+    containers {
+      # Placeholder — the deploy workflow replaces image, command, args,
+      # env, secret references, and probes via gcloud/manual release
+      # operations.
+      image = "us-docker.pkg.dev/cloudrun/container/hello"
```
muhlemmer, line 63: Why is this here?

infra/mgmt/mgmt.tfvars
muhlemmer: Not in this PR, because it cannot work yet — but worth deciding now.

`env_project_ids` drives five things in the mgmt tier, and every one of them needs a **real, existing** GCP project:

- `infra_apply_editor` — `roles/editor` for the OpenTofu CI identity
- `infra_apply_iam` — `roles/iam.securityAdmin` for the same
- `deploy_run_developer` — `roles/run.developer` for `github-deploy`
- `data.google_project.env` — a lookup that resolves **at plan time**
- `env_run_agent_reader` — AR reader for that project's Cloud Run service agent, whose email is built from the project *number* returned by that lookup

So uncommenting the line with `REPLACE_PROD_PROJECT` in it would fail the plan outright. Worse, it would fail it on `main`: the `mgmt-plan` job is wired to fail a push when the mgmt plan is non-empty, so a placeholder here breaks the build rather than sitting harmlessly.

And there is no prod project to point it at. The whole prod path is still the untouched template from #188 — `infra/environments/prod.tfvars` reads `project_id = "REPLACE_GCP_PROJECT"` and `base_domain = "zitadel.example.com"`.

The good news is that this half is cheap and safe whenever we want it. Adding prod to `env_project_ids` only grants CI identities access to a project; it creates no prod workload, no database, no load balancer. So it can land as soon as a prod project exists, well ahead of any prod infrastructure.

**If a prod project already exists, give me the ID and I will wire both `mgmt.tfvars` and `prod.tfvars` in a follow-up.** If not, I would leave this commented and treat standing up prod as its own piece of work — this PR is about getting dev under version control and deployable.

infra/modules/cloud-run/main.tf
muhlemmer: Because `containers.image` is a required field: OpenTofu cannot create a `google_cloud_run_v2_service` without one. But the deploy workflow, not OpenTofu, owns which release runs — so `image` is listed under `lifecycle.ignore_changes` a few lines below, along with `command`, `args`, `env` and the probes.

The net effect is that this value is a **create-time bootstrap only**. It is read once, when the service does not yet exist, and never again.

This PR's own plan CI demonstrates it. The live service is running the real nextgen image; the config in front of you says `hello`; and the plan proposes **no image change at all** — the only diff on that resource is the pre-existing cosmetic drift:

```
~ resource "google_cloud_run_v2_service" "zitadel" {
      name = "zitadel-dev"
    ~ template {
        ~ containers {
            - name = "nextgen-1" -> null
            ~ resources {
                ~ limits = { ~ "cpu" = "1000m" -> "1" }
```

Fair question to ask, though, and the old comment did not really answer it — it described what the deploy workflow does without saying why a placeholder had to exist, and it was written back when there *was* no deploy workflow. Rewritten in 8d9ef14c to say the actual reason, and I added a note on the migrate job that its placeholder is still what genuinely runs, since there is no `migrate` subcommand to point it at (#1138).

infra/modules/postgres/main.tf
```diff
@@ -0,0 +1,53 @@
...
+  name                = var.instance_name
+  database_version    = var.database_version
+  region              = var.region
+  deletion_protection = var.deletion_protection
+
+  settings {
+    tier              = var.tier
+    edition           = var.edition
+    availability_type = var.availability_type
+    disk_size         = var.disk_size_gb
```
fforootd, line 25: **Blocking.** `disk_size` is pinned while `disk_autoresize = true`. The locked provider 6.50.0 treats config smaller than state as a shrink and plans destroy and recreate (`ForceNewIfChange("settings.0.disk_size", IsDiskShrinkage)`); its docs warn about exactly this. Once dev autoresizes past 20 GB, the next merge touching `infra/` replaces `zitadel-dev`, and dev sets `deletion_protection = false`, so the apply job runs it.

```hcl
lifecycle {
  ignore_changes = [settings[0].disk_size]
}
```

---

infra/modules/cloud-run/main.tf
```diff
@@ -0,0 +1,125 @@
...
+        limits = {
+          cpu    = var.cpu
+          memory = var.memory
+        }
+      }
+    }
+  }
+
+  lifecycle {
+    ignore_changes = [
```
fforootd, line 121: **Blocking.** #1133 mounts the master key via `--set-secrets=<path>=secret:latest`, which is a secret volume plus a volume mount. Neither `template[0].volumes` nor `template[0].containers[0].volume_mounts` is ignored here, so the first infra apply after a deploy removes them and rolls a revision without the key file. The server then generates a throwaway key and existing KEKs stop decrypting. Add both, plus `template[0].containers[0].name`: the plan on this PR already rolls a revision just to null `nextgen-1`.

---

.github/workflows/infra.yml
```diff
@@ -0,0 +1,293 @@
...
+            base="${{ github.event.before }}"
+            head="${{ github.sha }}"
+          fi
+          # An all-zero "before" sha means the branch was created by this
+          # push; compare against origin/main instead.
+          if [ -z "$base" ] || [ "$base" = "0000000000000000000000000000000000000000" ]; then
+            git fetch --depth=1 origin main || true
+            base="$(git rev-parse origin/main^)"
+          fi
+          if git diff --name-only "$base" "$head" -- infra/mgmt/ | grep -q .; then
```
fforootd: **Blocking.** This can't work on the default checkout. Depth 1 has neither SHA, `git diff` fails with `fatal: bad object`, and `| grep -q .` inside the `if` turns that into `changed=false`. That is what the Mgmt plan job did on this PR even though it adds all of `infra/mgmt/`, and on push the same thing masks the drift gate. Use `fetch-depth: 0` or fetch both SHAs, and don't hide git's exit code behind grep.

---

.github/workflows/infra.yml
```diff
@@ -0,0 +1,293 @@
...
+        if: steps.changed.outputs.changed == 'true'
+        working-directory: ${{ env.WORKING_DIR }}
+        run: tofu init -backend-config=mgmt.tfbackend
+
+      - name: Plan
+        if: steps.changed.outputs.changed == 'true'
+        working-directory: ${{ env.WORKING_DIR }}
+        run: |
+          set -o pipefail
+          tofu plan -var-file=mgmt.tfvars -no-color -detailed-exitcode 2>&1 | tee /tmp/mgmt-plan.txt
```
fforootd, line 242: **Blocking.** Under the runner's `bash -e`, a detailed exit code of 2 aborts the step before `PLAN_EXIT` is written, and the comment and summary steps are skipped. A mgmt PR with changes is exactly the case that should comment.

```bash
set +e
tofu plan -var-file=mgmt.tfvars -no-color -detailed-exitcode 2>&1 | tee /tmp/mgmt-plan.txt
rc=${PIPESTATUS[0]}
set -e
echo "PLAN_EXIT=$rc" >> "$GITHUB_ENV"
[ "$rc" != 1 ]
```

---

infra/mgmt/main.tf
```diff
@@ -0,0 +1,169 @@
...
+resource "google_service_account" "infra_apply" {
+  account_id   = "infra-apply"
+  display_name = "OpenTofu CI"
+  description  = "GitHub Actions identity for infrastructure plan/apply"
+
+  depends_on = [google_project_service.apis]
+}
+
+# Mgmt project: read/write state bucket
+resource "google_project_iam_member" "infra_apply_storage" {
```
fforootd, line 54: `infra-apply` runs the mgmt plan but only holds `storage.admin` and `artifactregistry.reader` on `zitadel-ops`. Refreshing the SAs, WIF pool and provider, project IAM members and enabled services needs at least `roles/viewer` there, unless that was granted out of band. It hasn't surfaced only because of the detection bug in the workflow.

---

infra/mgmt/main.tf
```diff
@@ -0,0 +1,169 @@
...
+  oidc {
+    issuer_uri = "https://token.actions.githubusercontent.com"
+  }
+}
+
+# Bind infra-apply SA to the WIF pool
+resource "google_service_account_iam_member" "wif_infra_apply" {
+  service_account_id = google_service_account.infra_apply.name
+  role               = "roles/iam.workloadIdentityUser"
+  member             = "principalSet://iam.googleapis.com/${google_iam_workload_identity_pool.github.name}/attribute.repository/${var.github_repo}"
```
fforootd, line 173: Binding on `attribute.repository` alone means every workflow run in the repo can impersonate `infra-apply`, including the PR `plan` job, which has `id-token: write`. That is editor plus securityAdmin on dev and storage.admin over all state with no environment gate. Suggest mapping `attribute.ref` or `attribute.environment`, binding apply to `refs/heads/main` or the `dev` environment, and using a read-only identity for PR plans. Follow-up is fine, but worth an issue.

---

infra/modules/iam/main.tf
```diff
@@ -0,0 +1,81 @@
...
+
+# Metrics
+resource "google_project_iam_member" "run_metrics" {
+  project = var.project_id
+  role    = "roles/monitoring.metricWriter"
+  member  = "serviceAccount:${google_service_account.run.email}"
+}
+
+# Certificate Manager — runtime domain provisioning (infra.rs)
+resource "google_project_iam_member" "run_cert_manager" {
```
fforootd: `certificatemanager.editor`, `compute.loadBalancerAdmin` and `dns.admin` are for an `infra.rs` runtime that doesn't exist in this repo: no Rust, no such clients in `go.mod`. On the internet-facing SA they only add blast radius. Drop them and the `infra.rs` and `cloud.gcp` mentions in the README and outputs, or note what they're for.

---

infra/mgmt/main.tf
```diff
@@ -0,0 +1,169 @@
...
+  ])
+
+  project            = var.mgmt_project_id
+  service            = each.value
+  disable_on_destroy = false
+}
+
+# ─── Artifact Registry ──────────────────────────────────────────────────────
+
+resource "google_artifact_registry_repository" "zitadel" {
```
fforootd, line 25: Once #1133 deploys from `ghcr.io` directly, this repository, `deploy_ar_writer`, `env_run_agent_reader` with its `data.google_project` lookup, and `infra_apply_ar_reader` have no consumer, and the cross-project image validation the last one exists for no longer applies. Leave them in this PR so the mgmt plan stays empty; retire them in a follow-up mgmt PR after #1133 lands. The KEEP-only cleanup policy becomes moot at the same time.

---

fforootd (changes requested):
The secrets module is right: containers without values, accessor for the runtime, version-adder for deploy, and the README matches the runtime code. The blockers are in the surrounding automation. Two of them let the auto-apply job destroy dev data or the dev master key, and the mgmt drift job has never actually run. Each fix is a few lines; details inline.

Context for the mgmt tier: #1133 will deploy straight from GHCR, which Cloud Run has supported for public images since the July 14 GA. That leaves the Artifact Registry repo and its three bindings without a consumer. Not for this PR, since it imports live state and the mgmt plan should stay empty; see the note on the repo resource.

Non-blocking, fine as a follow-up issue: no `ssl_policy` on the HTTPS proxy so TLS 1.0 is accepted, the secrets module doesn't pass `deletion_protection`, `load_balancer` and `dns` miss `depends_on = [module.project]`, the cacheable path rule targets routes the server doesn't serve, `google-beta` is required but unused, and the README's step 1 path and variable list don't match the workflows: `GCP_DEV_PROJECT_ID` is missing, while `GCP_MGMT_PROJECT_ID` and `AR_REGION` only served the mirror step #1133 is dropping.

---

infra/modules/postgres/main.tf
muhlemmer: Fixed in 744aaa0c — `ignore_changes = [settings[0].disk_size]`.

Probed the project to see how close this was: `dataDiskSizeGb` is still **20** with `storageAutoResize=True`, `storageAutoResizeLimit=0` (unlimited) and `deletionProtectionEnabled=False`. So it was armed and had not fired — the first autoresize would have made the next `infra/` merge plan the replacement, with nothing to stop it.

infra/modules/cloud-run/main.tf
muhlemmer: Fixed in 744aaa0c, taking the cleaner of your two options: **the module declares the volume** rather than widening `ignore_changes`. A mount that is OpenTofu's own state cannot be stripped by OpenTofu.

That forced the ownership split you implied: since `--set-secrets` is declarative over volumes *and* env, a deploy passing it for the DSN alone would delete the module's mount. So the module owns the DSN env reference too, `env` came out of `ignore_changes`, and #1133 shrinks to `--image` only. Added `containers[0].name` as you asked.

One thing this exposed: OpenTofu now references secrets that have no versions, and a revision cannot start against an empty secret — so the very apply that creates the containers would have failed. The wiring is staged behind `runtime_secrets_ready` (79bad17f), default false, with the sequence written up under "First deploy" in the README. The flip is a one-line change whose plan shows the mount being added.

.github/workflows/infra.yml
muhlemmer: Fixed in 744aaa0c: `fetch-depth: 0`, and the diff now runs on its own line so git's exit code fails the step instead of being swallowed by `grep`.

Confirmed your read of it from the job record — on this PR, which adds all of `infra/mgmt/`, `Detect mgmt changes` concluded no-change and all seven following steps were skipped, while the job reported success. So the drift gate had never executed at all.

.github/workflows/infra.yml
muhlemmer: Fixed in 744aaa0c with your snippet. Exercised it against a stubbed `tofu`:

| tofu exit | step exit | `PLAN_EXIT` recorded |
| --- | --- | --- |
| 0 | 0 | 0 |
| 1 | 1 | 1 |
| 2 | 0 | 2 |

So a plan with changes now records and passes through to the comment and summary steps, and only a real error fails.

infra/modules/iam/main.tf
muhlemmer: Dropped in 744aaa0c, along with the `infra.rs` and `cloud.gcp` mentions in the README and outputs. Confirmed your reasoning both ways: no Rust anywhere in the repo, and no Certificate Manager, Compute or DNS client in `go.mod` — and the project's IAM policy shows all three live on `zitadel-run` today, so this was real exposure on the identity terminating internet traffic, not just stale config. Left a comment in the module saying what they were for, so a narrowed version can come back with the component that needs it.

infra/mgmt/main.tf
muhlemmer: Good catch, and I could not verify it — the probe I ran was scoped to the dev project only, so `zitadel-ops` IAM is still unchecked. Filed as part of the WIF follow-up context; the read-only command is `gcloud projects get-iam-policy zitadel-ops --flatten='bindings[].members' --filter='bindings.members:infra-apply'`.

Worth noting it stays masked until the `fetch-depth` fix lands, since the job has never actually reached `tofu init`. Once it does, a missing `roles/viewer` will surface as a plan failure rather than silence.

infra/mgmt/main.tf
muhlemmer: Filed as #1162 with the live evidence: the dev IAM policy confirms `infra-apply` holds `roles/editor` **and** `roles/iam.securityAdmin` there, so the blast radius is as you describe. Wrote it up with both halves — gating the binding on `attribute.ref`/`attribute.environment`, and giving the PR `plan` job a read-only identity, since a plan needs neither role.

infra/mgmt/main.tf
muhlemmer: Agreed, and kept them here so the mgmt plan stays empty. Filed as #1164, sequenced after #1133 has actually deployed from GHCR so the rollback path still exists.

One caveat I put in the issue: Google recommends an Artifact Registry *remote repository* for production availability rather than direct public-registry pulls, so retiring the standard repo for dev does not settle the prod question — a remote repository is a different resource.

infra/mgmt/main.tf
muhlemmer: Confirmed and fixed — you were exactly right, and it was masked as you said.

Once the `fetch-depth` fix made the job actually run, it failed on precisely this:

```
Error 403: Permission denied to list services for consumer container [projects/831790715403]
  permission: serviceusage.services.list
```

The policy before the fix held only what you predicted:

| member | role |
| --- | --- |
| `infra-apply@zitadel-ops` | `roles/artifactregistry.reader` |
| `infra-apply@zitadel-ops` | `roles/storage.admin` |

`roles/viewer` has since been granted, and the rerun is the first complete read of the mgmt tier: **20 resources refreshed** — the AR repo, both service accounts, the WIF pool and provider, every IAM binding and all five enabled APIs — ending in `No changes. Your infrastructure matches the configuration.`

Resolving: verified end to end.

infra/mgmt/main.tf
muhlemmer: Filed as #1162 with live evidence, per your "follow-up is fine, but worth an issue".

The dev IAM policy confirms the blast radius you described — `infra-apply` holds `roles/editor` **and** `roles/iam.securityAdmin` there — and the mgmt policy confirms the binding is repository-wide with no ref, environment or actor gate. The issue carries both halves of your suggestion: gating on `attribute.ref`/`attribute.environment`, and a read-only identity for PR plans.

Resolving here since it is tracked; the security question itself stays open on #1162.

infra/mgmt/main.tf
muhlemmer: Filed as #1164, and kept in this PR exactly as you asked so the mgmt plan stays empty — which it now does (`No changes`, 20 resources).

Sequenced after #1133 has actually deployed from GHCR, so the rollback path survives. One caveat I added to the issue: Google recommends an Artifact Registry *remote repository* for production availability rather than direct public-registry pulls, so retiring the standard repo for dev does not settle the prod question — a remote repository is a different resource.

Resolving here; the retirement itself lives on #1164.

muhlemmer:
Waiting for https://github.com/zitadel/nextgen/pull/1151 to land, so we can cleanup the masterkey regeneration warnings. Other comments are resolved.

muhlemmer:
### Note for review: this apply takes dev down until the first deploy

The plan removes two env vars from the running service and adds nothing back:

```
- env { NEXTGEN_DATABASE_POSTGRES     → zitadel-postgres-secret-dev }
- env { NEXTGEN_SERVER_ENCRYPTION_KEY → zitadel-encryption-key-secret-dev }
```

That is deliberate, and the downtime has been accepted. Recording why here so it does not read as an oversight in the plan output.

**Why it happens.** Moving the master-key mount into OpenTofu (the fix for the `--set-secrets` blocker) means OpenTofu has to own `env` too, since `gcloud run services update --set-secrets` is declarative over volumes *and* env — a deploy passing it for the DSN alone would delete the mount. So `env` came out of `ignore_changes`. While `runtime_secrets_ready = false` nothing is declared, so the apply strips what is there.

**Why it is not worked around.** `ignore_changes` has to be a static list, so `env` cannot be ignored only while the flag is false. And the secrets cannot be seeded before the containers exist, which is the same apply. The alternative — landing `env` ownership together with the flag flip — was considered and rejected as more moving parts for a dev environment mid-cutover.

**What is actually lost.** The serving revision is a June build (`0.0.1-snapshot-8c78c632`) reading `NEXTGEN_SERVER_ENCRYPTION_KEY`, a configuration key the server no longer has. Data loss across this cutover has been accepted, so nothing is being preserved that anyone wants.

**The window closes at step 4:**

1. Merge this PR — secret containers and bindings created, service rolls without a database
2. Set the `dev` environment's variables and secrets
3. Merge #1133 and run *Deploy / Cloud Run* with `sync_secrets: true`
4. Flip `runtime_secrets_ready = true` — the mount and the DSN land, and that plan is worth reading before it applies

Steps 1 and 4 are the two applies; everything between them is dev being down on purpose.

PR #1133, opened 2026-09-03
ci(deploy): deploy released images to Cloud Run

muhlemmer:
## Summary

Part of #734.

Adds the workflow the infra tier always assumed existed. `infra/README.md` describes a "GitHub Actions deploy workflow" owning the Cloud Run image, env and secrets — those fields sit under `lifecycle.ignore_changes` in the `cloud-run` module precisely so OpenTofu does not fight it — but the workflow was never written. That is why nothing has been deployed since #188 was closed; the abandoned PR said so itself: *"For now I only moved over the infra pipeline as we have no releases yet in this repo."* We have releases now.

Three things it has to do that are not obvious from the outside:

- **Mirror the image.** Releases publish to `ghcr.io`, and Cloud Run pulls only from Artifact Registry or GCR. `docker buildx imagetools create` copies the multi-arch manifest as-is; a plain pull/push would flatten it to the runner's architecture.
- **Mount the master key as a file, not an env var.** `server.master_keys` is a map keyed by key ID, and Viper cannot populate map keys from the environment — `NEXTGEN_SERVER_MASTER_KEYS_*` is silently ignored. The server *does* discover keys in its master-key directory by filename. Without a provisioned key it generates a throwaway one per instance and logs that it is "for local/dev only"; on Cloud Run, where that directory is ephemeral, project KEKs would end up wrapped by a key the next instance does not have.
- **Leave the migrate job alone.** The released binary has no `migrate` subcommand — its only command is `server`, which runs `pool.Migrate(ctx)` at startup — so running `zitadel-migrate-<env>` would start a server that never exits and fail on the job's 600s timeout.

Secrets never touch the repository. Values live in the environment's GitHub secrets and are pushed to Secret Manager as new versions, gated behind a `sync_secrets` input so ordinary deploys do not churn versions. They are piped over stdin, never passed as argv.

Manual dispatch on purpose: a `release: published` trigger would not fire, because the release is created by `release-publish.yml` using `GITHUB_TOKEN`, and events raised by that token do not start new workflow runs.

## Validation

- Parsed the workflow YAML and asserted the trigger, inputs, env and step list.
- Exercised the secret-sync shell against a stubbed `gcloud`: the value arrives on stdin (65 bytes of PEM, newlines intact) and never as an argv entry; a missing environment secret fails loudly with an `::error::` annotation.
- Verified the master-key mechanism against the real published image, `ghcr.io/zitadel/nextgen:1.0.0-alpha.20`: with a key placed in the master-key directory the server reaches `server listening for requests` and logs `migrate keys to latest master key`, with no dev-only warning. The stock image with no key logs `created server master key file … (generated for local/dev only)`.

Not run: the workflow itself. It needs the GCP identity and the secret containers from the companion PR, and this environment has no `gcloud` and no credentials.

## Release notes / changeset

No changeset required — no shipped behavior changed. Deployment wiring only, which is why the title is `ci` rather than `feat`.

## Notes

- **Depends on the companion infra PR** for the Secret Manager containers and the `secretVersionAdder` binding. Merge that one first; this workflow will fail at the sync step without it.
- **First run needs `sync_secrets: true`**, and the `dev` environment needs two secrets set: `MASTER_KEY_PEM` (an RSA private key in PEM, the same PKCS#1 shape the server generates) and `DATABASE_POSTGRES_DSN` (built from the `postgres_private_ip` and `postgres_database_name` outputs).
- **Ingress is internal-load-balancer only**, so a deployed revision is not reachable at its `run.app` URL — test through the load balancer at `dev.zitadel.io`.
- The `GCP_DEV_PROJECT_ID` repository variable is new; the others (`GCP_WIF_PROVIDER`, `GCP_WIF_SERVICE_ACCOUNT`, `GCP_MGMT_PROJECT_ID`, `AR_REGION`) already exist. The first step fails with a clear message if any are missing.
- Follow-ups worth their own issues: auto-deploying dev on release via `workflow_run`; a real `migrate` subcommand so the migration job can be wired; and Spanner support, which needs the `spanner_integration` build tag question settled before it can run on a released image at all.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,168 @@
+name: Deploy / Cloud Run
+
+# Deploys a published release image to the Cloud Run service that infra/
+# provisions. The infrastructure itself is owned by .github/workflows/infra.yml;
+# this workflow owns only the release-specific fields (image, env, secrets),
+# which the OpenTofu cloud-run module leaves under `lifecycle.ignore_changes`.
+#
+# Manual dispatch on purpose. A `release: published` trigger would not fire:
+# the release is created by release-publish.yml using GITHUB_TOKEN, and events
+# raised by that token do not start new workflow runs. Auto-deploying dev on
+# every release is a reasonable follow-up via `workflow_run`, but a shared
+# environment is worth a deliberate button for now.
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Released version to deploy, e.g. 1.0.0-alpha.20 (no leading v)."
+        required: true
+        type: string
+      environment:
+        description: "Target environment."
+        required: true
+        default: dev
+        type: choice
+        options:
+          - dev
+      sync_secrets:
+        description: "Push the environment's GitHub secrets to Secret Manager as new versions. Required on the first deploy and after a rotation."
+        required: false
+        default: false
+        type: boolean
+
+permissions:
+  contents: read
+  id-token: write
+
+concurrency:
+  group: deploy-${{ inputs.environment }}
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    name: Deploy ${{ inputs.version }} to ${{ inputs.environment }}
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.environment }}
+    env:
+      VERSION: ${{ inputs.version }}
+      ENVIRONMENT: ${{ inputs.environment }}
+      PROJECT_ID: ${{ vars.GCP_DEV_PROJECT_ID }}
+      REGION: ${{ vars.AR_REGION }}
+      MGMT_PROJECT_ID: ${{ vars.GCP_MGMT_PROJECT_ID }}
+      SOURCE_IMAGE: ghcr.io/zitadel/nextgen
+
+    steps:
+      - name: Check required repository variables
+        run: |
+          missing=()
+          for v in PROJECT_ID REGION MGMT_PROJECT_ID; do
+            [ -n "${!v}" ] || missing+=("$v")
+          done
+          if [ ${#missing[@]} -gt 0 ]; then
+            echo "::error::Missing repository variables: ${missing[*]}. See infra/README.md — they come from \`tofu output\` in infra/mgmt."
+            exit 1
+          fi
+
+      - uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: ${{ vars.GCP_WIF_PROVIDER }}
+          service_account: ${{ vars.GCP_WIF_SERVICE_ACCOUNT }}
+
+      - uses: google-github-actions/setup-gcloud@v2
+
+      # ── Image ────────────────────────────────────────────────────────────
+      # Cloud Run pulls only from Artifact Registry or GCR, so the released
+      # GHCR image is mirrored into the mgmt project's registry.
+      # `imagetools create` copies the multi-arch manifest as-is; a plain
+      # pull/push would flatten it to the runner's architecture.
+      - name: Mirror the release image into Artifact Registry
+        run: |
+          set -euo pipefail
+          AR_IMAGE="${REGION}-docker.pkg.dev/${MGMT_PROJECT_ID}/zitadel/nextgen"
+          echo "AR_IMAGE=$AR_IMAGE" >> "$GITHUB_ENV"
+
+          if ! docker manifest inspect "${SOURCE_IMAGE}:${VERSION}" >/dev/null 2>&1; then
+            echo "::error::${SOURCE_IMAGE}:${VERSION} does not exist. Check the version input against the published releases."
+            exit 1
+          fi
+
+          gcloud auth configure-docker "${REGION}-docker.pkg.dev" --quiet
+          docker buildx imagetools create \
+            --tag "${AR_IMAGE}:${VERSION}" \
+            "${SOURCE_IMAGE}:${VERSION}"
+
+      # ── Secrets ──────────────────────────────────────────────────────────
+      # GitHub environment secrets are the source of truth; Secret Manager is
+      # the delivery mechanism Cloud Run requires. The containers and their
+      # IAM bindings are created by OpenTofu (infra/modules/secrets), so this
+      # only ever adds a version — the deploy identity holds
+      # secretVersionAdder and cannot read a value back.
+      - name: Sync secrets to Secret Manager
+        if: ${{ inputs.sync_secrets }}
+        env:
+          MASTER_KEY_PEM: ${{ secrets.MASTER_KEY_PEM }}
+          DATABASE_POSTGRES_DSN: ${{ secrets.DATABASE_POSTGRES_DSN }}
+        run: |
+          set -euo pipefail
+          for pair in \
+            "MASTER_KEY_PEM:zitadel-master-key-${ENVIRONMENT}" \
+            "DATABASE_POSTGRES_DSN:zitadel-database-postgres-${ENVIRONMENT}"; do
+            var="${pair%%:*}"
+            secret="${pair#*:}"
+            if [ -z "${!var:-}" ]; then
+              echo "::error::Environment secret $var is not set on the '${ENVIRONMENT}' environment."
+              exit 1
+            fi
+            # printf into gcloud over a pipe: the value never becomes an argv
+            # entry and never lands in a temporary file.
+            printf '%s' "${!var}" |
+              gcloud secrets versions add "$secret" --project="$PROJECT_ID" --data-file=- >/dev/null
+            echo "added a new version of $secret"
+          done
+
+      # ── Deploy ───────────────────────────────────────────────────────────
+      # The master key is mounted as a file rather than passed as an env var:
+      # `server.master_keys` is a map keyed by key ID and Viper cannot fill map
+      # keys from the environment, but the server discovers keys in its
+      # master-key directory by filename. Without it the server silently
+      # generates a throwaway key on every instance, which would leave project
+      # KEKs wrapped by a key the next instance does not have.
+      #
+      # Migrations run at server startup — the binary has no `migrate`
+      # subcommand, so the zitadel-migrate-<env> job stays unwired.
+      - name: Deploy to Cloud Run
+        run: |
+          set -euo pipefail
+          RUN_SA="zitadel-run@${PROJECT_ID}.iam.gserviceaccount.com"
+          DATA_DIR=/var/lib/zitadel/nextgen-data
+
+          gcloud run services update "zitadel-${ENVIRONMENT}" \
+            --project="$PROJECT_ID" \
+            --region="$REGION" \
+            --image="${AR_IMAGE}:${VERSION}" \
+            --service-account="$RUN_SA" \
+            --port=8080 \
+            --set-env-vars="NEXTGEN_SERVER_DATA_DIR=${DATA_DIR}" \
+            --set-secrets="NEXTGEN_DATABASE_POSTGRES=zitadel-database-postgres-${ENVIRONMENT}:latest,${DATA_DIR}/master-keys/master-key.pem=zitadel-master-key-${ENVIRONMENT}:latest"
+
+      # `services update` already fails when the revision does not become
+      # ready, so this reports which revision won rather than re-checking it.
+      - name: Report the serving revision
+        run: |
+          set -euo pipefail
+          revision="$(gcloud run services describe "zitadel-${ENVIRONMENT}" \
+            --project="$PROJECT_ID" --region="$REGION" \
+            --format='value(status.latestReadyRevisionName)')"
+          {
+            echo "### Deployed"
+            echo
+            echo "| | |"
+            echo "|---|---|"
+            echo "| Version | \`${VERSION}\` |"
+            echo "| Image | \`${AR_IMAGE}:${VERSION}\` |"
+            echo "| Service | \`zitadel-${ENVIRONMENT}\` |"
+            echo "| Revision | \`${revision}\` |"
+            echo
+            echo "Ingress is internal-load-balancer only, so reach it through the load balancer, not the \`run.app\` URL."
+          } >> "$GITHUB_STEP_SUMMARY"
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,9 +1,11 @@
 name: Deploy / Cloud Run
 
 # Deploys a published release image to the Cloud Run service that infra/
-# provisions. The infrastructure itself is owned by .github/workflows/infra.yml;
-# this workflow owns only the release-specific fields (image, env, secrets),
-# which the OpenTofu cloud-run module leaves under `lifecycle.ignore_changes`.
+# provisions. This workflow owns the release and nothing else: it sets
+# `--image` and stops there. Configuration — env, secrets, volumes — belongs to
+# OpenTofu, because `gcloud run services update --set-secrets` is declarative
+# over volumes as well as env, so a deploy that touched them would strip the
+# master-key mount the infra module declares.
 #
 # Manual dispatch on purpose. A `release: published` trigger would not fire:
 # the release is created by release-publish.yml using GITHUB_TOKEN, and events
@@ -26,7 +28,7 @@ on:
         options:
           - dev
       sync_secrets:
-        description: "Push the environment's GitHub secrets to Secret Manager as new versions. Required on the first deploy and after a rotation."
+        description: "Push this environment's GitHub secrets to Secret Manager as new versions. Required once, before the environment's runtime_secrets_ready flag is flipped."
         required: false
         default: false
         type: boolean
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
+name: Deploy / Cloud Run
+
+# Deploys a published release image to the Cloud Run service that infra/
+# provisions. The infrastructure itself is owned by .github/workflows/infra.yml;
+# this workflow owns only the release-specific fields (image, env, secrets),
+# which the OpenTofu cloud-run module leaves under `lifecycle.ignore_changes`.
```
fforootd: Not true for the file mount: `--set-secrets` with a path is `template.volumes` + `containers[0].volume_mounts`, neither is in #1132's `ignore_changes`, and infra.yml applies on every push to `main` touching `infra/**`. The next apply drops the mount, the server mints a throwaway key under the same ID `master-key.pem`, and every KEK fails with `enc_key.decrypt_failed`. Cleanest: let the cloud-run module own the volume and set only `--image` here. Minimum: add both attributes to `ignore_changes`.

---

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
...
+        type: string
+      environment:
+        description: "Target environment."
+        required: true
+        default: dev
+        type: choice
+        options:
+          - dev
+      sync_secrets:
+        description: "Push the environment's GitHub secrets to Secret Manager as new versions. Required on the first deploy and after a rotation."
```
fforootd: "after a rotation" orphans every KEK: same path, same file name, same key ID. New instances load the rotated PEM as `master-key.pem`, old `kid`s resolve to it, and `MigrateToLatestMasterKey` skips because the IDs match. Rotation needs a new ID with the old key still mounted (`docs/operations/encryption-keys.md`, "Rotating the master key"). Pin versions and version the file name, or drop the rotation claim for now.

---

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
...
+jobs:
+  deploy:
+    name: Deploy ${{ inputs.version }} to ${{ inputs.environment }}
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.environment }}
+    env:
+      VERSION: ${{ inputs.version }}
+      ENVIRONMENT: ${{ inputs.environment }}
+      PROJECT_ID: ${{ vars.GCP_DEV_PROJECT_ID }}
+      REGION: ${{ vars.AR_REGION }}
```
fforootd: The job already runs under `environment: ${{ inputs.environment }}`. Environment-scoped `GCP_PROJECT_ID`/`GCP_REGION` would make the input real and stop borrowing `AR_REGION` as the Cloud Run region.

---

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
...
+      MGMT_PROJECT_ID: ${{ vars.GCP_MGMT_PROJECT_ID }}
+      SOURCE_IMAGE: ghcr.io/zitadel/nextgen
+
+    steps:
+      - name: Check required repository variables
+        run: |
+          missing=()
+          for v in PROJECT_ID REGION MGMT_PROJECT_ID; do
```
fforootd: Add `GCP_WIF_PROVIDER` and `GCP_WIF_SERVICE_ACCOUNT`. Also `GCP_DEV_PROJECT_ID` isn't a mgmt `tofu output`, it's an input.

---

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
...
+          RUN_SA="zitadel-run@${PROJECT_ID}.iam.gserviceaccount.com"
+          DATA_DIR=/var/lib/zitadel/nextgen-data
+
+          gcloud run services update "zitadel-${ENVIRONMENT}" \
+            --project="$PROJECT_ID" \
+            --region="$REGION" \
+            --image="${AR_IMAGE}:${VERSION}" \
+            --service-account="$RUN_SA" \
+            --port=8080 \
+            --set-env-vars="NEXTGEN_SERVER_DATA_DIR=${DATA_DIR}" \
```
fforootd: `NEXTGEN_SERVER_DATA_DIR` is already set by the Dockerfile, and `--service-account`/`--port` re-assert fields OpenTofu owns and doesn't ignore; `services update` keeps them when omitted. Once #1151 lands, `NEXTGEN_SERVER_GENERATE_MASTER_KEY=false` belongs here so a missing mount fails the revision instead of minting a key.

---

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
...
+            --set-secrets="NEXTGEN_DATABASE_POSTGRES=zitadel-database-postgres-${ENVIRONMENT}:latest,${DATA_DIR}/master-keys/master-key.pem=zitadel-master-key-${ENVIRONMENT}:latest"
```
fforootd: Verify on a real revision before merge: in gen2 root owns the volume and the image runs as uid 65532; and the key-dir scan skips only `IsDir()`, so a projected layout with a `..data` symlink would be read as a key file and fail startup. `ls -la` the mount.

---

fforootd (changes requested):
Two blockers on the master-key mount (L6, L29); the rest are nits. Copilot's `packages: read` and Buildx comments don't apply: the package is public and `imagetools create` needs no builder.

---

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
...
+
+          if ! docker manifest inspect "${SOURCE_IMAGE}:${VERSION}" >/dev/null 2>&1; then
+            echo "::error::${SOURCE_IMAGE}:${VERSION} does not exist. Check the version input against the published releases."
+            exit 1
+          fi
+
+          gcloud auth configure-docker "${REGION}-docker.pkg.dev" --quiet
+          docker buildx imagetools create \
+            --tag "${AR_IMAGE}:${VERSION}" \
+            "${SOURCE_IMAGE}:${VERSION}"
```
fforootd: "Cloud Run pulls only from Artifact Registry or GCR" is no longer true: public GHCR images deploy directly since the July 14 GA. Keep the existence check, drop the mirror:

```yaml
# Cloud Run imports public GHCR images directly, so nothing is mirrored.
# Release tags are immutable, so the one-hour tag cache Cloud Run keeps
# for public registries cannot serve a stale image.
- name: Check that the release exists
  run: |
    set -euo pipefail
    if ! docker manifest inspect "${SOURCE_IMAGE}:${VERSION}" >/dev/null 2>&1; then
      echo "::error::${SOURCE_IMAGE}:${VERSION} does not exist. Check the version input against the published releases."
      exit 1
    fi
```

Optional belt and braces: resolve the digest here with `docker buildx imagetools inspect --format '{{.Manifest.Digest}}'` and deploy `${SOURCE_IMAGE}@${DIGEST}`.

---

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
...
+  deploy:
+    name: Deploy ${{ inputs.version }} to ${{ inputs.environment }}
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.environment }}
+    env:
+      VERSION: ${{ inputs.version }}
+      ENVIRONMENT: ${{ inputs.environment }}
+      PROJECT_ID: ${{ vars.GCP_DEV_PROJECT_ID }}
+      REGION: ${{ vars.AR_REGION }}
+      MGMT_PROJECT_ID: ${{ vars.GCP_MGMT_PROJECT_ID }}
```
fforootd: `MGMT_PROJECT_ID` is only used by the mirror step; drop it here and from the required-variables loop. `REGION` is now just the Cloud Run region, so `AR_REGION` is a misnomer; fine to keep the variable, but the README on #1132 should say what it is for.

---

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
...
+      - name: Deploy to Cloud Run
+        run: |
+          set -euo pipefail
+          RUN_SA="zitadel-run@${PROJECT_ID}.iam.gserviceaccount.com"
+          DATA_DIR=/var/lib/zitadel/nextgen-data
+
+          gcloud run services update "zitadel-${ENVIRONMENT}" \
+            --project="$PROJECT_ID" \
+            --region="$REGION" \
+            --image="${AR_IMAGE}:${VERSION}" \
```
fforootd: `--image="${SOURCE_IMAGE}:${VERSION}"` once the mirror is gone. Same substitution in the summary table below so it reports the image that was actually deployed.

---

fforootd (commented):
Cloud Run imports public GHCR images directly since the July 14 GA, so the mirror step can go and with it the docker registry login, the mgmt project variable, and the deploy identity's need for `artifactregistry.writer`. The image stays public, so no remote repository is needed. Inline suggestions below; the master-key mount also needs the `ignore_changes` fix on #1132 or the next infra apply strips it.

---

.github/workflows/deploy.yml
muhlemmer: Fixed on #1132 (744aaa0c) rather than here, taking your "cleanest" option: the cloud-run module now declares the volume and the DSN reference, and this workflow sets only `--image`. Since `--set-secrets` is declarative over volumes as well as env, that was the only split that does not leave one side able to undo the other.

.github/workflows/deploy.yml
muhlemmer: Dropped the rotation claim in ab4f6914. You are right that it orphans everything: same path, same file name, same key ID, so old `kid`s resolve to the rotated key and `MigrateToLatestMasterKey` skips.

The input description now says what it actually does — seed the secrets once, before the environment's `runtime_secrets_ready` flag is flipped — and points at `docs/operations/encryption-keys.md`. Real rotation needs the new-ID-alongside-old shape and is worth doing deliberately rather than as a side effect of a deploy flag.

.github/workflows/deploy.yml
muhlemmer: Done in ab4f6914 — environment-scoped `GCP_PROJECT_ID` and `GCP_REGION`. Agreed that this is what makes the `environment` input mean something rather than decoration, and it stops `AR_REGION` doing double duty as the Cloud Run region.

.github/workflows/deploy.yml
muhlemmer: Both added to the preflight in ab4f6914. And you are right that `GCP_DEV_PROJECT_ID` was never a mgmt output — I had written it into the README's output list, which was wrong. The README on #1132 now separates the two: WIF pair from `tofu output`, project and region as environment variables you set.

.github/workflows/deploy.yml
muhlemmer: All three dropped in ab4f6914. `--service-account` and `--port` re-asserted fields OpenTofu owns, which is exactly the fight this split is meant to avoid, and `NEXTGEN_SERVER_DATA_DIR` was redundant with the Dockerfile.

On `NEXTGEN_SERVER_GENERATE_MASTER_KEY=false` once #1151 lands: agreed, and it is the right belt to this braces. Worth noting it becomes *more* valuable with the staged flag — if someone seeds the secrets but forgets to flip `runtime_secrets_ready`, today the server quietly generates a key; with that env var it would refuse to start.

.github/workflows/deploy.yml
muhlemmer: You are right, and this was my error — I asserted the Artifact Registry restriction without rechecking it. Verified against current Google docs: Cloud Run imports public GHCR and Docker Hub images directly, cached up to an hour.

Mirror dropped in ab4f6914, along with the registry login and the mgmt project variable. Took the digest suggestion too: the workflow resolves `{{.Manifest.Digest}}` and deploys `${SOURCE_IMAGE}@${DIGEST}`, so the cache cannot matter and the revision records exactly what shipped. Verified against 1.0.0-alpha.20 — resolves to `sha256:ac57b5f1…`, and a bogus tag fails the check.

.github/workflows/deploy.yml
muhlemmer: Both done in ab4f6914. `MGMT_PROJECT_ID` is gone from the env block and the preflight. `AR_REGION` is gone entirely rather than renamed — the region now comes from environment-scoped `GCP_REGION`, which removes the misnomer instead of documenting it.

.github/workflows/deploy.yml
muhlemmer: Done in ab4f6914, with the digest rather than the tag: `--image="$IMAGE"` where `IMAGE=${SOURCE_IMAGE}@${digest}`. The summary table reports the same variable, so it shows the digest actually deployed.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
...
+      sync_secrets:
+        description: "Push the environment's GitHub secrets to Secret Manager as new versions. Required on the first deploy and after a rotation."
+        required: false
+        default: false
+        type: boolean
+
+permissions:
+  contents: read
+  id-token: write
+
```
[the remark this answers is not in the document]
muhlemmer: Moot as of ab4f6914 — the mirror step and its registry login are gone, since Cloud Run imports public GHCR images directly. What remains is `docker manifest inspect` and `imagetools inspect` against a public package, which need no credentials.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
...
+      - name: Check required repository variables
+        run: |
+          missing=()
+          for v in PROJECT_ID REGION MGMT_PROJECT_ID; do
+            [ -n "${!v}" ] || missing+=("$v")
+          done
+          if [ ${#missing[@]} -gt 0 ]; then
+            echo "::error::Missing repository variables: ${missing[*]}. See infra/README.md — they come from \`tofu output\` in infra/mgmt."
+            exit 1
+          fi
```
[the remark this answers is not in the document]
muhlemmer: Fixed in ab4f6914 — the preflight now checks `GCP_WIF_PROVIDER` and `GCP_WIF_SERVICE_ACCOUNT` as well. `MGMT_PROJECT_ID` is gone from it entirely, along with the mirror step that was its only consumer. Exercised the check with values present and missing; a missing pair reports `::error::Missing Actions variables: PROJECT_ID WIF_PROVIDER`.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,168 @@
...
+
+      - uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: ${{ vars.GCP_WIF_PROVIDER }}
+          service_account: ${{ vars.GCP_WIF_SERVICE_ACCOUNT }}
+
+      - uses: google-github-actions/setup-gcloud@v2
+
```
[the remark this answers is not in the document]
muhlemmer: `imagetools create` is gone with the mirror step (ab4f6914). The workflow still calls `docker buildx imagetools inspect` to resolve the release digest, which — as noted in review — needs no builder instance and no auth for a public package. Verified it against `ghcr.io/zitadel/nextgen:1.0.0-alpha.20` on a runner-equivalent environment: resolves to `sha256:ac57b5f1…` with no `docker buildx create` step.

.github/workflows/deploy.yml
muhlemmer: Taken seriously, and it turned into its own PR: **#1161**.

I could not test a real revision — nothing is deployed with a mount yet — so I reproduced the layout locally instead, and the `..data` half of your concern is a genuine bug in the server today:

```
..2026_09_05_12_00_00.123456   DirEntry.IsDir()=true   os.Stat().IsDir()=true
..data                         DirEntry.IsDir()=false  os.Stat().IsDir()=true
master-key.pem                 DirEntry.IsDir()=false  os.Stat().IsDir()=false
```

`..data` is a symlink, so `DirEntry.IsDir()` reports false, the key-dir scan adopts it as a key named `..data`, and startup dies with `failed to read encryption key file ".../..data": is a directory`. Confirmed end-to-end by running the server against that directory.

#1161 skips dot-prefixed entries and stats through symlinks, with three tests each confirmed to fail beforehand. Worth having regardless of Cloud Run: any self-hoster projecting keys through a Kubernetes secret hits it, and a stray `.DS_Store` previously became the deployment's *only* master key, since a non-empty directory also suppresses generation.

Leaving this thread open for the ownership half, which I still cannot check: gen2 root owning the volume against uid 65532. Once a revision is up with the mount, `ls -la` on it settles that — but with #1161 in, the projection layout is no longer a merge gate either way.

muhlemmer: Update: **#1161 is merged**, so the `..data` half of this is fixed on `main` — the key-directory scan now skips dot-prefixed entries and stats through symlinks, with regression tests for the projected layout.

Leaving this thread open for the half I still cannot verify: whether gen2's root-owned volume is readable by uid 65532. That needs a live revision with the mount, which does not exist yet. It is no longer a merge gate — #1161 makes both projection layouts safe — but it is a real unknown that would surface as a failed revision on first deploy, so it seems worth keeping visible until someone can `ls -la` the mount.

PR #1143, opened 2026-09-03
fix: make local claims work end to end

fforootd:
## Summary

Three fixes found by running the published `@zitadel/cli@1.0.0-alpha.21` claim flow against the local runtime.

- `zitadel start` now sets `NEXTGEN_PLATFORM_BOOTSTRAP_PROJECT=true` for the binary and Docker runtimes. Without the platform project the console falls back to the app's own project, the claim page's sign-in is rejected on origin (`origin "http://localhost:8080" is not allowed for this project`), and `claim/complete` refuses every session while the CLI waits out the link. A CLI-owned data dir is the one place enabling it is safe.
- `zitadel claim --json` writes the claim link and its expiry to stderr; stdout stays the single JSON envelope. `--json` silences consola, so the documented agent invocation (`--non-interactive --json`) never surfaced the link at all: zero bytes on either stream until the timeout envelope. The wrong-origin warning reaches stderr the same way.
- `claimState()` reports a recorded owning team on any server; only the detached nudge stays cloud-only, because its "temporary" framing is only true on the cloud. Before, `status` omitted `data.project.claim` and `doctor` said "not on a server where teams apply" right after the CLI itself wrote `team_id` and `claimed_at` for a local claim.

SKILLS.md documents all three.

## Validation

- `moon run cli:lint` — 0 errors (27 pre-existing warnings)
- `moon run cli:test` — 113 files, 1108 tests pass; new or changed: launcher env (binary + Docker), link on stderr under `--json`, warning on stderr under `--json`, attached-on-local for `claimState`, `status`, and `doctor`
- Manual round trip with the source-built CLI driving the published alpha.21 server binary: `start` created the platform project with no env from the caller, `claim --json` printed the link on stderr with stdout empty until the envelope, the browser leg registered and completed the claim, and `status` / `doctor` reported the team on the local server
- oxfmt is not enforced in CI and none of the touched files were clean at HEAD, so the formatter was not run over whole files; every added line was checked to be oxfmt-clean on its own

## Release notes / changeset

- Changeset: `.changeset/cli-local-claim.md` — `zitadel start` provisions the platform project in the local runtime so the console sign-in and `zitadel claim` work locally; `zitadel claim --json` writes the link and expiry to stderr; `status` and `doctor` report a recorded team on any server, with only the nudge staying cloud-only. `@zitadel/cli` minor (fixed group).

## Notes

- Behavior change to be aware of: a CLI-started server now resolves `proj_platform` as its default project, so the embedded console signs into the platform project instead of the app's project. That is the shape Console ADR 0004 describes; the scaffolded app is unaffected because it carries its own publishable key, and nothing in the journey e2e reads the hosted login shell.
- Follow-ups from the same test run, not in this PR: `claim/init` should refuse on a deployment without a platform project instead of letting the CLI poll for ten minutes; the embedded console shows dev-proxy copy (`CONSOLE_PROJECT_SECRET`) on every management 401 until session-derived authorization (ADR 053) lands.

https://claude.ai/code/session_015pep3j1rKFMZ7GaGwujmGr

### DIFF
--- a/.changeset/cli-local-claim.md
+++ b/.changeset/cli-local-claim.md
@@ -1,0 +1,9 @@
+---
+"@zitadel/cli": minor
+---
+
+`zitadel start` now provisions the platform project in the local runtime, so the console's own sign-in and `zitadel claim` work against a local server the same way they do on Zitadel Cloud. Before, the claim page signed in against the app's own project and stopped on an origin error while the CLI waited out the link.
+
+`zitadel claim --json` now writes the claim link and its expiry to stderr (stdout stays the single JSON envelope), so an agent driving the command can hand the link to a human instead of blocking silently until it expires.
+
+`zitadel status` and `zitadel doctor` now report a recorded owning team on any server, not only on Zitadel Cloud; only the "attach it to a team" nudge stays cloud-only.
--- a/.changeset/cli-local-claim.md
+++ b/.changeset/cli-local-claim.md
@@ -2,7 +2,7 @@
 "@zitadel/cli": minor
 ---
 
-`zitadel start` now provisions the platform project in the local runtime, so the console's own sign-in and `zitadel claim` work against a local server the same way they do on Zitadel Cloud. Before, the claim page signed in against the app's own project and stopped on an origin error while the CLI waited out the link.
+`zitadel start` now provisions the platform project in the local runtime, so the console's own sign-in and `zitadel claim` work against a local server the same way they do on Zitadel Cloud. Before, the claim page signed in against the app's own project and stopped on an origin error while the CLI waited out the link. A local runtime started by an older CLI is restarted (its data is kept) the next time you run `zitadel start`, so an upgrade does not leave a server without the platform project running.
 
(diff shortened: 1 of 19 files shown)


fforootd:
Addressed the review findings in 4f1391d:

- **Stale runtimes reused after an upgrade** (binary.ts, docker.ts threads): runtime records now carry a `launch_contract`; `start` reuses a healthy binary or Docker runtime only when its record shows the current contract, and otherwise restarts it on the same data dir. Verified on a data dir created by the published alpha.21 CLI: the restart bootstrapped the platform project into the existing SQLite database and `claim` minted a link on the local origin. Covered by three new `start` tests plus parser round-trip tests.
- **Inherited `NEXTGEN_PLATFORM_PROJECT_ID`** (suppressed comment): the managed binary runtime no longer passes it through, with a test.
- **`new URL(claim_url).origin` can throw**: the warning goes through `originOf()`, which falls back to the raw value.
- **Expiry narrated verbatim**: `claimLinkNarration()` only names an expiry that parses, in the ISO form `claimDeadline()` honours.

---

Issue #965, opened 2026-08-25
Environments - Lifecycle and classes


vitorbari:
## Status quo

Environments are a fixed, hard-coded set. `DefaultEnvironmentNames` in [`internal/domain/environment.go`](https://github.com/zitadel/nextgen/blob/main/internal/domain/environment.go) is `{"dev", "staging", "prod"}`, and `EnvironmentService.SeedDefaults` creates all three inside the transaction that creates the project, so a project cannot exist without them and no code path adds a fourth or removes one.

The stored record is identity only — `(project_id, id, name, created_at)`, unique on `(project_id, name)`, the name restricted to a lowercase DNS-style label of at most 63 characters. No class, no state, no lifecycle column.

The API is read-only: `GET /environments` lists and `GET /environments/{name}` reads. There is no create, rename or delete operation, and `environment.created` is the only event type — no updated or deleted counterpart exists.

The CLI carries a second, different set. [`apps/cli/src/lib/environment.ts`](https://github.com/zitadel/nextgen/blob/main/apps/cli/src/lib/environment.ts) declares `z.enum(["development", "preview", "production"])`, which `apply`, `plan` and `schemas list` accept as `--environment`. It selects a branch of the `environments` object in `zitadel.json`, which today holds a `server` per environment and feeds server resolution (`resolveServer`, origin `config-env`). So `zitadel.json` already has an environments block, keyed by names that match neither the count nor the spelling of the server's set, and neither list is derived from the other.

## Lifecycle

**Can the set change at all.** Whether a developer may create an environment beyond the seeded three, and whether the seeded set is a replaceable default or a platform guarantee that other features may rely on.

**Rename.** Whether renaming preserves identity. The deployment API addresses environments by name in the URL path (#532) and deployments are append-only records that reference that name, so a rename either rewrites history, orphans it, or is disallowed. Whether a retired name can later be reused is the same question.

**Delete.** What happens to the release the environment is currently running, to anything living in it, and to its deployment history. Whether deletion is removal or retirement that keeps the record addressable.

**Interface.** Which surface owns the set. The other three questions hang off this one:

- *Declarative* — the environments are listed in `zitadel.json` and `zitadel apply` reconciles the platform to that list, matching how the rest of the project's configuration is delivered. Create, rename and delete become diffs, and the open question is what `apply` does with a removed entry.
- *Imperative* — environments are managed through the API and console, and `zitadel.json` keeps only the per-environment settings it holds today.

[...]

PR #1145, opened 2026-09-04
feat: variables

### LINKED ISSUE #967: Environments - Variables
Research how a value that must differ per environment reaches a running release.

A release is promoted between environments unchanged ([ADR 035](https://github.com/zitadel/nextgen/blob/main/docs/adrs/035-configuration-environments.md)), so such a value cannot live inside it. Identity provider connections are the first case (#851): a project registers a separate OAuth client per environment, so the client id and the client secret both differ — one public, one sensitive.

`docs/design/idp/schemas/github.json` in #919 shows the problem, and the fix:

```json
"client_id": "Iv1.abc123",
"client_secret_env": "GITHUB_CLIENT_SECRET"
```

```json
"client_id": "${env.GITHUB_CLIENT_ID}",
"client_secret": "${env.GITHUB_CLIENT_SECRET}"
```

Both `client_id` and `client_secret` need to be references — syntax to be decided here.

**Not operating system environment variables.** These are values the developer sets on their own project's environments, stored there and resolved when a request is served on that environment. GitHub Actions is the closest parallel — a repository declares environments, each holding its own secrets and variables, and one unchanged workflow reads `${{ secrets.X }}` or `${{ vars.X }}` per environment. Vercel does the same across Production, Preview and Development.

Open questions: where a value is stored and who may read it; how a resource references one, given that a field-name suffix like `_env` needs a twin for every referencing field; whether sensitive and non-sensitive values are one mechanism or two, since they differ in whether a value can be read back or shown in a deploy diff; and what a deploy does when a referenced value is missing in the target environment.

**Outcome.** An ADR deciding how per-environment values are stored, referenced and resolved, and how the sensitive ones are protected. No behaviour ships here; #851 depends on the implementation that follows.

Part of #528.

wim07101993:
## Summary

Both the ADR and the implementation for variables, used per environment now and for settings later.

Configuration values that differ per environment can now be stored as variables, and managed over the API. A variable belongs to a project, and optionally to one of its environments; a configuration document references one as `${{ NAME }}`, and the value entered at the owner serving the request is substituted in. A reference that is the whole field keeps the value's type, so `"${{ RETRY_COUNT }}"` resolves to `10` rather than `"10"`; a reference inside a longer string is rendered into it, so `"https://${{ HOST }}/callback"` resolves to a URL; and a reference nothing was entered for is left as it stands. A variable marked secret is encrypted with the project's own key and stays readable after that key is rotated.

The project level and each environment are **separate owners, not a hierarchy**: a variable is read, written and deleted at exactly the owner addressed, and nothing is inherited in either direction. Team, user-schema and user levels are deliberately not here — ADR 061 §3 says why, and the ADR's Scope section lists what is deferred with the trigger for each.

Four endpoints, all scoped by the usual `project_id` and addressing one of its environments with an optional `environment_name`:

- `GET /variables` — the variables entered at that owner, keyed by name.
- `PATCH /variables` — enter or replace them there; names absent from the body are untouched, and the body is applied whole or not at all.
- `GET /variables/{variable_name}` — one name at that owner.
- `DELETE /variables/{variable_name}` — remove what that owner entered.

**Secrets are write-only.** A read reports that a secret is held and nothing more (`{"secret": true}`), never the value. A secret is reached by name at the point of use (`GetDecryptedVariables`) rather than by resolving a whole document, so nothing decrypts that the caller did not ask for.

## Validation

- `go test ./...`
- `moon run server:check-generate`, `moon run server:openapi`, `moon run workspace:check-openapi-rules`

## Release notes / changeset

- Changeset: `.changeset/per-environment-variables.md` — minor for `@zitadel/server` and `@zitadel/api`. Per the decision table this is a `feat` reaching users: it adds a public API surface.

## Notes

- Closes: https://github.com/zitadel/nextgen/issues/967
- ADR: `docs/adrs/061-per-environment-variables-and-secrets.md`


docs/adrs/061-per-environment-variables-and-secrets.md
```diff
@@ -0,0 +1,300 @@
...
+
+A name is `^\w+$`, which is exactly what the reference syntax below can address.
+
+### 2. `${{ NAME }}` references, with no namespace
+
+A configuration document references a variable by placeholder:
+
+```json
+{
+  "client_id": "${{ GITHUB_CLIENT_ID }}",
```
grvijayan: Secret fields such as `client_secret` should accept only the whole-field variable form (e.g. `"client_secret": "${{ GOOGLE_CLIENT_SECRET }}"`), where the placeholder is the entire value. With the embedded form, any literal text around the placeholder becomes part of the resolved secret and is committed into an immutable revision (e.g. `"client_secret": "a8f3c1-${{ SECRET_TAIL }}"`), and a trailing space resolves to a wrong secret (e.g. `"client_secret": "${{ GOOGLE_CLIENT_SECRET }} "`). The ADR should state that a field whose value is a secret accepts only the whole-field form.

docs/adrs/061-per-environment-variables-and-secrets.md
```diff
@@ -0,0 +1,300 @@
...
+reaches the row.
+
+Decryption resolves the key **named in the value's own JWE header**, not
+whatever key is active at read time. A variable outlives the key that was active
+when it was written, unlike a cookie or a token whose ciphertext is short lived,
+so reaching for the active key would make every stored secret unreadable the
+first time a project's secret key is rotated. Key lookups are memoized per
+document, since one document usually holds several secrets under one key.
+
+Reads return the ciphertext. Decryption happens only where a value is being
```
grvijayan: > Decryption happens only where a value is being substituted into a document.

Substituting secrets into the document by default might not be a good idea. As acknowledged in section 9 below, substituting secrets into the document leaves it containing unmarked decrypted values, so a caller cannot redact them from logs or deploy diffs.

For example, it does not fit what we decided for IdPs, i.e., "secret resolution must never happen upstream of anything that is diffed, hashed, committed, or printed" ([Secrets and Environments](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/1-resource-model.md#secrets-and-environments)). In this case, the client secret is needed once during token exchange, so it should never be written into the resolved document. The secret should be fetched and decrypted when needed, which keeps the resolved document free of cleartext secrets and satisfies the above requirement. 

The ADR should state that secrets are consumed via fetch-by-name rather than substituted into a document.

wim07101993: not sure about this one. Why resolve all fields in a doc except the secrets? Imo variables should only be resolved when they are needed but that is something the `VariableService` cannot know.

Also, by making it explicitly fetch by name, we get higher amount of roundtrips to the database. Right now one roundtrip is done to get all variables and afterwards a roundtrip for each key that might need to be fetched to decrypt data (which is normally only one key).

It could be possible to replace the "{{ SECRET }}" value with an interface which allows to fetch the actual secrets, and fetch those secrets in "bulk" when needed. But that would increase the complexity quite a bit.

grvijayan: If all of the vars are being used at the same time, then fetching them all through substitution makes sense, but that doesn't apply in all cases. In the case of IdP connections, there is no single "when needed". The `client_id` is needed during the `authorize` step, and the `client_id` and `client_secret` are needed at the `callback` step. Substituting the whole document during `authorize` decrypts the secret which isn't used in that fetch, and the `callback` will need its own fetch anyway. The caller knows which fields it needs and when, so a fetch-by-name makes more sense in this case. 

As I understand it, substitution does one query for all the variables, then one lookup for the encryption key. A fetch-by-name could accept a list of names and do the same two steps. So fetching `client_id` during authorize, and `client_id` + `client_secret` during callback would cost the same as substitution, without decrypting the secret when it isn't needed. It only gets more expensive if a caller fetches one name at a time.

wim07101993: That is why there are multiple methods on the `VariableService` exposed:
- `GetVariables` fetches variables by name, without decrypting secrets.
- `ReplaceVariables` replaces all variables in a document with the actual values, decryption included.

So if you would want to only fetch the client id, you can only get the variable `CLIENT_ID` without having to resolve all other vars. No extra roundtrips or decryption required.

If you want both client id and secret, you can either get the variables with names `CLIENT_ID` and `CLIENT_SECRET` and decrypt the client secret using the `GetDecryptedValue` method. Or you can create a map `{"clientId": "${{CLIENT_ID}}", "clientSecret": "${{CLIENT_SECRET}}"}` and use that to call the `ReplaceVariables` method which automatically decrypts them.

There are some optimizations left on the table regarding fetching decryption keys and decrypting variables I think, but I left those out right now to keep the code "simpler". If there are concerns regarding performance I can look into those.

docs/adrs/061-per-environment-variables-and-secrets.md
```diff
@@ -0,0 +1,300 @@
...
+  broken. See §9 for where that decision belongs.
+
+**There is no `vars.` / `secrets.` namespace and no `${env.X}` prefix.** Every
+variable lives in one store under one flat name space, so a prefix would carry
+no routing information: it would only add syntax that every author has to get
+right and every reader has to strip. Whether a value is secret is a property of
+the value, not of the reference to it, and the document that references it does
+not change when the flag does.
+
+### 3. The owner is a set of independent levels
```
elinashoko: The core direction makes sense: values that differ between environments need to stay outside the release so the same release can be promoted unchanged.

From a product POV, though, I think we’re implementing too much of the longer-term model at once while leaving some essential behaviour unresolved:

- Let’s start with the smallest slice we need: Project- and environment-level values for the first IdP use case. We need to ensure this fits the wider release model, including how user schemas and other configuration resources are pinned and resolved. But that doesn’t appear to require variables themselves to be scoped to Teams, user schemas or individual users. It’s useful to mention those as possible future extensions, but let’s implement them through separate, equally small follow-ups once we have concrete use cases.
- The precedence model should follow actual product requirements. The ADR currently allows environment-, Team-, schema- and user-level values to overlap, then introduces ranking to decide which one wins. For example, what should happen when an environment-specific value and a user-specific value with the same name both apply? Until we have a real journey that answers that, I don’t think we should commit to this inheritance model.
- We need to settle the customer-facing security model for secrets before building IdPs on top of this. Who can add or replace a secret? Can anyone view it again after saving it? How do we prevent it from appearing in previews, logs or other output? My expectation is that production secrets are write-only: authorised people can add or replace them, and Zitadel can use them, but nobody can retrieve the value afterwards. Is that the model we want?
- Invalid environment configuration shouldn’t be able to go live. Missing values currently remain as placeholders, mistyped environment names create invisible values, and renaming or deleting an environment can orphan them. The detailed implementation may belong to deployment and environment lifecycle work, but the required product outcome should be clear here: validate the target environment before deployment and fail safely if required values can’t be resolved.

I’d narrow this implementation to the smallest complete IdP slice: store the required values per Project/environment, validate them before use, handle secrets according to the agreed security model, and define the behaviour when required configuration is missing. We can keep the broader settings model as future direction and add each scope once its use case and inheritance behaviour are properly defined.

elinashoko (changes requested):

docs/adrs/061-per-environment-variables-and-secrets.md
wim07101993: The problem with envronments/releases/deployments is, that we do not yet have any of that. So we cannot tap into that. Currently variables live in the storage just as all other resources do.

I can reduce the scope by removing some levels of owners/scope, but the only levels we currently want are the ones, that don't exist yet. Even environments are not yet implemented.

Maybe wait with this ticket until those are implemented or remove scope entirely for now?

elinashoko: Yeah.. I'd say: 
- Nail the ADR/ be clear on the exact scope we want now (when envs/releases land) vs later 
- Ready epic with tickets/ chunks 
- Then implement the smallest slice that we need for IDPs 
- The implementation of the rest comes later

docs/adrs/061-per-environment-variables-and-secrets.md
```diff
@@ -0,0 +1,316 @@
...
+Three limits keep one small document from rendering an enormous response, and
+one in-memory document from recursing forever:
+
+| Limit            | Value              | Why                                                                                      |
+|------------------|--------------------|------------------------------------------------------------------------------------------|
+| Value size       | 16 KiB per string  | a variable is a config value, not a payload                                              |
+| Expansion budget | 1 MiB per document | nothing caps how many places reference one name                                          |
+| Document depth   | 20                 | a document built in memory can contain itself; one that arrived as JSON is far shallower |
+
+### 9. What this ADR does not decide
```
vitorbari: We have to cover the runtime integration (in this ADR or a follow-up one). For example, it would be useful to define:
- How envs/secrets are loaded in the context of a request (single, batch, etc - when exactly)
- How env/secrets are resolved during read and write operations of resources. Most likely we only want to resolve variables during read operations (e.g. we don't resolve variables when an IDP is being written, only when retrieving it from DB)
- How does it integrate in the current stack? (e.g. we probably want domain objects with resolved variables - how does it integrate exactly between storage and domain objects?)

docs/adrs/061-per-environment-variables-and-secrets.md
grvijayan: yes, this makes sense
my point is that for secrets specifically, the ADR makes it seem like substituting cleartext secrets into the whole document is the way to go, but it should rather be fetch-by-name when needed

wim07101993: I'll rephrase some things

docs/adrs/061-per-environment-variables-and-secrets.md
```diff
@@ -0,0 +1,450 @@
...
+
+A domain object therefore carries references, not resolved values. The object
+written and the object read back are the same shape, a revision hash over it is
+stable, and a document that reaches a log or a diff carries `${{ NAME }}` rather
+than a value.
+
+**Whichever path the use case fits.** `GetVariables` / `GetDecryptedVariables`
+for one that knows its names — the default, and the only way to reach a secret
+(§7). `ReplaceVariables` for one that genuinely wants a whole document resolved
+at once.
```
vitorbari: I think this actually depends on which model we want to offer:

1. Developers can use variables anywhere (`ReplaceVariables`) - this is similar to helm templates
2. Developer can only use variables on pre-determined resource attributes (`GetVariables` / `GetDecryptedVariables`)

It would look like:

1 - variables anywhere:

```json
{
  "$schema": "./idp-connection.json",
  "slug": "github",
  "protocol": "${{ GITHUB_PROTOCOL }}", // can use vars anywhere in the resource definition
  "template": "github",
  "display_name": "GitHub - ${{ ANOTHER_VAR }}",
  "oauth2": {
    "authorization_endpoint": "https://github.com/login/oauth/authorize",
    "token_endpoint": "https://github.com/login/oauth/access_token",
    "userinfo_endpoint": "https://api.github.com/user",
    "client_id": "${{ GITHUB_CLIENT_ID }}",
    "client_secret_env": "${{ GITHUB_CLIENT_SECRET}}"
  }
}
```

2 - pre-determined attributes:

```json
{
  "$schema": "./idp-connection.json",
  "slug": "github",
  "protocol": "oauth2",
  "template": "github",
  "display_name": "GitHub - ${{ ANOTHER_VAR }}", // ERROR when saving - not allowed
  "oauth2": {
    "authorization_endpoint": "https://github.com/login/oauth/authorize",
    "token_endpoint": "https://github.com/login/oauth/access_token",
    "userinfo_endpoint": "https://api.github.com/user",
    "client_id": "${{ GITHUB_CLIENT_ID }}", // ok
    "client_secret_env": "${{ GITHUB_CLIENT_SECRET}}" // ok
  }
}
```

wim07101993: I don't see why we would only allow for variables in predetermined fields. IMHO each field should be parameterisable if we want to provide a flexible configuration to our customers. 

Currently the service allows for retrieving vars in the two ways, by name(s) or by resolving a map. But I think that has nothing to do with how variables can be used in different places of the application. It is rather a flexible variable storage right now which can evolve in the future. How the environments, releases, deployments,... will use them, is still up for debate I think. 

The idea is that this storage can later also be used for settings/configuration of the application itself.

vitorbari: Do you prefer to keep this ADR for the storage part only and create a follow up ADR for the usage of env vars and secrets? 

I'm still don't have a clear understanding of which application layer would resolve the placeholders.

wim07101993: >**On read, per use case.** Resolution belongs to the code about to use the
value, not to a layer between storage and the domain. A use case knows which
names it needs and when; nothing above it does. So there is no request-scoped
preload and no resolve-everything step on a resource read — either would have to
guess, and guessing means fetching names the request never uses and decrypting
secrets it never touches, which is what §7 exists to avoid.

As I stated there. There is not a specific layer in which all variable resolution should happen. If a part of the code needs some variables, it can resolve them using the `VariableService`. There are three methods provided to get variables:

- `GetVariables`: get variables by name, do not decrypt, safe to get and maybe log
- `GetDecryptedVariables`: get variables by name, do decrypte, not safe to log or send back in a response
- `ReplaceVariables`: receives a doc (`map[string]any`) and replaces all variables in that document. It does decrypt secrets, so also not safe to log or send back in a response.

If the IDPs need to get only a client-id, it can get that using `GetVariables`. If in a later stage maybe more variables are required (clientID, clientSecret, ...), it can use `ReplaceVariables` and provide a document in which all variables are replaced.

This adds a little overhead, but fetching all variables in one place could become slow when there are a lot of variables and certainly if they are secret.

vitorbari: Let's take for example the use case of listing IDP in the console. The only layer that knows which of the domain object fields are returned, is the API handler. It wouldn't be correct to resolve variables on that layer.

A completely different alternative would be to resolve all variables for each env when the deployment is created instead of during runtime (the resolved document per env would be stored somehow which gets used during runtime).

I think, at this point, let's split the concern out of this ADR. Can you please create a follow up ticket to cover the env var and secret resolution?

docs/adrs/061-per-environment-variables-and-secrets.md
```diff
@@ -0,0 +1,450 @@
...
+> promoted between environments unchanged; environments are runtime slots),
+> [ADR 029](029-cryptography-secrets-and-key-lifecycle.md) (envelope encryption,
+> per-project keys), [ADR 028](028-storage-v2-statements-and-dialects.md) and
+> [ADR 041](041-storage-statement-contract-tests.md) (statements per dialect,
+> shared contract tests).
+>
+> **Deviates from** [ADR 047](047-dialect-id-generation.md) §Prefix registry: a
+> variable has no minted PK (§6).
+
+## Context
```
elinashoko, line 22: Thanks, this is much clearer now!! 

Two things still stand out to me:

- We shouldn’t let someone save a variable against an environment that doesn’t exist. Right now, a typo means the API returns success, but the value is stored somewhere nothing will ever use it. The intended environment then remains unconfigured and the problem only shows up later. I think we should reject unknown environment names when the variable is saved. We can handle rename/delete behaviour separately with the wider environment work.

- I’m not convinced we need whole-document substitution for secrets yet. For the IdP journey, we only need to fetch the specific values when they’re actually used. Resolving the whole document could pull in secrets that aren’t needed and leave us with a document where we can’t tell which values are sensitive anymore. That creates a risk of them ending up in logs, errors, previews or diffs.

What’s the concrete use case for supporting this now? And if we keep it, how do we make sure those secrets can’t accidentally be exposed? If we don’t have a clear use case and a safe way to handle it yet, I’d keep fetching secrets by name for now and add whole-document substitution later.

The missing-variable check can come later with deployments, but can we make sure it’s tracked as a blocker for making the IdP journey live? We need to catch missing credentials before users hit a broken sign-in.

docs/adrs/061-per-environment-variables-and-secrets.md
vitorbari: Not sure if this use case I mentioned makes sense at all. When retrieving a resource via API, that is usually not in the context of a single env, it should return the raw definition with the placeholders.

docs/adrs/061-per-environment-variables-and-secrets.md
wim07101993: If we don´t allow environments that don't exist yet, I will remove the environment identifier from the implementation for now and add an issue to implement it later on (since we do not have environments yet). 

The whole-document substitution is not actually an exposed endpoint or anything like that. It is an internal function which developers can use to get a number of vars/secrets. The document is something the developer must specify and the resolved document should not be exposed to the user. Secrets will in general never be exposed to the client as specified: 

> The management API: GET/PATCH /variables and GET/DELETE /variables/{variable_name}, gated on variable.read / variable.write behind the same project-scoped check as every other management resource. A secret never reads back.

The usecase can be that a client has declared a variable which is not the entire value of a field. E.g. a redirect-uri like `"https://${{ ENV_SUB_DOMAIN }}.my-app.example.com"`. If a developer only has the `GetVariables` function which calls them by name, they would need to extract the variable-name from that value. The `ReplaceVariables` function allows a developer to pass a "document" (but read that as a GoLang `map[string]any`), and replace those variables so that the returned "document" contains `"https://dev.my-app.example.com"`. This centralizes the logic to extract variables from a "document".

But it does not change the fact that a developer should never write that resolved "document" in logs or an api response.

wim07101993: My bad, environments do exist. Will fix

docs/adrs/061-per-environment-variables-and-secrets.md
wim07101993: I thought that the api would not return the resolved idp documents. It would contain the variablenames. When comparing to platforms like GitHub, there you also view the action with the variable names, not the resolved variables. (Agreed that is in text form, but the argument still stands I think)

I do think snapshotting all variables for a deployment is the right call though. Just like how we have revisions of schemas/idps/... a deployment would be a revision of the variables. Will add that in the ADR.

wim07101993:
@elinashoko, @vitorbari, @grvijayan because there was much confusion on the ADR, I rewrote it completely. I you can have a look, I have addressed all concerns I think.

docs/adrs/061-per-environment-variables-and-secrets.md
```diff
@@ -0,0 +1,252 @@
...
+   allowed. So the entire field value is replaced with the value of the secret.
+5. **Non-existing variable**: Nothing to do, the field value is rendered as is.
+   We assume this was done on purpose.
+
+#### Examples
+
+| Example                                    | Variable value | Result                           | 
+|:-------------------------------------------|:---------------|:---------------------------------|
+| `"client_id": "abcd1234"`                  |                | `"client_id": "abcd1234"`        | 
+| `"retry_count": "${{ RETRY_COUNT }}`       | 5              | `"retry_count": 5,`              |
```
vitorbari: There should be no double quotes before `${{ RETRY_COUNT }}`

docs/adrs/061-per-environment-variables-and-secrets.md
```diff
@@ -0,0 +1,252 @@
...
+5. **Non-existing variable**: Nothing to do, the field value is rendered as is.
+   We assume this was done on purpose.
+
+#### Examples
+
+| Example                                    | Variable value | Result                           | 
+|:-------------------------------------------|:---------------|:---------------------------------|
+| `"client_id": "abcd1234"`                  |                | `"client_id": "abcd1234"`        | 
+| `"retry_count": "${{ RETRY_COUNT }}`       | 5              | `"retry_count": 5,`              | 
+| `"client_id": "${{ CLIENT_ID }}`           | 4321dcba       | `"client_id": "4321dcba"`        |
```
vitorbari: closing double-quote missing

docs/adrs/061-per-environment-variables-and-secrets.md
```diff
@@ -0,0 +1,252 @@
...
+rotation, that can be different keys.
+
+### API
+
+Variables and secrets can be managed using `PATH`/`GET`/`DELETE` endpoints on
+the api. There are no separate endpoints for secrets since they exist in the
+same storage so separating them out in the api would only create a
+mental/syntax overhead.
+
+`PATH` accepts a JSON-document in which the variables are listed to patch.
```
vitorbari: PATCH*

internal/service/variables.go
```diff
@@ -0,0 +1,248 @@
...
+	if err := s.v2Pool.Statements().DeleteVariable(ctx, owner, name); err != nil {
+		if _, ok := errors.AsType[*database.NoRowFoundError](err); ok {
+			return domain.ErrVariableNotFound().WithParent(err)
+		}
+		return domain.ErrInternal(err).WithMessage("failed to delete variable from database")
+	}
+	return nil
+}
+
+func (s *variableService) ReplaceVariables(ctx context.Context, owner domain.VariableOwner, doc map[string]any) (map[string]any, error) {
```
vitorbari: This function mutates the input `doc`. Better to perform a deep copy, or make it more explicit the mutation behavior.

example:

```go
// A. Mutating

// service
ReplaceVariablesInPlace(ctx context.Context, owner domain.VariableOwner, doc map[string]any) error

// caller
if err := h.variables.ReplaceVariablesInPlace(ctx, owner, conn.Config); err != nil {
    return err
}
```

```go
// B. Non-mutating

// service
ReplaceVariables(ctx context.Context, owner domain.VariableOwner, doc map[string]any) (map[string]any, error)

// caller
resolved, err := h.variables.ReplaceVariables(ctx, owner, conn.Config)
if err != nil {
    return err
}
```

internal/domain/variable.go
```diff
@@ -0,0 +1,212 @@
...
+	ret := make(Variables, len(vs))
+	for name, v := range vs {
+		value, err := v.GetDecryptedValue(decrypter)
+		if err != nil {
+			return nil, ErrFailedToDecryptVariable(err).WithDetails(map[string]any{"name": v.Name})
+		}
+		ret[name] = &Variable{
+			Name:  v.Name,
+			Owner: v.Owner,
+			Value: value,
```
vitorbari, line 210: Even after decrypting, it is important to know if it is secret.

wim07101993: maybe, but I explicitly did not add mark it as such because the `GetDecryptedValue` method on a `Variable` would break. It would try to decrypt an already decrypted value. 

I could add another flag whether it is already decrypted. But that feels wrong, a separate type would than be better but that would add more complexity to an already rather complicated piece of code.

WDYT?

internal/service/variables.go
vitorbari: @wim07101993 , I saw the new changes. If we are going it the "InPlace" option, we don't need it to return a `map`, only the `error`.

api/openapi/endpoints/variables/methods.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+patch:
+  operationId: updateVariables
+  summary: Update variables
+  description: |
+    Enters or replaces variables at the owner this request addresses.
+
+    Every name in the body is written at exactly that owner — the project, or
+    the environment named by `environment_name` — and reaches no other. Names
+    not in the body are untouched; nothing here removes a variable, which is
```
vitorbari: Should we implement deletion on explicit `null` value? That would allow for bulk deletion.

wim07101993: Good idea, will add

internal/storage/dialect/postgres/migration/sql/000021_variables.sql
```diff
@@ -0,0 +1,70 @@
...
+-- address of its own -- the project level -- and not a wildcard. project_id has
+-- to be set all the same: it carries the foreign key, and a variable with no
+-- project would belong to nothing.
+CREATE TABLE zitadel_nextgen.variables (
+    name             TEXT NOT NULL CHECK (name <> '')
+    , project_id     TEXT NOT NULL CHECK (project_id <> '')
+        REFERENCES zitadel_nextgen.projects (id) ON DELETE CASCADE
+    -- Scoped by environment name, not id, the way an environment is addressed
+    -- everywhere else, and the way a request serving an environment knows it.
+    , environment_name TEXT NOT NULL DEFAULT ''
```
vitorbari: Storing the `environment_name` will basically make it impossible to rename the environments.
It would be to use the env ID here. In the API we can accept both name or ID, or pick one of them.

The only place we definitely need env names is in the CLI commands, for example `zitadel variables list --env prod`, but if we do the name->ID conversion in the CLI or the API does not matter much IMO.

vitorbari: Relates to https://github.com/zitadel/nextgen/issues/965

vitorbari (approved):

docs/adrs/061-per-environment-variables-and-secrets.md
```diff
@@ -0,0 +1,262 @@
...
+>
+> **Builds on** [ADR 035](035-configuration-environments.md) (a release is
+> promoted between environments unchanged; environments are runtime slots),
+> [ADR 029](029-cryptography-secrets-and-key-lifecycle.md) (envelope encryption,
+> per-project keys).
+>
+> **Deviates from** [ADR 047](047-dialect-id-generation.md) §Prefix registry: a
+> variable has no minted PK (§6).
+
+## Context
```
elinashoko, line 22: Thanks, this is much clearer now. Keeping it to Projects and environments, removing inheritance, rejecting unknown environments and making secrets write-only addresses my earlier concerns.

I still think we might be adding quite a lot of flexibility, complexity and risk before we actually need it. The use case we have today is pretty focused: give an IdP the right values for each environment while keeping the same configuration across environments.

The ADR goes further and allows variables across configuration, including replacing values across a whole document. What’s the actual use case for supporting that now? It means the same release could behave quite differently in each environment, beyond the specific differences we intended.

I’m especially unsure about doing this with secrets. For the IdP journey, we already know exactly which credentials we need and when, so we can fetch only those. Resolving a whole document can pull secrets into places where they’re no longer clearly marked as sensitive, and then we’re relying on every developer to remember not to log or return them. How would we make sure they can’t accidentally show up in logs, errors or other output? If we don’t need this for the first journey, I’d leave it out for now and add it later when we have a clear use case and a safe way to handle it.

I’d also remove the suggestion that variables should later be tied to deployments. We haven’t decided whether credentials should be included in a deployment, changed separately, or affected by rollback. For example, rolling back a deployment shouldn’t accidentally bring back an old secret. Let’s decide that as part of the deployment and credential-rotation journeys rather than setting the direction here.

elinashoko (changes requested):

docs/adrs/061-per-environment-variables-and-secrets.md
vitorbari: > The ADR goes further and allows variables across configuration, including replacing values across a whole document. What’s the actual use case for supporting that now? It means the same release could behave quite differently in each environment, beyond the specific differences we intended.

There are many products that work in this way, for example, terraform, github actions, helm templates. You can use environment specific values basically anywhere in a resource definition.

If we want to restrict env vars and secrets to only very specific fields, we would have to enforce that via schemas and application logic.

These are two different models, and the transition between them is not simple. We have to pick the one that feels right for the product.

Regardless of that, the concrete part of this ADR is not changing. It introduces storage and APIs for the env vars and secrets which will work the same despite which level of flexibility we pick.

vitorbari: Some AI generated references:

<details>
<summary><b>Prior art — Terraform, GitHub Actions, Helm</b></summary>

**Terraform.** A variable reference is valid in any argument position:

```hcl
variable "client_secret" {
  type      = string
  sensitive = true
}

resource "okta_app_oauth" "example" {
  label         = "${var.env_name} login"
  redirect_uris = ["https://${var.host}/callback"]
  client_secret = var.client_secret
}
```

The relevant part is `sensitive = true`. Terraform tracks sensitivity through the graph: any value derived from a sensitive value is itself sensitive, and plan output prints `(sensitive value)` instead of the content. Flexibility about *placement* is paired with a propagating sensitivity marker, not with a list of approved fields.

**GitHub Actions.** `${{ vars.X }}` and `${{ secrets.X }}` are valid in nearly any expression position, and both resolve per environment:

```yaml
jobs:
  deploy:
    environment: production
    steps:
      - run: ./deploy.sh --host "${{ vars.API_HOST }}"
        env:
          TOKEN: ${{ secrets.DEPLOY_TOKEN }}
```

Two safety mechanisms, neither of which restricts where a reference may appear: the runner registers every secret value with a log masker, so any occurrence in job output is replaced with `***`; and secrets are withheld entirely from workflows triggered by forks. Note also the `secrets.` prefix — a reader can see at the reference site that the value is sensitive, without looking it up.

**Helm.** `{{ .Values.x }}` anywhere in a template, with per-environment values files (`helm install -f values-prod.yaml`). Helm has no notion of a sensitive value at all — which is exactly why the ecosystem grew sops, sealed-secrets and external-secrets on top of it. It is the cautionary case: templating without a sensitivity model is the part people end up retrofitting.

The common shape across all three: placement is unrestricted, and sensitivity is handled by a mechanism that travels with the value.

</details>

<details>
<summary><b>How each model looks in our IdP definition</b></summary>

**Variables anywhere:**

```json
{
  "slug": "github",
  "display_name": "GitHub (${{ ENVIRONMENT_NAME }})",
  "oidc": {
    "issuer": "${{ IDP_ISSUER }}",
    "authorization_endpoint": "https://${{ TENANT }}.okta.com/oauth2/v1/authorize",
    "client_id": "${{ GITHUB_CLIENT_ID }}",
    "client_secret": "${{ GITHUB_CLIENT_SECRET }}",
    "scopes": ["openid", "profile", "${{ EXTRA_SCOPE }}"]
  }
}
```

No schema change needed. Any field can differ per environment, including ones we have not anticipated. The cost is the one you name: the same release can behave differently per environment in ways beyond what we intended, and a reviewer reading the document cannot tell which of these resolve to secrets.

**Pre-declared fields only:**

```json
{
  "slug": "github",
  "display_name": "GitHub",
  "oidc": {
    "issuer": "https://github.com",
    "client_id": "${{ GITHUB_CLIENT_ID }}",
    "client_secret": "${{ GITHUB_CLIENT_SECRET }}",
    "scopes": ["openid", "profile", "email"]
  }
}
```

Enforcement is not free. Every field that may hold a reference needs a pattern admitting one, every field that may not needs a pattern rejecting one, and the substitution pass needs the same list so a reference that slipped past the schema is not resolved anyway. Two places to keep in step, per field, per resource type — and each new resource repeats the exercise.

</details>

wim07101993: I think the whole document vs by name is still a bit of a misunderstanding. Like I specified in the ADR, this is only a tool for us developers to get to a variable. They should never be exposed outside the server. It sits like this:

If I need to make a request to an idp, I need to get to some variables. Since I do not want to rewrite the entire mechanism to fetch variables each time there is a "service" which can be called (only from go code) with 3 functions:

1. `GetVariables`: you specify a list of variables to fetch and the function returns those variables. No modifications, decryption,... plain data as it is stored in the database. This is useful for an api request which needs to return a list of variables. The api layer can remove all the encrypted values and return the list.
2. `GetVariablesDecrypted`: you specify a list of variables to fetch and the function returns those variables but decrypted. This is useful for when you need the value of a variable, but don't want to call the decryption logic.
3. `ReplaceVariablesInPlace`: you specify a "document" and the function replaces all occurences it can find a variable.

When making that request to the IDP, I can use the `ReplaceVariablesInPlace` function. I pass the following data into it:
```json
{
  "clientId": "${{ IDP_CLIENT_ID }}",
  "clientSecret": "${{ IDP_CLIENT_SECRET }}",
  "url": "https://${{ IDP_ENVIRONMENT }}.myidp.example.com"
}
```

That resolved document can then be used to create the correct request for the IDP. Notice that this is a way easier method than just providing a list of variables to fetch. The `url` parameter for example is not just a variable, it is a value which contains a variable. The logic to find those variables and replace them, exists in this `ReplaceVariablesInPlace` not in the others since they only "get" variables.

As for the concern of secrets, legitimate concert if we were to send the document without any processing in the http-response. However, that also counts for the others. If we were to just send the encrypted values in the response, there would also be a problem. This is however not possible. Because we use an api-spec first approach, the correct responses are already modeled out and we only need to map values from one to the other.

As for the suggestion for deployments. I can leave it out, no problem.

wim07101993:
@copilot can you fix merge conflicts?

wim07101993:
@copilot can you solve the merge conflicts?

elinashoko (approved):

PR #1150, opened 2026-09-04
fix(login): stop the passkey action submitting the password field, keep flow state and auth attempt in step

vitorbari:
## Summary

Three sign-in dead ends in the flow engine. The first two are what the bug report showed; the third surfaced from review and shares a root cause with the second.

**"Sign in with a passkey" failed on the password box.** Browsers post the whole form, so the password field rides along with the passkey action — and both the value it carried and the value it lacked sank the leg before the WebAuthn prompt appeared: an empty box failed the required-field check (`error.x-auth-methods#password_required`), a filled one was verified as a password (`error.invalid_credentials`). Neither is a password submission. A passkey login leg consumes exactly one value — the identifier that scopes `allowCredentials` — so the rest of the form is now dropped before validation, dispatch, and collection, which then agree on what the leg was given.

**Back to the identifier, then re-entering an address, failed with "The user was already authenticated".** Identifying pins the user as a factor on the auth attempt, and `PrepareUserChallenge` refuses a second user challenge on a session-linked attempt — which every flow is, since the flow service links the attempt to the session it runs against. `processBack` left that factor in place. Back onto a step that re-collects the identifier now drops the resolved user and rotates the attempt in lockstep, the same rule the declared re-purpose path already follows. The collected password goes with it: `mergeCollected` banks it before the proof is checked, so a rejected password otherwise outlived its attempt and would reach `create_user` for whoever the flow resolved next.

**Signing up with an address that already has an account ended without a handoff token.** A successful identifier lookup pins the user on the attempt whichever purpose asked for it, but the register branch returned `user_already_exists` before `recordResolvedUser` ran, so flow state and attempt disagreed. `terminate` gates the handoff on a resolved user, so that sign-in verified the password, reached the terminal step, and returned nothing to exchange for a session. It also made the back fix above skip its rotation on this path — the shipped default flow routes straight into it. Both branches now record.

## Notes

Reported from browser traces of repeated passkey sign-up / sign-out cycles against the embedded console. The third bug came from @IAM-marco's review; two existing tests reached `done` without asserting a token and so never saw it.

Resolves the standing TODO on `validateAndMerge` about action-dependent required-field checks.

The integration tests need Docker (`postgres_integration` tag) and were not executed locally; they compile clean under `go vet -tags postgres_integration`.

Not addressed here: the login UI carries a typed value across a step swap (the email ends up in the box now named `x-auth-methods#password`), and `POST /projects/query` answers 401 to a valid session after claim. Both are separate follow-ups.


internal/domain/flow_state_machine.go
```diff
@@ -1106,20 +1143,38 @@ func (r *FlowStateMachineRuntime) advance(state *FlowState, prev *FlowDefinition
...
+	// The identifier is re-offered, so the user may submit a different one —
+	// and even an unchanged one re-runs the user challenge. Both need an
+	// attempt without a user factor on it; keeping the resolved user would
+	// also scope the next passkey ceremony's allowCredentials to whoever the
+	// abandoned leg resolved.
+	collectsIdentifier, err := r.stepCollectsIdentifier(pc.ctx, pc.state, prevStep)
+	if err != nil {
+		return FlowStepResult{}, err
+	}
+	if collectsIdentifier {
```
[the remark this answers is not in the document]
vitorbari: Fixed in d24d6ce7. `processBack` now resolves the target step once and uses that set for both the identifier check and the render; `stepCollectsIdentifier` became a pure `collectsIdentifier(resolved)`. Prefill and build still happen after the resolved user is dropped.

internal/domain/flow_state_machine.go
```diff
@@ -1106,24 +1143,46 @@ func (r *FlowStateMachineRuntime) advance(state *FlowState, prev *FlowDefinition
...
+	// The identifier is re-offered, so the user may submit a different one —
+	// and even an unchanged one re-runs the user challenge. Both need an
+	// attempt without a user factor on it; keeping the resolved user would
+	// also scope the next passkey ceremony's allowCredentials to whoever the
+	// abandoned leg resolved.
+	if collectsIdentifier(resolved) {
+		if err := r.dropResolvedUser(pc, "back to identification"); err != nil {
+			return FlowStepResult{}, err
+		}
+	}
```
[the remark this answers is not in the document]
vitorbari: Fixed in 5fa4299c. Confirmed the report: `mergeCollected` banks the password before the proof is checked, so a rejected one survived. Moved the clear into `dropResolvedUser` rather than adding it at the back call site — the declared re-purpose path was already clearing it separately, so both callers now get it from one rule. Nothing prefills from the collected password, so clearing costs the user nothing. Test asserts the rejected password is present before back and gone after.

internal/domain/flow_state_machine_test.go
```diff
@@ -3567,6 +3572,366 @@ func TestFlowStateMachine_Process_NavPurposeSwitchDropsResolvedUser(t *testing.T
...
+// TestFlowStateMachine_Back_ToIdentifierRotatesAuthAttempt replays the
+// browser trace behind a "The user was already authenticated" report:
+// identify, fail the password, go back, identify again. The first
+// identification pins the user as a factor on the attempt, and
+// PrepareUserChallenge refuses a second user challenge on a session-linked
+// attempt — which every flow is, since the flow service links the attempt to
+// the session it runs against. Back onto a step that re-collects the
+// identifier must rotate the attempt in lockstep with dropping the resolved
+// user, or the second identification dies.
+func TestFlowStateMachine_Back_ToIdentifierRotatesAuthAttempt(t *testing.T) {
```
IAM-marco, line 3620: Can you check if it's possible to make the mocks stricter by matching against the second parameter of the calls instead of using `gomock.Any()`?

internal/domain/flow_state_machine.go
```diff
@@ -542,6 +520,47 @@ func (r *FlowStateMachineRuntime) routeOutcome(pc *processCtx, resolved FlowReso
...
+// PrepareUserChallenge refuses a second user challenge on a session-linked
+// attempt — which every flow is, since the flow service links the attempt to
+// the session it runs against. Clearing the flow state without rotating
+// leaves the attempt one step ahead of the flow, and the next identifier
+// submission dies on "The user was already authenticated". The abandoned
+// attempt ages out like any abandoned flow. No resolved user → nothing on the
+// attempt to escape → no rotation (idle navigation must not mint attempt
+// rows). reason names the caller for the wrapped error.
+func (r *FlowStateMachineRuntime) dropResolvedUser(pc *processCtx, reason string) error {
+	hadResolvedUser := pc.state.CollectedData.UserID != ""
```
IAM-marco, line 546: This check decides whether going back rotates the auth attempt, and it misses one path that still dies the old way.

`hadResolvedUser` treats an empty `CollectedData.UserID` as "nothing is pinned on the attempt". But when someone on the sign-up tab enters an email that already has an account, the identifier dispatch calls `SubmitIdentifier` (which pins that user on the attempt server-side) and then returns `user_already_exists` before ever reaching `recordResolvedUser`:

https://github.com/zitadel/nextgen/blob/5fa4299cfb7c732283ae987afa9eac5d03f2d282/internal/domain/flow_state_machine.go#L756-L760

So the attempt knows the user, but the flow state does not. The shipped default flow walks straight into this: `user_already_exists` routes to the `password` step, with the sign-up step left on the back stack:

https://github.com/zitadel/nextgen/blob/5fa4299cfb7c732283ae987afa9eac5d03f2d282/packages/config/defaults/default-login.json#L112-L114

Concretely: sign-up tab, existing email, the flow flips to password verification. The user presses back to fix the address. This function sees no `UserID`, skips the rotation, and the next address they submit runs `SubmitIdentifier` on the old attempt and fails with "The user was already authenticated". That is the same dead end this PR eliminates for the plain login path, and the changeset currently states the back fix without this exception.

A direction rather than a patch: record the resolved user on the `user_already_exists` branch too. After the purpose flip the flow is verifying exactly that user, which is what `recordResolvedUser` describes, and it would make `UserID` an honest signal for every path that pins a factor. Tracking "the attempt carries a user factor" separately from `UserID` would also work. Rotating unconditionally is out, per this function's own rule that idle navigation must not mint attempt rows.

No current test catches this. `TestFlowStateMachine_Process_NavPurposeSwitchDropsResolvedUser` drives the sign-up-with-existing-email flip but stops before pressing back; extending that trace with back plus a re-identify would pin the fix.

internal/domain/flow_state_machine.go
```diff
@@ -1106,24 +1152,46 @@ func (r *FlowStateMachineRuntime) advance(state *FlowState, prev *FlowDefinition
 // processBack pops the previous BackStack entry and re-renders that
 // step, restoring the snapshotted purpose. CollectedData is preserved
 // (previous form prefills); PendingChallenge is dropped; History is
-// left intact.
+// left intact. Landing back on a step that collects the identifier drops
```
IAM-marco: Three doc comments now tell a story the code moved past, all from the same refactor. This one still opens with "CollectedData is preserved (previous form prefills)", and the sentence added here then says the resolved user gets dropped. The resolved user is part of `CollectedData` (`CollectedData.UserID`), and the drop also wipes `CollectedData.AuthMethods.Password`, which this comment never mentions. The PR's own test asserts both come back empty:

https://github.com/zitadel/nextgen/blob/5fa4299cfb7c732283ae987afa9eac5d03f2d282/internal/domain/flow_state_machine_test.go#L3647-L3653

Only `CollectedData.UserData` truly survives. Worth rewording to name what survives and what goes when the target step collects the identifier.

Two neighbours went stale the same way. `renderStep`'s doc still says callers "advance state (or pop for back)", but after this change no caller of `renderStep` pops:

https://github.com/zitadel/nextgen/blob/5fa4299cfb7c732283ae987afa9eac5d03f2d282/internal/domain/flow_state_machine.go#L1242-L1244

And `buildStep`'s doc lists exactly two calling patterns, while the new call in `processBack` is a third (it has a `processCtx` but passes the local `prevStep`):

https://github.com/zitadel/nextgen/blob/5fa4299cfb7c732283ae987afa9eac5d03f2d282/internal/domain/flow_state_machine.go#L1328-L1330

Small wording fixes, but this file leans on comments as contracts, and these three now contradict the tests and the call graph.

internal/domain/flow_state_machine.go
```diff
@@ -1187,6 +1255,17 @@ func (r *FlowStateMachineRuntime) renderStep(ctx context.Context, def *FlowDefin
 	return r.buildStep(state, step, resolved, nil, nil, nil), nil
 }
 
+// collectsIdentifier reports whether a resolved field set carries the
+// identifier challenge.
+func collectsIdentifier(resolved FlowResolvedFields) bool {
```
IAM-marco: Minor, from the over-engineering pass: eight lines and one caller for a question the standard library already answers. `slices.ContainsFunc(resolved.Fields, func(f FlowField) bool { return f.Challenge == FlowFieldChallengeIdentifier })` says the same thing at the call site, at the cost of importing "slices". For what it's worth, the rest of the diff came back lean on that pass: `verificationChainDefinition` is not duplicating an existing definition (the three-step navigate one has no fields), and `identifierFieldsOnly` and `dropResolvedUser` both earn their place.

IAM-marco (changes requested):

internal/domain/flow_state_machine.go
vitorbari: Confirmed and fixed in 0922a5ba — and it was hiding a second bug.

Your trace reproduces exactly. Probe before the fix:

```
after flip: step=password purpose=login userID="" attempt=attempt-1
after back: step=register                  attempt=attempt-1   <- no rotation
```

What I did not expect: the same blank UserID also gates the handoff. `terminate` skips minting when `CollectedData.UserID == ""`, so signing up with an address that already has an account verifies the password, reaches `done`, and returns no handoff token — the user cannot exchange for a session at all. `TestPurposeNavRotatesAuthAttempt` drives that exact path but only asserted the step name, so nothing caught it.

That settled the choice between your two directions. Recording on the `user_already_exists` branch fixes rotation and handoff with one edit; a separate "attempt carries a user factor" flag would have fixed only the rotation and left `done` tokenless. I checked the hazard the re-purpose comment warns about — a login-resolved user leaking into passkey registration — and `PreparePasskeyRegistrationChallenge` reads the factor off the attempt, not flow state, and the attempt already has it pinned on this path. So recording widens nothing; it stops flow state lying about what the attempt already knows.

Tests: `TestFlowStateMachine_Back_AfterUserAlreadyExistsRotatesAuthAttempt` walks sign-up tab → taken address → back → different address. The two existing flip tests now assert the handoff token. Reverting the one-line reorder fails all three. Integration side, `TestPurposeNavRotatesAuthAttempt` now asserts the token too.

Changeset updated to name all three fixes rather than overstating the back one.

internal/domain/flow_state_machine.go
vitorbari: All three corrected in 0922a5ba.

`processBack` now names what survives and what goes: `CollectedData.UserData` stays so the previous form prefills; a target that collects the identifier also drops the resolved user and the password collected for them, and rotates the attempt.

`renderStep` no longer claims a popping caller — after the single-resolution change its callers are Start, Render, routeOutcome and terminate, none of which pop.

`buildStep` now lists the third pattern (`processBack` passes `pc.state` with the back-stack step it popped to).

internal/domain/flow_state_machine.go
vitorbari: Taken — `collectsIdentifier` is gone, replaced by `slices.ContainsFunc` inline at its one call site in 0922a5ba.

Also folded in a Copilot nitpick on the neighbouring helper while I was there: `identifierFieldsOnly` said "The identifier itself stays in", which reads as though a passkey leg always keeps one. It keeps whatever the *current step* declares as an identifier field, so the password step submits nothing at all — that being the discoverable ceremony, where the assertion carries the user handle. Comment now says that.

internal/domain/flow_state_machine_test.go
vitorbari: Done in 0922a5ba. Added two matchers and used them across the rotation tests:

- `attemptFor(projectID, sessionID)` on `Start` — the session id is the load-bearing part, since exchange upgrades that session in place, so a rotation that dropped it would break sign-in with no test noticing.
- `identifiedBy(attemptID, attribute, value)` on `SubmitIdentifier` — a test pinning rotation can no longer pass on a call that went to the wrong attempt.

`SubmitPassword` in the back test now matches its attempt and plaintext too.

internal/domain/flow_state_machine.go
```diff
@@ -708,10 +759,17 @@ func (r *FlowStateMachineRuntime) dispatchChallenges(pc *processCtx, resolved Fl
...
 				return flowDispatchResult{}, fmt.Errorf("flow state machine: submit identifier: %w", err)
 			}
+			// A successful lookup pins the user on the attempt whichever
+			// purpose asked for it, so the flow records it either way.
+			// user_already_exists then flips to login and routes to
+			// verification, where the flow is verifying exactly this user — so
+			// recording is honest, and it keeps CollectedData.UserID a truthful
+			// signal of what the attempt carries. Left blank, the two fall out
+			// of step: back skips its rotation and terminate skips the handoff.
+			recordResolvedUser(state, userID)
```
IAM-marco, line 769: Not a blocker for this PR, but this line makes an existing hole easier to fall into, so flagging it here to get it into the right ticket.

Recording the user here is what fixes the back rotation and the missing token, and the shipped flow is fine: default-login.json sends `user_already_exists` to the password step, so a credential is always checked before the flow can finish:

https://github.com/zitadel/nextgen/blob/03083dedd05653adf5a55d7c9a0241ae1b81a3c3/packages/config/defaults/default-login.json#L112-L114

The trade shows up in flows a project authors badly. A definition can route `user_already_exists` straight to a terminal step. Before this commit such a flow ended with no session token: the user was looked up but never recorded, so `terminate` skipped the handoff. Broken, but broken in the locked-out direction. After this commit the same flow hands out a working login for nothing more than typing an email address that already has an account. No password, no passkey. The new assertion in `TestFlowDispatch_RegisterEntry_IdentifierAlreadyExists_Flips` pins exactly that shape: identify, land on `done`, token issued:

https://github.com/zitadel/nextgen/blob/03083dedd05653adf5a55d7c9a0241ae1b81a3c3/internal/domain/flow_state_machine_test.go#L1714-L1720

To be clear about scope: the login side has always had this hole. A definition that sends the identifier step straight to `done` has minted a token on lookup alone since before this PR, and `terminate`'s own comment says why: attempts start with empty `RequiredChecks`, so every attempt counts as complete and the token gets minted, with a note to remove the guard once the policy engine populates `RequiredChecks`:

https://github.com/zitadel/nextgen/blob/03083dedd05653adf5a55d7c9a0241ae1b81a3c3/internal/domain/flow_state_machine.go#L1234-L1238

So this commit does not create the problem, it adds one more definition shape that reaches it. Nothing to change in this PR. Worth a line in whatever tracks the RequiredChecks policy work, so "a terminal step sits behind a verification" is enforced server-side instead of trusted to every definition author.

IAM-marco (approved):

Issue #1044, opened 2026-08-28
CLI - cross-file SSO validator rules


grvijayan:
## Context

Currently, `idp-connection.json` (#1000) validates files one at a time. However, many rules span multiple files, such as a flow step naming a provider not enabled in the schema, or a connection that no schema lists.

These cross-file rules live in the TypeScript validator (`@zitadel/config/validate`), which the CLI executes during `plan` and `apply` (`apps/cli/src/lib/sync/flow-validation.ts`). The Go server mirror at flow-definition writes is deferred with #1014; until it lands these rules run only in the CLI.

While the validator currently covers passwords and passkeys, this ticket implements the **SSO cross-file rules** defined in the [IdP resource model](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/1-resource-model.md#validator-rules) and [Auth method selection](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/2-auth-method-selection.md#validation-rules) design, along with the necessary working-tree reader.

This ticket implements **error-grade** rules only. Warning-grade rules are deferred as a batch to #1083.

## Scope

**1. Inputs**
*   **Working Tree Reader:** The validator must read the entire local working tree (every connection under `.zitadel/idps/`, every user schema, and every flow definition). This allows it to trace a flow's `sso_providers` slug to the schema's `sso.providers` list and to the connection file.
*   Server-side state (like the Invalid Strategy Pointer check, #1029) is not needed here.

**2. Flow-to-Schema Rules (Errors)**

| Rule | Fires when |
| :--- | :--- |
| **Flow enables SSO** | A step has `sso_providers` but the pinned schema's `sso.enabled` is not `true`. Matches the existing password rule shape. |
| **Provider ID validity** | An `sso_providers` entry is missing from the pinned schema's `sso.providers` list. |
| **Cross-resource resolution** | A name in `sso.providers`, or an `sso_providers` entry on a step, has no matching connection file under `.zitadel/idps/`. The server's step-entry check is deferred with #1014; this schema-side check is CLI-only. |
| **Full outcome routing** | A step with `sso_providers` fails to route `callback`, `identity_unknown`, or `user_already_exists`. `callback` is enforced today; the other two are new. |
| **Collection-step conflict routing** | A step with `on_success: create_user_with_sso` fails to route `user_already_exists`. |
| **Empty `claim_mapping` intersection** | An offered provider's `claim_mapping` keys share zero properties with the flow's pinned schema. |
[...]

Issue #1037, opened 2026-08-28
Flow engine - SSO collection step and `create_user_with_sso`


grvijayan:
## Context

When identity resolution (#1035) raises `identity_unknown`, the flow moves to a data collection step. This occurs when the provider returns a new subject and the claims do not cover every required property, or if a required unique value arrives unverified.

The provider's data is waiting in the "parked identity" record (#1032), which holds the subject, mapped claims, and verification results. The collection step displays the schema's fields with the provider's values prefilled. The user completes any missing data, and submitting the form creates the account from the external identity using a new `create_user_with_sso` action. The [Social login flow](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/3-social-login-flow.md#new-users-prefill-and-confirm) defines the prefill rules and the `create_user_with_sso` action; the [CLI provider setup](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/4-cli-provider-setup.md#3-the-sso-collection-step-register-sso) scaffolds the step as `register-sso`, modeled on `register-password`.

This ticket implements the engine side: prefill on render and the `create_user_with_sso` submission handler. Scaffolding the `register-sso` step into the default flow is handled in #1047.

## Scope

**1. Prefill on Render**
*   **Data Mapping:** A step reached via `identity_unknown` renders its fields with the `value` populated from the parked identity's mapped claims, keyed by schema property name (`claim_mapping` decides which claim feeds which property; the mapping already ran in #1027). The wire shape remains unchanged, utilizing the existing `value` field in `field.yaml`.
*   **Missing Data:** Required or optional properties with no corresponding claim render empty. Required properties must be completed by the user (enforced by existing required-field validation).
*   **Editability:** Every prefilled field is editable. *(Note: `x-editable` is currently out of the dialect and will not be reintroduced in this iteration; editability is uniform across all fields).* 
*   **Missing Identity Record:** If the step renders without a parked identity (e.g., reached via browser `back` button, or after the record expires), it renders with empty values and no error. However, submitting it will fail with a **flow integrity error**, as `create_user_with_sso` requires an identity to consume.

**2. Edits and Verification**
*   **Verification Drop Rule:** The parked identity carries a verification result per claim. If a user submits a value that differs from the prefilled one, that property's verification result is immediately dropped (the provider vouched for its own value, not the user's edit). 
*   *Note:* While this rule is implemented now, the results are not yet persisted onto the user record (`x-verify` is not in the dialect for this release).

[...]

Issue #1031, opened 2026-08-28
Flow engine - render `sso_providers` from the connection


grvijayan:
## Context

Following the schema updates in #1000, a flow step's `sso_providers` is now simply a list of connection slugs. However, the login UI still expects a fully resolved object (`[{id, name, template}]`) for each provider.

This ticket bridges that gap at render time, per [Rendering from the Connection](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/2-auth-method-selection.md#rendering-from-the-connection): the engine resolves each slug to its underlying connection to emit the `id` (slug), `name` (`display_name`), and `template`. By doing this, the connection remains the single source of truth for branding, preventing flow definitions from going stale. This ticket also implements the [Social login flow](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/3-social-login-flow.md#constraints--edge-cases) rule that pattern environments must suppress provider buttons.

Handling the actual SSO submission is covered in #1030.

## Scope

**1. Stored and Domain Shape**
*   **Struct Updates:** The stored definition content (`internal/storage/flowdefinition/content.go`) and the domain step (`FlowSSOProvider` in `internal/domain/flow_state_machine.go`) must be updated to carry the slug list, dropping the legacy `{id, name, template}` struct.
*   **Validator:** The validator's "step with no inputs" check must continue counting `sso_providers` entries as valid inputs.

**2. Render-Time Resolution**
*   **Mapping:** When rendering a step with `sso_providers`, the engine resolves each slug (via #1003) and emits `{id: <slug>, name: <display_name>, template: <template>}` in the exact order defined by the step.
*   **Live Updates:** Because resolution happens at render, a `display_name` edit on the connection reflects on the next page load without requiring a flow-definition update.
*   **ID Strictness:** The rendered `id` is the connection slug. Instance-suffixed values (e.g., `google-1`) are no longer used.
*   **Dead Slugs:** If a slug resolves to no connection during render, it is silently dropped from the rendered list, and a diagnostic naming the step and slug is logged. The page renders the remaining providers and inputs normally. *(Note: #1014 prevents dead slugs at write-time; this handles edge cases like subsequent deletions).*
*   **Example Updates:** Update the YAML examples showing instance-suffixed IDs to slug IDs.

**3. Pattern Environments**
[...]

Issue #1014, opened 2026-08-27
[deferred] Service - SSO validation during flow-definition writes


grvijayan:
## Context

Cross-resource SSO validation occurs exclusively when creating or updating a flow definition, as it is the only document that pairs a `user_schema` with specific `sso_providers`.

By centralizing these error-grade checks on the server (mirroring the config package validator), we keep IdP connection writes entirely free of cross-resource validation (#1003). The *Invalid Strategy Pointer* rule requires only the connection document and the strategy registry, so it continues to run during connection writes (#1029).

**Deferred:** Server-side validation is deferred because the CLI syncer (currently the only flow writer) already enforces these rules during `plan` and `apply` (#1044). If a malformed flow bypasses the CLI, the runtime still fails safely by dropping unresolvable slugs (#1031), throwing flow-level errors on bad submissions (#1030), and halting on unrouted outcomes. This server-side mirror will only become a strict requirement when a second writer, such as Console flow editing (#998), is introduced.

## Scope

**1. Schema Reader Extension**
*   Update the Go `x-auth-methods` reader ([`internal/domain/flow_field_resolver_schema.go`](https://github.com/zitadel/nextgen/blob/main/internal/domain/flow_field_resolver_schema.go)) to return the `providers` list from the `sso` slot.

**2. Flow-Definition Rules (Create/Update)**

Write operations must be rejected with an error if they violate any of the following:
*   **SSO Enabled:** A step utilizing `sso_providers` requires the pinned schema to have `sso.enabled` set to true.
*   **Provider Allowed:** Every entry in `sso_providers` must be explicitly listed in the pinned schema's `sso.providers`.
*   **Slug Resolution:** Every entry must resolve to a valid, existing connection within the project (via slug lookup).
*   **Required Routes:** A step using `sso_providers` **must** route `callback`, `identity_unknown`, and `user_already_exists`.
*   **Creation Route:** A step where `on_success` is set to `create_user_with_sso` **must** route `user_already_exists`.

**3. Connection vs. Pinned Schema Rules**

For every connection that an `sso_providers` entry resolves to, the following cross-resource checks apply (both are error-grade to ensure `plan` and `apply` agree; doc changes tracked in #1044):
*   **Empty `claim_mapping` Intersection:** Fails if the connection's `claim_mapping` keys share zero properties with the pinned schema.
*   **Unmapped Verification Key:** Fails if a key in `verified_claims` has no corresponding entry in `claim_mapping`.
[...]

PR #1156, opened 2026-09-04
feat(idp): add the IdP connection schema and SSO vocabulary to the config dialect

### LINKED ISSUE #1000: Config contracts - IdP connection schema, `sso` auth-method schema, flow-definition SSO vocabulary
## Context

Social login modifications require updating three committed-config contracts:
1. **IdP Connection:** A new schema to create provider connections.
2. **User Schema:** Update `sso` to list the specific provider connections (slugs) users are permitted to sign in with.
3. **Flow Definition:** Update `sso_providers` to list the slugs offered in the flow, and introduce the SSO outcome and actions.

All subsequent tickets in this epic depend on this ticket.

## Scope

*   **IdP Connection Schema (`idp-connection.json`):**
    *   Add to the dialect schema set in two locations: embedded in the server and shipped in the config package.
    *   Add the file to the existing drift test so both copies remain identical.
    *   Update the setup process to copy the schema into `.zitadel/meta/` alongside existing schemas.
    *   The receipt suite validates its fixtures against this shipped file rather than the draft in `docs/design/idp/schemas/`.
*   **User Schema SSO Slot (`sso-auth-method.json`):**
    *   Add the new schema and update the `sso` slot of `auth-methods.json` to reference it.
    *   Enforce validation: `providers` is **required and non-empty** when `enabled` is `true`, and **forbidden** when `enabled` is `false`. (The other four slots remain unchanged).
    *   *Note:* This intentionally rejects the stored shape `sso: {"enabled": true}` (missing providers) on re-publish. Stored schema revisions are immutable and unaffected; only re-publishing or editing an affected schema body triggers the error, and schemas without an `sso` slot are unaffected (area 2, Breaking schema migrations).
*   **Flow Definition Schema Updates:**
    *   Refactor `sso_providers` to strictly accept a list of connection slugs.
    *   Add `create_user_with_sso` as an accepted `on_success` value.
*   **Engine & Validators (`identity_unknown`):**
    *   Introduce `identity_unknown` as a reserved outcome that triggers a `login` to `register` purpose flip.
    *   Apply this addition at all enumeration sites: the TypeScript validator, the Go validator, the engine flip table, and the `transitions` description within the flow definition schema.
*   **Documentation & Examples:**
    *   **ADR 017:** Amend the SSO note to clarify that #851 adds `identity_unknown` (instead of reusing `user_not_found`) and intentionally omits `user_link_required` because account linking is out of scope.
    *   **Runtime Examples:** Replace the instance-suffixed provider id (`google-1`) with the connection slug (`google`) in all four flow component examples: `sso-provider.yaml`, `flow-submit-request.yaml`, `flow-step.yaml`, and `flow-definition-step.yaml`. The last one also moves to the slug-list shape.

## Acceptance Criteria

- [ ] The receipt suite runs against the shipped `idp-connection.json` (not the draft in `docs/design/idp/schemas/`) and passes.
- [ ] User schema validation enforces logic: `sso: {"enabled": true}` without `providers` fails; `sso: {"enabled": false, "providers": [...]}` fails; `sso: {"enabled": false}` passes.
- [ ] A flow step containing `sso_providers: ["google"]` and `on_success: "create_user_with_sso"` successfully passes flow definition schema validation, while the legacy shape `[{id, name, template}]` fails.
- [ ] Both Go and TypeScript validators accept `identity_unknown` as a transition key, and the engine correctly flips the purpose (just as `user_not_found` does).
- [ ] The server-embedded and config-package copies of `idp-connection.json` are byte-identical, and this is actively covered by the drift test.
- [ ] No runtime example carries an instance-suffixed provider id: `sso-provider.yaml`, `flow-submit-request.yaml`, `flow-step.yaml`, and `flow-definition-step.yaml` use the connection slug (`google`), and `flow-definition-step.yaml` uses the slug-list shape.

## Resources

*   [Area 1: The connection schema & Forward Compatibility](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/1-resource-model.md#the-connection-schema)
*   [Area 2: Schema Location for `providers` & State Agreement](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/2-auth-method-selection.md#schema-location-for-providers)
*   [Area 3: Resolution Branches](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/3-social-login-flow.md#resolution-branches)
*   [Area 4: Dependencies](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/4-cli-provider-setup.md#dependencies)
*   [ADR 017: Flow Engine Auth Attempt Dispatch](https://github.com/zitadel/nextgen/blob/main/docs/adrs/017-flow-engine-auth-attempt-dispatch.md)

grvijayan:
## Summary

Closes #1000.

Social login needs three configuration contracts before the rest of #851 can be built. This PR ships them.

**1. A schema for connection files.** `idp-connection.json` describes a provider connection under `.zitadel/idps/`. It lives in `api/openapi/endpoints/schemas/`, which the server embeds and compiles at startup and the config package copies at build time. `zitadel setup` copies it into `.zitadel/meta/` so editors validate connection files. Credentials follow ADR 061: `client_secret` is a whole-value `${{ NAME }}` reference, and `client_id` is a literal or a reference.

**2. A shape for the `sso` slot in a user schema.** `sso-auth-method.json` replaces the generic slot in `auth-methods.json`. A user schema lists the connection slugs its users may sign in with under `sso.providers`. The list is required when `enabled` is true and rejected when it is false.

**3. SSO vocabulary in the flow-definition dialect.**

- `sso_providers` is a list of connection slugs. Each must match a connection's `slug`.
- `on_success` accepts `create_user_with_sso`, for the step that collects what the provider did not supply.
- `identity_unknown` is a reserved outcome. The engine emits it during login when the provider's subject matches no user and, under the connection's `creation: auto`, the claims do not fill every required property. It switches the flow to register, as `user_not_found` does.

The receipt suite now validates the connection and sso-auth-method claims against the shipped copies, and the scaffold flow validates against the shipped flow-definition schema.

ADR 017 gains an amendment explaining why `identity_unknown` is a separate outcome from `user_not_found`, and why no `user_link_required` outcome is added.

## What this PR does not do

Each item has an owner ticket.

- **Wire and storage keep the object shape for `sso_providers`.** `flow-definition-step.yaml`, the Go domain step, and the stored revision JSON still carry `{id, name, template}` objects. Only the example id changed to `google`. A flow file that still carries the object shape is now flagged in the editor, and a file rewritten to the slug list fails `zitadel plan` until #1031. #1031 owns rendering providers from a connection and the slug list at those layers. This is a deliberate narrowing of the ticket's last bullet.
- **`create_user_with_sso` exists in the dialect only.** The Go `FlowOnSuccess` enum and the handler belong to #1037.
- **The CLI rejects both of the above for now.** `zitadel plan` validates flow files with the Zod schema generated from the wire yaml, which still has object `sso_providers` and `on_success: create_user` only. That resolves with #1031 and #1037.
- **No rule yet that an SSO step must route `identity_unknown`.** That is #1044 for the CLI and #1014 for the server. Adding it to the purpose-flip coverage tables now would reject the shipped default flow, so both tables carry a comment instead.

## Implementation notes

- **Why the conditional in `sso-auth-method.json` sits at the top level.** The "providers required when enabled, forbidden when disabled" rule is written as top-level `if`/`then`/`else` instead of inside an `allOf`. Same JSON Schema meaning, but ogen ignores the conditional and types the object from `properties`, giving a typed `OptSSOAuthMethod`. Inside an `allOf`, ogen fell back to an untyped field it always emits. Because `CreateSchema` re-marshals the decoded body before validating it, a body without `sso` became `"sso":}` and was rejected as invalid JSON. `TestUserSchemaRoundTrip` in `internal/api` pins the round trip.
- **Redocly** flags `required` inside `then`/`else` as undefined properties. Two pointers are added to `.redocly.lint-ignore.yaml` with a note.
- **ADR 061.** The connection schema and the Google/GitHub examples use the `client_secret: "${{ NAME }}"` shape from #1145 instead of the earlier `client_secret_env` twin. The design drafts were updated to match.

## Validation

- `go build ./...` and `go test ./internal/domain/ ./internal/api/ ./internal/service/ ./api/...`
- New Go cases in `schema_validator_test.go` cover the four `sso` slot combinations plus non-slug providers.
- `moon run config:test`: offline `$ref` check, validator drift audit, receipt suite.
- CLI unit tests for setup and orca patchers.
- `moon run server:openapi`: Redocly clean apart from the two explicit ignores above.
- `go generate ./api` and orval regenerate; only the ogen output changed.

## Release notes

`.zitadel/meta/` now contains `idp-connection.json` and `sso-auth-method.json`. A user schema with `sso.enabled: true` lists the connection slugs its users may sign in with under `sso.providers`. In a flow definition, `identity_unknown` is a reserved transition outcome that switches a login flow to register. The editor schema also describes `sso_providers` as connection slugs and `on_success: create_user_with_sso`; the CLI and API accept those once #1031 and #1037 land.


packages/config/src/validate.test.ts
```diff
@@ -243,7 +243,7 @@ describe("steps", () => {
   it("rejects sso_providers without transitions.callback", () => {
     const def = flow();
     const s = step(def, "identifier");
-    s.sso_providers = [{ id: "idp", name: "IdP", template: "generic" }];
+    s.sso_providers = ["idp"];
```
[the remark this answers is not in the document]
grvijayan: fixed

internal/domain/flow_state_machine.go
```diff
@@ -718,12 +718,13 @@ func identifierFieldsOnly(resolved FlowResolvedFields, values map[string]any) ma
...
-// login + user_not_found → register; register + user_already_exists → login.
-// Recovery never flips.
+// applyOutcomeFlip flips CurrentPurpose on resolution outcomes:
+// login + user_not_found → register; login + identity_unknown → register;
+// register + user_already_exists → login. Recovery never flips.
 func applyOutcomeFlip(state *FlowState, outcome string) {
 	switch {
-	case state.CurrentPurpose == FlowDefinitionPurposeLogin && outcome == FlowImplicitOutcomeUserNotFound:
+	case state.CurrentPurpose == FlowDefinitionPurposeLogin &&
+		(outcome == FlowImplicitOutcomeUserNotFound || outcome == FlowImplicitOutcomeIdentityUnknown):
```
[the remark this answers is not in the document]
grvijayan: will be added during the implementation

packages/config/src/meta-schemas.test.ts
```diff
@@ -43,6 +43,8 @@ describe("meta-schemas", () => {
       "property-name.json",
       "auth-methods.json",
       "auth-method.json",
+      "sso-auth-method.json",
+      "idp-connection.json",
```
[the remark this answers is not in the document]
grvijayan: fixed

api/openapi/endpoints/schemas/idp-connection.json
```diff
@@ -0,0 +1,396 @@
...
+        "corp_idp"
+      ]
+    },
+    "protocol": {
+      "type": "string",
+      "enum": [
+        "oidc",
+        "oauth2"
+      ],
+      "description": "Selects which protocol block is required and which engine handles the connection. Closed: every value has a server implementation."
```
wim07101993: Do we need to mention that every value has a server implementation? I would assume that is the default

api/openapi/endpoints/schemas/idp-connection.json
```diff
@@ -0,0 +1,396 @@
...
+            "auto"
+          ],
+          "default": "auto"
+        }
+      }
+    },
+    "oidc": {
+      "type": "object",
+      "additionalProperties": false,
+      "description": "OIDC connection details. Endpoints come from discovery, but any may be supplied to override it, or to serve a provider that exposes no .well-known document.",
```
wim07101993, line 111: should we put the oidc/oauth2 config in a separate schema? 
Also normally an oidc config can be fetched from the discovery url. I think we should provide an option to provide the discovery url and client id/secret

api/openapi/endpoints/schemas/idp-connection.json
```diff
@@ -0,0 +1,396 @@
...
+            "type": "string"
+          },
+          "description": "Extra provider-specific authorize parameters (for example `prompt`). Engine-owned protocol parameters are reserved and rejected by `propertyNames`: the engine composes them itself, and an override of `state` or `nonce` would silently defeat CSRF and token binding. `client_secret` and `client_assertion` are rejected by the same guard: values here are committed to git and appended to the public authorize URL, so a credential pasted one level too deep leaks twice."
+        },
+        "supplementary_fetch": {
+          "type": "string",
+          "enum": [
+            "github_primary_email"
+          ],
+          "description": "Provider-specific follow-up call after userinfo, implemented in the engine. Closed: every value is implemented behaviour. New values are non-breaking."
```
wim07101993: Same comment as before: do we need to specify that what we write in the spec is implemented behavior?

api/openapi/endpoints/schemas/idp-connection.json
```diff
@@ -0,0 +1,396 @@
...
+      "minLength": 1,
+      "maxLength": 256,
+      "description": "Shown to the user on the sign-in button."
+    },
+    "subject_claim": {
+      "type": "string",
+      "minLength": 1,
+      "description": "Claim carrying the provider's stable subject identifier. Optional for OIDC (the engine uses `sub` per spec; override where `sub` is pairwise, e.g. Entra `oid`). Required for OAuth2 \u2014 no `sub` guarantee."
+    },
+    "claim_mapping": {
```
wim07101993, line 56: I thought we were going to use an expression language for these mappings. Will that be  possible once this lands?

api/openapi/endpoints/schemas/idp-connection.json
```diff
@@ -0,0 +1,396 @@
...
+        "issuer",
+        "client_id",
+        "client_secret",
+        "scopes"
+      ],
+      "properties": {
+        "issuer": {
+          "type": "string",
+          "format": "uri",
+          "pattern": "^https://[^/\\s]+|^http://(localhost|127\\.0\\.0\\.1)(:[0-9]+)?(/|$)",
```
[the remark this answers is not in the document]
grvijayan: Fixed in 07cb96ad and eda0f270. Every URL field is grouped and end-anchored, the issuer rejects query and fragment per OpenID Connect Discovery, and the endpoints reject a fragment per RFC 6749. Full URI validity stays with `format`, which the server does not assert yet; that goes with the connection endpoint work.

internal/domain/flow_field_resolver.go
```diff
@@ -242,6 +242,11 @@ const FlowImplicitOutcomeUserNotFound = "user_not_found"
 // FlowImplicitOutcomeUserAlreadyExists drives the register → login flip.
 const FlowImplicitOutcomeUserAlreadyExists = "user_already_exists"
 
+// FlowImplicitOutcomeIdentityUnknown drives the login → register flip when
+// SSO resolution finds no subject. Raised only by ceremony resolution, never
+// by a typed identifier, so a shared entry step can route the two apart.
```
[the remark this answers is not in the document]
grvijayan: Fixed in 07cb96ad. The comment now says no user was found for the provider's subject.

api/openapi/endpoints/schemas/idp-connection.json
grvijayan: yeah, we don't need to mention that, will update.

api/openapi/endpoints/schemas/idp-connection.json
grvijayan: yeah, we don't need to mention that, will update.

api/openapi/endpoints/schemas/idp-connection.json
grvijayan: I didn't see the need to split them (as we don't need to reuse these protocol blocks at the moment), so for the sake of simplicity I left them inline. When we do add more protocols in the future we can improve readability by extracting them and referencing them from here.

The issuer field is the discovery URL. The oidc block only requires `issuer`, `client_id`, `client_secret`, and `scopes` (lines 112-116). All the other endpoint fields, `jwks_uri`, `authorization_endpoint`, `token_endpoint`, `userinfo_endpoint`, are optional, they're only there in case a provider doesn't publish a well-known document for whatever reason (legacy servers?). The [example](https://github.com/zitadel/nextgen/blob/2536af080c0678b9829dbc426b65cab2bef71bfd/docs/design/idp/schemas/google.json) for Google sets the issuer and not the optional endpoints.

api/openapi/endpoints/schemas/idp-connection.json
grvijayan: For the MVP, we decided to keep it minimal. So, for Google and GitHub, a plain claim mapping should suffice. Regarding supporting expression language in the future - it shouldn't be a problem. We also have a ticket to support advanced identity provider claim mappings here: https://github.com/zitadel/nextgen/issues/996

api/openapi/endpoints/schemas/idp-connection.json
wim07101993: The problem is that there is no real way to know what the discovery endpoint is. The discovery exists in under the `.well-known` resource. But that `.well-known` resource is not as well-known as one would think. 

In zitadel for example we use `{tenant}.{region}.zitadel.cloud/.well-known/openid-configuration`. This is correct both according to the oidc spec (issuer is `https://{tenant}.{region}.zitadel.cloud`) and RFC8615 which defines `/.well-known` to be at the top of the path. 

But e.g. keycloak is older than that RFC and it hosts the discovery document under `https://{host}/realms/{realm-name}/.well-known/openid-configuration`. Correct according to oidc but not the RFC since it implementation predates that.

Because there was ambiguity, yet another standard was proposed: RFC8414 in which is stated that instead of appending the issuer, the path segment should be inserted. Which makes the issuer url `https://example.com/.well-known/oauth-authorization-server/issuer1`.

So in the case of a spec-compliant oidc server I would rather have a customer provide the discovery url, which contains the issuer, than have them provide the issuer and us guess/try where the discovery document lives.

If the server does not provide a discovery document, they can still opt for using plain oauth2, which is in the end the underlying protocol.

grvijayan: Yes, the issuer is the discovery base rather than the discovery URL itself. 
According to the [spec](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig), it should be `issuer + /.well-known/openid-configuration`. The Keycloak example you used abides by this spec. It is the same in Zitadel as well: https://zitadel.com/docs/guides/integrate/identity-providers/generic-oidc.

The config we currently have should suffice for standard providers. If we need customization for a provider that publishes the document elsewhere, we can add it additively as and when the need arises. What do you think?

wim07101993 (approved):

PR #1176, opened 2026-09-08
feat: per project password hashing algorithm

### LINKED ISSUE #504: Crypto: Per-project configurable hashing method (FIPS-aware)
ADR requirement (§Hashing): "A project admin should be able to configure which hashing method should be used… so that a tenant can have specific requirements such as FIPS."

Current state: Hashing algorithm is a single process-global setting in `cmd/server/server.go`. There is no per-project override and no admin surface to set it.

Tasks:
- [ ] Add a hashing-policy field to the project domain model + storage (`internal/domain/project.go`, project dialect in `internal/storage/v2/dialect/postgres/project.go`).
- [ ] Resolve the effective hasher per project at hash time (fallback to server default).
- [ ] API + validation so a project admin can set the algorithm/params; reject non-approved algorithms when a compliance profile is active (ties into Issue x).
- [ ] Migration for the new column.

Acceptance: A project can be configured to hash with argon2id vs bcrypt; password hashing for that project uses the configured method; verification remains backward compatible.

wim07101993:
## Summary

Let a project choose the method its passwords are hashed with (ADR 029 §Hashing). `PATCH /projects/{project_id}` takes a `password_hash` of an algorithm and its cost parameters, and `null` hands the project back to the server default. Verification is unchanged and deployment-wide, so stored passwords keep working across a change; only the next password written moves. A method is accepted only if the deployment can verify it and its cost sits inside the configured limits, which now cover scrypt, pbkdf2 and sha2 as well as bcrypt and argon2.

## Validation

- `go test ./...`

## Release notes / changeset

- Changeset: .changeset/per-project-password-hashing.md — per project password hashing algorithm (list @zitadel/server for server changes)

## Notes

- Closes: https://github.com/zitadel/nextgen/issues/504


wim07101993:
@copilot  can you fix the merge conflicts?

docs/adrs/029-cryptography-secrets-and-key-lifecycle.md
```diff
@@ -33,6 +33,24 @@ This is so that a tenant can have specific requirements such as FIPS.
...
+deployment-wide. A project's stored passwords predate whatever it picks today,
+so every hash the server could read before a project chooses stays readable
+after, and changing the choice does not invalidate what was written under the
+old one. Nothing rehashes on a policy change either: a password moves to the new
+method the next time its owner sets one.
+
+The choice is bounded by the deployment rather than free. A project may name
+only an algorithm the server can both hash with and verify -- hashing with a
+method the deployment cannot read back would lock out every user whose password
+was written that way -- and only with cost parameters inside the configured
```
muhlemmer: Check

internal/service/project.go
```diff
@@ -384,6 +442,26 @@ func (s *projectService) Update(ctx context.Context, id, name string) (*domain.P
 	return project, nil
 }
```
muhlemmer: Update() has become a rather large function. Is it possible to split some things out into helpers, so that it becomes a bit easier to read?

For example this nested `if` can be done in a helper

https://github.com/zitadel/nextgen/blob/438cc843bf17998963cb5951d015df0eea00b79d/internal/service/project.go#L391-L399

And I'm aware this isn't from this PR, but the error handling could also go into a helper (which may be reused elsewhere):

https://github.com/zitadel/nextgen/blob/438cc843bf17998963cb5951d015df0eea00b79d/internal/service/project.go#L433-L441

I just think the function became a little bit too cluttered after this addition.

internal/storage/dialect/sqlite/migration/sql/000008_project_password_hash.sql
muhlemmer: Duplicate migration number with `000008_variables.sql`

muhlemmer (commented):
Solid implementation. Just one cosmetic comment and a merge conflict.

Issue #1049, opened 2026-08-28
CLI - development-runtime secret join


grvijayan:
## Context

A connection file references a secret as `client_secret: "${{ NAME }}"`, a whole-value reference to a variable, but never contains the value itself; the actual value lives locally in `.env.local` (#1045). The engine requires this value only once, precisely at token exchange (#1027).

Per the [IdP resource model](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/1-resource-model.md#secrets-and-environments), resolving this secret anywhere upstream of the diffing, hashing, committing, or printing phases is strictly prohibited. While the production secret store remains out of scope for this epic, we must support the local development scenario: when the CLI starts the local runtime, the variables declared by the project's connections must be securely passed to the engine process. The engine then resolves the name to the value at exchange time.

This ticket implements the CLI side of the **[dev-runtime secret join](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/4-cli-provider-setup.md#dependencies)**: the injection at spawn and its reporting. The engine-side resolver that reads the injected value at the callback is #1032.

## Scope

**1. Injection at Spawn**
*   **Resolution Chain:** `zitadel start` reads the variable names referenced by `client_secret` and by any `${{ NAME }}` `client_id` from all `.zitadel/idps/*.json` files. It resolves them in order: `.env.local` → `.env` → the CLI's own environment (the sources the setup presence check reads, [Missing Secret Value](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/4-cli-provider-setup.md#missing-secret-value)).
*   **Injection Strategy:** It passes strictly those names into the runtime environment.
    *   *Binary Backend:* Passed via the spawn environment.
    *   *Docker Backend:* Passed via `--env NAME` arguments (reading from the host env, not via assignment).
*   **Missing Values:** A name lacking a local value triggers a warning naming the variable and the file (with fixes in `next_actions`), but the runtime still starts. Under `--json`, the payload lists injected and missing names (never values).
*   **Runtime Updates:** A running runtime is not hot-updated. Developers must restart after adding a provider/value (explicitly noted in the missing-value warning).
*   **Status Reporting:** `zitadel status` lists connection variable names and whether each was successfully injected at the last start.

**2. Safety & Security Strictness**
[...]

PR #1187, opened 2026-09-09
feat: let the project's env files configure the local server

mridang:
## Summary

The developer's env files now configure the local server. `zitadel start` reads `.env.local`, then `.env`, and hands every `NEXTGEN_*` variable to the server it launches. That is the server's own configuration namespace, so anything the server accepts from the environment can be set per project without touching the CLI.

- `.env.local` wins over `.env`; empty values are skipped; nothing outside the prefix is passed. The address, data directory and public base the CLI sets itself always win.
- Binary backend: child environment. Docker: bare `--env NAME`, value via the docker client's environment. Nothing reaches argv, `runtime.json`, logs or `--json`.
- `start` and `status` report the injected names under `runtime.env.injected`.
- `setup` writes a comment at the top of the scaffolded `.env.example` and `.env.local` explaining this.

This does not inspect `.zitadel/` to decide what the runtime needs. Once resources declare the variables they reference (the IdP work), that introspection lands here and narrows what is injected. Part of #1049.

## Validation

- `moon run cli:test`
- `zitadel start` against a server built from this commit, and against `ghcr.io/zitadel/nextgen:1.0.0-alpha.22` on Docker, with a `NEXTGEN_*` variable in `.env.local`: value present in the process, absent from argv and every local file.

## Release notes / changeset

`@zitadel/cli` minor: `.changeset/cli-dev-runtime-env-join.md`.

## Notes

A committed `.env` can now set local server config for everyone who clones; for a dev server that is the intent. A local server binary built before #1152 rejects `--migrate`; rebuild with `moon run server:build`.


apps/cli/src/lib/local-server/env-vars.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export const EMPTY_SUMMARY: EnvSummary = { injected: [], missing: [] };
+export const EMPTY_ENV: ResolvedEnv = { ...EMPTY_SUMMARY, values: {} };
+
+/**
+ * Names a project file may reference but must never be allowed to set on the
+ * spawned process: the loader and the CLI's own server settings. A checked-in
+ * `.env` overriding `NODE_OPTIONS` would run code on `zitadel start`.
+ */
+export const RESERVED_ENV = /^(PATH|NODE_OPTIONS|LD_.*|DYLD_.*|NEXTGEN_SERVER_.*)$/;
```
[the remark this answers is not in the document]
mridang: Good catch. `RESERVED_ENV` now carries the `i` flag, so `Path`, `node_options` and the rest are refused too, with a comment noting Windows treats them as one variable. Test extended with mixed-case names.

apps/cli/src/lib/flows/env-refs.ts
```diff
@@ -1,34 +1,83 @@
...
-          refs.add(child);
-        } else {
-          visit(child);
-        }
-      }
+export const projectEnvRefs = async (cwd: string): Promise<string[]> => {
+  const entries = await readdir(join(cwd, ZITADEL_DIR), {
+    recursive: true,
+    withFileTypes: true,
+  }).catch((error: unknown) => {
```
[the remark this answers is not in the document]
mridang: Agreed. Replaced the recursive `readdir` with an explicit walk that skips `local/` before entering it, so the runtime data directory is never enumerated.

mridang:
Simplified per discussion: `start` no longer scans `.zitadel/` for references. It reads `.env.local` and `.env` and hands the runtime every `ZITADEL_*` variable. The reference scanner, the reserved-name guard and the missing-variable warnings are gone with it; `data.runtime.env.injected` still lists the names.

apps/cli/src/lib/local-server/env-vars.ts
```diff
@@ -0,0 +1,82 @@
...
+ */
+import { readFile } from "node:fs/promises";
+import { join } from "node:path";
+import { parseEnv } from "node:util";
+
+import { ZitadelError } from "../errors";
+import { isObject } from "../json";
+
+/** Highest priority first. */
+export const ENV_FILES = [".env.local", ".env"] as const;
```
vitorbari, line 21: Should we add `.env.local` or `.env.example` to the repo as a placeholder?

mridang: Good shout, and sorry for the delay in getting to this.

While looking into it I noticed the server reads `NEXTGEN_*` variables, not `ZITADEL_*`, so the PR has been reworked: `start` now forwards `NEXTGEN_*` from the env files, which is the server's own configuration namespace, and skips the three keys the CLI sets itself.

On the placeholder: those files are created during `setup` in the scaffolded app rather than in this repo, and they now start with a commented explanation that any `NEXTGEN_*` variable in them configures the local server, with an example line. See 187189ead.

apps/cli/src/commands/start.ts
```diff
@@ -101,6 +102,10 @@ export default class Start extends BaseCommand {
 
     const paths = await ensureLocalState(this.meta.cwd);
     const existingRuntime = await readRuntimeMetadata(this.meta.cwd);
+    // ZITADEL_* variables from .env.local and .env are read before any runtime
```
vitorbari: There is some confusion between `ZITADEL_*` vs `NEXTGEN_*`, both are mentioned.
I think we should stick to `ZITADEL_*`

mridang: The two stale comments are fixed in 80545a471. `NEXTGEN_*` is deliberate though: `setup` already writes five `ZITADEL_*` variables into `.env.local` for the app's SDK (project id, project secret, issuer, URL, environment). Under a `ZITADEL_*` rule all of those would be injected into the server, which reads none of them, and the project secret would land in the container's environment for no reason. `NEXTGEN_*` is the namespace the server actually reads its config from, so only server settings get through and no list is needed.

vitorbari: Fine for now, but we need to find another prefix convention once nextgen moves to zitadel repo.

vitorbari (approved):

apps/cli/src/commands/start.ts
mridang: Valid point. I'll make a note.

PR #1189, opened 2026-09-09
feat(console): show login flows and their definitions

### LINKED ISSUE #1184: Console: login flows list and detail screens
The console has a `/flow-definitions` list and detail pair, but both are placeholder screens: the list is a three-card summary strip over a Name/Status/Created table, the detail renders `JSON.stringify` of the whole response, and neither is reachable from the sidebar. Design has since drawn both, so this replaces them with the designed screens.

**List** — the MVP directory-row layout, one card row per flow: name over its purposes (`Login + Register`), the flow's step names as badges, the user schema, and the last-change date, with a row menu. The nav gains a `Login journeys` entry, which is the label the frame uses. Rows link to the detail.

**Detail** — the login-flow detail panel: title with the user-schema badge, a meta card carrying the flow id and created date, a `STEPS` table of step / fields / actions, and the JSON/YAML document viewer with a copy button and the CLI apply hint. That viewer already exists for user schemas and generalises rather than being rebuilt, which keeps the two configuration screens the same panel.

Both are read-only. Flow definitions are configuration applied through the CLI, so the screens show the document and name the command, matching how the schema screens already behave.

### Backend

`GET /flow_definitions` covers the list, including the schema name via `expand=user_schema`, and `GET /flow_definitions/{id}` covers the detail. Three gaps, none blocking:

- The detail frame shows an `EXPIRES AT` value. Nothing on a flow definition expires, and the frame's id prefix suggests it was carried over from the runtime flow resource. Shipping without it; needs a design decision.
- `getFlowDefinition` takes no `expand`, so the detail resolves its schema badge with a second `GET /schemas/{id}`. Worth adding the parameter for symmetry with the list.
- A definition's `name` is a slug (`default-login`) and there is no display-name field, so the frames' `Default login` is a client-side humanisation of it.

bastionstack:
## Summary

- The console's `/flow-definitions` list and detail were placeholders — a three-card summary strip over a Name/Status/Created table, and `JSON.stringify` of the whole response — and neither was reachable from the sidebar. Both are now the designed screens, behind a new `Login flows` nav entry.
- **List**: one row per flow, showing the name over its purposes (`Login + Register`), the definition's step names as chips, the user schema, and the last change, with a row menu. Purposes render in a fixed order rather than the JSON's so two rows stay comparable; the name is humanised from the slug, since a definition has no display-name field.
- **Detail**: the title with a badge for the user schema, a card carrying the flow id and created date, a steps table of step / fields / actions, and the definition rendered as JSON or YAML with a copy button and the CLI apply hint.
- Both are read-only. Flow definitions are configuration applied through the CLI, so the screens show the document an operator would edit and name the command that applies it — the same posture the schema screens already take.
- `SchemaDocumentViewer` becomes a shared `DocumentViewer` rather than being rebuilt for this screen, so the two configuration screens draw the same panel.
- The user schema is a link wherever it is named, on both screens.

## Validation

- `moon ci :lint :typecheck :build :test`
- `moon run console:typecheck console:test`
- `moon run console-e2e:e2e`
- `moon run console-e2e:e2e-real`
- Browser check against `moon run console:dev-real` at 1280 and 390, light and dark, plus a full-frame pixel diff against both design frames (list 1.13%, detail 1.43%; the remainder is real-vs-mock data, the sidebar rows we do not ship, the context switcher removed by design decision D6, and the omitted `EXPIRES AT`). There is no light-theme diff because the design supplies no light frame for these screens.

## Release notes / changeset

- Changeset: `.changeset/console-login-flow-screens.md` — `@zitadel/server` minor. The console ships inside the server container, and this adds a screen a user reaches.

## Notes

- **`EXPIRES AT` is drawn on the detail frame but not implemented.** Nothing on a flow definition expires: the response carries only `created_at` and `updated_at`, and the frame's `flw_` id prefix suggests it was carried over from the runtime flow resource. Needs a design decision.
- **`getFlowDefinition` takes no `expand`**, unlike the list, so the detail resolves its schema badge with a second `GET /schemas/{id}`. Worth adding the parameter for symmetry.
- The list reads the schema name out of the `expand=user_schema` embed, which is the same `{ id, schema, metadata }` envelope `GET /schemas/{id}` returns — the name is one level in. Reading the envelope directly finds no `title` and silently falls back to printing the raw id under a heading that says `USER SCHEMA`. `apps/console-e2e/src-real/login-flows.spec.ts` guards that against a live server, because a jsdom fixture only encodes the same assumption as the code.
- The schema detail screen moves too, in its own commit. It and this one are the same composition — a single `detailPanel` card in the page wrapper — but carried 36/32 and 24 respectively. Both frames put the panel at 24px on every side and 16px on the sides when narrow, so the schema screen matched neither; they now share `DETAIL_PANEL_PAGE`. Resource detail screens (users, teams, projects) are a different composition and keep their own measured insets.
- One deviation left deliberately: `MetaValue` renders ids at 13/19 where the detail frame draws `FLOW ID` at 12/16. That frame is internally inconsistent — its own `CREATED` is 13/19 — and changing the component would move the user and team detail cards, so it is worth a design answer rather than a unilateral edit.
- The step chips include terminal steps (`done`, `register-password`) because the design annotation specifies `steps[].name` in definition order. The mock only ever drew user-facing steps, so a real flow reads busier than the frame.
- Closes #1184

## Reproduce locally

Copy this to an agent:

```
Review the console login-flow screens on branch feat/console-login-flows.

1. From the nextgen repo root:
   git fetch origin
   git worktree add ../nextgen-login-flows feat/console-login-flows
   cd ../nextgen-login-flows
   corepack pnpm install
   moon run cli:build

2. Free the ports first — 5174 and 8094 collide across worktrees:
   moon run workspace:cli -- stop --all

3. Boot a seeded real instance as a background job (it does not exit):
   moon run console:dev-real
   Note the `project` id printed in the banner.

4. Sign in at http://localhost:5174/login as dev@zitadel.local / Console-dev-1

5. The instance bootstraps a single flow, so the list has one row. To populate
   it, seed two more AS DRAFT. This matters: the engine selects the newest
   active definition matching a purpose, so an active `login` flow with an
   empty audience is chosen ahead of default-login and changes how sign-in
   behaves. Run in the browser devtools console, replacing PROJECT_ID:

   const p = "PROJECT_ID";
   const sch = (await (await fetch(`/api/flow_definitions?project_id=${p}`)).json())
     .flow_definitions[0].flow_definition.user_schema;
   for (const d of [
     { name: "password-login", purposes: { login: "identifier" }, steps: [
         { name: "identifier", fields: ["email"], actions: [{ name: "submit", kind: "submit", primary: true }], transitions: { submit: { target: "password" } } },
         { name: "password", fields: ["x-auth-methods#password"], actions: [{ name: "submit", kind: "submit", primary: true }], transitions: { submit: { target: "done" } } },
         { name: "done", complete: "show" } ] },
     { name: "passkey-login", purposes: { login: "identifier" }, steps: [
         { name: "identifier", fields: ["email"], actions: [{ name: "passkey", kind: "passkey", primary: true }], transitions: { passkey: { target: "done" } } },
         { name: "done", complete: "show" } ] },
   ]) console.log(d.name, (await fetch("/api/flow_definitions", { method: "POST",
     headers: { "content-type": "application/json" },
     body: JSON.stringify({ project_id: p, flow_definition: { ...d, status: "draft", user_schema: sch } }) })).status);

   Two 201s, then reload.

6. Look at /flow-definitions, click a row to open the detail, switch the
   JSON/YAML tab and use both copy buttons, click the user schema on each
   screen, toggle the theme with the control at the top right, and narrow the
   window to about 390px.

7. Run the gates:
   moon ci :lint :typecheck :build :test
   moon run console-e2e:e2e
   moon run console-e2e:e2e-real   # stop console:dev-real first, both boot Postgres

Expect these to differ from the design frames and not be defects: real data
rather than the mock's, the sidebar listing only shipped screens, one context
switcher rather than two, no EXPIRES AT on the detail, and rows ordered
newest-first because that is the order the API returns.
```


mridang (approved):

apps/console/src/routes/_authed/flow-definitions/index.tsx
```diff
@@ -1,76 +1,219 @@
...
+ * Five columns — name over its purposes, the definition's step names as chips,
+ * the user schema, the last change, and the row menu. The whole row is the
+ * click target via a stretched link, so there is still exactly one focusable
+ * control for the destination and middle-click still opens a tab; the menu
+ * sits above that overlay.
+ *
+ * Below `lg` the columns stack. The MVP frame draws a narrow variant of the
+ * screen but not of this row, so the stack is the console's reading of it.
+ */
+function FlowRowItem({ id, definition, updatedAt, schemaName, schemaId }: FlowRow) {
```
mridang, line 99: Do we need to show the status of the flow definitions - active or draft?

mridang (approved):

apps/console/src/routes/_authed/flow-definitions/index.tsx
```diff
@@ -1,76 +1,219 @@
...
-    <Page>
-      {/* action / tabs parked: the create button is a disabled stub and the tab
-          strip is non-functional hardcoded chrome. Restore when create flows and
-          sub-view navigation are wired. */}
-      <PageHeader
-        title="Login flows"
-        description="Flow definitions that drive your login journeys. Push changes via the CLI."
-      />
+// Mapped rather than spread so the wire's snake_case stops at the loader.
+function toFlowRow(entry: {
```
mridang: Could we not use ListFlowDefinitions200FlowDefinitionsItem type (or a aliased type) here? Maybe the API is clunky and I might need to clean that up. Using the genrated types will help you remove this "as" and unknown casts.
We could also look into switching to camelCase for the API client so you wouldn't need to convert from snake_case to camelCase

mridang (approved):

apps/console/src/routes/_authed/flow-definitions/index.tsx
bastionstack: Done in bfe71c4c — `lib/flow-definition.ts` aliases the generated types, which removed every cast and the `unknown`. Agreed on camelCase; worth its own issue.

apps/console/src/routes/_authed/flow-definitions/index.tsx
bastionstack: Worth showing, I think, but it is a change from the frame so I would rather you called it.

The design does not draw status on this screen. It matters more than the frame implies though: `active` and `draft` decide which definition the engine actually selects, and two flows are indistinguishable in this list today. While testing I seeded two extra flows as `active` to populate the list and silently hijacked sign-in — the newest active definition matching `login` won, and its first step offered only a passkey. Nothing in the UI said why.

Cheapest version is the existing `StatusBadge` beside the name, `draft` only, so an active list stays exactly as drawn. Say the word and I will add it, or I will raise it as a design question instead.

bastionstack: Added in c5b3a51e — drafts only, so an all-active list renders as drawn. It cannot show which of several *active* definitions wins; nothing exposes that, raised as #1202.

PR #1192, opened 2026-09-09
feat(api): accept user and team locators on grants

adlerhurst:
## Summary

Breaking wire update of the grants HTTP API (alpha). Create, get, and list name the bound person as `user` (`user_id`) or `team` (`team_id`) instead of `principal_type` / `principal_id`. Create accepts an id or a unique locator (`user.identifier` / `team.name`) and resolves it in the platform project. Storage (`authz_assignments`) and `authz.granted` / `authz.revoked` events stay `(principal_type, principal_id)`.

`POST /grants` with `user.identifier` is accepted with 201 whether or not a user matched, so the HTTP status cannot enumerate whether an address belongs to an active platform user. A miss or several matches still return a Grant and write nothing. A duplicate returns the existing row. Granting the session caller's own resolved user is `grant.invalid`. `user_id` and team locators keep 404 / 409.

`expand: ["principal"]` still exists as one flag for both kinds. Extra fields are copied onto the same `user` / `team` object (no sibling `principal`). Missing principals stay a degraded ref. GET-by-id takes the same expand query param and the same dual `user.read` + `team.read` gate.

Console add-admin stays picker-by-id; the request body is the mechanical `{ user: { user_id }, relation: "admin" }` compile fix.

## Validation

Console Settings → Admins against `moon run console:dev-real` (`dev@zitadel.local`): create a grant for `ada.lovelace@example.com` (`POST /grants` with `{ user: { user_id }, relation: "admin" }`), the list shows the row at level Admin, then remove via `DELETE /grants/{id}` and the table returns to “No admins yet.”

[console_admins_create_list_delete_grant.mp4](https://cursor.com/agents/bc-88736a34-63c5-4d95-a3c2-10e62e7a54e2/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fconsole_admins_create_list_delete_grant.mp4)

- `go test ./internal/service -count=1 -run TestGrantService_Create` — pass (identifier miss/ambiguous/duplicate → 201; self-grant → grant.invalid; user_id/team unchanged)
- `go test ./internal/api -count=1 -run 'TestCreateGrant|TestGrantCaller|TestGrantResponse'` — pass
- `moon run server:generate` and `moon run server:check-generate` — pass
- Spanner `ci / full-pr` failed on `TestGrantCreateLocators/unknown_identifier_is_accepted_without_a_row` racing sibling parallel creates; isolated that subtest onto its own project. Docker is unavailable here so the Spanner lane was not re-run locally.

## Release notes / changeset

Changeset: `.changeset/grant-api-locators.md` — `@zitadel/server` **minor**. Callers can create grants by user identifier or team name; create/get/list drop `principal_type` / `principal_id`. Creating by identifier always returns 201 except self-grant; it does not reveal whether the address matched.

No additional changeset for the console test isolation fix — no shipped behavior changed.

## Notes

- Nested OpenAPI `oneOf` without a discriminator made ogen skip grant operations, so create/get/list stay a single object and XOR is enforced in the handler/service as `grant.invalid`.
- Add-admin-by-email UX is out of scope (later PR).
- Identifier create opacity is HTTP-status only: a GET of a synthetic `asgn_` id still 404s. Volume/timing probes and rate limits stay on #1229.
- QueryUsers allowlist cleanup and Console people-picker replacement stay on #1229 after Console drops the picker.

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#my-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-88736a34-63c5-4d95-a3c2-10e62e7a54e2?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-88736a34-63c5-4d95-a3c2-10e62e7a54e2&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


internal/api/grant.go
```diff
@@ -107,66 +111,130 @@ func (h *Handler) DeleteGrant(ctx context.Context, params api.DeleteGrantParams)
...
+		} else {
+			input.Identifier = identifier
+		}
+		return input, nil
+	}
+	teamID, hasID := team.TeamID.Get()
+	name, hasName := team.Name.Get()
+	if hasID == hasName {
+		return input, domain.ErrGrantInvalid().WithDetails("team requires exactly one of team_id or name")
+	}
```
IAM-marco, line 144: Simplification: the handler runs the same exactly-one checks that the service runs again immediately after.

What changed: `createGrantInput` rejects a request where both or neither of `user` and `team` are set, and inside each object where both or neither of its two locator fields are set.

Why this location matters: every create request flows from here straight into `GrantService.Create`, whose first step is `validateCreateGrant`, and that function counts the four locator fields and returns the same `grant.invalid` error when the count is not exactly one: https://github.com/zitadel/nextgen/blob/76d1b958f9f7f73669d7993046ccbae59a04b6d9/internal/service/grant.go#L170-L204

What follows: one rule enforced twice in two shapes. If a fifth locator ever arrives, both places must change together, and the error subtests in the handler test duplicate what `TestGrantService_CreateLocators` already covers.

One secondary effect, kept separate: dropping the handler checks makes the error detail text less specific (the service message lists all four fields instead of naming the one object the caller sent), while the error code stays `grant.invalid` either way. If the more specific texts matter, that is the one reason to keep the handler version.

Suggested direction: map the optional values straight onto the service input (`input.UserID = string(userID)` and so on for whichever fields are present) and let `validateCreateGrant` be the single place that decides validity.

Test coverage: the service tests already lock the rule in, so the refactor stays protected; the handler subtests that only exercise the duplicated checks can leave with them.

internal/api/grant.go
```diff
@@ -107,66 +111,130 @@ func (h *Handler) DeleteGrant(ctx context.Context, params api.DeleteGrantParams)
...
+		lifecycleOwnerTeamID.SetTo(teamID)
+	} else {
+		lifecycleOwnerTeamID.SetToNull()
+	}
+	out.Metadata.SetTo(api.UserMetadata{
+		CreatedAt:            u.Metadata.CreatedAt,
+		UpdatedAt:            u.Metadata.UpdatedAt,
+		Status:               api.UserMetadataStatus(u.Metadata.Status),
+		LifecycleOwnerTeamID: lifecycleOwnerTeamID,
+	})
```
IAM-marco: Simplification: this metadata conversion is a line for line copy of the one in user.go.

What changed: `grantUserResponse` builds the `lifecycle_owner_team_id` optional (set or null) and fills an `api.UserMetadata` literal when expand loaded the user.

Why this location matters: `userResponse` in user.go does the same dance with the same fields: https://github.com/zitadel/nextgen/blob/76d1b958f9f7f73669d7993046ccbae59a04b6d9/internal/api/user.go#L305-L323 The API layer now holds two copies of the rule for how a domain user's metadata becomes wire metadata.

What follows: the next metadata field, or a change to how the owning team is derived, must be edited in both places, and nothing warns when only one copy is updated.

Suggested direction: extract a small shared helper, something like `userMetadataToAPI(u *domain.User) api.UserMetadata`, and call it from both spots.

Test coverage: both call sites have tests today (the grant internal tests and the user handler tests), so a shared helper is exercised from both sides the moment it exists.

api/openapi/endpoints/grants/grant-user.yaml
```diff
@@ -0,0 +1,29 @@
...
+          The schema that defines `attributes`. Present only when the request
+          asked for `expand: ["principal"]` and the user could be loaded.
+      attributes:
+        type: object
+        additionalProperties: true
+        description: |
+          The user's content, satisfying the schema named by `schema`. Property
+          names and types are determined entirely by that schema. Present only
+          when the request asked for `expand: ["principal"]` and the user
+          could be loaded.
```
IAM-marco: Simplification: this inline attributes schema mints a second generated type that is structurally identical to the User one.

What changed: `GrantUser` declares `attributes` inline as `type: object` with `additionalProperties: true`.

Why this location matters: user.yaml declares `attributes` with exactly the same shape and nearly the same wording: https://github.com/zitadel/nextgen/blob/76d1b958f9f7f73669d7993046ccbae59a04b6d9/api/openapi/endpoints/users/user.yaml#L29-L36 Code generation turns each inline object into its own named type, so Go now has `api.GrantUserAttributes` next to `api.UserAttributes`, identical in shape but incompatible in the type system.

What follows: the grant response code needs its own `convertUsingJson[api.GrantUserAttributes]` call instead of reusing the user conversion, and the integration test can only compare the two shapes through JSON: https://github.com/zitadel/nextgen/blob/76d1b958f9f7f73669d7993046ccbae59a04b6d9/internal/api/integration_test/grant_test.go#L591-L593

Suggested direction: move the attributes schema into a shared component (a user-attributes.yaml next to user-ref.yaml), reference it from both user.yaml and here, and regenerate. The grant response code then reuses the User attributes conversion, and the JSON comparison in the test can go back to a plain equality check. The presence caveat specific to grants can stay in this file's surrounding description.

Test coverage: the existing expand tests (`TestGrantQueryExpand` and the internal handler tests) would catch any behavior change during the swap.

internal/service/grant.go
```diff
@@ -163,17 +168,31 @@ func (s *GrantService) Revoke(ctx context.Context, projectID, id string) error {
...
+	}
+	if identifier != "" {
+		n++
+	}
+	if teamID != "" {
+		n++
+	}
+	if teamName != "" {
+		n++
+	}
```
IAM-marco: Simplification: the four locator fields are trimmed and counted one by one here, then trimmed again in every arm of the switch below.

What changed: `validateCreateGrant` creates four `strings.TrimSpace` locals and four if blocks just to count how many locator fields are set, and `resolveLocator` repeats `strings.TrimSpace` on the same fields in each of its case arms (lines 215 to 246).

Why this location matters: these two functions are the only consumers of the four fields, and each re-derives the same normalized values independently, so the same information is computed roughly a dozen times per create.

What follows: mechanical noise now, and every future locator field adds another trim local, another if block, and more trim calls in the switch.

Suggested direction: normalize once at the top of `Create` (write the trimmed values back onto the input), count with a small loop over the four values, and let both functions read the fields directly.

Test coverage: `TestGrantService_CreateLocators` pins both validation and resolution, so the cleanup is protected by tests that already exist.

internal/service/grant.go
```diff
@@ -184,6 +203,161 @@ func validateCreateGrant(input CreateGrantInput) error {
...
+	case strings.TrimSpace(input.TeamID) != "":
+		teamID := strings.TrimSpace(input.TeamID)
+		home, err := s.resolvePrincipalHome(ctx, stmts, domain.AuthzPrincipalTypeTeam, teamID)
+		if err != nil {
+			return "", "", err
+		}
+		if err := s.loadPrincipal(ctx, stmts, home, domain.AuthzPrincipalTypeTeam, teamID); err != nil {
+			return "", "", err
+		}
+		return domain.AuthzPrincipalTypeTeam, teamID, nil
```
IAM-marco: Simplification: the user id arm and the team id arm are the same three step sequence written twice.

What changed: `resolveLocator` resolves a direct id by calling `resolvePrincipalHome`, then `loadPrincipal`, then returning the type and id. It does this once for users (lines 215 to 224) and once for teams (lines 231 to 240), identical apart from the principal type constant.

Why this location matters: this switch is the single entry point for all four locator kinds, so its shape is the pattern the next locator kind will copy.

What follows: ten duplicated lines now, and any change to the id path (an extra check, a different error) has to be made twice.

Suggested direction: one small helper along the lines of `resolveByID(ctx, stmts, principalType, id)` used by both arms; each case then becomes a single line.

Test coverage: the id based create paths are covered by the existing service tests, so running them verifies the extraction.

internal/service/grant.go
```diff
@@ -184,6 +203,161 @@ func validateCreateGrant(input CreateGrantInput) error {
...
+func (s *GrantService) resolveUserByIdentifier(ctx context.Context, stmts AllStatements, home, identifier string) (string, error) {
+	urlsByKey, err := s.designatedIdentifierKeys(ctx, stmts, home)
+	if err != nil {
+		return "", err
+	}
+	found := map[string]struct{}{}
+	for key, urls := range urlsByKey {
+		if len(urls) == 0 {
+			continue
+		}
```
IAM-marco: Simplification: this guard can never fire.

What changed: the loop over `urlsByKey` skips entries whose URL list is empty.

Why this location matters: `designatedIdentifierKeys` below is the only producer of that map, and it only creates an entry by appending a non empty URL after skipping schemas with an empty key or URL (lines 331 to 335). An entry with an empty list cannot exist.

What follows: three lines of dead code that imply a case the data model cannot produce, sending the next reader hunting for how it could happen.

Suggested direction: delete the guard. If the invariant is worth recording, a one line comment on the producer says it more cheaply than unreachable code.

Test coverage: no test exercises this branch, and none can, which is itself the signal that it can go.

internal/service/grant.go
```diff
@@ -184,6 +203,161 @@ func validateCreateGrant(input CreateGrantInput) error {
...
+				slog.String("home_project_id", home),
+				slog.Int("matches", len(found)),
+			)
+		}
+		return "", domain.ErrGrantPrincipalNotFound()
+	}
+	for id := range found {
+		return id, nil
+	}
+	return "", domain.ErrGrantPrincipalNotFound()
```
IAM-marco: Simplification: a loop to extract the single key of a map, plus a final return that cannot be reached.

What changed: after the `len(found) != 1` guard, the code iterates the map to return its only key, then falls through to a `return "", domain.ErrGrantPrincipalNotFound()` that no path reaches.

Why this location matters: the guard three lines up already guarantees the map holds exactly one entry, so the loop is a roundabout way to read one value and the trailing return is dead code that still looks like a real failure path.

Suggested direction: track the matched id in a plain `matchID` string alongside the dedup map while looping over the lookups, then `return matchID, nil` after the guard.

Test coverage: the identifier lookup tests in grant_test.go, including the ambiguity cases, already pin this function's behavior.

internal/service/grant.go
```diff
@@ -184,6 +203,161 @@ func validateCreateGrant(input CreateGrantInput) error {
...
+			Pagination: database.Page[domain.JSONSchemaField]{
+				Limit:  refSchemaPageSize,
+				Cursor: cursor,
+				OrderBy: database.OrderBy[domain.JSONSchemaField]{
+					Columns:   []database.Column[domain.JSONSchemaField]{database.Col(domain.JSONSchemaFieldURL)},
+					Direction: database.OrderAsc,
+				},
+			},
+		}, JSONSchemaQueryOptions{})
+	}
```
IAM-marco: Simplification: this schema listing closure is a copy of the one in user_ref.go.

What changed: `designatedIdentifierKeys` builds a `list` closure over `ListJSONSchemas` with the same filter (project id plus user schema kind), the same page size, and the same ordering as `designatingSchemas`: https://github.com/zitadel/nextgen/blob/76d1b958f9f7f73669d7993046ccbae59a04b6d9/internal/service/user_ref.go#L111-L128

Why this location matters: both functions answer neighboring questions about the same schema set, and the pagination boilerplate is the bulk of both.

What follows: sixteen duplicated lines, and a future change (a status filter, a different page size) has to find both copies to stay correct.

Suggested direction: one package level helper, say `listUserSchemas(ctx, stmts, projectID)`, returning the iterator both callers consume.

Test coverage: both callers have their own tests, so a shared helper is exercised from two directions immediately.

apps/console/src/routes/_authed/settings/admins.tsx
```diff
@@ -159,32 +160,33 @@ function AdminsScreen() {
...
-  if (!grant.principal) return undefined;
-  const principal = grant.principal as unknown as Record<string, unknown>;
-  if (grant.principal_type === "team") return field(principal, "name");
-  return field(principal, "display") ?? field(principal, "identifier");
+  if (grant.team) {
+    return grant.team.name;
+  }
+  if (!grant.user) return undefined;
+  const user = grant.user as unknown as Record<string, unknown>;
+  return field(user, "display") ?? field(user, "identifier");
```
IAM-marco: Simplification: the cast and the `field()` indirection work around a type that already carries these properties.

What changed: `principalName` casts `grant.user` to `Record<string, unknown>` and reads `display` and `identifier` through the `field()` helper, with the comment above defending the defensive read.

Why this location matters: `GrantUser` is generated from grant-user.yaml, which builds on the user-ref component, and that component types `identifier` and `display` as optional strings: https://github.com/zitadel/nextgen/blob/76d1b958f9f7f73669d7993046ccbae59a04b6d9/api/openapi/components/schemas/user-ref.yaml#L16-L38 Line 185 already trusts the generated type when it reads `grant.team.name` directly; the user branch is the odd one out.

What follows: the cast throws away type checking exactly where the compiler could verify the field names, so a renamed wire field would slip through as a silent undefined at runtime instead of failing the typecheck.

Suggested direction: `return grant.team?.name ?? grant.user?.display ?? grant.user?.identifier;`, then drop the `field` import and the part of the comment defending the cast. If the generated type turns out not to carry these fields, that is a generation problem worth fixing at the source rather than casting around here.

Test coverage: admins.spec.tsx covers the name fallback chain and `moon run console:typecheck` is the gate that makes the direct reads trustworthy.

api/openapi/endpoints/grants/by_id/getGrant-error-response.yaml
```diff
@@ -43,3 +47,13 @@ content:
...
+      team.permission_denied:
+        summary: team.permission_denied
+        value:
+          code: team.permission_denied
+          message: 'the team management API requires the project secret'
+      user.permission_denied:
+        summary: user.permission_denied
+        value:
+          code: user.permission_denied
+          message: 'the user management API requires the project secret'
```
IAM-marco, line 59: Did we settle somewhere to call the Team and User APIs as Team/User management API? If not, I would just say 

```suggestion
          message: 'the team API requires the project secret'
      user.permission_denied:
        summary: user.permission_denied
        value:
          code: user.permission_denied
          message: 'the user API requires the project secret'
```

api/openapi/endpoints/grants/query/grant-expand.yaml
```diff
@@ -1,17 +1,17 @@
...
+    `metadata`) on `user`, or Team `status` / `created_at` / `updated_at`
+    on `team`. The ref (`user_id` / `team_id`) is always present. When the
+    principal cannot be loaded, the object stays a degraded ref — the same
+    shape as not expanding.
 
     Requires `user.read` and `team.read` in addition to `project.read`.
     Both are checked on the whole request before the list, because a mixed
     page is the common case. A caller who may not read either resource
-    receives 403 rather than a silently missing `principal`.
+    receives 403 rather than silently omitting extras.
```
IAM-marco: Can you rephrase this with bro skill?

internal/api/grant_internal_test.go
IAM-marco, line 1: Please use assert/require instead of t.Fatal

internal/service/grant_test.go
```diff
@@ -219,6 +210,283 @@ func TestGrantService_Create(t *testing.T) {
...
+	t.Parallel()
+
+	userID := "user_grant_ident"
+	teamID := "team_grant_name"
+	schemaDoc := `{"x-identifier":"email"}`
+
+	t.Run("identifier locates active user", func(t *testing.T) {
+		t.Parallel()
+		svc := newMockedGrantService(t, grantPlatformProjID, func(s *servicemocks.MockAllStatements) {
+			s.EXPECT().ListJSONSchemas(gomock.Any(), gomock.Any(), gomock.Any()).Return(&database.ListResult[*domain.JSONSchema]{
```
IAM-marco: Please go through all mocked calls and see if it's possible to enforce the second param instead of using `gomock.Any()`: this way we make the mock stricker and verification stronger

IAM-marco (changes requested):
A general comment: make sure that the returned errors do not lead to user enumration attacks (or more generally to object enumeration attacks).

---

Two-axis review of this branch, staged as inline comments below.

## Code review

Five parallel checks ran against the diff from main: repository guideline compliance, a bug scan, git history, past review comments, and code comment accuracy. Four of the five found nothing: wire naming, ID rules, audit events, generated files, and console conventions all follow the documented guidelines, no concrete defect was found on the create/get/list/expand paths, and every piece of earlier Copilot and Cursor feedback was verified as actually fixed at this head.

The git history check raised one candidate: PR #1118 deliberately made `expand: ["principal"]` return an explicit null when the principal was asked for but could not be loaded, so callers could tell "asked and it is gone" from "did not ask" (ADR 059 rule 3). This branch collapses both cases into the same degraded ref. Confidence scoring dropped the finding (25 of 100, threshold 80): the branch amends ADR 059 with an explicit grant exception documenting exactly this loss, and the PR body states it, so it is a decided and visible tradeoff rather than an oversight. No code review finding survived the filter.

## Ponytail

Nine simplification findings, all staged as inline comments. The theme: the locator feature works, but several pieces exist twice. The handler re-validates what the service already validates, the user metadata conversion copies user.go, the schema listing closure copies user_ref.go, and the inline attributes schema mints a second generated type identical to the User one (which is what forced the JSON comparison workaround in the integration test). A few smaller blocks can also shrink or go: an unreachable guard, a loop that extracts a single map key, repeated TrimSpace calls, and a console cast that discards type checking the generated client already provides. Roughly 60 lines could come out with no behavior change.

## Tally

Code review: 0 findings (1 candidate from the history check, filtered at confidence 25 of 100).

Ponytail: 9 findings. The most impactful one is the inline `attributes` schema in grant-user.yaml, because the duplicate generated type it mints leaks into the Go conversion code and the integration test assertions.

IAM-marco:
While this PR is still open, we want to change the locator-create contract. This is now tracked in #1229; raising it here because changing it pre-merge is free and post-merge it is a breaking change.

What we want instead, on the identifier path only:

- Zero matches returns the same success-shaped response as a created grant. The server silently does nothing and keeps the reason in the logs, like it already does for ambiguous matches.
- `grant.already_exists` is also not distinguishable on this path: same success shape, nothing written twice.
- Granting to yourself (the resolved user equals the caller) is rejected with a visible error.

Why: the current contract ("zero or several resolve as not found") collapses the reasons, but it still tells the caller whether a grant was created, which answers "does this email belong to an active user". The recorded rationale for keeping 409 distinguishable was that a caller who can create grants can also list them. That covers people who already hold a grant, but create-by-identifier also answers the question for addresses that hold no grant at all, which is exactly the enumeration surface the Console flow needs closed: it shows a fixed "if the user exists in our system, they have been granted access to your project" message, and that message is only honest if the API does not leak the difference.

The explicit `user_id` path keeps its current errors. If this PR absorbs the change, #1229 shrinks to a small allowlist cleanup.

IAM-marco (changes requested):
See https://github.com/zitadel/nextgen/pull/1192#issuecomment-5683600373

IAM-marco (approved):
Approving. The two-axis review is fully addressed and CI is green.

- All prior review threads are resolved: the eight simplification points were fixed as suggested, and the one kept item (the handler-side exactly-one check) has a sound rationale (it checks field presence where the service checks trimmed values).
- The anti-enumeration follow-up landed: identifier create returns a success shape on a miss, ambiguity, or duplicate, and a session caller granting themselves is rejected. The residual body-shape leak is tracked in #1229.
- CI is green on this head. Getting there fixed two flaky infra issues and one real bug that the flakes had been masking: the Spanner-emulator lane timeout was raised (its slow authz list test runs in ~8m under the new budget), and two real-instance helpers that still sent the old grant body now send the user locator shape.

Issue #1205, opened 2026-09-10
Reconcile the branding design notes with the shipped appearance contract


bastionstack:
The branding design notes under `docs/design/branding/` describe the appearance surface as a proposed extension and describe a path where Liquid templates emit their own `<style>` block. Both are behind the code: the appearance blocks ship on `POST /branding`, and the orchestrator owns theming through the token bridge, with templates structuring HTML only.

`schema.md` carries most of it. `tokens.md` still opens on the same framing, and its Path A / Path B split no longer describes how a revision reaches `--zl-*`. A reader following either one today would build against a contract that is already settled somewhere else.

Reconcile the notes with the shipped OpenAPI and the bridge, or retire the parts the contract has replaced and point at it.

Issue #1065, opened 2026-08-31
Decide the fate of the split and hero layout chrome stylesheet


bastionstack:
`layout-chrome.css` paints the split and hero page layouts and is the densest consumer of `--zl-*` tokens in the components package. Those layouts now belong to page chrome, which the embedding application owns, so the widget will not grow new page-level splits.

Revisions that already publish them keep rendering, so the file cannot simply be deleted. Decide whether it stays as a frozen legacy render path with a removal horizon, or moves out of the orchestrator shadow root.

Context: the login customization strategy under `docs/design/branding/`.

PR #1206, opened 2026-09-10
feat(login): paint the appearance a branding revision publishes

### LINKED ISSUE #1203: Align the login widget branding with the shipped revision contract
`POST /branding` now persists `theme` with independent light and dark sides, `typography` with a font URL and scale, and `shape` with a radius, density and logo scale. The orchestrator's token bridge and its branding type predate that contract, so a published revision reaches the widget and paints the wrong thing or nothing.

`branding.ts` still declares itself a client-only extension: a top-level `palette`, a `theme` carrying only `dark.palette`, an `assets` block with no counterpart on the resource, and separate heading and mono faces the schema deliberately excludes. `branding-to-tokens.ts` reads that shape.

- [ ] Each side's palette applies to its own theme only, with no cross-fallback from light to dark
- [ ] Resolve the mark from the resolved side; top-level `logo_url` is used only when neither side sets one
- [ ] Theme resolution stops defaulting to dark when a revision states no mode, so an unpublished side is never used
- [ ] `shape.radius` accepts a pixel integer as well as a preset
- [ ] `radius: none` reaches `--zl-radius-xs` and `--zl-radius-sm`, which atoms draw fields, checkboxes and selects with
- [ ] `typography.scale` reaches the text size and leading tokens, not a `--zl-font-scale` nothing reads
- [ ] `shape.logo_scale` multiplies the existing logo height caps
- [ ] The client-only keys are dropped or promoted, and the type is derived from the wire model

Supersedes #1060, #1061 and #1062. Part of #678.

bastionstack:
## Summary

- Each theme side keeps its own palette. A dark key no longer falls through to the light value.
- The mark comes from the resolved side. Top-level `logo_url` applies only when neither side names one.
- A revision that publishes one side renders that side only, whatever the element's `theme` or the operating system asks for.
- `shape.radius` takes a pixel value and scales the whole corner ramp, so `none` reaches fields and checkboxes.
- `shape.logo_scale` multiplies the logo height caps; `typography.scale` reaches the text sizes and their leading.
- `Branding` and its members are now the wire types. `BrandingAssets` and the `assets` block are removed: never rendered, and per-side logos replace them.
- Storybook presets, `branding.example.json` and the affected rows of `tokens.md` move to the shipped shape.

## Validation

- `moon run components:test components:test-browser`
- `moon ci :lint :typecheck :build :test :test-browser`
- `moon run :typecheck`
- `moon run demo-next-e2e:e2e`
- Storybook in Chrome: both sides, per-side marks, logo scale

## Release notes / changeset

- Changeset: `.changeset/branding-widget-paints-the-revision.md` — major on `@zitadel/components`, patch on `@zitadel/server`.

## Notes

- `layout-chrome.css` gains `calc()` wrappers because the card mark is drawn there. That file is #1065.
- Docs are corrected only where they described this mapping. The rest is #1205.

Closes #1203


packages/components/src/index.ts
```diff
@@ -77,12 +77,14 @@ export {
...
   getCurrentStep,
+  publishedSides,
+  resolveLogoUrl,
   type Branding,
-  type BrandingAssets,
   type BrandingAttribution,
   type BrandingPalette,
   type BrandingShape,
   type BrandingTheme,
+  type BrandingThemeSide,
```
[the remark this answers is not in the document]
bastionstack: Reclassified the changeset as major for `@zitadel/components` rather than keeping a compatibility export: `assets` was validated and never rendered, so an alias would preserve a type with nothing behind it. The changeset now names the removed keys and what replaces them.

packages/components/src/orchestrator/branding.ts
```diff
@@ -1,90 +1,86 @@
...
-  muted?: string;
-  border?: string;
-  text?: string;
-  text_muted?: string;
-  link?: string;
-  success?: string;
-  warning?: string;
-  error?: string;
-};
+export type BrandingPalette = CreateFlow201BrandingThemeLightPalette;
```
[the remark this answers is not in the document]
bastionstack: Same as the note on `src/index.ts` — the bump is major now, and the changeset lists `logo_dark` (replaced by the per-side `theme.*.logo_url`), `favicon` and `background_image` (no successor), plus the heading and mono faces the revision does not model.

packages/components/src/orchestrator/branding-to-tokens.ts
```diff
@@ -6,30 +6,50 @@
...
+// The corner ramp a brand's single radius value scales. Ratios come from the
+// design system's own steps rather than being spelled out, so a change to the
+// ramp in Figma moves the branded corners with it.
+const RADIUS_STEPS = ["xs", "sm", "md", "lg", "xl"] as const;
+
+const RADIUS_RATIOS: Record<(typeof RADIUS_STEPS)[number], number> = (() => {
+  const base = remToNumber(tokens.radius.md);
+  const ratios = {} as Record<(typeof RADIUS_STEPS)[number], number>;
+  for (const step of RADIUS_STEPS) {
+    ratios[step] = remToNumber(tokens.radius[step]) / base;
```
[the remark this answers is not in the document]
bastionstack: Not the case here: `tokens` is the resolved tree (`tokens.radius.md === "0.5rem"`), and `cssVars` is the one holding `var(--zl-radius-md)`. The radius tests assert literal lengths and pass, which they could not against `NaNrem`. Added a case asserting no `NaN` reaches the emitted sheet so the mistake would fail loudly if anyone made it.

packages/components/src/orchestrator/templates/layout-chrome.css
```diff
@@ -105,14 +105,16 @@ img[data-zl-asset-broken] {
...
+   is the tenant's `shape.logo_scale`; it multiplies whatever cap is in force so
+   a host override of the cap keeps working underneath it. Bounded the way the
    split pane's mark is (`img.zl-split__compact`): a tenant asset carries no
    intrinsic size the layout can trust, and an SVG with only a viewBox stretches
    to fill whatever box it lands in — which here is the full page width. */
 .zl-card-logo {
   width: auto;
   max-width: min(12rem, 60%);
-  max-height: var(--zl-card-logo-max-height, 2.5rem);
+  max-height: calc(var(--zl-card-logo-max-height, 2.5rem) * var(--zl-logo-scale, 1));
```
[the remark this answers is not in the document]
bastionstack: Fixed, and a bit wider than described: the compact split mark and the hero brand mark both scale now, and the hero height budget subtracts the scaled logo cap rather than the base one — at `logo_scale: 2` the mark grew while the hero still reserved the old height. Hero image caps stay unscaled.

docs/design/branding/tokens.md
```diff
@@ -191,17 +191,16 @@ Asset URLs live on branding; the host loads them (img, link, background). They a
 
 | Source                               | Applied as                                                                         |
 | ------------------------------------ | ---------------------------------------------------------------------------------- |
-| `logo_url` (baseline)                | `<img>` in the default template; `<zl-logo>` atom reads the attribute if bundled   |
+| `logo_url` (baseline)                | Single-mark fallback, used only when neither side names one                        |
 | `hero_url` (baseline)                | `background-image` on `:host` when the `split` layout is active                    |
 | design-system default font           | Loaded by the orchestrator as `<link rel="stylesheet">` (`applyDefaultFont`, default Arimo) so the brand face paints with no branding; dropped when `font_url` is set. See [ADR 025](../../adrs/025-default-brand-font-loading.md) |
 | `font_url` (baseline)                | Tenant override; injected by the orchestrator as `<link rel="stylesheet">` before the widget paints, replacing the default font |
```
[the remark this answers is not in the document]
bastionstack: Renamed to `typography.font_url`, along with the default-font row that referenced it.

packages/components/src/orchestrator/zitadel-login.ts
```diff
@@ -466,6 +467,20 @@ export class ZitadelLogin extends ZitadelSurface {
     return rendered + html;
   }
 
+  /**
+   * The branding a template renders against, with `logo_url` already resolved
+   * to the mark for the active side. Templates read one logo field and get the
+   * right file per surface; the per-side URLs stay on `theme` for a template
+   * that wants to reach them itself.
+   */
+  private brandingForTemplate(): Branding | Record<string, never> {
```
peintnermax, line 476: **Theme flip mid-step now drops typed input**

`brandingForTemplate()` resolves `logo_url` from `themeController.theme`, so with a two-sided revision that has different marks per side the Liquid output is a different string per theme. `ThemeController` is live — it subscribes to `prefers-color-scheme` in `auto` mode and reacts to the element's `theme` property — so a theme change while the visitor is on a step re-runs `render()`, `unsafeHTML` re-parses the subtree, and the form is rebuilt. `applyValuesToFields()` only runs from `updated()` when `response` changed, so whatever was typed is gone (focus too).

Before this change the flip was harmless: the palette lives in CSS under `data-theme` and the single `logo_url` kept the rendered string byte-identical, so `unsafeHTML` was a no-op.

The trigger is narrow (two distinct marks + a live theme change mid-step), but an embedding app with its own light/dark toggle next to the widget is not exotic, and tenants who author two marks are exactly the ones likely to ship one.

Suggested fix: track the theme the last render used and, in `updated()`, call `applyValuesToFields()` when it differs — same as the `response` path. Focus restoration could stay a known limitation. Alternatively, if this is an accepted trade-off, a browser test that flips the theme mid-typing and asserts the field still holds its value would make it a deliberate choice rather than an accident.

packages/components/src/orchestrator/theme-controller.ts
```diff
@@ -73,6 +75,15 @@ export class ThemeController implements ReactiveController {
   }
 
   private refresh(): void {
+    // A revision that publishes one side has no other surface to offer: the
+    // embedding page's `theme` and the visitor's operating-system preference
+    // can both ask for the other one, and there are no colours behind it.
+    const [only, second] = publishedSides(this.branding);
+    if (only && !second) {
```
peintnermax, line 82: **Single-sided revision now overrides the element's `theme` property — worth calling out as a behaviour change**

This short-circuit runs before `explicitMode` is read, so a revision that publishes one side ignores the embedding page's `theme` attribute entirely. The reasoning in the changeset is sound (there are no colours behind the other side), but until now the contract — also in the doc comment on `setModePreference` — was "the element wins", and existing embedders rely on it: `apps/demo-next` hardcodes `theme="light"`, and a tenant who publishes only `dark` will now get a dark widget there.

Could the PR body / migration note state this explicitly, alongside the `BrandingAssets` removal? Something like: "An element `theme` only selects among the sides the revision publishes; with one published side it has no effect." That's the sentence an integrator hitting this will search for.

packages/components/src/orchestrator/branding-to-tokens.spec.ts
```diff
@@ -33,83 +33,179 @@ describe("buildBrandingStylesheet", () => {
...
+        theme: {
+          light: { palette: { background: "#FFFFFF" } },
+          dark: { palette: { background: "#0A0A0A" } },
+        },
+      });
+      expect(css).toContain(':host([data-theme="light"]) {\n  --zl-background: #FFFFFF;');
+      expect(css).toContain(':host([data-theme="dark"]) {\n  --zl-background: #0A0A0A;');
+    });
+
+    it("does not fall a dark key through to the light value", () => {
```
peintnermax, line 72: **This test doesn't cover the cascade it's named after**

It only inspects the isolated `:host([data-theme="dark"])` block. What actually decides whether a key leaks is the "upfront" block on `BRANDING_SELECTOR` (`branding-to-tokens.ts:154`), which stamps the *resolved* side's palette onto both `data-theme` selectors. If `resolvedTheme` is `dark` and the dark palette sets a key the light palette doesn't (say `link`), the light surface inherits the dark value from that block — there's nothing in the light block to override it.

Today that can't happen, but only because `surface.ts` rebuilds the sheet in every `willUpdate` with the current `themeController.theme` right before stamping `data-theme`, so the upfront side always equals the active side. That coupling is invisible from this spec.

Two options:
- Pin it: build with `resolvedTheme: "dark"`, a dark-only key, and assert the light block wins for every key the light side *does* set and that nothing else from dark lands under `[data-theme="light"]` — or a browser test that toggles the theme and reads the computed `--zl-link`.
- Simplify: since the sheet is rebuilt on every theme change anyway, the upfront copy on both attribute selectors is redundant; emitting the resolved side on bare `:host` only (and the per-side blocks under their own attributes) removes the leak path instead of depending on call order.

peintnermax (changes requested):

packages/components/src/orchestrator/zitadel-login.ts
bastionstack: Fixed: values are restored on a theme change as well as a step change, focus left alone. Browser test types, flips, and asserts the field still holds its value — red without the fix.

packages/components/src/orchestrator/theme-controller.ts
bastionstack: Added to the changeset and the `setModePreference` doc comment. `apps/demo-next` is unaffected as it stands: the mock branding publishes no sides, so the element still wins there.

packages/components/src/orchestrator/branding-to-tokens.spec.ts
bastionstack: Took the second option. No palette on the shared block, so there is no path to the other side, and `resolvedTheme` is gone from the builder. Unit test for the dark-only key, browser test flips and reads computed `--zl-link`.

peintnermax (approved):
look good to me now 👍

Issue #528, opened 2026-07-14
Environments


vitorbari:
**An environment is a runtime slot on a project.** A project has several, and each one runs a single configuration release at a time. They exist so that a change can be tried somewhere before it reaches the people using the product, and so that the exact artifact that was tried is the one that moves on.

What an environment *is* beyond that slot is undecided, and four other pieces of work are waiting on the answer.

| Waiting | Because |
|---|---|
| Releases — #529 | A release is deployed to an environment and promoted between them. #529 needs environments to exist and be addressable by name, and nothing more; #534 provides exactly that much. |
| Social login — #851 | An identity provider connection carries a client id and a client secret, and those differ per environment. A release is promoted unchanged ([ADR 035](https://github.com/zitadel/nextgen/blob/main/docs/adrs/035-configuration-environments.md)), so such a value cannot travel inside it — the environment has to hold it. |
| Runtime resolution — #536 | Serving a request means resolving which revision of the configuration applies. That takes the environment as given, so something has to decide which environment the request belongs to first. |
| Authorization — #958 | Whether a grant is made at project scope or per environment depends on whether environments differ in kind at all (#965). |

The open questions are split into research tickets, one per aspect, each landing an ADR. Execution tickets follow as those ADRs are written, and belong here too.

| Aspect | Ticket | Waited on by |
|---|---|---|
| Lifecycle and classes — create, rename, retire, what a project starts with, and whether production and ephemeral environments differ in kind | #965 | #529's environment commands, #958 |
| Data isolation — what is shared between environments and what is not | #966 | nothing yet |
| Variables and secrets — per-environment values a release cannot carry | #967 | #851 |
| Runtime selection — how a request resolves to one environment | #968 | #536 |

#534 is the first of the execution tickets, and runs ahead of the research: it implements the environment as an identity, with read endpoints and a seeded default set, so the rest of the milestone has something to point at while the questions above are still open.

Relates to PR #478

vitorbari:
Relates to #478

IAM-marco:
## What #528 has to decide

The plan is to split this issue into 3 ADRs:

- **051 Environment Data Isolation** (Acceptance Criteria 2)
- **052 Environment Selection at Runtime** (Acceptance Criteria 4)
- **053 Environment Lifecycle and Classes** (Acceptance Criteria 1 and 3)

Below is what each one has to settle. Three items are marked **needs an answer**; the rest carry a recommendation and will be written that way unless someone objects.

[...]

PR #1211, opened 2026-09-11
docs: propose environment lifecycle and classes (ADR 061)

### LINKED ISSUE #965: Environments - Lifecycle and classes

vitorbari:
Decides whether the environment set is fixed or configurable, and which lifecycle operations exist. Closes #965; part of #528.


docs/adrs/061-environment-lifecycle-and-classes.md
```diff
@@ -0,0 +1,434 @@
...
+  [ ] prod
+  [ ] Other — enter a name
+```
+
+`dev` is preselected; at least one selection is required.
+
+Each selected environment then gets an issuer, because a production environment
+is invalid without one. `dev` is prefilled with the detected local development
+origin — the dev-server port the wizard already asks for — and the rest are
+asked for, blank for a non-production environment and required for `prod`:
```
IAM-marco: Not sure if this is a good default, because it creates an exception just for an environment that we decided to call prod.

If a user wants their prod environment to be called productions, this automation won't happen.

I would make the behaviour consistent across environments.

docs/adrs/061-environment-lifecycle-and-classes.md
```diff
@@ -0,0 +1,434 @@
...
+```
+$ zitadel environments apply
+  ~ staging → qa   env_01KX…
+  renamed .zitadel/environments/staging.json → qa.json
+```
+
+#### Why the field, and not the filename
+
+`.zitadel/state.json` keys resources by **file path**, and the resource's handle lives in the file body — that is how flows and schemas already work. Environments follow the same pattern, so a file at a stable path whose `name` changed is an update, not a delete and a create.
+
```
IAM-marco: This part is not clear to me: it shows that both the filename and the `name` field are changing, contrary to what the paragraph title says.

docs/adrs/061-environment-lifecycle-and-classes.md
```diff
@@ -0,0 +1,345 @@
...
+**Deleting requires that the environment is neither `production` nor the last one. The delete is a hard-delete, the row and its deployment records are removed, and the name becomes reusable.**
+
+Removing `.zitadel/environments/<name>.json` and running `zitadel environments apply` deletes the environment, subject to two guards enforced on the server, not only in the CLI:
+
+- **A `production` environment cannot be deleted.** It must first be declared `production: false` in its file and synchronized, and only then removed.
+- **The last environment cannot be deleted.** A project always has at least one.
+
+### 8. Environments can be marked as `production`
+
+**`production: true` rejects localhost issuers and shared-hosting wildcards. The toggle belongs to the environment, not the project.**
```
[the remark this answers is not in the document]
IAM-marco: On the shared hosting wildcard: if we go for auto-generated domain names, isn't this going to be a problem? Or maybe not clear to me what you mean by this.

docs/adrs/061-environment-lifecycle-and-classes.md
```diff
@@ -0,0 +1,434 @@
...
+  `http://[::1]:3000`.
+- Shared-hosting wildcards — `https://*.vercel.app`, `https://*.netlify.app`,
+  `https://*.pages.dev`, and the rest of the server's known-safe suffix list.
+- Custom-domain wildcards — `https://*.preview.acme.com`.
+- No declaration at all: the environment is inert until one is added, and the CLI may infer one from the local dev server.
+
+A **production** environment accepts:
+
+- `https://` origins — the same explicit form, and the one production is expected to use.
+- Custom-domain wildcards — `https://*.acme.com`. Accepted as declared; proving
```
IAM-marco: How can you tell a shared hosting wildcard from a custom domain wildcard?

docs/design/api/system-permission-catalog.md
```diff
@@ -129,6 +129,9 @@ the target catalog for the planned paths in
 | `domain.read` | Read/list `/projects/{id}/domains` | Domain context known to Zitadel; does not imply DNS/TLS infrastructure management. |
 | `domain.write` | Add/update a domain | Create + manage. |
 | `domain.delete` | Remove a domain | |
+| `environment.read` | `GET /environments`, `GET /environments/{name}` | Read/list a project's environments. |
+| `environment.write` | `POST /environments`, `PATCH /environments/{name}` | Create + manage, including renaming and toggling the `production` class ([ADR 061](../../adrs/061-environment-lifecycle-and-classes.md)). |
+| `environment.delete` | `DELETE /environments/{name}` | Server refuses the last environment and a production one regardless of the grant. |
```
IAM-marco, line 134: These new rows are fine, but two other places in this same file still don't know environments exist. The bullet near the top ([L20-L24](https://github.com/zitadel/nextgen/blob/e84eb916e279351e89492b3f7be987eda97805db/docs/design/api/system-permission-catalog.md#L20-L24)) lists which project settings get their own permissions: branding, domains, features, allowed origins, signing keys, webhooks. Environments belong in that list now. And the class table ([L81-L96](https://github.com/zitadel/nextgen/blob/e84eb916e279351e89492b3f7be987eda97805db/docs/design/api/system-permission-catalog.md#L81-L96)) has a row for each of those families but none for environments, even though they fit the same `manage` shape as `domain`, `allowed_origin`, and `webhook` on L91 (write covers create plus update, delete is separate). As it stands, someone reading only the top of the file would conclude environments don't have their own permissions. Suggest adding `environment.*` to the bullet and an `environment` row to the class table.

IAM-marco (changes requested):

docs/adrs/061-environment-lifecycle-and-classes.md
vitorbari: What makes `issuer` required is the `production: true` flag, not the name.

```json
{
  "$schema": "../meta/environment.json",
  "name": "prod", // or production, live, the name does not matter
  "issuer": ["https://acme.com", "https://app.acme.com"],
  "production": true
}
```

However, there is indeed some implicit naming convention on the question the CLI asks. If the user choses `Other — enter a name`, then call it `production`, there is nothing that tells the CLI to generate the environment JSON file with `"production": true`.

I will rework this part.

vitorbari: Please check 2d8e6747b368d7d93f584be7ac0cc59457127883.

I tried to find a balance so that the setup never produces invalid environment files, but also account for different use cases.

docs/adrs/061-environment-lifecycle-and-classes.md
vitorbari: Clarified. Please check: 50d7aa22c575d58cfad3738ede4301bf9aeba04b

docs/adrs/061-environment-lifecycle-and-classes.md
vitorbari: One idea is to use https://pkg.go.dev/golang.org/x/net/publicsuffix#PublicSuffix. 

Anyways, I removed all references to custom-domain vs shared-domain, because it is out of scope for this ADR.

I will create a follow up issue for that.

docs/design/api/system-permission-catalog.md
vitorbari: Fixed

IAM-marco (changes requested):
Merge conflicts

PR #1214, opened 2026-09-11
feat(database): persist deployments and the environment's current one

vitorbari:
This PR adds the storage layer for deployments.

A deployment is the act of activating a release on an environment.

There are two concerns:

1. **Which release an environment currently runs** — solved by adding `current_deployment_id` to the `environments` table (the deployment row it names carries the release).
2. **The history of deployments** — solved with a `deployments` table, an append-only log.

Both are written in one transaction: the insert and the pointer swap succeed or fail together, so an environment always runs exactly the release its pointer says.

The statement layer also guarantees:

- **Idempotent writes** — deploying the release the environment already runs writes nothing and hands back the deployment that made it live (`created=false`).
- **Optimistic locking** — an optional expected current deployment id makes the swap conditional; a stale expectation returns `dep.conflict` carrying the actual current deployment and release.


internal/storage/dialect/spanner/deployment.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+		}
+		if err := deployment.CheckExpectedCurrent(current, expectedCurrentDeploymentID); err != nil {
+			return err
+		}
+		// Idempotent on the running release: deploying what already runs is
+		// answered with the deployment that made it live, and nothing is
+		// written.
+		if current != nil && current.ReleaseID == entity.ReleaseID {
+			*entity = *current
```
wim07101993: This mutates the entity inside a retryable transaction callback, so an aborted attempt can re-INSERT with another deployment's ID. 

Spanner re-runs that closure on ABORTED and the mutation is never undone, so attempt 1 can take the idempotent branch (overwriting entity.ID and entity.Metadata), get aborted by a concurrent pointer swap, and attempt 2 then INSERTs with another deployment's ID, giving UniqueError instead of a new deployment.

Violates AGENTS.md:84. Same lines in postgres/sqlite, latent (no retry there).

(Claude)

internal/storage/dialect/spanner/deployment.go
```diff
@@ -0,0 +1,211 @@
...
+		row, err := tx.ReadRow(ctx, "environments", spanner.Key{entity.ProjectID, entity.EnvironmentID},
+			[]string{"current_deployment_id"})
+		if err != nil {
+			return err
+		}
+		var currentID spanner.NullString
+		if err := row.Columns(&currentID); err != nil {
+			return wrapError(err)
+		}
+		// The current row is read whole because both branches below need it:
```
wim07101993: Missing current deployment row is a hard error at postgres/deployment.go:58, spanner/deployment.go:66, sqlite/deployment.go:65, so a cascade from postgres/migration/sql/000022_deployments.sql:39 wedges the environment forever.

1. environments.current_deployment_id points at dep_A.
2. Someone deletes the release, the ON DELETE CASCADE removes the dep_A row, but the environment pointer still names it.
3. Every later deploy to that environment reads that pointer first, finds no row, and returns NoRowFoundError.

Issue #1003, opened 2026-08-27
Service - IdP connection service and handlers


grvijayan:
## Context

Building on the IdP API definition (#1001) and connection storage (#1002), this ticket implements the core API endpoints for Identity Provider connections. This encompasses input validation, slug uniqueness, immutability rules for identity fields, response formatting, audit events, and authorization boundaries.

## Scope

**1. Create**
*   **Validation:** Validate the payload against `idp-connection.json` prior to storage.
*   **Responses:** Return `409 idp.already_exists` for duplicate slugs. Return `201` with the shared response schema on success.

**2. Revise (Update)**
*   **Validation:** Validate the payload against `idp-connection.json` prior to storage.
*   **Immutability:** Refuse modifications to the slug or core identity fields (`protocol`, `subject_claim`, and `issuer` for OIDC; `token_endpoint` and `userinfo_endpoint` for OAuth2).
    *   *Reasoning:* Identity links key on the connection. Changing the authority would hand existing accounts over to whoever holds the same subjects there.
    *   *Error Response:* Return `400 idp.field_immutable` naming the specific field.
*   **Success:** On a valid update, store and serve the update as a new revision.

**3. Reads (Get & Query)**
*   **Endpoints:** Implement Get by ID, Get by Slug, and Query.
*   **Payload Shape:** Return the stored document nested under the `definition` key, with `revision_id` identifying the revision returned.
*   **Query Capabilities:** Support filtering and sorting exclusively by `slug` and `created_at`. Return `501` for any unsupported filter operations.

**4. Authorization**
*   **Scopes:** Gate endpoints using `idp.read` and `idp.write`.
*   **Errors:** Return `404 idp.not_found` for nonexistent or foreign-project IDs.
*   **Query Gate:** Route queries through the same shared authz gate as other project resource lists:
    *   A caller with a partial grant sees only their permitted rows (an empty `200`, never a `403`).
    *   A caller with no access to the project gets a `404`.
*   **Indexing:** Register connections and revisions in the resource scope index (ADR 033) upon creation so the gate can map their IDs to a project.

**5. Audit Events**
*   **`idp.created`**: Carries `slug`, `protocol`, `template`, `display_name`, and `revision id`.
*   **`idp.updated`**: Carries `revision id` and the changed display fields.
*   **Security Constraint:** The document body and `client_secret_env` must **never** appear in payloads.

**Out of Scope**
[...]

Issue #1002, opened 2026-08-27
Storage - IdP connections and revisions


grvijayan:
## Context

To ensure stability during active processes (like auth attempts or releases), IdP connections are strictly revisioned. Every edit creates a new, immutable copy of the connection document, allowing processes to safely pin a specific version without fear of it changing mid-flight.

*   **Identity links** key on the connection ID, which persists across revisions.
*   **Schemas and flow definitions** reference connections by their project-unique slug.

## Scope

*   **Persistence (all three dialects):**
    *   Each connection is assigned a project-scoped ID and a project-unique slug.
    *   Every edit appends a new revision with its own ID, storing the connection document it is given.
    *   Revisions are immutable and ordered chronologically; the newest revision is always the one actively served.
    *   Concurrent revises: last write wins. Revisions are immutable and auth attempts pin one at the start, so concurrent edits never disrupt an in-flight ceremony.
*   **Statements:**
    *   Implement create, revise (append a new revision), get by id, get by slug, get revision, and list.
    *   *Constraint on get revision:* This is exposed exclusively to the flow engine. An auth attempt pins a revision at the start, and the callback loads that exact revision (Area 3, `state` record). A public API for reading specific revisions is not planned for this iteration.
    *   *List:* Uses ADR 027 cursor-based pagination (ordered by `created_at`). Supports filtering by `slug` and `created_at`, and returns the same fields as get by id.
*   **IDs & Errors:**
    *   Mint connection and revision IDs via the dialect minter, registering prefixes per ADR 047.
    *   Map slug-uniqueness violations to a typed error. The service reads by slug before writing and picks create or revise, so the constraint is a guard against a race, not how a duplicate is detected.
*   **Out of Scope:**
    *   Deletion logic is deferred to #1013.

## Acceptance Criteria

- [ ] Create successfully stores the connection and its first revision atomically; attempting a second create with an identical slug in the same project returns the typed uniqueness error.
- [ ] Revise appends a new revision and leaves all prior revisions unmodified; standard get operations serve this newest revision.
- [ ] Get by id, get by slug, get revision, and list accurately return the stored data.
- [ ] The new list statement is integrated into the shared cursor-pagination drain test, and the new tables are included in the nullability audit.
- [ ] All criteria above pass across `postgres`, `spanner`, and `sqlite`.

## Depends on

- #1001

## Resources

[...]

PR #1217, opened 2026-09-11
feat(idp): define the connection API

### LINKED ISSUE #1001: API - IdP connection endpoints
## Context

No IdP endpoint exists yet. The connection resource, the CLI syncer, the console, and the engine all need the same API. This ticket defines it in OpenAPI and regenerates the clients, so consumers can be built against a stable contract. The handlers, storage, and service follow in their own tickets.

A connection has two ids:
1. **Connection id:** a unique id generated once during connection creation; referenced by identity links.
2. **Revision id:** each edit stores a new immutable revision under the connection with its own id, which are pinned by auth attempts and releases.

This is the model the IdP resource model decides ([Connection Lifecycle](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/1-resource-model.md#connection-lifecycle), [Dependencies](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/1-resource-model.md#dependencies); README scope table, "Connection resource"). It differs from #530 by design: #530 removes `PUT` for flow definitions and correlates revisions by handle, while identity links (#1033) need an id that survives revisions and a slug reuse. Connections keep the stable id and revise under it.

## Scope

*   **Operations (`/idps`, tag "Identity Providers"):**
    *   Five operations: create, query, get by id, get by slug, update (revise).
    *   Delete is deferred to #1013: its contract (success and conflict responses) depends on the deletion semantics decided there, and no 851 journey exercises it. A revision read is deferred until a consumer exists; both are additive for generated clients.
    *   Path parameter `{id}`; reads scope by a `project_id` query parameter.
    *   Every operation wires its generated default error response.
*   **Scopes:**
    *   `idp.read`, `idp.write`, declared in both OAuth2 flow blocks. `idp.delete` follows with #1013.
*   **Error codes (`idp` prefix):**
    *   `idp.not_found`, `idp.already_exists`.
    *   `idp.field_immutable`: a 400 naming the field when a revise changes the slug or an identity field.
    *   Codes are `<prefix>.<code>` via `ResourcePrefix.ErrorCodePrefix`.
*   **Response shape (shared by create, revise, get, get by slug):**
    *   Connection id, the revision id currently served, slug, `created_at`, `updated_at`, and the stored document verbatim under `definition`. No `status` field: the merged design gives a connection no state, and #1013 adds one only if deletion needs it.
    *   Branding's revision response keys on the resource noun, so `definition` is a deliberate departure.
*   **Query (`POST /idps/query`):**
    *   Follows the teams shape: request body with `filter`, `sorting`, `limit`, and `page_token`; cursor paginated per ADR 027 and ADR 031.
    *   Default order `created_at` ascending, as teams and projects default.
    *   Filter and sort fields: `slug`, `created_at`; `id` is the cursor tie-break, as in teams.
    *   Row shape: connection id, served revision id, `slug`, `display_name`, `protocol`, `template`, `created_at`, `updated_at`. No `definition` per row; the console list (#1051) needs only these, and a consumer fetches the document by id or slug.
    *   Unsupported filter operations return 501, as the shared list code does.
    *   No `GET /idps`.
*   **Request body (create and update):**
    *   Follows flow definitions: a wrapper `{project_id, idp}` where `idp` is an `idp-connection.yaml` OpenAPI component mirroring `idp-connection.json`, so clients get typed bodies.
    *   The server still validates every body against `idp-connection.json` before storing; a new test keeps `idp-connection.yaml` and `idp-connection.json` in step.
*   **Clients:**
    *   Regenerate the CLI and console clients.

## Acceptance Criteria

- [ ] The generated CLI and console clients expose the five operations with a typed connection body.

## Depends on

- #1000

## Open Questions

None.

## Resources

*   [ADR 027: Cursor-based Pagination](https://github.com/zitadel/nextgen/blob/main/docs/adrs/027-cursor-based-pagination.md)
*   [ADR 031: OpenAPI Querying](https://github.com/zitadel/nextgen/blob/main/docs/adrs/031-openapi-querying.md)
*   [Area 1: Connection Lifecycle](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/1-resource-model.md#connection-lifecycle) (the two ids)
*   [Area 1: Dependencies](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/1-resource-model.md#dependencies) (get-by-slug, uniqueness, validate-before-store rows)

grvijayan:
## Summary

Closes #1001.

Defines the identity provider connection endpoints in OpenAPI and regenerates the
CLI and console clients.

Five operations:

- **`POST /idps`** creates or revises a connection document. If the slug does not
  already exist, a new connection is created. If a connection with that slug
  already exists, a revision is created. A revision gets a new `revision_id` and
  leaves `id` alone, so identity links keep resolving while releases and auth
  attempts stay pinned to the revision they captured.
- **`POST /idps/query`** pages through a project's connections with a cursor.
  Each row is the same body the get endpoint returns: the connection id, a
  revision id, the slug, the two timestamps, and the connection document that
  revision holds. Each connection appears once, at its newest revision. Filter
  and sort fields are `slug` and `created_at`.
  Config files reference a connection by slug, and the query resolves one with
  `equals` on `slug`, as teams resolve a name, so there is no second read path.
- **`GET /idps/{id}`** reads one connection at its newest revision.
- **`GET /idps/{id}/revisions`** lists every revision of one connection, newest
  first, with the same cursor pagination as the query. The order is fixed, so
  there is no sorting.
- **`GET /idps/revisions/{revision_id}`** reads one revision by the value an
  auth attempt or a release pins. It takes no connection id because an auth
  attempt holds none. Both routes follow ADR 062 (#1245).

The pair `/idps/{id}/revisions` and `/idps/revisions/{revision_id}` trips
redocly's `no-ambiguous-paths` rule, which this repo sets to error. The rule
compares templates and does not know the id format: the single request
`/idps/revisions/revisions` fits both. No minted id is the literal `revisions`,
and the generated router matches the literal before `{id}`, so the entry is
pinned in `.redocly.lint-ignore.yaml` with that reason.

Identity fields are fixed for the life of a connection and a revision that
changes one is rejected: `protocol`, `subject_claim`, and the endpoints that
name the authority (`issuer` for OIDC, `token_endpoint` and
`userinfo_endpoint` for OAuth 2.0). Those values decide which provider account
a stored subject belongs to, so changing one would repoint existing identities
rather than reconfigure them. The slug is not one of them: it selects create
against revise, so no revision can carry a changed one. `subject_claim` is
compared by effective value, since OIDC uses `sub` when the field is absent.
Two error codes: `idp.not_found` and `idp.field_immutable`.

The permission catalog is corrected to match: `idp.write` maps to `POST /idps`
alone, `idp.read` covers the query, the get and the two revision reads, and `idp.delete` is marked not
yet exposed.

## Why there is no PUT or PATCH

A connection is a chain of immutable revisions, so an update creates a revision
rather than editing one in place. `POST /idps` already has to decide between
create and revise from the document's slug, and a separate verb would have to
make the same decision from the same input. There is no 409 for the same
reason: submitting a slug that exists is the normal way to revise.

## The mirror and how it is kept honest

`endpoints/idps/idp-connection.yaml` mirrors `idp-connection.json`, the schema the
server validates a document against, so generated clients get a typed body rather
than an opaque object. That makes two hand-maintained files describing one
contract, so `idp_connection_mirror_test.go` holds them together.

The test parses both files and applies a list of the twelve places they are
allowed to differ, naming each by its exact path. It would be shorter to ignore
whole keywords instead, but then a new `allOf` could appear anywhere and the test
would stay green. Naming the two that exist today means a third one fails.

It then compares the property names, and each property in turn. Per property
rather than whole document, so a failure names the field and prints that subtree
alone instead of both copies of the schema.

Each entry also checks its key is actually present before removing it. So when a
difference goes away upstream, the test fails and names the entry to delete,
rather than passing while the list slowly goes stale.

The twelve, and why each has to differ:

- **The root `allOf` and `oauth2.allOf`.** ogen types the whole enclosing object
  as `any` when either is present. The server still enforces both against the
  schema.
- **`verified_claims`.** An `anyOf` of consts aborts generation with "sum types
  with same names not implemented" and drops the operation from the client
  entirely. The mirror widens it to the two value classes ogen can carry.
- **`format: uri` on the eight endpoint fields.** With it, ogen decodes into
  `url.URL` and discards every pattern on the field, which would take the
  https-only rule and the fragment ban from #1000 with it. Without it all eight
  URL validators are generated and the server keeps the string as sent.

`contains` on the OIDC scopes and `propertyNames` on both
`static_authorize_parameters` are present in both files. ogen ignores them
without error, so there is no reason to drop them from the mirror.

## Validation

- `moon run server:openapi`: valid; one pinned ignore for `no-ambiguous-paths`,
  see above.
- `go generate ./api`: error schemas, OpenAPI errors, ogen.
- `moon run api:generate`: orval; the output is gitignored under `packages/api`.
- `go build ./...`
- `go test ./api/... ./internal/domain/`
- Mutation-checked the parity test five ways, each reverted: rewording one
  description in the mirror, adding a property to the schema, and retiring three
  different deviations. All five go red and name what broke.
- The nine connection documents used as examples were validated against
  `idp-connection.json` with ajv. The revision examples reuse two of them.

## Release notes / changeset

Changeset: `.changeset/idp-connection-api.md`, minor for `@zitadel/api` and
`@zitadel/server`. The server entry is there because the generated server
interface gains the five operations. Until the handlers ship they return 500
with the `internal` code: ogen's not-implemented error is a bare `errors.New`,
so it never matches `domain.Error` and takes the default branch of the mapper.

## Notes

- Error responses list only the boundary codes (`internal`, `req.invalid`,
  `auth.unauthorized`) because the generator infers each operation's error set
  from its handler and there is none yet. They fill in with #1003, which
  covers the handlers for all five operations. The list-revisions statement is
  #1252.
- The revision id prefix appears in examples as `idprev_`. The real prefix is
  registered in #1002.
- The design docs still name a get-by-slug read (`docs/design/idp/README.md`,
  `1-resource-model.md`). They are corrected in a docs PR together with ADR 062
  (#1245), which came out of review.
- `docs/design/idp/schemas/idp-connection.json` duplicates the shipped schema and
  is removed in a follow-up PR. Only one markdown link in area 1 points at it.


api/openapi/endpoints/idps/by_id/methods.yaml
```diff
@@ -0,0 +1,86 @@
...
+      content:
+        application/json:
+          schema:
+            $ref: ../idp-response.yaml
+          examples:
+            revised:
+              summary: A connection that has been revised since it was created.
+              value:
+                id: idp_01KWH3B72K7M7F0N9WD3P2E4YM
+                revision_id: idprev_01KWHF2XY6RN7F0N9WD3P2E4YM
```
wim07101993, line 34: to keep things consistent, maybe we need an ADR on how we version resources.

We already have schemas where each individual revision is a resources descriminated by `objectType`, this pr adds a `revisionId` and with the changes which will come with environments/releases/reployments, more resources will become versioned. I think we need a single way to version everything so that the api is better understandable. 

Don't know how that will influence this pr though...

api/openapi/endpoints/idps/slug/methods.yaml
```diff
@@ -0,0 +1,91 @@
+get:
```
wim07101993: do we need this endpoint if the query endpoint already provides the same functionality?

api/openapi/endpoints/idps/idp-connection.yaml
```diff
@@ -0,0 +1,245 @@
...
+      client_secret:
+        type: string
+        pattern: '^\$\{\{ *[A-Za-z0-9_]+ *\}\}$'
+        description: "A reference to the variable holding the client secret, never the secret itself. A secret may only be referenced as the entire value: text around the placeholder would be rendered into the resolved credential, so it is rejected. The engine resolves it against the environment serving the request."
+      token_endpoint_auth_method:
+        type: string
+        enum:
+          - client_secret_basic
+          - client_secret_post
+        default: client_secret_basic
```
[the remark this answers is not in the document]
grvijayan: Whether the handler passes the raw body or the decoded document to storage is decided in #1003

api/openapi/endpoints/idps/methods.yaml
```diff
@@ -0,0 +1,186 @@
...
+
+    The document is validated against the `idp-connection.json` meta-schema
+    before anything is stored.
+  tags:
+    - Identity Providers
+  security:
+    - oauth2:
+        - idp.write
+  parameters:
+    - $ref: ../../components/parameters/project-id.yaml
```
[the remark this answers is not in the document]
grvijayan: Intentional. The three read operations take project_id as a query parameter, and create takes it the same way, as branding and schemas do. Recorded as a deviation on #1001.

api/openapi/endpoints/idps/query/query-idps-response.yaml
```diff
@@ -0,0 +1,18 @@
+type: object
+description: |
+  Paginated list of identity provider connections.
+required:
+  - idps
+properties:
+  idps:
+    type: array
+    items:
+      $ref: ../idp-response.yaml
```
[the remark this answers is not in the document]
grvijayan: Intentional. Each row is the shared response shape, definition included, so a consumer that lists connections does not fetch each one again. Recorded as a deviation on #1001

api/openapi/endpoints/idps/methods.yaml
```diff
@@ -0,0 +1,221 @@
+post:
+  operationId: createIdp
+  summary: Create or revise an identity provider connection
```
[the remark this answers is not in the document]
grvijayan: Intentional. POST /idps creates when the slug is new and revises when it exists, so there is no separate update operation. Recorded as a deviation on #1001.

.changeset/idp-connection-api.md
```diff
@@ -0,0 +1,11 @@
+---
+"@zitadel/api": minor
+---
```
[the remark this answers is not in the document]
grvijayan: Added in 5264a080.

api/openapi/endpoints/idps/slug/methods.yaml
```diff
@@ -0,0 +1,91 @@
...
+  operationId: getIdpBySlug
+  summary: Get an identity provider connection by slug
+  description: |
+    Returns a connection by slug.
+
+    A slug is unique within a project and fixed for the life of the
+    connection, so this resolves the same connection as `GET /idps/{id}` and
+    returns the same body. It exists because configuration files reference a
+    connection by slug, so the CLI holds a slug rather than an id.
```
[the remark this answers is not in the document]
grvijayan: Fixed in 5264a080.

internal/domain/idp_connection.go
```diff
@@ -0,0 +1,21 @@
+package domain
+
+// PrefixIDPConnection namespaces connection ids ("idp_01KWH3B..."), the id a
+// release pins and an identity link references. Revisions carry their own
+// prefix, registered with the storage layer that allocates them.
```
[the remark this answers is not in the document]
grvijayan: Fixed in 5264a080. The comment now says an identity link references the connection id, and attempts and releases pin the revision id.

api/openapi/endpoints/idps/slug/methods.yaml
grvijayan: that's a good question

according to the initial framing of the ticket, the query endpoint didn't return the definition, so a get-by-slug made sense. Now that the query endpoint also returns the definition, it doesn't make sense to have an explicit endpoint to get-by-slug. Will remove it.

grvijayan: removed in https://github.com/zitadel/nextgen/pull/1217/commits/4bdb5b2ab5c5ea8dac6d3f1d36a974e1d26972ef

api/openapi/endpoints/idps/by_id/methods.yaml
grvijayan: yes, that makes sense.

I see that [ADR 035](https://github.com/zitadel/nextgen/blob/main/docs/adrs/035-configuration-environments.md#releases) already mentions that resources that can change should produce new immutable revisions and cites `schemas` as the example to follow, but there's nothing specific beyond that.

For the revisioned resources so far:
1. Schemas: objectType, revision id
2. Flow definitions: name, revision id
3. IdPs: slug, revision id, a fixed connection id

In case of flow definitions, the name remains the same and the revision ids change, which is similar to IdPs in this PR, but IdPs require the fixed connection id for identity linking.

We can add an amendment to this ADR. What do you think?

wim07101993: As far as I know, schemas don't have a revision id. Same for the flow definitions, they only have an `ID`. I don't see revision ids anywhere but in the release layer, but that is completely different from the resource layer.

grvijayan: I see what you mean. 

So, in schemas and flow definitions the `id` changes with every revision, so it is essentially the revision id.
In case of connections though, there's a need for two ids: a revision id that is pinned by releases and auth attempts (similar to flow definitions, for example) and a unique, stable connection id for identity links. So we can't really define a single way to version everything, because it varies depending on the resource.

To be in line with the other resources (and the ADR), IdPs can define `connection_id` and `id` fields (instead of a `revision_id` field), with `id` serving the same purpose as in other resources. What do you think?

wim07101993: That would be a solution. The other solution is to have and ID for the resource and revision ids on top of that to version them. That would make it easier to correlate resources together. However, that won't play nicely with schemas as they use their url as the schema ID right now. Which is confusing since it isn't necessarily a URL. 

I do not know which one will be easiest to use for a customer

grvijayan: > The other solution is to have and ID for the resource and revision ids on top of that to version them

I would go with `id` and `connection_id`.

The solution you suggest changes how the existing kinds are versioned: schemas and flow definitions have one id per revision and no fixed resource id, so both would have to migrate, and the schema id being a URL is its own problem on top. With `id` per revision, `id` means the same thing for every kind and a release pointer pins `id` everywhere. `connection_id` is just an extra column on connections that only identity links use, in the same way a `slug` or an `objectType` is a per-kind column.

grvijayan: cc: @vitorbari, wanted to check with you regarding revisioning ids.

grvijayan: fyi @IAM-marco as you're working on the next ticket.

grvijayan: during the id/connection_id/revision_id discussions, @vitorbari mentioned that it would be useful to have id/revision id for other resources as well. In that case, should we simply call them `id` and `revision_id`. It would also make it easier to understand the API doc as well.

wim07101993: In that case I come back to my initial comment. Wouldn't it make sense to use the same revision system everywhere? Wouldn't it be easier to understand from a customer perspective?

grvijayan: > Wouldn't it make sense to use the same revision system everywhere? Wouldn't it be easier to understand from a customer perspective?

Yes. If we are all on the same page, I can update the ADR accordingly.

grvijayan: ADR PR: https://github.com/zitadel/nextgen/pull/1245

wim07101993 (approved):

PR #1220, opened 2026-09-14
ci: enforce the single egress-client rule with a go/analysis check

mridang:
## Summary

#1208 gives the server one hardened egress client, but nothing stops the next code path from reaching for `http.Get` on a user-supplied URL: Go has no in-process permission model, and the repo's only Go static analysis is the built-in `go vet` set, which knows nothing about which client you construct. This PR adds that missing guard as a custom `go/analysis` analyzer and makes it a CI gate.

- New `tools/analyzers/egresslint`: reports any outbound HTTP or TCP client built outside `internal/httputil`. It is type-aware rather than text-matching, so it catches every construction path: composite literals, `new`, zero-value declarations, by-value struct fields (named or embedded), defined types over `http.Client` and conversions back, generic type arguments, `http.DefaultClient`/`http.DefaultTransport`, the package-level `http.Get`/`Head`/`Post`/`PostForm`, `net.Dial*`, `tls.Dial*`, `net.Dialer`/`tls.Dialer`, `syscall.Socket`/`Connect`, zero-value clients inside slices, arrays, maps, channels and `make()`, and imports of known third-party dialing libraries (HTTP client wrappers, websocket clients, gRPC), matched by module root so `grpc/codes` and `grpc/status` stay allowed.
- Exemptions are explicit and audited. Test files and generated files are skipped by default (`-egresslint.check-tests`, `-egresslint.check-generated` to include them). A single site is exempted with an `//egress:allow <reason>` directive on the same line or the line above; a directive without a reason exempts nothing and is itself reported at the construction it failed to cover.
- New `tools/analyzers/cmd/analyzers`: a `multichecker` bundle so further analyzers join by adding one import and one line, with flags namespaced per analyzer. It runs standalone or as a `go vet -vettool`.
- New moon task `server:lint` running the bundle over `./...` once for the default build and once per integration build tag (`postgres_integration`, `spanner_integration`, `sqlite_integration`), since tagged files are invisible to the default run and the tags cannot be combined. Added to `server:test`'s dependencies alongside `vet` and picked up by `moon ci :lint` by name. The built-in vet analyzers keep running through the existing `vet` task; passing the bundle as `-vettool` would replace them rather than add to them.
- Three directives on the existing intentional sites: the audit export shipper (operator-configured URL, already excluded by design in #1208), the integration-test harness client, and the Spanner test database helper's gRPC dial to the local emulator.

<details>
<summary>What it deliberately does not catch</summary>

The analyzer inspects the packages it is given and never the bodies of dependencies, so a library that builds its own client internally is only caught when its import path is on the `-egresslint.third-party` list. It also does not model process execution (`curl` via `os/exec`) or DNS-only egress through `net.Resolver`; those are network-policy concerns rather than client-choice ones. The rule's job is to make "I need to fetch a URL" have exactly one obvious, reviewed answer in this codebase. This scope statement is in the package doc.
</details>

## Validation

- `go test ./tools/...`: fixtures cover every construction rule, both directive placements, the missing-reason case, the sanctioned package, generated files, and the test-file flag.
- `go run ./tools/analyzers/cmd/analyzers ./...` over the whole repo, and once more under each integration tag: clean with the three directive-exempted sites and no false positives in `api/generated` or `internal/httputil`. Removing the shipper's directive makes the scan fail with exit status 3.
- Probed against eleven bypass patterns in a throwaway package (type alias, defined type plus conversion, embedding, generics, library-built client, library-performed GET, `os/exec`, raw socket, `net.Resolver`, `net.ListenConfig`, transport wrapper defaulting to `http.DefaultTransport`): eight caught, three documented as out of scope above. The alias case was a real miss in the first cut, since Go 1.23+ represents aliases as a distinct type; fixed with `types.Unalias`.
- `moon task server:lint` and `moon task server:test` resolve with the new dependency. Not yet executed end to end through moon locally, since the task's dependencies build both UI bundles first; the underlying command is the one run above.

## Release notes / changeset

## Notes

- Originally stacked on #1208 because the shipper directive sits on a line that PR adds. #1208 has merged, so this now targets `main` and the branch was rebuilt on it with the same three commits.
- The task is named `lint` rather than after the analyzer so it stays correct once a second analyzer joins the bundle. The directory is `analyzers`, the framework's own word for what the binary contains.
- `go mod tidy` on `main` wants to promote `spf13/pflag` from indirect to direct; that drift predates this PR and is left alone.


tools/analyzers/egresslint/egresslint.go
```diff
@@ -0,0 +1,369 @@
...
+		"Connect": "calls syscall.Connect",
+		"Socket":  "calls syscall.Socket",
+	},
+}
+
+// Analyzer is the egresslint analyzer.
+var Analyzer = &analysis.Analyzer{
+	Name:     "egresslint",
+	Doc:      "reports outbound HTTP or TCP clients constructed outside the hardened egress package (ADR 061)",
+	URL:      "https://github.com/zitadel/nextgen/blob/main/docs/adrs/061-egress-policy.md",
```
IAM-marco: This link 404s: the ADR file is `061-egress-policy-user-injectable-urls.md`, not `061-egress-policy.md`. Worth pinning to the exact name since the analyzer prints this URL in every finding.

tools/analyzers/egresslint/egresslint.go
```diff
@@ -0,0 +1,369 @@
...
+			st.skip = true
+			continue
+		}
+		if !flagCheckGenerated && ast.IsGenerated(f) {
+			st.skip = true
+			continue
+		}
+		for _, cg := range f.Comments {
+			for _, c := range cg.List {
+				if !strings.HasPrefix(c.Text, Directive) {
```
IAM-marco: Prefix matching makes `//egress:allowed ...` count as a directive with reason "ed", silently exempting the line. Requiring the directive to be exactly `//egress:allow` followed by a space or end-of-comment closes that off.

tools/analyzers/egresslint/egresslint.go
```diff
@@ -0,0 +1,369 @@
...
+				if what, ok := forbiddenTypeOf(pass.TypesInfo.TypeOf(f.Type)); ok {
+					report(pass, current.allowLines, f.Pos(), "%s as a by-value struct field outside the hardened egress package", what)
+				}
+			}
+		case *ast.IndexExpr:
+			// zero[http.Client]() — explicit type argument to a generic.
+			checkTypeArgs(pass, current.allowLines, []ast.Expr{n.Index})
+		case *ast.IndexListExpr:
+			checkTypeArgs(pass, current.allowLines, n.Indices)
+		case *ast.ValueSpec:
```
IAM-marco, line 304: The zero-value checks cover `var c http.Client` and by-value struct fields, but `make([]http.Client, 1)`, array types (`var a [2]http.Client`), and map values still yield usable zero-value clients unseen. Contrived, so either teaching the composite-type walk about element types or adding one line to the scope statement works; right now it is in neither.

tools/analyzers/egresslint/egresslint.go
```diff
@@ -0,0 +1,369 @@
...
+	// Directive is the comment prefix that exempts one construction site.
+	Directive = "//egress:allow"
+
+	defaultSanctioned = "github.com/zitadel/nextgen/internal/httputil"
+)
+
+// defaultThirdPartyClients lists import paths of HTTP client libraries that
+// build their own transports and therefore bypass the hardened client. The
+// list is matched by prefix so versioned module paths (/v2) are covered.
+var defaultThirdPartyClients = []string{
```
IAM-marco, line 70: Optional: websocket libraries (`github.com/gorilla/websocket`, `github.com/coder/websocket`) and `google.golang.org/grpc` dial on their own too and are likelier arrivals in this codebase than most of the HTTP wrappers listed. gRPC would presumably become a sanctioned path when it arrives, but landing on this list first means that arrival is a visible decision instead of a silent one.

IAM-marco (approved):
Worth having: this enforces ADR 061 decision 1 (one shared mechanism, no feature-local checks) the same way erroranalysis enforces the error contract, the directive design fails closed (no reason, no exemption), and the repo-wide scan needing exactly two directives matches the ADR's sanctioned exceptions precisely. Let's land it once #1208 merges and this retargets to main.

Four small things inline, none blocking: a broken doc link, two coverage nits worth either fixing or writing into the scope statement, and optional third-party list additions.

tools/analyzers/egresslint/egresslint.go
mridang: Fixed in 54013b061, pinned to the exact filename.

tools/analyzers/egresslint/egresslint.go
mridang: Good catch. The directive is now matched exactly: `//egress:allow` followed by a space or the end of the comment. Fixtures cover `//egress:allowed` and `//egress:allow-ish`, neither of which exempts anything now. 54013b061.

tools/analyzers/egresslint/egresslint.go
mridang: Taught the type walk about slice, array, map and channel elements, plus `make()`. Pointer elements (`[]*http.Client`, `map[K]*http.Client`) stay allowed since they construct nothing. Fixtures for each. 54013b061.

tools/analyzers/egresslint/egresslint.go
mridang: Added gorilla, coder, nhooyr and x/net websocket plus `google.golang.org/grpc`, matched by module root and `/vN` so `grpc/codes` and `grpc/status` still pass. The one real root-gRPC import is the Spanner test database helper, now behind a directive. Doing this exposed that files behind the integration build tags were invisible to the scan, so the lint task now also runs once per tag (53317ba48).

IAM-marco:
Pipeline tests were failing 🫨

Issue #1221, opened 2026-09-14
CLI - CRUD operations on resources


mridang:
The CLI can describe configuration but cannot operate on the resources behind it. There is no way to create a user, list sessions, revoke one, or inspect an audit event without dropping to `curl` — which means rediscovering the auth header, the pagination shape and the error format each time, and keeping that knowledge in shell history rather than in a tool. Agents hit the same wall harder, since they have no way to learn the surface at all.

The CLI should let a developer perform the full lifecycle on the resources the API exposes: create, read, list, update and delete, for users, teams, sessions, grants, projects and events. The operations available per resource follow the API — where an endpoint does not exist, neither does the command.

Requirements:

- Consistent across resources. The same flags, the same output shape, the same errors, whichever resource is addressed, so learning one teaches the rest.
- Usable by scripts and agents, not only humans — machine-readable output and a way to discover what exists without reading help text.
- Safe on destructive operations, and never requiring a credential on the command line.

Configuration resources — schemas, environments, releases, flow definitions, branding — stay declarative: they are authored as files and shipped through a release, so the CLI may read them to show what the server currently holds, but must not write them. Two writers over the same state would fork the source of truth.

Related: partially overlaps #537, which asks for `releases list`. #537 and #540 assume environments are covered by `zitadel status` and exclude an `env list`, which is worth confirming if read-only access to environments lands here.

Issue #538, opened 2026-07-15
CLI - add `releases create` and `deploy` commands


vitorbari:
**Context.** The developer-facing entry point for shipping configuration. Construction and deployment are separate CLI commands so a release can be built once and rolled through multiple environments unchanged — the CI shape "build on merge, walk the same id through dev → staging → prod." Backs [ADR 035 §CLI — deploy](https://github.com/zitadel/nextgen/blob/main/docs/adrs/035-configuration-environments.md#releases-create-and-deploy).

**Description.**

Commands:

- `zitadel releases create` — packages `.zitadel/` and posts to `POST /configuration-releases`. Prints the `release_id`. Does **not** touch any environment.
- `zitadel deploy --release <release_id> --env <name>` — deploys an existing release to an env. No packaging, no construction.
- `zitadel deploy [--env <name>]` (no `--release`) — convenience path: constructs from local, then deploys the returned id to the target env. Primary developer workflow.

`--env` is required in non-interactive mode. In interactive mode the CLI lists environments and prompts. There is no implicit default environment.

**Bundle assembly (releases create / deploy without --release).**

The CLI walks `.zitadel/` and serializes it into the bundle shape accepted by `POST /configuration-releases`:

```json
{
  "audit": {
    "message": "...",
    "git_sha": "...",
    "git_dirty": false
  },
  "schemas": [ /* ... */ ],
  "flows":   [ /* ... */ ]
}
```

Audit metadata is derived from local state:
- `git_sha` — HEAD of the current repo.
- `git_dirty` — `true` when the working tree has uncommitted changes.
- `message` — where the CLI sources this from (git commit subject / `-m` / interactive prompt / config) is deferred to the CLI-ergonomics follow-up. MVP can require `-m` and surface a follow-up TODO.

**Removal preview + confirmation.**

Before creating the deployment, the CLI reads the target env's current release, computes handles present in the current release but absent from the incoming one, and prints the summary. Interactive mode prompts; non-interactive mode requires `--confirm-removals` to proceed.

**Same-content re-run (no-op).**

If the constructed release id matches the target env's current release, the CLI skips the `POST /deployments` call and prints "nothing to do." `--force` overrides to record a new deployment anyway.

**Concurrency check.**

The CLI passes `expected_current_deployment_id` on `POST /deployments` (the deployment id observed when computing the preview). On `409 Conflict`, it refreshes state, recomputes the preview, and re-prompts.

**Sequence (interactive convenience path).**

```mermaid
[...]

PR #1222, opened 2026-09-14
docs: propose CLI resource commands (ADR 064)

mridang:
## Summary

Proposes ADR 064: the decisions behind a CRUD-style resource surface for the CLI (#1221). The implementation is #1210, which currently carries a copy of this ADR; splitting it out so the decisions can be reviewed on their own.

- **Which resources, and which verbs.** Runtime resources (`users`, `teams`, `sessions`, `events`, `grants`, `projects`) get the verbs their endpoints support. Configuration resources (`schemas`, `environments`, `releases`, `flow-definitions`, `branding`) get `list` and `get` only, because they are authored as files and shipped through a release (ADR 035) and a write verb would be a second writer over the same state.
- **One grammar over two transports.** Every list takes `--filter field=operation:value`, whether the endpoint is `POST /<resource>/query` (ADR 031) or a `GET` with query parameters. Each field declares the operations it accepts, so the CLI abstracts the transport without ever inventing a capability: an operation the endpoint cannot honour is refused locally, naming the field.
- **Removal is spelled one way.** `delete` everywhere, with what the server actually did reported beside the id (`deleted`, `revoked`, `deactivated` per ADR 024) rather than renaming the command per resource.
- **Four facts are contract for agents.** The `--json` envelope, exit codes, error codes (ADR 030) and cursor field names (ADR 027) are what agents parse, so changing any of them is breaking even though no type signature moves.
- **Credentials never reach the command line** (ADR 036), enforced in code rather than documented as advice.

The ADR also records six open questions for the API, including the three collections that list through `GET` instead of `POST /query`, and the fact that nothing carries a "this field is sensitive" signal end to end.

## Note on the number

Originally numbered 061, renumbered to 062 after #1207 merged and took that number. `scripts/check-adrs-index.mjs` requires contiguous numbering, so a number cannot be reserved ahead of merge and the same clash can recur: #1119, #1145 and #1211 each still propose an ADR numbered 061. Happy to renumber again if one of them lands first.

## Testing

- `node scripts/check-adrs-index.mjs` passes.
- Every relative link in the ADR resolves on this branch.


docs/adrs/062-cli-resource-commands.md
```diff
@@ -0,0 +1,368 @@
...
+typo suggestions on unknown commands, no network timeout, and no pager. They
+are real, but fixing them here would mean changing commands this surface does
+not own.
+
+## Open questions for the API
+
+These are the places where the CLI surfaced something the API should decide.
+Recorded here because §12 makes them the API's problem, not the CLI's:
+
+- `events`, `schemas` and `flow-definitions` read through `GET` with query
```
vitorbari, line 315: https://github.com/zitadel/nextgen/blob/main/docs/adrs/031-openapi-querying.md

`GET /<resource>` and `POST /<resource>/query`. The `/query` one allows for advanced filtering and sorting. For now, it depends on the use case

docs/adrs/062-cli-resource-commands.md
```diff
@@ -0,0 +1,368 @@
...
+  parameters while every other list uses `POST /<resource>/query`. The CLI
+  hides the difference (§11), but these endpoints accept only `equals`, sort by
+  at most one implicit field, and spell a range as two parameters — so the
+  capability gap is real even where the spelling is not. Either the deviation
+  is intended and should be written down, or these should move.
+- `GET /releases` answered 500 on a project with no releases, on the prebuilt
+  server this branch was tested against. It may already be fixed; it is
+  recorded because the CLI is how it was noticed.
```
vitorbari, line 323: It works on main. This branch probably had an intermediate state of `GET /releases`

docs/adrs/062-cli-resource-commands.md
```diff
@@ -0,0 +1,368 @@
...
+- Three filter operations are advertised by the query contract but answer 501.
+  The CLI offers them because the contract does; today they fail at runtime.
+- `users` cannot be filtered by email, which is the field a human most often
```
vitorbari, line 326: For now, we don't support search based on flexible schema attributes, but it will definitely be needed at some point.

docs/adrs/062-cli-resource-commands.md
```diff
@@ -0,0 +1,368 @@
...
+and default to this API's (ADRs 027 and 031) rather than being written into the
+factory, so "platform-agnostic" is a property a test can demonstrate rather
+than a label. Adding a resource is a table entry; changing a
+convention changes it everywhere at once. Hand-authoring a command file per
+verb is not allowed — that is how surfaces drift, and it is the failure this
+ADR exists to prevent.
+
+### 5. Commands are registered from a table, not discovered from the filesystem
+
+A generated command has no file, so oclif's directory-pattern discovery cannot
```
IAM-marco, line 90: I'm not acquainted with oclif. Is it useful to include this? If so, maybe we should have a vocabulary somewhere (top of the doc?) that explains these terms

docs/adrs/062-cli-resource-commands.md
IAM-marco, line 1: Can you run this through the bro skill? Some parts are hard to understand as a non-native speaker

docs/design/cli/resource-commands.md
```diff
@@ -0,0 +1,329 @@
...
+through a release, so the CLI reads them to show what the server currently
+holds and never writes them here:
+
+| Resource            | Verbs     | Backing endpoints                                    |
+| ------------------- | --------- | ---------------------------------------------------- |
+| `schemas`           | list, get | `GET /schemas`, `GET /schemas/{id}`                  |
+| `environments`      | list, get | `GET /environments`, `GET /environments/{name}`      |
+| `releases`          | list, get | `GET /releases`, `GET /releases/{id}`                |
+| `flow-definitions`  | list, get | `GET /flow_definitions`, `GET /flow_definitions/{id}` |
+| `branding`          | list, get | `GET /branding`, `GET /branding/{id}`                |
```
vitorbari, line 56: `environments`: PR #1211 is introducing `environments apply` based on locally defined environment JSON files. There is no conflict with the read operations added here.

`releases`: similar situation as `environments`, there are plans to introduce `releases create`, which creates a release based on locally defined configuration resources. Different than the other `create` for runtime resources. https://github.com/zitadel/nextgen/issues/538. No conflicts with the read operation added here.

All the other ones, are revisioned resources, so we might need a different approach.

`schemas`, `flow-definitions`, and soon `idps`; these are revisioned with a correlation ID, for example, for schemas, the correction ID is called `objectType`.

The important question is if `<resource> list` should return all revisions, or only the latest revision of each object-type, effectively, it is lists the different object-types. In the console, we chose the latter - the api supports that with the query parameter `revisions=latest OR all`: https://github.com/zitadel/nextgen/blob/b8e846278c0715091f1211d2b5a7482fb09d8ac3/api/openapi/endpoints/schemas/methods.yaml#L80

As a full example:

id (revision) | kind | object-type
-- | -- | --
1 | user-schema | human-user
2 | user-schema | machine-user
3 | user-schema | human-user
4 | user-schema | human-user
5 | user-schema | machine-user

`revisions=all` returns all entries, `revisions=latest` return only  IDs 4 and 5.

The commands could be:

```bash
$ zitadel schemas list

- 4 human-user
- 5 machine-user

(uses the revisions=latest param)
```

```bash
$ zitadel get human-user

(get by object-type, not by ID)

// returns the latest revision contents
```

```bash
$ zitadel get human-user revisions

// returns all revision for "human-user"
```

--

For `branding`, it is slightly different because correlation ID or object type, it is just a list of revisions

Some ideas:

```bash
$ zitadel branding revisions -> all revisions
$ zitadel branding latest -> latest revision
$ zitadel branding get <ID>
```

IAM-marco (changes requested):

docs/adrs/064-cli-resource-commands.md
```diff
@@ -0,0 +1,368 @@
...
+> **Context:** [ADR 007](007-gitops-configuration-surface.md) left runtime resources to server APIs and the [CLI design overview](../design/cli/README.md) reserved "a planned one-shot imperative CLI surface" for them; this ADR is that surface
+> **Relates to:** [ADR 004](004-agent-contract-and-agents-md.md), [ADR 027](027-cursor-based-pagination.md), [ADR 030](030-error-model-mapping-and-reporting.md), [ADR 031](031-openapi-querying.md), [ADR 035](035-configuration-environments.md), [ADR 036](036-api-credential-planes.md)
+
+Reference documentation for the surface lives in
+[`docs/design/cli/resource-commands.md`](../design/cli/resource-commands.md).
+This ADR records the choices behind it and why the alternatives were rejected.
+
+## Decision
+
+### 1. Configuration stays declarative; runtime resources become imperative
```
elinashoko, line 14: A few product-level points I’d like us to review:

* The ADR describes `plan` / `apply` as the configuration path in several places. Since these will be sunset once environments and releases are in place, let’s describe the durable model instead: configuration is authored locally and deployed through releases.

* I’m not convinced that removal should always be called `delete` when that isn’t what the user is actually doing. `sessions delete` is much less intuitive than `sessions revoke`, and `teams delete` suggests permanent removal even though the Team is only deactivated and remains available. Consistent CRUD verbs are useful, but meaningful IAM lifecycle actions such as `revoke` and `deactivate` should remain explicit.

* The declarative versus imperative split makes sense for the resources covered here. I’d just scope it explicitly to the current resource set rather than establishing it as a permanent rule for every future resource. Some future resources, such as customer-managed enterprise SSO connections, may need both a configuration path and an API-driven path depending on who manages them.

* One small clarification on `--json`: if it’s used in an interactive terminal, can the command still prompt? Since JSON is the contract for agents and automation, I’d expect it to always return structured output and never open an interactive prompt, even when the terminal technically supports one.

elinashoko (changes requested):

docs/adrs/062-cli-resource-commands.md
mridang: You're right, and the ADR was wrong to call it a deviation. ADR 031 says it plainly: `GET` is allowed alongside `POST /<resource>/query` and "won't have filter/sort functionality".

I've dropped that open question and rewritten §13. It now says those six lists genuinely have less capability by contract, and that the CLI reports exactly that — fewer operations per field, no `--sort` where the endpoint cannot sort — rather than presenting it as something the API should fix. Whether any of them should gain `/query` is a use-case question, as you say.

docs/adrs/062-cli-resource-commands.md
mridang: Confirmed, and thanks — I built the server from current main to check rather than take it on trust, and `GET /releases?project_id=…` on a project with no releases answers `200 {"releases":[]}`.

The 500 came from the prebuilt `apps/server-darwin-arm64/bin/nextgen` I had been testing against, which predates recent work. I've removed that open question from the ADR.

docs/adrs/062-cli-resource-commands.md
mridang: Thanks — I've reworded it so it reads as a known limitation rather than an oversight: email lives in the user's `attributes`, those are defined by the project's own schema, and search over schema-defined attributes isn't supported yet.

I kept it in the list because email is the field a human most often has in hand, so it is probably the case that forces attribute search when the time comes.

docs/adrs/062-cli-resource-commands.md
mridang: Agreed, oclif is an internal implementation choice and the ADR shouldn't require knowing it. I've rewritten that section without any framework vocabulary: it now says most CLI frameworks find commands by scanning a directory of files, that a generated command has no file so the CLI keeps an explicit list instead, what that costs, and that generated commands keep their internals out of `zitadel commands --json`.

No glossary needed in the end. Those were the only framework terms in the ADR and there are none left. The concrete names (the `COMMANDS` table, the bundle entry) live in `docs/design/cli/resource-commands.md`, which is where an implementer would look.

docs/adrs/062-cli-resource-commands.md
```diff
@@ -0,0 +1,368 @@
...
+`--all` suggests nothing because it already drained. A refused deletion carries
+the same command with `--force`. Handing back a bare `next_page_token` and
+expecting the caller to rebuild the command is how a paging loop silently drops
+a filter on page two.
+
+### 9. `--json` is the contract; human output adapts to the terminal
+
+`--json` emits the standard envelope (`status`, `data`, and on failure `code`,
+`message`, `hint`, `next_commands`) and nothing else, so `jq` is the intended
+consumer and never has to strip a banner.
```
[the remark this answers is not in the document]
mridang: Fixed. The envelope now names `cli_version`, `command` and `source` alongside `status` and `data`, and says it is ADR 004's envelope rather than a shape this surface invents.

docs/design/cli/resource-commands.md
```diff
@@ -0,0 +1,329 @@
...
+- `--all` drains every page in order. It is exclusive with `--page-token`.
+- While a page remains, the envelope also carries `data.next_commands` with a
+  runnable next-page command. A cursor is only valid alongside the sorting and
+  filters that issued it, so the suggestion repeats the whole invocation rather
+  than handing back a bare token; `--all` suggests nothing, having drained
+  everything.
+- The envelope is identical for every resource:
+
+  ```json
+  { "status": "ok", "data": { "items": [ … ], "count": 42, "next_page_token": "…" } }
```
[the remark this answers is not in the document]
mridang: Fixed. The example envelope in the design doc now carries `cli_version`, `command` and `source` too.

docs/adrs/062-cli-resource-commands.md
```diff
@@ -0,0 +1,368 @@
...
+
+The line is between transport and meaning. How a filter travels — a query body
+or a query parameter, one parameter or two — is plumbing, and §11 hides it so
+the surface stays learnable. What a filter *can do* is meaning, and that is
+never invented: an operation the endpoint lacks is refused, a verb it lacks is
+absent, a field it cannot sort by gets no `--sort`.
+
+So `events`, `schemas` and `flow-definitions` reading through `GET` while every
+other list uses `POST /<resource>/query` is smoothed over in the spelling and
+recorded as an open question for the API — three resources, not one, so it is a
```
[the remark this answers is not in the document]
mridang: Fixed. It is six of eleven that read through `GET` (`events`, `schemas`, `environments`, `releases`, `flow-definitions`, `branding`), not three. Reworded, and per vitorbari this is ADR 031's contract rather than a deviation, so the open question is gone.

docs/adrs/062-cli-resource-commands.md
```diff
@@ -0,0 +1,368 @@
...
+  parameters while every other list uses `POST /<resource>/query`. The CLI
+  hides the difference (§11), but these endpoints accept only `equals`, sort by
+  at most one implicit field, and spell a range as two parameters — so the
+  capability gap is real even where the spelling is not. Either the deviation
+  is intended and should be written down, or these should move.
+- `GET /releases` answered 500 on a project with no releases, on the prebuilt
+  server this branch was tested against. It may already be fixed; it is
+  recorded because the CLI is how it was noticed.
+- Three filter operations are advertised by the query contract but answer 501.
+  The CLI offers them because the contract does; today they fail at runtime.
```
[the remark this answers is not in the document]
mridang: Fixed. §11 now states its own limit: the registry can only declare what the contract says a field accepts, so an operation the contract advertises but the server answers 501 for fails at the server, not locally. The 501 bullet says the same thing from the other side.

docs/design/cli/resource-commands.md
```diff
@@ -0,0 +1,329 @@
+# CLI Resource Commands
+
+> **Status:** Shipped — eleven resources; see the table below.
```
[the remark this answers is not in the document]
mridang: Fixed. The design doc now reads `Status: Proposed`, and points at the ADR for the decisions and #1210 for the implementation.

docs/design/cli/resource-commands.md
```diff
@@ -0,0 +1,329 @@
...
+> **Context:** The imperative surface for runtime resources that
+> [README.md](README.md#what-lives-in-zitadel-and-what-doesnt) reserves for
+> data the developer does not own in git. Configuration resources are readable
+> here but are written only through `plan` / `apply`
```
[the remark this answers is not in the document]
mridang: Fixed. The context now describes the durable model: configuration is authored locally and deployed through releases (`deploy`, ADR 035), with `plan` / `apply` named only as what ADR 007 originally specified.

docs/adrs/064-cli-resource-commands.md
```diff
@@ -0,0 +1,368 @@
...
+
+| Resource | Verbs |
+|---|---|
+| `users` | `list` `get` `create` `update` `delete` |
+| `teams` | `list` `get` `create` `update` `delete` |
+| `grants` | `list` `get` `create` `delete` |
+| `projects` | `list` `get` `update` |
+| `sessions` | `list` `get` `delete` |
+| `events` | `list` `get` |
+| `schemas`, `environments`, `releases`, `flow-definitions`, `branding` | `list` `get` |
```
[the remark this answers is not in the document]
mridang: Recorded rather than silently decided, since it is a public command contract. Per vitorbari, #1211 introduces `environments apply`, so the plural spelling is already in use and there is no real clash; the ADR notes the two spellings and asks for one owner's decision.

docs/adrs/007-gitops-configuration-surface.md
```diff
@@ -6,6 +6,8 @@
 
 > **Superseded in part.** ADR 035 (Environment Releases for Configuration Resources) replaces the per-resource `plan` / `apply` orchestration described here with atomic release construction (`POST /configuration-releases`) and a separate deployment step (`POST /environments/{env}/deployments`). `zitadel apply` becomes `zitadel deploy`; `zitadel plan` is removed. The higher-level premise — repo files describe configuration, secrets stay out of VCS — is unchanged.
 
+> **Runtime resources.** [ADR 061](061-cli-resource-commands.md) adds the imperative `zitadel <resource> <verb>` surface for runtime resources this ADR leaves to server APIs. The experimental status of IdP and app management below is unchanged.
```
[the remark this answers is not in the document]
mridang: Fixed. The note in ADR 007 now says ADR 062 *proposes* the surface and is not accepted yet.

docs/adrs/README.md
```diff
@@ -66,3 +66,4 @@ This directory contains architecture decision records (ADRs) for nextgen.
...
+| [061](061-cli-resource-commands.md)              | CLI Resource Commands | Proposed | Runtime resources (`users`, `teams`, `sessions`, `events`, `grants`, `projects`) get imperative `<resource> <verb>` commands generated from one registry by a platform-agnostic factory; configuration resources (`schemas`, `environments`, `releases`, `flow-definitions`, `branding`) get `list` and `get` only, because they are authored as files and shipped through a release (ADR 035) and a write verb would be a second writer. One `--filter field=operation:value` grammar covers every list: the registry declares what each field accepts and how it reaches the wire, so a structured query endpoint and a `GET` list read the same while an operation the endpoint lacks is refused locally. Paging, unpaged collections, and lists that must drain are declared the same way. Verbs exist only where an endpoint does, apart from two bootstrap endpoints `setup` owns. Credentials never reach a command line (ADR 036); the `--json` envelope, exit codes, error codes, and cursor field names are contract for agents (ADR 004). IdP and app management stay experimental under ADR 007. |
```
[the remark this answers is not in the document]
mridang: Fixed. The index summary no longer claims an absolute guarantee: it says credential-looking values are refused on the command line by a name heuristic the ADR records as a stopgap.

docs/design/cli/resource-commands.md
```diff
@@ -0,0 +1,329 @@
...
+zitadel <resource> get     <id>
+zitadel <resource> create  --data '<json>' | --file <path|->
+zitadel <resource> update  <id> --data '<json>' | --file <path|->
+zitadel <resource> delete  <id> [--force]
+```
+
+The verbs are generated from one registry
+([`apps/cli/src/commands/resources.ts`](../../../apps/cli/src/commands/resources.ts))
+by a platform-agnostic factory
+([`apps/cli/src/lib/oclif/crud/`](../../../apps/cli/src/lib/oclif/crud/)).
```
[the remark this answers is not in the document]
mridang: Fixed. Those two paths are plain code spans now rather than links, so nothing points at files that are absent from this branch. You were right that the PR's link-validation claim only covered the ADR's own relative links.

docs/design/cli/resource-commands.md
```diff
@@ -0,0 +1,329 @@
...
+| `crud/types.ts`                           | The registry contract, generic over the platform connection `Ctx`  |
+
+`buildResourceCommands` binds each registry entry to its operation classes and
+returns oclif command classes keyed by id (`users:list`).
+
+## Discovery
+
+`zitadel resources` answers, in one call, what an agent otherwise pieces
+together from eight `--help` invocations: which resources exist, the verbs each
+exposes, what its list can filter and sort on, and the body fields of its
```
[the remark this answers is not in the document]
mridang: Fixed. It is thirty-three commands across eleven resources, not eight.

docs/design/cli/resource-commands.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Open decisions
+
+1. **Bulk delete.** One id per invocation today. A multi-id form should report
+   per id rather than stop at the first failure.
+2. **Filter value typing.** Every value is a string. Numeric or boolean filter
+   fields would need a per-field cast in the registry.
+3. **Config-resource reads.** `flows`, `schemas`, `branding`, `releases`, and
+   `environments` could gain `list` / `get` through the same factory; their
+   write path stays `apply` / `deploy`.
```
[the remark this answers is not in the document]
mridang: Fixed. That open decision is gone, since the five configuration resources are defined as `list` / `get` earlier in the same document; the stale `flows` spelling went with it.

docs/design/cli/resource-commands.md
mridang: Hey Vitor, thanks, this is the useful kind of question. Taking the main one first.

I agree latest should be the default for `schemas list`. It is a small change because the filter already exists, it is only a question of which value we send when you do not say:

```bash
$ zitadel schemas list

- 4 human-user
- 5 machine-user
```

```bash
$ zitadel schemas list --filter revisions=all

- 1 human-user
- 2 machine-user
- 3 human-user
- 4 human-user
- 5 machine-user
```

The history of one schema is close to your `get human-user revisions`:

```bash
$ zitadel schemas list --filter object_type=human-user --filter revisions=all

- 1 human-user
- 3 human-user
- 4 human-user
```

```bash
$ zitadel schemas get 4

(the contents of that revision)
```

One wrinkle worth deciding on. Flow definitions have no `revisions` parameter, they are grouped by `name` instead, so the closest thing today is:

```bash
$ zitadel flow-definitions list --filter name=default-login

(that flow's revisions, newest first)
```

And branding has no grouping at all, it is just a flat list of revisions. So schemas can match the console, but the other two cannot follow the same rule until the API gives them the same switch. If we want one rule across all revisioned resources, that is an API question first.

On `zitadel get human-user`, one thing to flag. Everything in the CLI is `zitadel <resource> <verb>`, so that one would be the only command that reads the other way round. The equivalent that fits would be `zitadel schemas get human-user`, but today `get` takes a revision id and there is no endpoint to fetch by object type, so the CLI would have to list everything and pick the newest itself. If the API can do it, the CLI picks it up like any other resource and we get it for free.

On the branding ideas, we can add custom verbs, the commands are generated so writing one is not much work. What I worry about is ending up with a different vocabulary per resource, where branding has `latest`, schemas has `list --filter`, and people have to remember which is which.

Another idea instead of custom verbs. Typing just the resource name gives you the most useful view. For most resources that is simply the list, since only schemas, flow definitions and branding are versioned:

```bash
$ zitadel users

- user_01M29PW  ada@example.com    active
- user_01M29PX  grace@example.com  active
```

```bash
$ zitadel events

- evt_01M2A9T  user.created  2026-09-16
- evt_01M2A9S  team.created  2026-09-16
```

And for the versioned ones it is the latest of each:

```bash
$ zitadel schemas

- 4 human-user
- 5 machine-user

(latest of each, run `zitadel schemas --help` for the rest)
```

```bash
$ zitadel branding

(the current revision)
```

So the rule stays the same everywhere. The bare name shows what you most likely wanted, and versioning only changes what that means for the three resources that have it. Nothing new to learn per resource.

For this PR I think the only thing that needs settling is the schemas default, which I am happy to change. The rest is worth its own issue if you want to pin down how revisioned resources should look.

docs/adrs/064-cli-resource-commands.md
mridang: Thanks, all four are good points. Taking them in order.

**plan / apply.** Agreed, and changed. The ADR and the design doc now describe the durable model, configuration is authored locally and deployed through releases, with `plan` / `apply` mentioned only as what ADR 007 originally specified.

**Removal naming.** You convinced me, and I have changed it. I had unified everything on `delete` for consistency, but you are right that it tells the user something untrue. I went and checked what `gh` does, and it is strict about the word rather than about uniformity: secrets, keys and labels are deleted, while a pull request is closed, locked or merged, and `gh pr delete` does not exist at all. So the rule is now that `delete` means the resource is gone, and anything that changes state while leaving it readable is named after what it does:

```bash
$ zitadel sessions revoke sess_01M2A9T
Revoked session sess_01M2A9T

$ zitadel teams deactivate team_01M29PW
Deactivated team team_01M29PW

$ zitadel users delete user_01M29PX
Deleted user user_01M29PX
```

Note that `teams deactivate` is new, it used to be `teams delete` even before this change, so your point improved something that was already wrong. The JSON still reports what happened beside the id, so agents read `{ "id": "...", "deactivated": true }` rather than inferring from the verb.

**Scoping the split.** Agreed, and reworded. The ADR now says the declarative versus imperative boundary describes the eleven resources it covers rather than being a law for everything later, and calls out exactly your case, that something like a customer-managed SSO connection could need both paths depending on who owns it, and should be decided on its own merits.

**`--json` and prompts.** It already behaves the way you expect. A command is treated as non-interactive when any of `--non-interactive`, a non-TTY stdout, or `--json` is true, so `--json` on a real terminal never prompts. A destructive verb without `--force` returns a structured error instead:

```bash
$ zitadel teams deactivate team_01M29PW --json
{
  "status": "error",
  "code": "E_VALIDATION",
  "message": "Deactivate requires --force in non-interactive mode",
  "hint": "Pass --force to deactivate team team_01M29PW.",
  "next_commands": ["zitadel teams deactivate team_01M29PW --force"]
}
```

All of this is pushed to both this PR and the implementation PR (#1210).

docs/adrs/064-cli-resource-commands.md
mridang: I do not have that skill set up, so I went at it differently. I measured the ADR instead of rewriting it, and the problem turned out to be concentrated rather than everywhere.

Out of 100 sentences, 18 ran past 35 words and 14 stacked two or more dashes, usually one sentence doing three jobs at once. Those were the hard ones. I split all of them into shorter sentences without changing any of the decisions.

Before:

> `--force` is declared by each command that honours it rather than inherited globally, because what it permits differs — overwrite a managed file, destroy a resource — and a global flag would let a habit formed on the harmless one carry into the destructive one.

After:

> Each command that honours `--force` declares its own, rather than inheriting one. What it permits differs: on `setup` it overwrites a managed file, on a destructive verb it destroys a resource. A single global flag would let a habit formed on the harmless one carry into the destructive one.

It is now 0 sentences over 35 words and 0 with stacked dashes.

I also took out the implementation detail you flagged. The section about how commands are registered with the framework is gone, since that is not a decision anyone needs the ADR for, and it lives in the design doc instead. There is no mention of oclif anywhere in the ADR now.

If specific parts are still hard to follow, point at them and I will fix those directly.

docs/design/cli/resource-commands.md
vitorbari: > Flow definitions have no revisions parameter, they are grouped by name instead, so the closest thing today is

We will add `revisions` to flow-definitions API to make the behavior consistent with the schemas.

> On zitadel get human-user, one thing to flag. Everything in the CLI is zitadel <resource> <verb>, so that one would be the only command that reads the other way round. The equivalent that fits would be zitadel schemas get human-user

My bad, I meant `zitadel schemas get human-user`.

mridang: Thanks, both of those are in now.

`schemas list` sends `revisions=latest`, so it shows the current schemas rather than every revision:

```bash
$ zitadel schemas list

- 4 human-user
- 5 machine-user
```

```bash
$ zitadel schemas list --filter revisions=all

- 1 human-user
- 2 machine-user
- 3 human-user
- 4 human-user
- 5 machine-user
```

And `get` takes the object type as well as a revision id, which is your `schemas get human-user`:

```bash
$ zitadel schemas get human-user

(the current revision of that object type)
```

```bash
$ zitadel schemas get sch_04

(that exact revision)
```

Same for flow definitions by name:

```bash
$ zitadel flow-definitions get default-login

(the newest revision of that flow)
```

One thing to be upfront about. There is no endpoint that fetches by object type, so the CLI does cusrom filtering under the hood, and it decides which kind of reference you gave it from the shape of the argument, `sch_` or a URI means a revision, anything else means an object type. That works but it is a rule we made up, not something the API promised. If we get a real route for it, that logic goes away and it becomes a normal `get` like every other resource. I have written both up as open questions in the ADR, along with adding `revisions` to flow definitions so their list can default to latest the same way. Either way, this works now and was a good suggestion from you.

vitorbari: @mridang ,

We have added `revisions=all|latest` to `GET /flow_definitions`, ref: #1246

mridang: Thanks, that is in now. `flow-definitions list` sends `revisions=latest` the same way `schemas list` does, so a bare list is the current flows rather than every revision of each:

```bash
$ zitadel flow-definitions list

- flowdef_9  default-login  active
```

```bash
$ zitadel flow-definitions list --filter revisions=all

(every revision of every flow)
```

```bash
$ zitadel flow-definitions list --filter name=default-login --filter revisions=all

(that flow's revisions, newest first)
```

I have dropped the open question about the asymmetry from the ADR, since the two behave the same now. The remaining one is fetching by correlation value, so `schemas get human-user` and `flow-definitions get default-login` still resolve with a filtered list under the hood rather than a real route.

vitorbari:
Relates to #1245

After the changes proposed by #1245, the endpoints for revisioned resources will be modified and standardized. The CLI will have to adapt to the new contract (just a heads up).

vitorbari (approved):

PR #1223, opened 2026-09-14
feat(idp): store connections and their revisions

### LINKED ISSUE #1002: Storage - IdP connections and revisions

IAM-marco:
Closes #1002. Stacked on #1217 (the API contract); #1217 does not depend on this PR.

## What this adds

The storage layer for revisioned IdP connections, across postgres, spanner, and sqlite.

- Two tables per dialect: a minimal `idp_connections` row (id, slug, head pointer, timestamps) and immutable `idp_connection_revisions` rows holding the full JSON document. The identity fields live only inside the document; nothing filters on them.
- "Latest" is a head pointer: revise appends a revision and moves `latest_revision_id` in one transaction, so concurrent revises are last-write-wins and every superseded revision stays readable for the auth attempt or release that pinned it.
- Statements: create (connection + first revision + resource-scope row, one transaction), revise, get by id, get by slug, get revision (by project and revision id alone, flow-engine only per the ticket), and an ADR 027 cursor-paginated list that serves each connection with its latest document.
- IDs are dialect-minted per ADR 047: `idp_` for connections, new `idprev_` for revisions; both rows added to the registry (the `idp` row is a backfill, the constant already existed).
- `UNIQUE (project_id, slug)` is the table's only user-reachable unique constraint, because spanner reports empty constraint names: any uniqueness violation therefore maps unambiguously to the new `idp.already_exists` error. This PR ships the domain constructor and generated wire schema; the service mapping lands with the service layer.

<details>
<summary><h3>Design decisions and their reasons</h3></summary>

- **Head pointer instead of ordering** (the json_schemas precedent resolves latest by `created_at` ordering): the contract's `updated_at` means revise touches the parent row anyway, so the pointer is free, reads become a plain join, and a same-timestamp concurrent revise wins by commit order instead of erroring.
- **Minimal connection table**: the immutable-field check is the service's job (it already reads the latest document before choosing create or revise), and the query contract only filters on slug and created_at, so denormalized identity columns would carry no read.
- **Revise mints the revision id unconditionally** rather than through `ensureManagedID`: the entity usually arrives from a get carrying the revision it was read at, and Ensure would keep that value and rewrite the revision instead of appending one.
- **Default list order is ascending** (flowdef and releases default descending): the contract fixes created_at ascending with id tiebreak; a login screen renders connections in the order they were added.
- **No FK on `latest_revision_id`**: the revision row points back at the connection, and closing the loop would make the paired insert order impossible. The revisions table cascades from the connection, which cascades from the project.
</details>

<details>
<summary><h3>Tests</h3></summary>

- Portable `stmttest` suite (all dialects): create atomicity including the resource-scope row, duplicate slug in the same project fails typed while the same slug in another project succeeds, revise appends and the prior revision stays byte-identical, every get miss is `NoRowFoundError` including wrong-project scoping, list is project-scoped and oldest-first and serves the latest document, project delete cascades through to revisions.
- `battleIDPConnections` in the shared cursor drain: every seeded connection has a superseded revision, so a stale row surfacing at a page boundary means the head join and keyset predicate fell out of step. A second drain covers slug order.
- Both tables audited by `TestSchemaNullabilityMatchesDDL` through the join schema.
- A postgres compile-shape test proves the authz list predicate names the join alias `c` rather than the table name, a branch no drain reaches because they all run unrestricted.

Run locally: sqlite and postgres integration lanes green; spanner lane in progress at PR time.
</details>

https://claude.ai/code/session_01RRASjUAtUZ3nrZAJmwyZ96


internal/service/statement.go
```diff
@@ -146,6 +147,33 @@ type ReleaseStatements interface {
...
+	// writes the connection row, that revision, and the resource-scope index
+	// row in one transaction. CreatedAt equals UpdatedAt on a fresh connection.
+	// A slug already taken in the project surfaces as *database.UniqueError.
+	CreateIDPConnection(ctx context.Context, entity *domain.IDPConnection) error
+	// ReviseIDPConnection appends a revision holding entity.Document and moves
+	// the connection's head pointer to it, in one transaction. The revision id
+	// is always freshly minted, so passing back an entity read from a get does
+	// not overwrite the revision it was read from. Concurrent revisions are
+	// last-write-wins on the head; both revisions stay readable by id.
+	ReviseIDPConnection(ctx context.Context, entity *domain.IDPConnection) error
```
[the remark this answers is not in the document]
IAM-marco: Added TestIDPConnectionStatements_ConcurrentReviseLastWriteWins in faac0b8c: two writers revise the same connection concurrently, each on its own entity copy (the transaction callback writes the minted revision back onto the entity, and Spanner replays the callback on an ABORTED retry, so a shared struct would race). Both calls must succeed, all three revisions stay readable by id with their own document, and the head has to name one of the two new revisions and serve its document. Which writer wins is commit order, so the test asserts membership rather than a specific winner. Verified with -race and -count=5 on the sqlite and postgres integration lanes; the spanner lane exercises the retry path.

api/openapi/components/schemas/errors/idp-already_exists.yaml
```diff
@@ -0,0 +1,14 @@
+# Code generated by gen_error_schemas; DO NOT EDIT.
+description: |
+  Identity provider connection: a connection with this slug already exists in the project
```
grvijayan, line 3: While working on the API definition ticket, I realized that we could use `POST /idps` endpoint for both create and revise (similar to flow definitions), so, it doesn't make sense to have `already_exists` as error anymore.

internal/storage/dialect/spanner/migration/sql/000025_idp_connections.sql
```diff
@@ -0,0 +1,70 @@
...
+    ON idp_connections (project_id, created_at, id)
+-- +goose StatementEnd
+-- +goose StatementBegin
+-- One revision of one connection; rows are never updated. See the postgres
+-- migration for why the document is opaque and why there is no project FK here.
+CREATE TABLE idp_connection_revisions (
+    project_id      STRING(MAX) NOT NULL,
+    id              STRING(MAX) NOT NULL,
+    connection_id   STRING(MAX) NOT NULL,
+    document        JSON        NOT NULL,
```
IAM-marco: Spec note: This table stores the revision document without checking that it is a JSON object. The postgres and sqlite versions of the same table both refuse anything that is not an object: see https://github.com/zitadel/nextgen/blob/3f7112ff90835bde77ca447f41c6740ee0aed9f7/internal/storage/dialect/postgres/migration/sql/000022_idp_connections.sql#L41-L43. So the three databases no longer enforce the same rule.

What follows: a caller that stores a bare JSON string or array gets it accepted on Spanner and rejected on the other two, and the difference only shows up after switching databases. The service will validate documents against the API schema before storing, but the table-level guard is exactly the belt the other two dialects chose to keep.

Suggested direction: add the same object check here if Spanner accepts it in a check constraint (JSON_TYPE(document) = 'object'), and if the emulator refuses that, leave a comment in this file saying the skip is deliberate so the next reader does not re-open the question.

Test coverage: no test stores a non-object document today, so nothing would catch the difference.

internal/storage/dialect/postgres/migration/sql/000022_idp_connections.sql
```diff
@@ -0,0 +1,63 @@
...
+    -- connection.
+    , CONSTRAINT fk_idp_connection_revisions_connection
+        FOREIGN KEY (project_id, connection_id)
+        REFERENCES zitadel_nextgen.idp_connections (project_id, id)
+        ON DELETE CASCADE
+);
+
+-- Reading a connection's history walks the child rows by parent.
+CREATE INDEX idx_idp_connection_revisions_connection
+    ON zitadel_nextgen.idp_connection_revisions (project_id, connection_id);
```
IAM-marco: Spec note: The new history endpoint pages a connection's revisions newest first, ordered by the revision's creation time and then its id (see https://github.com/zitadel/nextgen/blob/3f7112ff90835bde77ca447f41c6740ee0aed9f7/internal/storage/idpconnection/list.go#L34-L43), but this index only covers (project_id, connection_id).

Why this spot matters: it is the only index on the revisions table, and the history list walks exactly this table.

What follows: every page of the history list finds the connection's rows through this index and then sorts them without index support. That is harmless while a connection has a handful of revisions, but this is a public paginated endpoint and the sort cost grows with every revise.

Suggested direction: extend this index, and its spanner and sqlite twins, to (project_id, connection_id, created_at, id) so the newest-first walk reads straight off the index; an ascending index serves a descending scan fine.

Test coverage: correctness is covered by the drain test; nothing measures the query plan, which is normal here.

internal/storage/dialect/postgres/idp_connection.go
```diff
@@ -0,0 +1,261 @@
...
+	)
+
+	return &database.ListResult[*domain.IDPConnection]{
+		Items:      items,
+		NextCursor: nextCursor,
+	}, nil
+}
+
+// ListIDPConnectionRevisions implements [service.IDPConnectionStatements].
+func (s idpConnectionStatements) ListIDPConnectionRevisions(ctx context.Context, projectID, connectionID string, page database.Page[domain.IDPConnectionField]) (*database.ListResult[*domain.IDPConnection], error) {
```
IAM-marco: Simplification: This method repeats the whole body of ListIDPConnections just above it (https://github.com/zitadel/nextgen/blob/3f7112ff90835bde77ca447f41c6740ee0aed9f7/internal/storage/dialect/postgres/idp_connection.go#L179-L210): compile, query, collect rows, build the next cursor. The only real differences are the base query and the options.

What follows: two copies of the same plumbing to keep in step in this file, and the same twin exists in the spanner and sqlite files.

Suggested direction: one small private helper taking the base query and the compiled options, with both public list methods reducing to a line or two each. Roughly a hundred lines fall away across the three dialects, counting the matching get change below.

internal/storage/dialect/spanner/idp_connection.go
```diff
@@ -0,0 +1,294 @@
...
+	)
+
+	return &database.ListResult[*domain.IDPConnection]{
+		Items:      items,
+		NextCursor: nextCursor,
+	}, nil
+}
+
+// ListIDPConnectionRevisions implements [service.IDPConnectionStatements].
+func (s idpConnectionStatements) ListIDPConnectionRevisions(ctx context.Context, projectID, connectionID string, page database.Page[domain.IDPConnectionField]) (*database.ListResult[*domain.IDPConnection], error) {
```
IAM-marco: Simplification: Same twin list bodies as the postgres file; one private list helper collapses them. See the comment on the postgres version.

internal/storage/dialect/sqlite/idp_connection.go
```diff
@@ -0,0 +1,270 @@
...
+	)
+
+	return &database.ListResult[*domain.IDPConnection]{
+		Items:      items,
+		NextCursor: nextCursor,
+	}, nil
+}
+
+// ListIDPConnectionRevisions implements [service.IDPConnectionStatements].
+func (s idpConnectionStatements) ListIDPConnectionRevisions(ctx context.Context, projectID, connectionID string, page database.Page[domain.IDPConnectionField]) (*database.ListResult[*domain.IDPConnection], error) {
```
IAM-marco: Simplification: Same twin list bodies as the postgres file; one private list helper collapses them. See the comment on the postgres version.

internal/storage/dialect/postgres/idp_connection.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	entity, err := pgx.CollectExactlyOneRow(rows, scanIDPConnection)
+	if err != nil {
+		return nil, wrapError(err)
+	}
+	return entity, nil
+}
+
+// GetIDPConnectionRevision implements [service.IDPConnectionStatements].
+func (s idpConnectionStatements) GetIDPConnectionRevision(ctx context.Context, projectID, revisionID string) (*domain.IDPConnection, error) {
```
IAM-marco: Simplification: This get hand-writes its statement and row handling while getOne just above (https://github.com/zitadel/nextgen/blob/3f7112ff90835bde77ca447f41c6740ee0aed9f7/internal/storage/dialect/postgres/idp_connection.go#L146-L163) already does the same work through the query compiler, and the storage guide itself prefers the compiled path when a field schema exists.

Suggested direction: let getOne take the base query as a parameter and call it here with the revision-side query and an equality filter on project id and revision id; the hand-written constant then goes away. The join makes the project id on either table equivalent, so the filter can stay on the schema's bound columns.

internal/storage/dialect/sqlite/idp_connection.go
```diff
@@ -0,0 +1,270 @@
...
+	defer rows.Close()
+	entity, err := collectExactlyOneRow(rows, scanIDPConnection)
+	if err != nil {
+		return nil, wrapError(err)
+	}
+	return entity, nil
+}
+
+// GetIDPConnectionRevision implements [service.IDPConnectionStatements].
+func (s idpConnectionStatements) GetIDPConnectionRevision(ctx context.Context, projectID, revisionID string) (*domain.IDPConnection, error) {
```
IAM-marco: Simplification: Same hand-written get next to a getOne that could serve it; see the comment on the postgres version. The spanner file already routes this through its shared query helper, so only these two files carry the extra copy.

internal/storage/stmttest/cursor_battle_test.go
```diff
@@ -679,3 +682,82 @@ func battleReleases(t *testing.T, d dialect) {
...
+	revisions := []domain.IDPConnection{*entity}
+	for i := range 5 {
+		entity.Document = idpConnectionDocument("https://v" + string(rune('a'+i)) + ".example.com")
+		require.NoError(t, d.stmts.ReviseIDPConnection(t.Context(), entity))
+		revisions = append(revisions, *entity)
+	}
+	want := idpRevisionIDsOldestFirst(revisions)
+
+	orderAsc := idpconnection.RevisionsNewestFirst()
+	orderAsc.Direction = database.OrderAsc
```
IAM-marco: Guideline note: This drain builds its ascending order by taking RevisionsNewestFirst and flipping the direction, so a reader sees a name that promises newest-first produce oldest-first. Nothing is wrong functionally, and reusing the column list is the right instinct; this file is also where the next resource's author copies patterns from.

Suggested direction: either give the column list a direction-neutral name with the direction applied by callers, or a one-line comment saying the flip is deliberate because the drain needs both directions. Fine to leave as is if neither feels worth it.

api/openapi/components/schemas/errors/idp-already_exists.yaml
IAM-marco: Agreed, and it is done: the error constructor and this generated schema were removed when the branch was realigned with the merged contract (3f7112ff). Storage now surfaces only the database uniqueness error on the create race, and the service layer decides what to do with it.

internal/storage/idpconnection/list.go
```diff
@@ -0,0 +1,60 @@
...
+
+// RevisionsNewestFirst is the revision-list order: the revision's own
+// created_at DESC, its id DESC. Both columns sit on the revision row, and the
+// id breaks the tie between two revises that landed on the same instant.
+func RevisionsNewestFirst() database.OrderBy[domain.IDPConnectionField] {
+	return database.OrderBy[domain.IDPConnectionField]{
+		Columns: []database.Column[domain.IDPConnectionField]{
+			database.Col(domain.IDPConnectionFieldUpdatedAt),
+			database.Col(domain.IDPConnectionFieldRevisionID),
+		},
```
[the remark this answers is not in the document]
IAM-marco: The race was real under the head-pointer shape. Resolved by conforming to ADR 063 in 3f471b79: nothing records the head anymore, the newest revision is by definition the row with the greatest created_at, the unique (project_id, connection_id, created_at) index makes a same-instant second write fail loudly, and the concurrent test now asserts that the history's first row equals what get-by-id serves.

internal/service/statement.go
```diff
@@ -153,6 +154,46 @@ type ReleaseStatements interface {
 	ListReleases(ctx context.Context, filter *database.ListOptions[domain.ReleaseField]) (*database.ListResult[*domain.Release], error)
 }
 
+// IDPConnectionStatements persists identity provider connections and their
+// revisions. Connections are strictly revisioned: an edit never rewrites a
+// revision, so an auth attempt or a release can pin one and keep reading the
+// configuration it started with.
+type IDPConnectionStatements interface {
+	Statements
```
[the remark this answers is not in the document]
IAM-marco: Deliberate, matching repo precedent: ReleaseStatements and EnvironmentStatements are also absent from the mockgen list, so interfaces reached only through AllStatements get their methods on MockAllStatements and no standalone mock. Happy to add it the day a consumer needs MockIDPConnectionStatements on its own.

internal/storage/dialect/postgres/migration/sql/000022_idp_connections.sql
```diff
@@ -0,0 +1,66 @@
...
+-- Identity links key on the connection id; schemas and flow definitions
+-- reference the slug.
+CREATE TABLE zitadel_nextgen.idp_connections (
+    project_id TEXT COLLATE "C" NOT NULL
+        REFERENCES zitadel_nextgen.projects (id) ON DELETE CASCADE
+    , id TEXT COLLATE "C" NOT NULL CHECK (id <> '')
+    , slug TEXT COLLATE "C" NOT NULL CHECK (slug <> '')
+    -- Head pointer; deliberately no FK: the revision row points back here and
+    -- closing the loop would make the paired insert order-impossible.
+    , latest_revision_id TEXT COLLATE "C" NOT NULL CHECK (latest_revision_id <> '')
```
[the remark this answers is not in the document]
IAM-marco: Right call, thank you. The implementation now conforms in 3f471b79: the latest_revision_id column is gone, revise is a single append of the revision row, the newest revision is resolved by created_at with the ADR's unique tie index, and reads use the same no-newer-sibling shape flow definitions already use.

internal/storage/stmttest/idp_connection_test.go
```diff
@@ -0,0 +1,457 @@
...
+	return ids
+}
+
+func TestIDPConnectionStatements_CreateAndGet(t *testing.T) {
+	forEachDialect(t, func(t *testing.T, d dialect) {
+		projectID := ensureProject(t, d.stmts)
+		slug := "google-" + uniqueSuffix(t)
+		document := idpConnectionDocument("https://accounts.google.com")
+
+		entity := createIDPConnection(t, d.stmts, projectID, slug, document)
```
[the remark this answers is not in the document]
IAM-marco: Added in 3f471b79: a create with a valid JSON document that is not an object now runs in every dialect, the revision check rejects it, and the test asserts the connection, its slug, the minted revision id, and the resource-scope row are all absent afterwards.

internal/service/statement.go
```diff
@@ -153,6 +154,58 @@ type ReleaseStatements interface {
...
 }
 
+// IDPConnectionStatements persists identity provider connections and their
+// revisions. Connections are strictly revisioned: an edit never rewrites a
+// revision, so an auth attempt or a release can pin one and keep reading the
+// configuration it started with.
+//
+// Nothing records which revision is newest (ADR 063 §7). The newest revision of
+// a connection is the one with the greatest creation time among its own, which
+// is what every head read here serves.
```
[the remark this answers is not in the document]
IAM-marco: Both halves are addressed. The PR description now describes the ADR 063 shape (no head pointer, newest by created_at, same-instant tie fails loudly with superseded revisions staying readable), and 44e04cea also adds the portable slug and created_at filter coverage from the review overview. On the changeset note: this PR intentionally ships no customer-facing behavior (no handler exists yet), so the changeset stays with the PR that ships the endpoints.

internal/storage/dialect/postgres/migration/sql/000022_idp_connections.sql
```diff
@@ -0,0 +1,67 @@
+-- +goose Up
+-- An identity provider connection is strictly revisioned: every edit appends a
+-- revision, so an in-flight auth attempt or a release can pin the revision it
+-- started on and keep reading it unchanged. This table carries identity only:
+-- nothing here records which revision is newest, per ADR 063 section 7.
+-- Identity links key on the connection id; schemas and flow definitions
+-- reference the slug.
+CREATE TABLE zitadel_nextgen.idp_connections (
```
wim07101993: why put the connection and revision in separate tables? We were tying to move to a more consistent way of versioning resources (ADR 63). 

That asside, this table does not contain any actual info except for ids as I see it. Can't it be a single table with a constraint on `slug`+`revision_id`?

internal/service/statement.go
```diff
@@ -153,6 +154,58 @@ type ReleaseStatements interface {
...
+	// caller tells that apart from the slug collision CreateIDPConnection
+	// reports by which of the two it called.
+	ReviseIDPConnection(ctx context.Context, entity *domain.IDPConnection) error
+	GetIDPConnectionByID(ctx context.Context, projectID, id string) (*domain.IDPConnection, error)
+	GetIDPConnectionBySlug(ctx context.Context, projectID, slug string) (*domain.IDPConnection, error)
+	// GetIDPConnectionRevision serves the pinned revision rather than the head:
+	// RevisionID is the revision asked for and Document is its document. The
+	// revision id is unique within the project, so the connection it belongs to
+	// is not part of the lookup.
+	GetIDPConnectionRevision(ctx context.Context, projectID, revisionID string) (*domain.IDPConnection, error)
```
adlerhurst: Can we change that to a single GetIDPConnection and use options to filter like in GetTeam for example.

internal/service/statement.go
```diff
@@ -153,6 +154,58 @@ type ReleaseStatements interface {
...
+	// back from the insert, while CreatedAt is left as the caller had it,
+	// because the entity normally arrives from a get that already carries the
+	// connection's birth.
+	//
+	// A connection that does not exist is a *database.NoRowFoundError. Two
+	// revisions of one connection created in the same instant have no newest,
+	// so the second fails with *database.UniqueError rather than one of the two
+	// being picked at random; Spanner reports empty constraint names, so a
+	// caller tells that apart from the slug collision CreateIDPConnection
+	// reports by which of the two it called.
```
adlerhurst: Maybe it's just me but I have more questions when reading the comments of the interface and methods than just reading the method headers. It feels like the comment is too detailed and written in a complex way.

internal/storage/dialect/postgres/idp_connection.go
```diff
@@ -0,0 +1,224 @@
...
+	// connection.
+	createIDPConnectionRevisionStmt = `INSERT INTO zitadel_nextgen.idp_connection_revisions ` +
+		`(project_id, id, connection_id, document) VALUES ($1, $2, $3, $4) RETURNING created_at`
+
+	// One row per revision, carrying the connection's identity alongside it. The
+	// aliases `c` and `r` are the ones idpconnection.Schema qualifies its column
+	// names with, and the trailing r.created_at is the revision's birth, served
+	// as UpdatedAt. A head read narrows this with newestIDPRevision; the history
+	// reads page it as it stands, so a row there stands for the revision it was
+	// read at.
```
adlerhurst: It might be worth to deslop the comments. I know what aliases are used for 😄

adlerhurst (commented):
After I stoped reading the comments I started understanding what the code actually does :D
Nice work. The only thing I would do is to deslop the comments.

IAM-marco:
@adlerhurst I will address your comments but hold the merge: Vitor and Wim are discussing weather we want 1 table (data+revision) or 2 separate ones.

internal/service/statement.go
IAM-marco: Done in 060f90d5. GetIDPConnectionByID and GetIDPConnectionBySlug are now one GetIDPConnection(ctx, filter), same shape as GetTeam. I kept GetIDPConnectionRevision separate on purpose: it reads a pinned revision, possibly an old one, while GetIDPConnection always reads the newest. Folding both into one method would make the filter silently change what the query means.

internal/service/statement.go
IAM-marco: Fair point. The doc comments are trimmed in 060f90d5 to one or two short sentences each: what the method does and which errors it returns. The longer background moved out of the interface.

internal/storage/dialect/postgres/idp_connection.go
IAM-marco: Trimmed in 060f90d5, here and in the other IdP files. Comments now only state what the code cannot show, like why the no-newer-sibling condition exists and why the tie index is unique. The alias explanations and the SQL narration are gone.

mridang:
@IAM-marco i noticed you removed me as a reviewer, please let me know if you need one. Happy to help.

adlerhurst (approved):

adlerhurst:
> @adlerhurst I will address your comments but hold the merge: Vitor and Wim are discussing weather we want 1 table (data+revision) or 2 separate ones.

Okay. My two cents is keep the two tables for the moment. Because we can reference the object (without a specific revision) or a specific revision on other tables as needed.

PR #1224, opened 2026-09-14
feat(config): mirror the API in the editor schemas and add $file references

mridang:
Follow-up to #1196. The generated editor schemas carried `x-dialect` overrides that let a file on disk take shapes the API rejects, and they missed rules the server enforces. This PR removes the overrides, moves every rule JSON Schema can express into the OpenAPI YAML, and replaces the branding-only template key with a generic `$file` reference.

## Overrides removed

- **`sso_providers` as a slug list.** #1156 described it in the editor schema as a list of connection slugs ahead of the engine work, while the API, the Go domain type and stored revisions still take `{id, name, template}` objects. Nothing converts one into the other, so a flow file written in the slug shape passed the editor and then failed `zitadel plan`. The editor schema is back to the object shape. Moving every layer to slugs together is #1031.
- **`on_success: create_user_with_sso`.** Same situation: declared in the editor schema, rejected by the API and the CLI, and not wired in the engine. It is removed until #1037 adds the handler and the API value together.
- **`$schema` on a user property.** The hand-written meta-schema required it to equal the 2020-12 draft URL when present. Nothing sets it, the document is already pinned to 2020-12 at the top level, and the key is still accepted and validated as a URI through the draft `allOf`, so the rule is dropped. The server validates uploaded user schemas against this meta-schema, so `user-property-schema-keyword` is the changeset for that change.

The design doc scaffold in `docs/design/idp/schemas/default-login.scaffold.json` still uses the planned SSO shapes, so its test goes back to the form it had before #1156: it fails against the shipped schema only on those two deltas and passes once they are patched in. The unreleased `idp-config-contracts` changeset no longer announces the removed vocabulary.

## `$file` references replace `liquid_template_file`

- **In the YAML**, a string field is marked `x-local: file`. OpenAPI tooling ignores `x-` keys, so the API contract and the generated clients are unchanged. `liquid_template` in `branding.yaml` is the only field marked today.
- **In the generated editor schema**, a marked field accepts either its value or `{ "$file": "<path>" }`, as `anyOf` its own schema and `#/$defs/LocalFileReference`, and keeps the annotation. The generator no longer supports `x-dialect` and rejects an `x-local` kind it does not know.
- **In the CLI**, the new `lib/local-files` module resolves references generically: before upload every `$file` is replaced by the file's content, paths resolve relative to the file holding them and must stay inside the project, and after a write the server's canonical value goes back into the referenced file while the JSON keeps the reference. The branding syncer and setup use it instead of the hardcoded key.

A branding descriptor now reads `"liquid_template": { "$file": "./login.liquid" }`. `branding eject` and `setup --design` scaffold that form, and a descriptor that still carries `liquid_template_file` fails `zitadel plan` with a hint showing the replacement. The shipped branding README, the CLI docs page, `SKILLS.md` and the design docs use the new form, ADR 040 gains an amendment reflected in its status line and index row, and `branding-file-references` is the changeset.

## Server rules now stated in the YAML

Each rule was tried in isolation against orval, ogen, the Go build and redocly before it went in.

| Rule | How the YAML states it | orval | ogen |
| --- | --- | --- | --- |
| Asset URLs (`logo_url`, `hero_url`, `theme.*.logo_url`, `typography.font_url`) carry no `user:password@` | The https pattern requires the authority to contain no `@` | Regex updated | Regex table updated; URL fields were already validated in Go |
| `typography.font_url` needs `typography.font_family` | `dependentRequired` | Ignored | Ignored |
| A transition never sets both `purpose` and `action` | `not: { required: [purpose, action] }` | Ignored | Ignored |
| A step with `sso_providers` defines `transitions.callback`; a terminal step carries nothing else; a non-terminal step does something | One `not` with an `anyOf` of the forbidden step shapes | Ignored | Ignored |

The step rules are deliberately a single `not`: the same rules written as `allOf` with `if`/`then`/`else` made ogen type the whole step as `any`, which broke the Go build, and made orval rename 18 exported names. Where a generator ignores a rule, the Go validator already enforces it, so the API behaves as before and editors now flag the same files the server rejects.

`no-required-schema-properties-undefined` flags `required` names inside the `not` arms that are defined at the step root, so those pointers are ignore-listed with a note, following the existing `sso-auth-method` entry. That existing entry still pointed at `sso-auth-method.json`, which #1196 replaced with a YAML source, so the two warnings it suppressed had resurfaced on main; the entry now names the YAML file, and redocly is back to zero warnings.

`zitadel plan` also measured asset URLs in characters while the server counts bytes, so a URL with non-ASCII characters could pass plan and fail on apply. The CLI gate now counts UTF-8 bytes, and `editor-schema-structural-rules` is the changeset for these rules.

## Rules that stay in code

JSON Schema cannot express these, so the editor does not flag them. `zitadel plan` and the server both enforce them.

- **Byte limits.** `maxLength` counts characters, while the template cap (131072) and the URL cap (2048) are bytes.
- **Cross-references inside a flow.** Transition targets and `purposes` entries must name existing steps, action names must have matching transitions, step and action names must be unique by value, every step must be reachable, and no step may be trapped.
- **References to the user schema.** Every field must exist in the referenced user schema, `x-auth-methods#…` fields must name an enabled method, required user properties must be collected, and `on_success` must have its credentials collected upstream.
- **Purpose wiring.** An entry step must wire the counter-outcome when a flow serves both login and register, and a re-purposing transition must target that purpose's entry step.

## Verification

Run locally on this branch:

- `go generate` updates exactly the regex table in `oas_cfg_gen.go` and the two editor meta-schemas; orval and ogen succeed, and the Go build and the Go domain, service, api and openapi tests pass.
- redocly reports no warnings, and the changeset and ADR index checks pass.
- The editor meta-schema tests cover every new rule with accepting and rejecting cases, including an `@` in a URL path, empty lists on terminal steps, a callback-only step and a transition with a null `action`; the shipped login flow and every preset still validate.
- The config package (286 tests), the CLI (1142 tests), the generator's 8 fixture tests and the 12 `local-files` unit tests pass, with no new lint warnings and no added lines over 100 characters.


apps/cli/src/lib/local-files.ts
```diff
@@ -0,0 +1,189 @@
...
+    Object.keys(value).length === 1 &&
+    typeof value[FILE_REFERENCE_KEY] === "string"
+  );
+}
+
+/** The absolute path a reference points at; E_VALIDATION when it leaves the project. */
+export function resolveFileReference(context: FileReferenceContext, ref: string): string {
+  const absolute = resolve(context.cwd, context.baseDir, ref);
+  const rel = relative(context.cwd, absolute);
+  if (rel === ".." || rel.startsWith(`..${sep}`) || isAbsolute(rel)) {
```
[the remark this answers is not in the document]
mridang: Agreed, fixed in ff0ef6748. The containment check now applies to the real target: every reference is resolved through symlinks, symlinked parent directories included, and must land inside the real project root. That also covers a file write-back would create, and a symlink that resolves nowhere is refused, since a write through it would land wherever the link names. A symlink that stays inside the project still works. New tests cover a file symlink and a directory symlink pointing outside, a dangling symlink, an in-project symlink, and both inlining and write-back through an outside symlink, which leave the external file untouched.

.changeset/editor-schema-structural-rules.md
```diff
@@ -0,0 +1,7 @@
+---
+"@zitadel/api": minor
+"@zitadel/cli": minor
+"@zitadel/config": minor
```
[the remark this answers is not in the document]
mridang: Added in ff0ef6748. `editor-schema-structural-rules` now lists `@zitadel/server`, and so does `branding-file-references`, which also changes `api/openapi/**` and an embedded schema.

api/openapi/components/flows/flow-definition-step.yaml
```diff
@@ -70,24 +104,6 @@ properties:
...
-        pattern: "^[a-z0-9][a-z0-9_-]*$"
-        maxLength: 64
-      example: [google, github]
-      description: |
-        List of connection slugs offered on this step. Each entry must match
-        the `slug` of a connection in the project; the connection supplies the
-        display name and template at render time. When non-empty, the step
-        must define a `transitions.callback` entry — the engine handles the
-        reserved `sso` action transparently and fires `callback` on this step
-        when the IdP returns.
```
mridang, line 90: @grvijayan quick heads-up, this undoes part of #1156.

The editor schema had `sso_providers` as a list of slugs:

```json
"sso_providers": ["google", "github"]
```

but the API and the engine still expect objects:

```json
"sso_providers": [
  { "id": "google", "name": "Google", "template": "google" }
]
```

Same with `on_success`: the editor allowed `"create_user_with_sso"`, the API only knows `"create_user"`.

So a flow written with slugs looked fine in the editor and then failed on `zitadel plan`. I switched the editor back to what the API accepts. The idea is to move everything to slugs and add `create_user_with_sso` in one go with #1031 and #1037. Your design docs and the scaffold test still describe the slug version.

OK with you?

grvijayan: thank you for the heads-up

yes, we can do those changes together in the schema and the api.
i missed that it'd break plan. sorry about that.

mridang: No problem at all Gayathri. Thsnk you for the eyes. :)

api/openapi/components/flows/flow-definition-step.yaml
```diff
@@ -1,6 +1,40 @@
 type: object
 required: [name]
 additionalProperties: false
+# A step is rejected when it matches any of these shapes.
```
vitorbari, line 4: We can re-write these validations to provide better error message, for example:

Before:

```json
"not": {
  "anyOf": [
    {
      "required": ["complete"],
      "anyOf": [
        { "required": ["fields"],  "properties": { "fields":  { "minItems": 1 } } },
        { "required": ["actions"], "properties": { "actions": { "minItems": 1 } } },
        { "required": ["sso_providers"], "properties": { "sso_providers": { "minItems": 1 } } },
        { "required": ["gates"], "properties": { "gates": { "minProperties": 1 } } },
        { "required": ["transitions"], "properties": { "transitions": { "minProperties": 1 } } }
      ]
    }
  ]
}
```

Image

-- 

After:

```json
"allOf": [
  {
    "if": { "required": ["complete"] },
    "then": {
      "properties": {
        "fields": {
          "not": { "minItems": 1 },
          "errorMessage": "A terminal step (complete is set) must not collect fields."
        },
        "actions": {
          "not": { "minItems": 1 },
          "errorMessage": "A terminal step (complete is set) must not define actions."
        },
        "sso_providers": {
          "not": { "minItems": 1 },
          "errorMessage": "A terminal step (complete is set) must not offer sso_providers."
        },
        "gates": {
          "not": { "minProperties": 1 },
          "errorMessage": "A terminal step (complete is set) must not define gates."
        },
        "transitions": {
          "not": { "minProperties": 1 },
          "errorMessage": "A terminal step (complete is set) must not define transitions."
        }
      }
    }
  }
]
```

Image

mridang: Thanks for the suggestion, the messages would be much nicer. I tried it locally, but it doesn't work with our code generators yet.

As in you r example, the nicer messages need `allOf`, and orval can't handle `allOf `next to normal properties. The step schema turns into `unknown().and(...)` and a bunch of exports get renamed. That still happens on the latest orval. ogen had the same problem, but newer versions fixed it. `errorMessage` also isn't a standard keyword, so redocly rejects it.

Before this PR, the editor didn't check these step rules at all. You only found out when you ran the CLI or hit the server. Now the editor catches the mistake, even if the message is vague, so it's still a step forward.

For now this is as good as it gets. I'll look into fixing this in orval, and once that lands we can switch to if/then with proper messages. Okay? :)

mridang:
@vitorbari I accidentally merged it. I enabled Auto-Merge and it got merged once the checks passed. I didn't know that there was no require-approvals rule. I'm a creature of habit and I often do that on other repos. 🤦🏽 Anyway, let me know if there are still comments here and I can follow up right away. My bad.

vitorbari:
> @vitorbari I accidentally merged it. I enabled Auto-Merge and it got merged once the checks passed. I didn't know that there was no require-approvals rule. I'm a creature of habit and I often do that on other repos. 🤦🏽 Anyway, let me know if there are still comments here and I can follow up right away. My bad.

It is all good. I tested locally, and it is working fine. If we can improve the error message later, that would be great.

PR #1142, opened 2026-09-03
feat(api): accept Console session as user principal on grant APIs

adlerhurst:
## Summary

**TL;DR:** Grant create, get, query, and revoke, plus `POST /users/query`, accept a Console `__nextgen_session` cookie as the signed-in **user**. Authorization is `project.viewer` / `editor` / `admin` on the *target* project for grants (home may differ). User query is bound to the credential's home project. Project secrets still need `project.write`. CSRF is not this PR.

First slice of [#1022](https://github.com/zitadel/nextgen/issues/1022) (ADR 053 §5), now including `POST /grants/query` from [#1118](https://github.com/zitadel/nextgen/pull/1118) and `POST /users/query` so Console Settings → Admins can create (people picker), list, and remove. Stacked on [#1117](https://github.com/zitadel/nextgen/pull/1117) — retarget to `main` after that merges.

This revision addresses Copilot review comments: keep oauth2 `ActorContext` when both credentials are present; skip anonymous building sessions on grant/user-query (`ErrSkipServerSecurity` → 401, not 404); unmarshal revoke events as `AuthzRevokedPayload`. CSRF, QueryGrants expand skip, and ADR 053 §8 authorization metadata stay out of this slice (#1140 / locked product decisions).

Wim's empty-session-scope note is expected, not a 403: `Session.Token()` never mints `Scope`; `credentialCeiling` skips `project.write` for `PrincipalType=user`; Check uses the user principal. Comments landed on those two sites. The thread is left open for confirmation. We are not minting scopes onto session tokens.

### Behavior

- OpenAPI: grant create/get/delete/query and **user query** accept `oauth2` **or** `nextgenSession`.
- `HandleNextgenSession` mints `ScopeContext` for user-bound sessions (`PrincipalType=user`, home = session project), and does **not** replace an already-minted oauth2 `ScopeContext` or `ActorContext` (console e2e-real sends both the Vite-injected secret and the session cookie).
- Anonymous building sessions Skip grant/user-query `nextgenSession` so cookie-only callers 401; `/sessions/me` still accepts them. Dual-scheme oauth2 plus a leftover building cookie still succeeds.
- `credentialCeiling` fails closed on empty home, then skips `project.write` for users. Secrets still need write. Preview stays 403.
- Check stays `project.viewer` / `editor` / `admin` (not `grant.*`). Query uses the same `requireProjectAccess(..., grantAccess, opRead)` as Get.
[...]

PR #1225, opened 2026-09-15
feat: sign in to the local console as a CLI-created admin

mridang:
## Summary

Local servers now come with a local admin, so a developer exists on their own server from the first `zitadel start`, signs in to the console with a link instead of a password, and every local project is owned from the start. Claiming an anonymous project stays a Zitadel Cloud journey.

**CLI**

- `zitadel start` boots the server with the platform project (`NEXTGEN_PLATFORM_BOOTSTRAP_PROJECT=true`) and a local admin, on both the binary and the Docker runtime. The admin signs in as `admin@zitadel.localhost`; a generated password is kept in `.zitadel/local/admin.json` (gitignored with the rest of `.zitadel/local/`) and never printed. The server receives only a PBKDF2 hash, through the existing `--user-file` bootstrap import.
- Once the server is healthy, `start` signs the admin in through the auth-attempt API (identifier, then password, then handoff — see **Server** below) and prints a one-time console sign-in link (`/ui/console/login?handoff=…`), also returned as `data.console.sign_in_url`. If no link can be minted, `start` still succeeds and reports why in `data.console.error`.
- New `zitadel console` mints a fresh one-time link and opens it in interactive runs (`--no-open` to only print it).
- `zitadel setup --server local` attaches the new project to the local admin's team through the existing claim endpoints and records `team_id`/`claimed_at` in `.zitadel/secret`. `zitadel claim` then reports the project as already owned, and setup no longer suggests claiming.

**Console**

- `/login?handoff=<token>` exchanges the token for the session cookie before deciding whether to show the sign-in widget — only when the console is served from loopback (`localhost`, `127.0.0.0/8`, `[::1]`). The link is a bearer credential, so honouring it on a deployed console would let a crafted link sign a visitor into someone else's session.

**Server**

- Users imported with `--user-file` register unique attributes from their user schema's `x-unique` annotations, the same rule user creation applies. Previously only `username` was registered, so a bootstrap user on a schema that identifies users by email (the default schema) could not be found at sign-in and the login flow fell through to registration.
- The auth-attempt REST API could not identify anyone: an identifier proof carried a value but no way to say which attribute it was, so the lookup ran against attribute `""` and matched nobody. `IdentifierProof` gains a required `attribute_name`. This is what lets the CLI authenticate without driving the login flow. Previously opened separately as #1256, which has the full write-up and a worked example of the call sequence.
- Completed-factor payloads declare the `method` they are discriminated on. The server always sent it but the schema forbade it, so a generated client could not decode a completed password factor. Nothing on the wire changed.
- `NEXTGEN_PLATFORM_BOOTSTRAP_PROJECT=true` pins the deployment's default project to `proj_platform`, which moves the console's `console_project_id` and the hosted-login default. Harnesses that seed users into their own project opt out with `false`: both console e2e lanes, `console:dev-real` outside claim mode, and `demo-next-e2e`.

## Validation

- `moon run cli:lint cli:test`: 0 lint errors, 118 test files pass. New: `tests/unit/lib/local-server/admin.test.ts` (credential files are owner-only, the stored PBKDF2 hash recomputes, the credential is stable across starts) and `tests/unit/commands/console.test.ts`. Updated start, Docker and root-help tests for the new next command, the platform env var, and server arguments after the image.
- `go test -tags postgres_integration ./internal/bootstrap/users/` against Postgres: all pass, including the new `TestImport_uniqueAttributesFollowSchema` and the `TestBuildCreateAttributes_*` unit tests.
- `moon run workspace:journey -- --framework next` locally: 5 passed. `claim.spec.ts` now covers the local journey: the prepared project is owned, `claim` skips as already owned, and the `console` link signs the admin in.
- Manually, in fresh app directories on both runtimes (binary on a locally built server, Docker on a locally built image):

```
$ zitadel start
Console: signed in as admin@zitadel.localhost. Open this link (works once):
http://localhost:8080/ui/console/login?handoff=handoff_…
$ zitadel setup --server local --framework next
✔ Project owned by admin@zitadel.localhost (team team_localadmin)
$ zitadel claim
ℹ This project already belongs to team team_localadmin.
$ zitadel console --no-open
Signed in to the local console as admin@zitadel.localhost:
http://localhost:8080/ui/console/login?handoff=handoff_…
```

  In the browser, a link signs in once; reusing it after signing out shows the sign-in widget, and a fresh `zitadel console` link signs in again.

## Release notes / changeset

- `.changeset/local-admin-console-sign-in.md`: minor for `@zitadel/cli` (local admin on `start`, console sign-in link, `zitadel console`, local `setup` creates owned projects) and patch for `@zitadel/server` (bootstrap import honours `x-unique`).
- `.changeset/auth-attempt-identifier-proof.md`: **major** for `@zitadel/server` and `@zitadel/api`. `attribute_name` and the payload `method` are required, so a client generated from the old spec must be regenerated. Neither shape could have had a working consumer.

## Notes

- The console shows no data after signing in yet: management endpoints do not accept the console session until #1022 lands (the grants slice shipped in #1142). The sign-in itself works.
- The admin is `user_localadmin` in `team_localadmin`, and its user document names the default user schema under the server's default `schema.builtin_public_base`, which CLI-managed servers do not override.
- A local data directory created before this change has no local admin; `start` reports that and points at `zitadel reset --force`.
- **Open question for an auth reviewer.** The server takes the identifying attribute from the caller. That is more permissive than the login flow, which raises an identifier challenge only for the schema's designated `x-identifier` ([ADR 058 §5](https://github.com/zitadel/nextgen/blob/main/docs/adrs/058-user-identity-designation-and-references.md) retired the "any `x-unique` property can identify" rule). The alternative is for the server to read the designation itself and ignore the caller — which removes the widening and the breaking change, at the cost of wiring a schema lookup into the attempt service. The CLI works either way.


apps/cli/src/lib/claim-state.ts
```diff
@@ -38,6 +38,37 @@ function formatDeadline(deadline: Date): string {
...
+  }
+  try {
+    const res = await fetch(new URL("/console/runtime.json", server), {
+      signal: AbortSignal.timeout(timeoutMs),
+    });
+    if (!res.ok) {
+      return false;
+    }
+    const doc = (await res.json()) as { console_project_id?: unknown };
+    return doc.console_project_id === "proj_platform";
```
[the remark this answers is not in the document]
mridang: The `claim` gate this was reviewing is gone: the PR now boots every CLI-managed server with the platform project and a local admin instead of refusing `claim`, so `claim-state.ts` is no longer part of the change. The same concern applied to the new local `setup`, which uses the runtime document to decide whether to attach the project to the local admin: a deployment that pins `proj_platform` without the platform bootstrap would make that attach fail after setup had already written the app files. Fixed in 277093f64: attaching is now best-effort, so on failure setup warns, leaves the project unclaimed and falls back to the usual claim nudge. Covered by the new `keeps the project unclaimed when attaching it to the local admin fails` test.

apps/cli/SKILLS.md
```diff
@@ -216,6 +216,11 @@ the CLI's help layer, not the envelope.
   fresh link does **not** help — only a fresh `setup` yields a claimable
   project (the old one can no longer be claimed; it stays temporary and its
   data may be lost).
+  A claim needs a server that hosts the platform plane: Zitadel Cloud always
+  does, and a local or self-hosted server only when it was started with
+  `platform.bootstrap_project` (`NEXTGEN_PLATFORM_BOOTSTRAP_PROJECT=true`).
+  Against any other server the command exits `E_VALIDATION` before minting a
+  link, instead of opening a claim page that cannot complete.
```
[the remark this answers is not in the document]
mridang: No longer applicable: the refusal condition was removed along with the `claim` gate, so `SKILLS.md` no longer documents it and `claim` keeps its existing behaviour. The generated `README.md` was regenerated with `moon run cli:readme` and now also documents the new `zitadel console` command.

apps/cli/src/lib/local-server/admin.ts
```diff
@@ -0,0 +1,310 @@
...
+ */
+export async function ensureLocalAdmin(
+  cwd: string,
+): Promise<{ admin: LocalAdmin; userFile: string }> {
+  await mkdir(join(cwd, LOCAL_RUNTIME_DIR), { recursive: true, mode: 0o700 });
+  let admin = await readLocalAdmin(cwd);
+  if (!admin) {
+    admin = {
+      // A fixed, clearly-local identity rather than anything inferred from
+      // Git config, which can carry a stale, work, or noreply address.
```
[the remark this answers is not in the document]
mridang: Fixed in 5e748bfb1. `admin.json` is now created with an exclusive `wx` write, and a run that loses the race re-reads the winner's credential instead of minting its own, so the bootstrap document always carries the hash of the password the CLI kept. Covered by a unit test that runs two `ensureLocalAdmin` calls concurrently.

apps/cli/src/lib/local-server/admin.ts
```diff
@@ -0,0 +1,310 @@
...
+  const init = await fetchJson(`${serverUrl}/projects/${project}/claim/init?project_id=${project}`, {
+    method: "POST",
+    headers: { authorization: `Bearer ${projectSecret}`, origin },
+  });
+  if (init.status === 409 && isObject(init.body) && isObject(init.body.details)) {
+    const teamId = init.body.details.team_id;
+    if (typeof teamId === "string") {
+      return { team_id: teamId, claimed_at: new Date().toISOString() };
+    }
+  }
```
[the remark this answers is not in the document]
mridang: Handled one layer up, in 5e748bfb1. `claimProjectAsAdmin` reports the owning team as the API returns it, and the 409 branch deliberately returns no `claimed_at` because the response does not carry one. `setup` then writes `.zitadel/secret` only when the record is complete and the team is the local admin's; otherwise it logs that the project belongs to another team and leaves the local record alone, so the claim nudge stays.

internal/bootstrap/users/import.go
```diff
@@ -124,6 +136,39 @@ func buildCreateAttributes(attrs map[domain.AttributeKey]json.RawMessage) (domai
...
+	scope, _ := maputil.GetNested[string](schema, []string{"properties", string(key), domain.SchemaAnnotationUnique})
+	switch scope {
+	case domain.SchemaUniqueScopeProject:
+		return domain.AttributeUniquenessProject
+	case domain.SchemaUniqueScopeTeam:
+		return domain.AttributeUniquenessTeam
+	default:
+		return domain.AttributeUniquenessUnspecified
+	}
+}
```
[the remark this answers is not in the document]
mridang: Fixed in 5e748bfb1. The lookup now walks `key.Nodes()`, building the `properties/<node>` path at each level exactly as `CreateAttributesFromMap` does, so `profile.email` resolves through the nested schema. There is a nested-schema regression test, plus one for a team-scoped attribute imported without a team.

apps/cli-journey-e2e/src/claim.spec.ts
```diff
@@ -1,146 +1,104 @@
...
+  test.skip(metadata.framework !== "next", "local ownership is framework-independent; next lane only");
   test.skip(
     process.env.JOURNEY_PREEXISTING_APP === "1",
-    "the fresh-app lane already proves claim; keep the preexisting lane lean",
+    "the fresh-app lane already proves local ownership; keep the preexisting lane lean",
   );
   test.skip(
     metadata.runtime === "docker",
-    "the harness bootstrap env does not reach the docker container, so claim completion 401s there",
+    "the binary lane proves local ownership; the docker lane keeps to the scaffold journey",
```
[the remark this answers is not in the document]
mridang: Not in this PR. The Docker path was verified by hand end to end — the mount, the initial import, the sign-in link, and a second link from `zitadel console` — and the unit test pins the argv, the mount and the env the runtime passes. A container-backed journey lane is a real gap, but adding one is its own change: the journey harness has no Docker runtime today, and standing one up here would grow a CLI fix into an e2e infrastructure change.

mridang:
## How the local admin and claiming work: a full local run

A full local run of this branch at `6a8a3d02a`, after the ADR 058 identifier change, with the server, console and CLI all built from that commit. Two apps: `app-a` starts the server and holds the local admin; `app-b` is a second app set up against the same server. Port 7970 was used because 8080 was taken on the machine. The full output is in the collapsible block below.

### What happens

1. **`zitadel start`** boots the server with the platform project and creates the local admin, `admin@zitadel.localhost`. The CLI keeps a generated password in `.zitadel/local/admin.json`; the server only ever receives a PBKDF2 hash of it, through `--user-file`. Once the server is healthy, `start` signs the admin in through the auth-attempt API and prints a one-time console link. The identifier proof carries only the admin's email, and the server resolves it through the platform schema's designated identifier (`email`), as ADR 058 §5 defines.
2. **`zitadel setup` in `app-a`** creates the project and claims it for the admin straight away: `✔ Project owned by admin@zitadel.localhost (team team_localadmin)`. `.zitadel/secret` records `team_id` and `claimed_at`.
3. **`zitadel claim` on that project** skips with `already-claimed`.
4. **`zitadel setup` in `app-b`**, a folder without a local admin, leaves the project unclaimed and shows the usual "claim your project" prompt.
5. **`zitadel claim` in `app-b`** prints a claim link and waits. Opened in a browser already signed in as the admin through a `zitadel console` link, the page said *Project claimed*, and the waiting command finished with `✔ Project claimed`.
6. **After that claim**, `zitadel claim` in `app-b` skips as `already-claimed`, reopening the used claim link shows *Already claimed*, and the console's project picker lists both `app-a` and `app-b`.
7. **Each sign-in link works once.** The console's `POST /sessions/exchange` returned `200` the first time. After logging out, reopening the same link got `400`, and the console fell back to the sign-in widget.

### Known, not addressed here

- After signing in, the console shows *Console API not authorized*. Its management endpoints don't accept a console session until #1022 lands. Sign-in itself works.

Sign-in tokens and project secrets are redacted below, and so is the claim challenge id.

<details>
<summary>Full output (11 steps)</summary>

````text
Zitadel local claim run — #1225 at 6a8a3d02a (after the ADR 058 identifier change)
Server, console and CLI built from this commit. Port 7970 (8080 is in use by another process on this machine). Two apps: app-a (starts the server, has the local admin) and app-b.


════════════════════════════════════════════════════════════════════
  1. Start Zitadel locally
════════════════════════════════════════════════════════════════════

$ zitadel start --cwd=app-a --port=7970 --non-interactive

Console: signed in as admin@zitadel.localhost. Open this link (works once):
http://localhost:7970/ui/console/login?handoff=handoff_…[redacted]
Local Zitadel server is ready.
(server: localhost:7970)

Next:
  Console: you are admin@zitadel.localhost. Open http://localhost:7970/ui/console/login?handoff=handoff_…[redacted] (works once).
  From your app directory, run setup; the CLI will detect the framework or ask when needed.
  Setup installs dependencies when needed; then start your app dev server.
  $ npx @zitadel/cli@1.0.0-alpha.22 setup --server local
  $ npx @zitadel/cli@1.0.0-alpha.22 console

[exit 0]


════════════════════════════════════════════════════════════════════
  2. Set up an app in app-a  →  claimed for the local admin automatically
════════════════════════════════════════════════════════════════════

$ zitadel setup --cwd=app-a --server=local --framework=next --skip-install --non-interactive

◐ Detecting framework in app-a
✔ Detected next (dev port 3000)
◐ Creating project on http://localhost:7970
✔ Created project proj_01M33Q7BBK5T7FQ4ENY7HDSW7E
◐ Patching project files
ℹ Wrote the project's .gitignore additions (.gitignore)
ℹ Wrote the local project secret (.zitadel/secret)
ℹ Wrote the Zitadel project configuration (zitadel.json)
ℹ Wrote the .env example template (.env.example)
ℹ Wrote the local development environment variables (.env.local)
ℹ Wrote the sync state file (.zitadel/state.json)
ℹ Wrote the flow dialect spec (editor $schema) (.zitadel/meta/flow-definition.json)
ℹ Wrote the user-schema dialect spec (.zitadel/meta/user-schema.json)
ℹ Wrote the user-property dialect spec (.zitadel/meta/user-property.json)
ℹ Wrote .zitadel/meta/property-name.json
ℹ Wrote .zitadel/meta/auth-methods.json
ℹ Wrote .zitadel/meta/auth-method.json
ℹ Wrote .zitadel/meta/sso-auth-method.json
ℹ Wrote .zitadel/meta/idp-connection.json
ℹ Wrote the branding dialect spec (.zitadel/meta/branding.json)
ℹ Wrote the agent guidance (golden journey + config dialect) (AGENTS.md)
ℹ Wrote the README's Zitadel section (README.md)
ℹ Wrote the login page (app/login/page.tsx)
ℹ Wrote the registration page (app/register/page.tsx)
ℹ Wrote the profile page (app/profile/page.tsx)
ℹ Wrote the Next.js middleware (middleware.ts)
ℹ Wrote the web-component type declarations (custom-elements.d.ts)
ℹ Wrote package.json with the SDK dependency (package.json)
ℹ Left .zitadel unchanged (already matches target)
ℹ Wrote the editable default human user schema (.zitadel/schemas/default-human-user.json)
ℹ Wrote the editable default login flow (.zitadel/flows/default-login.json)
ℹ Wrote the schemas folder README (.zitadel/schemas/README.md)
ℹ Wrote the flows folder README (.zitadel/flows/README.md)
✔ Patched 27 files (1 unchanged)
✔ Project owned by admin@zitadel.localhost (team team_localadmin)

 ╭─────────────────────────────Zitadel is ready───────────────────────────────╮
 │                                                                            │
 │  DETECTED                                                                  │
 │  ✓ Framework   Next.js 15.0.0 · npm                                        │
 │                                                                            │
 │  INSTALLED                                                                 │
 │  ✓ Package         @zitadel/sdk-next   → package.json                      │
 │  ✓ Login page      app/login/page.tsx                                      │
 │  ✓ Register page   app/register/page.tsx                                   │
 │  ✓ Profile page    app/profile/page.tsx                                    │
 │  ✓ Middleware      middleware.ts                                           │
 │  ✓ Env vars        .env.local                                              │
 │                                                                            │
 │  CUSTOMIZE                                                                 │
 │  ✓ User schema   .zitadel/schemas/   → see its README.md                   │
 │  ✓ Login flow    .zitadel/flows/   → see its README.md                     │
 │                                                                            │
 │  PROJECT                                                                   │
 │  ✓ Project id     proj_01M33Q7BBK5T7FQ4ENY7HDSW7E                          │
 │  ✓ Server         http://localhost:7970                                    │
 │  ✓ App will run   http://localhost:3000                                    │
 │  ✓ Login design   built-in template                                        │
 │                                                                            │
 │  Install dependencies:                                                     │
 │                                                                            │
 │    npm install                                                             │
 │                                                                            │
 │  Start your project (then open http://localhost:3000/login):               │
 │                                                                            │
 │    npm run dev                                                             │
 │                                                                            │
 │  Verify auth in the browser: register a user, log out, log in again with   │
 │  the same user, and confirm /profile shows Signed in.                      │
 │                                                                            │
 │  Once login works, this shows your next steps (customizing is covered in   │
 │  your README's Zitadel section):                                           │
 │                                                                            │
 │    npx @zitadel/cli@1.0.0-alpha.22 status                                  │
 │                                                                            │
 ╰────────────────────────────────────────────────────────────────────────────╯


[exit 0]


════════════════════════════════════════════════════════════════════
  3. The local claim record  app-a/.zitadel/secret
════════════════════════════════════════════════════════════════════
{
  "claimed_at": "2026-09-22T04:50:42Z",
  "created_at": "2026-09-22T04:50:42Z",
  "preview_origins": [
    "http://localhost:3000"
  ],
  "preview_secret": "eyJhbGciOiJBMjU2\u2026[redacted]",
  "project_id": "proj_01M33Q7BBK5T7FQ4ENY7HDSW7E",
  "project_secret": "eyJhbGciOiJBMjU2\u2026[redacted]",
  "team_id": "team_localadmin"
}


════════════════════════════════════════════════════════════════════
  4. Claim the already-owned project
════════════════════════════════════════════════════════════════════

$ zitadel claim --cwd=app-a --non-interactive

ℹ This project already belongs to team team_localadmin.
Skipped: already-claimed (server: localhost:7970)

[exit 0]


════════════════════════════════════════════════════════════════════
  5. Set up a second app in app-b  →  no local admin here, so the project stays unclaimed
════════════════════════════════════════════════════════════════════

$ zitadel setup --cwd=app-b --server=http://localhost:7970 --framework=next --skip-install --non-interactive

◐ Detecting framework in app-b
✔ Detected next (dev port 3000)
◐ Creating project on http://localhost:7970
✔ Created project proj_01M33Q7C7XK32CTKV82CRFCFQ4
◐ Patching project files
ℹ Wrote the project's .gitignore additions (.gitignore)
ℹ Wrote the local project secret (.zitadel/secret)
ℹ Wrote the Zitadel project configuration (zitadel.json)
ℹ Wrote the .env example template (.env.example)
ℹ Wrote the local development environment variables (.env.local)
ℹ Wrote the sync state file (.zitadel/state.json)
ℹ Wrote the flow dialect spec (editor $schema) (.zitadel/meta/flow-definition.json)
ℹ Wrote the user-schema dialect spec (.zitadel/meta/user-schema.json)
ℹ Wrote the user-property dialect spec (.zitadel/meta/user-property.json)
ℹ Wrote .zitadel/meta/property-name.json
ℹ Wrote .zitadel/meta/auth-methods.json
ℹ Wrote .zitadel/meta/auth-method.json
ℹ Wrote .zitadel/meta/sso-auth-method.json
ℹ Wrote .zitadel/meta/idp-connection.json
ℹ Wrote the branding dialect spec (.zitadel/meta/branding.json)
ℹ Wrote the agent guidance (golden journey + config dialect) (AGENTS.md)
ℹ Wrote the README's Zitadel section (README.md)
ℹ Wrote the login page (app/login/page.tsx)
ℹ Wrote the registration page (app/register/page.tsx)
ℹ Wrote the profile page (app/profile/page.tsx)
ℹ Wrote the Next.js middleware (middleware.ts)
ℹ Wrote the web-component type declarations (custom-elements.d.ts)
ℹ Wrote package.json with the SDK dependency (package.json)
ℹ Wrote the editable default human user schema (.zitadel/schemas/default-human-user.json)
ℹ Wrote the editable default login flow (.zitadel/flows/default-login.json)
ℹ Wrote the schemas folder README (.zitadel/schemas/README.md)
ℹ Wrote the flows folder README (.zitadel/flows/README.md)
✔ Patched 27 files

 ╭─────────────────────────────Zitadel is ready───────────────────────────────╮
 │                                                                            │
 │  DETECTED                                                                  │
 │  ✓ Framework   Next.js 15.0.0 · npm                                        │
 │                                                                            │
 │  INSTALLED                                                                 │
 │  ✓ Package         @zitadel/sdk-next   → package.json                      │
 │  ✓ Login page      app/login/page.tsx                                      │
 │  ✓ Register page   app/register/page.tsx                                   │
 │  ✓ Profile page    app/profile/page.tsx                                    │
 │  ✓ Middleware      middleware.ts                                           │
 │  ✓ Env vars        .env.local                                              │
 │                                                                            │
 │  CUSTOMIZE                                                                 │
 │  ✓ User schema   .zitadel/schemas/   → see its README.md                   │
 │  ✓ Login flow    .zitadel/flows/   → see its README.md                     │
 │                                                                            │
 │  PROJECT                                                                   │
 │  ✓ Project id     proj_01M33Q7C7XK32CTKV82CRFCFQ4                          │
 │  ✓ Server         http://localhost:7970                                    │
 │  ✓ App will run   http://localhost:3000                                    │
 │  ✓ Login design   built-in template                                        │
 │                                                                            │
 │  Install dependencies:                                                     │
 │                                                                            │
 │    npm install                                                             │
 │                                                                            │
 │  Start your project (then open http://localhost:3000/login):               │
 │                                                                            │
 │    npm run dev                                                             │
 │                                                                            │
 │  Verify auth in the browser: register a user, log out, log in again with   │
 │  the same user, and confirm /profile shows Signed in.                      │
 │                                                                            │
 │  Once login works, this shows your next steps (customizing is covered in   │
 │  your README's Zitadel section):                                           │
 │                                                                            │
 │    npx @zitadel/cli@1.0.0-alpha.22 status                                  │
 │                                                                            │
 │  Claim your Project before Oct 6, 2026, 03:50 PM GMT+11 to make it         │
 │  permanent and start collaborating with your team. Until then, your        │
 │  Project is temporary and its data may be lost.                            │
 │                                                                            │
 │    npx @zitadel/cli@1.0.0-alpha.22 claim                                   │
 │                                                                            │
 ╰────────────────────────────────────────────────────────────────────────────╯


[exit 0]


════════════════════════════════════════════════════════════════════
  6. The unclaimed record  app-b/.zitadel/secret
════════════════════════════════════════════════════════════════════
{
  "created_at": "2026-09-22T04:50:43Z",
  "preview_origins": [
    "http://localhost:3000"
  ],
  "preview_secret": "eyJhbGciOiJBMjU2\u2026[redacted]",
  "project_id": "proj_01M33Q7C7XK32CTKV82CRFCFQ4",
  "project_secret": "eyJhbGciOiJBMjU2\u2026[redacted]"
}


════════════════════════════════════════════════════════════════════
  7. Claim the unclaimed app-b project  (prints a link, then waits for the browser)
════════════════════════════════════════════════════════════════════

$ zitadel claim --cwd=app-b --server=http://localhost:7970 --non-interactive


 ╭─────────────────────────Finish in your browser───────────────────────────╮
 │                                                                          │
 │  Create an account or sign in to claim your Project, make it permanent   │
 │  and start collaborating.                                                │
 │                                                                          │
 ╰──────────────────────────────────────────────────────────────────────────╯

http://localhost:7970/ui/console/claim?challenge_id=ch_…[redacted]&project_id=proj_01M33Q7C7XK32CTKV82CRFCFQ4
ℹ Open the link above to continue.
◐ Waiting for the browser step to finish
✔ Project claimed
Your Project is now permanent.
(server: localhost:7970)

Next:
  Manage your Project and collaborate in the Console:
  http://localhost:7970/ui/console/projects/proj_01M33Q7C7XK32CTKV82CRFCFQ4

[exit 0]


════════════════════════════════════════════════════════════════════
  8. Mint a console sign-in link
════════════════════════════════════════════════════════════════════

$ zitadel console --cwd=app-a --no-open --non-interactive

Signed in to the local console as admin@zitadel.localhost:
http://localhost:7970/ui/console/login?handoff=handoff_…[redacted]
ℹ Open the link above. It works once; run this command again for a new one.
Local console sign-in link.
(server: localhost:7970)

Next:
  The link signs you in once. Run `zitadel console` again for a fresh one.

[exit 0]


════════════════════════════════════════════════════════════════════
  9. app-b/.zitadel/secret after the browser claim  (team + time recorded)
════════════════════════════════════════════════════════════════════
{
  "claimed_at": "2026-09-22T04:51:29Z",
  "created_at": "2026-09-22T04:50:43Z",
  "preview_origins": [
    "http://localhost:3000"
  ],
  "preview_secret": "eyJhbGciOiJBMjU2\u2026[redacted]",
  "project_id": "proj_01M33Q7C7XK32CTKV82CRFCFQ4",
  "project_secret": "eyJhbGciOiJBMjU2\u2026[redacted]",
  "team_id": "team_localadmin"
}


════════════════════════════════════════════════════════════════════
  10. Claim app-b again  (now already claimed)
════════════════════════════════════════════════════════════════════

$ zitadel claim --cwd=app-b --server=http://localhost:7970 --non-interactive

ℹ This project already belongs to team team_localadmin.
Skipped: already-claimed (server: localhost:7970)

[exit 0]


════════════════════════════════════════════════════════════════════
  BROWSER — what the console showed (localhost:7970)
════════════════════════════════════════════════════════════════════

 a. Opened the sign-in link from step 8
    → Signed in as admin@zitadel.localhost; project picker shows "app-a".
    → The link was minted by the CLI proving only the admin's email; the server
      resolved it through the platform schema's designated identifier (email).
    → "Console API not authorized" in the main area is the known #1022 limitation.
    → Network: POST /sessions/exchange?project_id=proj_platform → 200 OK

 b. Opened the claim link printed in step 7
    → "Project claimed — Your project is now permanent." Step 7's waiting command
      then printed "✔ Project claimed".

 c. Reopened the same, now used, claim link
    → "Already claimed — the project is already claimed by a team".

 d. Console project picker after the claim
    → Lists both "app-a" and "app-b": the server confirms the admin owns both.

 e. Logged out, then reopened the step 8 sign-in link a second time
    → Sign-in widget; the link works only once.
    → Network: POST /sessions/exchange?project_id=proj_platform → 400 Bad Request


════════════════════════════════════════════════════════════════════
  11. Tear down
════════════════════════════════════════════════════════════════════

$ zitadel stop --cwd=app-a --non-interactive

Local Zitadel server stopped.
(server: localhost:7970)

[exit 0]

````

</details>

File: apps/cli/src/commands/start.ts [hunk 7]
```diff
@@ -107,6 +110,18 @@ export default class Start extends BaseCommand {
...
+    // server imports the local admin, and the console signs them in by link.
+    // The admin is a user of the platform project, so the two travel together:
+    // a caller that opts out of the platform bootstrap (a test harness wanting
+    // a bare single-project server) gets neither. The opt-out is read from the
+    // same merged environment the server receives — the project's env files
+    // over the shell — so a `false` in `.env.local` turns both off rather than
+    // being overridden by the user file forcing the bootstrap back on.
+    const local = platformBootstrapEnabled({ ...this.meta.env, ...env.values })
+      ? await ensureLocalAdmin(this.meta.cwd)
+      : undefined;
```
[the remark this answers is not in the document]
mridang: Fixed in 6b46cd150. The CLI now treats a `NEXTGEN_PLATFORM_PROJECT_ID` pin other than `proj_platform` as an opt-out, read from the same merged environment the server receives, so a pinned server starts without a local admin instead of failing validation. Covered by `start creates no local admin when the server is pinned to its own project`.

apps/cli/src/lib/local-server/admin-credential.ts
```diff
@@ -0,0 +1,191 @@
...
+const ADMIN_USER_ID = "user_localadmin";
+const ADMIN_TEAM_ID = "team_localadmin";
+
+/**
+ * The user schema the platform project is seeded with, derived from the same
+ * default `schema.builtin_public_base` the server uses rather than spelled out
+ * here. A document naming a schema the server does not have is bootstrapped
+ * against an empty placeholder, and the admin can then never sign in.
+ */
+const DEFAULT_USER_SCHEMA_URL = defaultHumanUserSchemaUrl();
```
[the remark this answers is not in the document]
mridang: Fixed in 6b46cd150. `start` passes the effective `NEXTGEN_SCHEMA_BUILTIN_PUBLIC_BASE` (shell overlaid by the project's env files) into `ensureLocalAdmin`, and the bootstrap document names the default user schema under that base. Covered by `names the schema under the server's configured schema base`.

apps/cli/src/lib/local-server/admin-credential.ts
```diff
@@ -0,0 +1,191 @@
...
+};
+
+/** Reads the local admin credential, or `undefined` when `start` never wrote one. */
+export async function readLocalAdmin(cwd: string): Promise<LocalAdmin | undefined> {
+  let raw: string;
+  try {
+    raw = await readFile(join(cwd, LOCAL_ADMIN_FILE), "utf8");
+  } catch {
+    return undefined;
+  }
```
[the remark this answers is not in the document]
mridang: Fixed in 6b46cd150. Only `ENOENT` now means there is no admin. Any other read error is reported as an unreadable credential rather than replaced, so the imported password and `admin.json` cannot drift apart. Covered by `refuses an unreadable credential instead of replacing it`.

apps/cli/src/lib/local-server/admin-credential.ts
```diff
@@ -0,0 +1,191 @@
...
+  const admin = (await readLocalAdmin(cwd)) ?? (await createLocalAdmin(cwd));
+
+  // Readable beyond the owner, unlike admin.json: the docker runtime mounts
+  // this document into a container running as the host user, and an unreadable
+  // file fails the server's bootstrap import at startup. It carries a hash.
+  const userFile = join(cwd, LOCAL_ADMIN_USER_FILE);
+  await writeFile(userFile, `${JSON.stringify(bootstrapUserDocument(admin), null, 2)}\n`, {
+    mode: 0o644,
+  });
+  await chmod(userFile, 0o644).catch(() => undefined);
```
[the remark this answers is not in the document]
mridang: Fixed in 6b46cd150. The document is written to a uniquely named staging file and renamed into place. The staging name is random rather than per-process, since two starts inside one process would otherwise rename each other's file away; the existing concurrent-start test caught that on the first attempt.

apps/cli/src/lib/local-server/sign-in.ts
```diff
@@ -0,0 +1,161 @@
...
+ * A rejected identifier means this server never imported the admin — an older
+ * data directory, say — which no retry fixes, so say that rather than letting
+ * the generic proof error send the reader hunting.
+ */
+function proofError(method: string, res: JsonResponse): ZitadelError {
+  const code = isObject(res.body) && typeof res.body.code === "string" ? res.body.code : undefined;
+  if (method === "identifier" && code === "att.proof_rejected") {
+    return new ZitadelError("E_AUTH", "The local server has no local admin user", {
+      hint: "The local data directory predates the local admin. Run `zitadel reset --force`, then `zitadel start`.",
+      nextCommands: ["zitadel reset --force", "zitadel start"],
```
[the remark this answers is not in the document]
mridang: Fixed in 6b46cd150. The message no longer asserts a cause: it names both the pre-admin data directory and a failed lookup, and points at `zitadel logs` first. The destructive `zitadel reset --force` stays in the human-readable hint and is no longer in `next_commands`, so an agent following the envelope cannot run it on a transient failure. A distinct server-side no-match code would be the fuller fix, but that belongs with the auth-attempt resolution work.

apps/cli/src/lib/local-server/sign-in.ts
```diff
@@ -0,0 +1,161 @@
...
+ * adds the rendering. With no page to render, the CLI states the factors
+ * outright: open an attempt, prove the identifier, prove the password, take the
+ * one-time `handoff_token`. That token becomes either a console sign-in link or,
+ * exchanged, a session cookie.
+ */
+
+/** A console URL that signs the admin in once, via a fresh handoff token. */
+export async function consoleSignInUrl(serverUrl: string, admin: LocalAdmin): Promise<string> {
+  const { handoffToken } = await signIn(serverUrl, admin);
+  return `${serverUrl}/ui/console/login?handoff=${encodeURIComponent(handoffToken)}`;
```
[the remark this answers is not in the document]
mridang: Leaving this one. A CLI-managed local server is started with the default console path, and moving it means plumbing the effective path through `start`, `console` and the runtime metadata for a setting a local developer has no reason to change. If custom console paths become a supported local configuration, the sign-in URL should come from the runtime metadata rather than being rebuilt from env.

internal/bootstrap/users/import.go
```diff
@@ -65,7 +66,11 @@ func importFile(
 		return fmt.Errorf("check existing user: %w", err)
 	}
 
-	attrs, err := buildCreateAttributes(doc.Attributes)
+	schema, err := loadUserSchema(ctx, v2Pool.Statements(), doc.Header)
+	if err != nil {
+		return err
+	}
+	attrs, err := buildCreateAttributes(doc.Attributes, schema, doc.Header.TeamID)
```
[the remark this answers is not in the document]
mridang: Leaving this for a separate change. It affects only users imported by an earlier version, and the local admin this PR adds is new, so there are none of those among them. Reconciling an existing user's unique-attribute registry on every startup is a migration concern, not an import one, and it should get its own change rather than ride along here.

internal/api/auth_attempt.go
```diff
@@ -134,8 +134,14 @@ func verifyRequestToProof(req *api.VerifyChallengeRequest) (service.Proof, error
...
 		p := req.GetOneOf().IdentifierProof
+		// Which attribute identifies a user is the project's decision, not the
+		// server's: a flow step collects `email`, `username` or whatever else
+		// the schema registers unique, and the proof says which one this value
+		// is. Assuming a default here would sign the wrong deployment's users
+		// out of their own instance.
 		return service.UserProof{
-			LoginName: p.GetLoginName(),
+			LoginName:     p.GetLoginName(),
+			AttributeName: p.GetAttributeName(),
```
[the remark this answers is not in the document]
mridang: Reworked in 18aeb8506 and 6a8a3d02a to follow ADR 058 §5. The request carries only `login_name`, and the service resolves it against each user schema's `x-identifier` and requires exactly one match; the full detail is on #1256, which carries the same change. The CLI now sends just the admin's email, which resolves through the platform project's default schema designating `email`. Verified end to end on a local server: start, setup auto-claiming as `team_localadmin`, claim reporting it already owned, and a `zitadel console` link producing an active session identified by `email`.

IAM-marco:
> ### Known, not addressed here
> 
>  After signing in, the console shows _Console API not authorized_. Its management endpoints don't accept a console session until [authz: accept Console session as user principal on management APIs #1022](https://github.com/zitadel/nextgen/issues/1022) lands. Sign-in itself works.

Hi @mridang 

The linked issue is just a wrapper for child issues. I left the [CSRF one](https://github.com/zitadel/nextgen/issues/1140) out of the claiming milestone as I don't see it necessary for having a functional experience in local. Do you agree? WDYT? It can be added on a second iteration

apps/cli/src/lib/local-server/http.ts
```diff
@@ -0,0 +1,53 @@
+import { ZitadelError } from "../errors";
+import { isObject } from "../json";
+
+/**
+ * The small HTTP helpers the local-admin calls share. Every request here runs
+ * against a server that already answered `/healthz`, so a stall means a wedged
+ * server rather than a slow start — hence the bound, which keeps
+ * `zitadel start` from hanging on a socket that accepts and never answers.
+ */
+const REQUEST_TIMEOUT_MS = 10_000;
```
IAM-marco: Simplification: this file hand-rolls an HTTP layer the repo already has.

**What changed.** The new `http.ts` adds `fetchJson`, `postJson` and `apiError`, and `sign-in.ts` and `claim-as-admin.ts` build their calls to the auth-attempt and claim endpoints on top of them, each followed by manual checks on the shape of the response body (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/lib/local-server/sign-in.ts#L33-L44, https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/lib/local-server/claim-as-admin.ts#L40-L56).

**Why this spot matters.** The CLI already talks to the server through the generated client. `apply.ts`, `status.ts` and `claim.ts` all call `createZitadelClient` from `@zitadel/api/client`, and the OpenAPI spec defines every operation these helpers reach: `createAuthAttempt`, `issueChallenge`, `verifyChallengeProof`, `createHandoff`, `initClaim`, `completeClaim`. The generated client throws a typed `ApiError` on non-2xx and passes a `RequestInit` through, so the `origin` header and the `AbortSignal.timeout` bound both fit (`status.ts` already sets a timeout this way). The responses come back typed, so the hand-written body checks go too.

**What follows.** Two HTTP layers now live in the CLI, with two error shapes and two timeout rules to keep in step.

**Kept separate.** Only `sessions/exchange` needs the raw `Set-Cookie` header from the response, so one direct `fetch` for that call in `sign-in.ts` is fine.

**Suggested direction.** Replace the helpers with a `createZitadelClient({ baseUrl: serverUrl, token: publishableKey })` in `sign-in.ts` and `claim-as-admin.ts`, and let `toZitadelError` map the `ApiError`. The whole of `http.ts` and the body-shape checks should then disappear.

**Coverage.** The unit tests in `sign-in.test.ts` and `claim-as-admin.test.ts` stub `fetch`, so they will need to stub the client's fetch instead, the way the `status` tests do.

apps/cli/src/lib/local-server/sign-in.ts
```diff
@@ -0,0 +1,163 @@
...
+
+/** Posts to the auth-attempt API with the browser-plane credential it expects. */
+function attemptClient(serverUrl: string, publishableKey: string): AttemptClient {
+  const origin = new URL(serverUrl).origin;
+  return async (path, body) =>
+    postJson(`${serverUrl}${path}`, body, {
+      authorization: `Bearer ${publishableKey}`,
+      origin,
+    });
+}
```
IAM-marco: Simplification: `AttemptClient` and its closure are a two-line client in disguise.

**What changed.** `attemptClient` wraps `postJson` so each auth-attempt call carries the publishable key and the `origin` header, and the `AttemptClient` type names that closure (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/lib/local-server/sign-in.ts#L102-L112).

**Why this spot matters.** With the generated client (see the note on `http.ts`), the same thing is `const client = createZitadelClient({ baseUrl: serverUrl, token: publishableKey })` plus one shared `const init = { headers: { origin } }` passed to each call. No custom type, no closure.

**What follows.** One more abstraction for a reader to decode, with a single caller.

**Suggested direction.** Drop the type and the factory once the calls go through the generated client.

apps/cli/src/lib/local-server/claim-as-admin.ts
```diff
@@ -0,0 +1,65 @@
...
+export type ProjectOwner = {
+  team_id: string;
+  /**
+   * Absent when the project was already claimed: the 409 names the owning team
+   * but not when it happened, and a local timestamp would disagree with the
+   * grant this record mirrors.
+   */
+  claimed_at?: string;
+};
+
```
IAM-marco: Simplification: the "already claimed" branch handles a case this code path cannot hit.

**What changed.** `ProjectOwner.claimed_at` is optional so that a 409 from `claim/init` can be returned as an owner without a timestamp (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/lib/local-server/claim-as-admin.ts#L7-L16, https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/lib/local-server/claim-as-admin.ts#L61-L65), and `setup` then has an extra branch for that shape (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/commands/setup/index.ts#L469-L472), with a dedicated unit test.

**Why this spot matters.** `setup` creates the project a few lines earlier in the same run, and the whole block is guarded by `!dryRun`. A project created seconds ago by this process is not already claimed by someone else. The one case the branch covers is a race nobody has.

**What follows.** A wider type, an else branch in `setup`, and a test, all for an outcome the surrounding code rules out.

**Suggested direction.** Make both `ProjectOwner` fields required. If a 409 ever does arrive, let it fall into the `catch` that `setup` already has right below (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/commands/setup/index.ts#L484-L488): a warning plus the usual claim nudge is the right outcome for it anyway.

**Coverage.** The "already claimed" test in `claim-as-admin.test.ts` goes with the branch; the remaining tests still cover the happy path and the error path.

apps/cli/src/lib/local-server/admin.ts
```diff
@@ -0,0 +1,23 @@
...
+export {
+  ensureLocalAdmin,
+  readLocalAdmin,
+  LOCAL_ADMIN_EMAIL,
+  LOCAL_ADMIN_FILE,
+  LOCAL_ADMIN_USER_FILE,
+  type LocalAdmin,
+} from "./admin-credential";
+export { consoleSignInUrl } from "./sign-in";
+export { claimProjectAsAdmin, type ProjectOwner } from "./claim-as-admin";
```
IAM-marco: Simplification: a barrel file that only re-exports three neighbours.

**What changed.** `admin.ts` re-exports everything from `admin-credential.ts`, `sign-in.ts` and `claim-as-admin.ts` (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/lib/local-server/admin.ts#L1-L23).

**Why this spot matters.** There are four importers: `console.ts`, `setup/index.ts`, `start.ts` and `admin.test.ts`. Each of them can import the one module it actually uses. The rest of `apps/cli/src/lib/local-server` has no barrel either; callers import `runtime.ts`, `docker.ts` or `binary.ts` directly.

**What follows.** One extra hop when a reader follows an import, and one more file to keep in sync when a symbol is added or renamed.

**Suggested direction.** Delete the file and point the four imports at the module that defines the symbol.

apps/cli/src/lib/local-server/platform.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * The well-known id of a deployment's platform project (`domain.PlatformProjectID`,
+ * `internal/domain/project.go`). A CLI-managed local server bootstraps it, which
+ * is what gives the developer a team to own their projects with.
+ */
+export const PLATFORM_PROJECT_ID = "proj_platform";
```
IAM-marco: Simplification: a module for one constant.

**What changed.** `platform.ts` holds only `PLATFORM_PROJECT_ID` (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/lib/local-server/platform.ts#L1-L6).

**Why this spot matters.** `runtime.ts` already holds every other local-server constant, and `admin-credential.ts` imports `LOCAL_RUNTIME_DIR` from there. The new constant belongs next to them.

**Suggested direction.** Move the one export line, with its comment, into `runtime.ts` and delete `platform.ts`.

apps/cli/src/commands/console.ts
```diff
@@ -0,0 +1,105 @@
...
+    "<%= config.bin %> <%= command.id %> --no-open",
+  ];
+
+  static override flags = {
+    "no-open": Flags.boolean({ description: "Print the sign-in link instead of opening a browser." }),
+    // Inherited from BaseCommand, but this command only ever signs in to the
+    // runtime started from this directory, so the override would be a lie in
+    // the help and the README. Hidden rather than removed: passing it is a
+    // clear error below instead of oclif's "Nonexistent flag".
+    server: Flags.string({ char: "s", hidden: true }),
```
IAM-marco: Simplification: a hidden `--server` flag that exists only to be rejected.

**What changed.** `console` declares a hidden `server` flag (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/commands/console.ts#L33-L38) and then throws a validation error when it is set (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/commands/console.ts#L50-L56), with a unit test for the rejection.

**Why this spot matters.** The sibling commands that work on the runtime started from this directory (`stop`, `reset`, `logs`, `status`) pass `resolveServer: false` and simply ignore `--server` (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/commands/stop.ts#L32-L35). `console` does the same `resolveServer: false` call, so it already matches them; the extra flag and error make it the one command that behaves differently.

**What follows.** Three code blocks and a test to maintain for a flag the command does not take.

**Suggested direction.** Drop the hidden flag, the `serverFlag` check, and its test, and let `console` follow the sibling pattern.

apps/cli/src/lib/local-server/admin-credential.ts
```diff
@@ -0,0 +1,200 @@
...
+  cwd: string,
+  options: {
+    /**
+     * The server's `schema.builtin_public_base`, when the environment overrides
+     * it. The platform project's default user schema is published under that
+     * base, and a document naming a schema the server does not have is
+     * bootstrapped against an empty placeholder — the admin could never sign in.
+     */
+    builtinSchemaBase?: string;
+  } = {},
```
IAM-marco: Simplification: an options object with a single field.

**What changed.** `ensureLocalAdmin` takes an `options` object whose only member is `builtinSchemaBase` (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/lib/local-server/admin-credential.ts#L101-L109), and the one caller in `start.ts` builds that object inline (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/commands/start.ts#L122-L126).

**Why this spot matters.** An options bag earns its place when there are several optional inputs or when more are expected. Here there is one, and the doc comment can sit on a plain parameter just as well.

**Suggested direction.** `ensureLocalAdmin(cwd: string, builtinSchemaBase?: string)`, and the call site passes the value directly.

apps/cli/src/lib/local-server/sign-in.ts
```diff
@@ -0,0 +1,163 @@
...
+        nextCommands: ["zitadel stop", "zitadel start"],
+        details: {
+          server_url: serverUrl,
+          console_project_id: isObject(runtime) ? runtime.console_project_id : undefined,
+        },
+      },
+    );
+  }
+  return runtime.publishable_key;
+}
```
IAM-marco: Simplification: two readers of the same runtime document, run back to back.

**What changed.** `platformPublishableKey` fetches `/console/runtime.json`, checks that `console_project_id` is the platform project, and returns the publishable key (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/lib/local-server/sign-in.ts#L138-L163). `localServerHostsPlatform` in `setup` fetches the same document and makes the same check (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/src/commands/setup/index.ts#L641-L657). `setup` calls the second and then, through `claimProjectAsAdmin`, the first, so the document is read twice per run.

**Why this spot matters.** Both functions decide the same question from the same bytes, with separate timeouts and separate error handling.

**What follows.** A change to the runtime document, or to what "hosts the platform" means, has to be made in two places.

**Suggested direction.** One `readPlatformRuntime(serverUrl)` next to the constant that returns `{ publishable_key }` or `undefined`. `localServerHostsPlatform` becomes `Boolean(await readPlatformRuntime(server))`, and the sign-in path takes the key from the same call.

apps/console/src/routes/login.tsx
```diff
@@ -54,6 +65,40 @@ export const Route = createFileRoute("/login")({
...
+      headers: {
+        "content-type": "application/json",
+        ...(publishableKey ? { authorization: `Bearer ${publishableKey}` } : {}),
+      },
+      body: JSON.stringify({ handoff_token: handoffToken }),
+    });
+  } catch {
+    // Fall through to the widget.
+  }
+}
```
IAM-marco, line 98: Simplification: a hand-written `fetch` for a call the generated client already provides.

**What changed.** `exchangeLinkHandoff` builds the URL, the query string, the headers and the body for `POST /sessions/exchange` by hand (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/console/src/routes/login.tsx#L83-L100).

**Why this spot matters.** The console's typed client (`api` in `apps/console/src/api/zitadel.ts`) has `exchangeHandoff(body, params, init)`, and the login widget calls exactly that (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/packages/components/src/orchestrator/api-client.ts#L82-L88). `auth/session.ts` uses the same client for `getMySession` on this very session surface. The SDK config also accepts a `publishableKey`, so a handle that carries the runtime-discovered key is available, the same way `ZitadelLogin` gets one further down this file.

**What follows.** Any change to how the console authenticates or reports errors on the generated client will not reach this one call.

**Suggested direction.** `await api.exchangeHandoff({ handoff_token }, { project_id: projectId }, { credentials: "include", headers: ... }).catch(() => undefined)`, about six lines, on a client configured with the publishable key.

internal/service/identifier.go
```diff
@@ -0,0 +1,114 @@
...
+
+// identifierUsers selects which users an identifier lookup may resolve to.
+// Each caller states it explicitly: whether an inactive user can be reached by
+// identifier is decided per surface, not inherited from a shared default.
+type identifierUsers uint8
+
+const (
+	anyUserStatus identifierUsers = iota
+	activeUsersOnly
+)
```
IAM-marco: Simplification: a two-value enum standing in for a boolean.

**What changed.** `identifierUsers` is a `uint8` type with two constants, `anyUserStatus` and `activeUsersOnly` (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/internal/service/identifier.go#L17-L25), used once to decide whether to add the status filter (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/internal/service/identifier.go#L46-L48).

**Why this spot matters.** The comment says each caller must state the choice explicitly, which a named parameter already does. There are two callers: `grant.go` wants active users only, `auth_attempt.go` wants any status.

**Suggested direction.** `activeOnly bool`; `grant.go` passes `true`, `auth_attempt.go` passes `false`. Same explicitness at the call site, six fewer lines.

internal/service/identifier.go
```diff
@@ -0,0 +1,114 @@
...
+			getLoggingContext(ctx, area).Info("identifier lookup matched no user",
+				slog.String("project_id", projectID),
+			)
+		}
+		return nil, errIdentifierUnresolved
+	}
+	for _, user := range found {
+		return user, nil
+	}
+	return nil, errIdentifierUnresolved
```
IAM-marco: Simplification: a loop to fetch the single entry of a map, plus an unreachable return.

**What changed.** After `len(found) != 1` is ruled out, the function ranges over the one-entry map to return its value, then returns `errIdentifierUnresolved` on a line that cannot run (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/internal/service/identifier.go#L81-L84).

**Why this spot matters.** The pre-refactor code in `grant.go` tracked the match directly. Keeping `var match *domain.User` next to `found`, set inside the loop, lets the function end with `return match, nil`, and the dead line goes.

**Suggested direction.** Track the last match alongside the map and return it; drop the trailing loop and the unreachable return.

internal/bootstrap/users/import.go
```diff
@@ -124,6 +147,47 @@ func buildCreateAttributes(attrs map[domain.AttributeKey]json.RawMessage) (domai
...
+	switch scope {
+	case domain.SchemaUniqueScopeProject:
+		return domain.AttributeUniquenessProject
+	case domain.SchemaUniqueScopeTeam:
+		return domain.AttributeUniquenessTeam
+	default:
+		return domain.AttributeUniquenessUnspecified
+	}
+}
+
```
IAM-marco, line 183: Simplification: a third copy of the `x-unique` to `AttributeUniqueness` switch.

**What changed.** `uniqueScopeFromSchema` walks to the property's `x-unique` annotation and maps `project` and `team` to the two uniqueness values (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/internal/bootstrap/users/import.go#L174-L190).

**Why this spot matters.** The same switch already exists in `internal/domain/attribute.go` (inside `CreateAttributesFromMap`, https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/internal/domain/attribute.go#L226-L232) and in `internal/domain/flow_field_resolver_schema.go` as `deriveUnique` (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/internal/domain/flow_field_resolver_schema.go#L226-L233). The function's own comment says it applies "the same rule" as the domain, but it does not call it.

**What follows.** A new uniqueness scope, or a renamed annotation value, must be changed in three files.

**Suggested direction.** Export one `domain.UniqueScopeOf(annotation string) AttributeUniqueness` and call it from all three places. Here the body becomes the `GetNested` walk plus that one call.

apps/cli/tests/unit/commands/local-runtime.test.ts
```diff
@@ -502,6 +514,85 @@ describe("local runtime commands", () => {
...
+    });
+
+    expect(result.exitCode).toBe(0);
+    const envelope = parseJson(result.stdout) as { data: { runtime: { pid: number }; console?: unknown } };
+    binaryPids.push(envelope.data.runtime.pid);
+    expect(envelope.data.console).toBeUndefined();
+    await expect(readFile(join(cwd, ".zitadel/local/admin.json"), "utf8")).rejects.toThrow();
+    const commandLine = (await readRuntimeMetadata(cwd)) as { command?: string } | undefined;
+    expect(commandLine?.command ?? "").not.toContain("--user-file");
+  });
```
IAM-marco: Simplification: three near-identical tests for the platform opt-out.

**What changed.** Three `it` blocks set up the same fake server and assert the same four things (no console data, no `admin.json`, no `--user-file` in the command line), differing only in how the opt-out is expressed: an env var, a line in `.env.local`, or a pinned project id (https://github.com/zitadel/nextgen/blob/91e3c94316f50d616db926109fdfc28db743a26c/apps/cli/tests/unit/commands/local-runtime.test.ts#L520-L594).

**Why this spot matters.** The file already has long tests; three copies of the same body triple the maintenance when an assertion changes.

**Suggested direction.** One `it.each([{ name, env, envFile? }])` with a single body that writes `.env.local` when the case gives one. Same three cases, one set of assertions.

IAM-marco (changes requested):
Two-axis review of 91e3c943 against main (4e61fd2b). Generated code and changeset files were skipped.

## Code review

No finding reached the posting bar (80 out of 100), so this axis has no inline comments. Eight candidates were checked in the code and scored; they are listed here so nothing is lost. Scores are from the calibration rubric, not a judgement of the PR.

- 75, guideline note: `apps/console/src/routes/login.tsx` exchanges the handoff token with a hand-written `fetch` instead of the console's typed client, which Console ADR 0002 asks for. The same spot carries a Ponytail comment below.
- 70, past review note: `internal/bootstrap/users/import.go` only registers unique attributes from the schema when the user is first created, so a bootstrap user imported by an older binary keeps its old registry state. Deferred on the Copilot thread, no follow-up issue filed.
- 55, bug: `apps/cli/src/commands/start.ts` calls `ensureLocalAdmin` without a guard, so a malformed `admin.json` fails `start` before the server boots. The branch has a commit that shapes that error on purpose, and `setup` guards the same read, so this is a design choice worth a look rather than a defect.
- 55, past review note: `apps/cli-journey-e2e/src/claim.spec.ts` still skips the Docker runtime, which Copilot asked to cover. Deferred by the author.
- 50, past review note: the flow path in `internal/service/auth_attempt.go` still looks the login name up project-wide rather than within the bound schema. Pre-existing, disclosed in the new comment, no follow-up issue filed.
- 50, guideline note: `internal/bootstrap/users/import.go` repeats the `x-unique` mapping the domain already has. Covered by a Ponytail comment below.
- 25, comment guidance: ADR 046 still says claiming is cloud-only while local setup now runs the claim endpoints; the ADR text is incomplete rather than wrong.
- 25, guideline note: the "user file forces the platform bootstrap" rule is written once per backend in `binary.ts` and `docker.ts`.

## Ponytail

Thirteen inline comments, all about size and duplication, none about correctness. The largest is the hand-rolled HTTP layer in `apps/cli/src/lib/local-server/http.ts`: the generated client already covers every endpoint it reaches, and dropping it also removes the manual response-shape checks in `sign-in.ts` and `claim-as-admin.ts`. The rest are a barrel file, a one-constant module, a hidden flag that exists only to be rejected, an options object with one field, two readers of the same runtime document, a two-value enum for a boolean, an unreachable return, a third copy of the `x-unique` switch, a 409 branch for a project the same run just created, and three near-identical tests.

Estimated net: about 260 lines fewer if every comment is taken.

## Tally

| Axis | Findings posted | Worst |
| --- | --- | --- |
| Code review | 0 (8 scored below 80) | Console handoff exchange bypasses the typed client (75) |
| Ponytail | 13 | Hand-rolled HTTP layer duplicating the generated client |

apps/cli/src/lib/local-server/http.ts
mridang: Done in d46a8adae: `http.ts` is gone. Sign-in and claim now call the generated client; the one direct `fetch` left is `sessions/exchange`, because the session only comes back as a `Set-Cookie` header the client doesn't expose.

apps/cli/src/lib/local-server/sign-in.ts
mridang: Done in d46a8adae: `AttemptClient` is gone; the auth-attempt calls use `createZitadelClient` with the publishable key.

apps/cli/src/lib/local-server/claim-as-admin.ts
mridang: Done in d46a8adae: `ProjectOwner` fields are required, and the 409 branch, setup's "already belongs" branch and its test are removed.

apps/cli/src/lib/local-server/admin.ts
mridang: Done in d46a8adae: barrel deleted, importers point at the modules directly.

apps/cli/src/lib/local-server/platform.ts
mridang: Done in d46a8adae: `PLATFORM_PROJECT_ID` lives in `runtime.ts`; `platform.ts` is deleted.

apps/cli/src/commands/console.ts
mridang: Done in d46a8adae: the hidden flag, the rejection and its test are gone; the README now lists the inherited `--server` like the sibling commands.

apps/cli/src/lib/local-server/admin-credential.ts
mridang: Done in d46a8adae: `ensureLocalAdmin(cwd, builtinSchemaBase?)`.

apps/cli/src/lib/local-server/sign-in.ts
mridang: Done in d46a8adae: one `readPlatformRuntime` in `runtime.ts`, used by both setup's check and sign-in.

apps/console/src/routes/login.tsx
mridang: Done in d46a8adae: `exchangeHandoff` on a client carrying the publishable key, errors still fall through to the widget. Checked the sign-in link against a local server.

internal/service/identifier.go
mridang: Done in d46a8adae: `activeOnly bool`.

internal/service/identifier.go
mridang: Done in d46a8adae: the match is tracked in the loop and returned directly.

internal/bootstrap/users/import.go
mridang: Done in d46a8adae: one `domain.UniqueScopeOf`, used by all three call sites.

apps/cli/tests/unit/commands/local-runtime.test.ts
mridang: Done in d46a8adae: folded into one `it.each`.

IAM-marco (approved):
Second look at d46a8ada. All 13 threads from the first round are addressed in the code, not only in the replies: the hand-written HTTP layer is gone and the sign-in and claim calls go through the generated client (the SDK fetch keeps the origin and cookie headers and adds the bearer token only when the caller did not), the barrel file, the one-constant module, the hidden --server flag and the 409 branch are removed, the runtime document has one reader, the resolver uses a plain boolean and returns the tracked match, the x-unique switch lives once in domain.UniqueScopeOf, and the three opt-out tests are one it.each. Net about 200 lines fewer.

CI is green on this head, including the binary fresh-app journey, which proves the local claim and the console sign-in link end to end against a real server.

Two small notes, neither blocking:

- The branch is behind main; a merge or rebase is needed before the squash.
- Three items deferred on earlier threads still have no follow-up issue: scoping the flow-path identifier lookup to the bound schema (ADR 058 section 5), a Docker variant of the ownership journey, and re-registering unique attributes for bootstrap users imported by an older binary. Filing them would keep them from getting lost.

mridang:
@IAM-marco thanks for the review. Yes we can do this in the second iteration https://github.com/zitadel/nextgen/issues/1140

PR #1239, opened 2026-09-16
feat: accept console session to read teams

### LINKED ISSUE #1227: Accept the Console session on team reads
Part of #1022.

## Goal

Let a signed-in Console session read teams, so the page the claim flow lands on actually loads.

## Current state

After completing a claim, the "Open the console" button links to `/`, which redirects to `/teams?status=active` (`apps/console/src/routes/_authed/index.tsx:20-25`). That page's only data call is `POST /teams/query`, and that endpoint only accepts a project secret (`api/openapi/endpoints/teams/query/methods.yaml`, `security: oauth2: [team.read]`). The browser only holds the `__nextgen_session` cookie, so the call fails with 401 in production. It works in development only because the Vite proxy injects a project secret (`apps/console/vite.config.mts:130-135`).

PR #1142 taught the server to accept the session cookie as a user principal, but only for grant operations and user queries: the allowlist in `internal/api/security.go:110-116` is exactly `CreateGrant`, `GetGrant`, `DeleteGrant`, `QueryGrants`, `QueryUsers`. Teams are missing.

- Scope is the reads the post-claim console actually performs: `queryTeams` and `getTeam`. This is not a blanket "sessions can call everything" change.
- Secret-based callers keep working unchanged.
- CSRF protection for cookie-authenticated calls is tracked separately in #1140; these are reads, not mutations, so nothing here depends on it.

## Dependencies

- PR #1142 (merged). The session-as-principal mechanism this extends.

## Test first (TDD)

Write API tests first: a session-authenticated `POST /teams/query` scoped to the platform project returns the caller's personal team; a session-authenticated `GET /teams/{id}` returns a team the caller can read; a secret-authenticated call behaves exactly as before; a session hitting a team operation not on the allowlist is still refused. Then implement.

## Tasks

- Add `nextgenSession: []` to the security blocks of `api/openapi/endpoints/teams/query/methods.yaml` and the `getTeam` operation in `api/openapi/endpoints/teams/by_id/methods.yaml`, and regenerate.
- Extend the session op allowlist in `internal/api/security.go` with the two team read operations.
- Check whether the embedded-console e2e lane (`apps/console-e2e/src-embedded/embedded-console.spec.ts`) can drop its "fails closed" expectation for the teams list.

## Reference

`internal/api/security.go:100-116` (the two allowlists); PR #1142 (how grants were added); ADR 053 §5.

wim07101993:
## Summary

Team query and get accept a Console session cookie as the signed-in user, so the team list the claim flow lands on loads without a project secret. Team writes are unchanged and still require one.

## Validation

- `go test ./...`

## Release notes / changeset

- Changeset: .changeset/session-team-reads.md — accept console session to read teams (list @zitadel/server for server changes)

## Notes

Closes https://github.com/zitadel/nextgen/issues/1227


wim07101993:
@copilot can you fix the merge conflicts?

internal/api/integration_test/team_session_test.go
```diff
@@ -0,0 +1,186 @@
...
+	})
+
+	t.Run("cannot list teams of a project it has no foothold in", func(t *testing.T) {
+		resp, err := session.QueryTeams(t.Context(), &api.QueryTeamsRequest{},
+			api.QueryTeamsParams{ProjectID: api.ProjectID(otherProject.ID)})
+		require.NoError(t, err)
+		if listed, ok := resp.(*api.QueryTeamsResponse); ok {
+			assert.False(t, teamListed(listed.Teams, foreignTeamID),
+				"a foreign project's teams must never appear")
+		}
```
IAM-marco: Bug: this list check cannot fail.

What changed: the subtest lists the teams of a project the signed-in user has no foothold in, then only asserts when the response is a successful team list.

Why here: this caller has no team in that project at all. For a caller with no foothold, the list gate in `internal/api/authz.go` (`requireProjectListAccess`) returns a not-found decision, which the handler answers as a 404, a `QueryTeamsNotFound` value. So the type check on line 172 is always false and the assert inside never runs.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L168-L176

What follows: the subtest is green as long as the request completes. If a future change let a stranger list a foreign project's teams with a 200, this test would still pass, which is the exact regression its name promises to catch.

Fix: require the response type directly, as the getTeam case three lines above does with `GetTeamNotFound`. If the intended contract is "either 404 or an empty list", assert both branches explicitly instead of skipping one.

Coverage: this is the only test of a session user listing a project they have no foothold in, so nothing else would catch it.

internal/api/integration_test/team_session_test.go
```diff
@@ -0,0 +1,186 @@
...
+		assert.Empty(t, listed.Teams,
+			"a signed-in user with no grant must not see any team, including its own")
+	})
+
+	t.Run("cannot read its own team without a grant", func(t *testing.T) {
+		resp, err := session.GetTeam(t.Context(), api.GetTeamParams{TeamID: api.TeamID(strangerTeamID)})
+		require.NoError(t, err)
+		_, readable := resp.(*api.TeamResponse)
+		assert.False(t, readable, helpers.MustMarshal(t, resp))
+	})
```
IAM-marco: Bug (judgement call): pin the exact denial.

What changed: this subtest asserts only that the response is not a `TeamResponse`.

Why here: any error type satisfies that, including a 500 or a validation error. The foreign-project case just below requires `GetTeamNotFound` exactly, so the two neighbours hold the code to different standards.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L155-L166

What follows: a regression that swaps the intended refusal for a different error code stays invisible.

Fix: require the specific type this design intends, 403 permission denied or 404 not found. The caller does have a foothold here, so it may differ from the foreign case. Whichever it is, name it.

Coverage: no other test reads a session user's own team without a grant.

internal/api/integration_test/team_session_test.go
```diff
@@ -0,0 +1,186 @@
...
+func TestTeamReadsAcceptSession(t *testing.T) {
+	t.Parallel()
+
+	console := harness.EnsurePlatformProject(t)
+	claimerID, claimerTeamID := harness.CreateUserOwnedByTeam(t, console.ID)
+	harness.SeedOwningTeam(t, console.ID, claimerTeamID)
+
+	session, err := helpers.NewApiClient(harness.EnsureTestServer(t).URL)
+	require.NoError(t, err)
+	session.SetSessionToken(platformSessionCookie(t, claimerID).Value)
```
IAM-marco: Guideline note: this helper already exists.

What changed: three lines build a client and set the session cookie for a user.

Why here: the same package already has `sessionClientForUser` in `my_projects_test.go` (lines 21 to 28) doing exactly this, used seven times there. The same three lines appear again at lines 134 to 136 of this file.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L33-L35

What follows: two more copies to keep in step if the cookie name or client setup changes.

Fix: call `sessionClientForUser(t, claimerID)` here and `sessionClientForUser(t, strangerID)` in the second test.

Coverage: not a behaviour change, no test needed.

apps/console-e2e/src-embedded/embedded-console.spec.ts
```diff
@@ -45,10 +45,12 @@ test("signs in end to end against the embedded API", async ({ page, seed }) => {
...
-  // The shell, not the screen: this lane carries no project secret, so the list
-  // itself fails closed until session-derived authorization lands (Console ADR
-  // 0003). The sidebar renders either way, and it is what shows the signed-in
-  // console was reached.
+  // The shell, not the screen. #1227 taught queryTeams to accept the session
+  // cookie, so the list no longer 401s -- but this lane's seeded user has no
+  // team and no grant, so it has no foothold in the console project and the
+  // call answers 404 instead. Rows need a grant, which is what `src-real/`
+  // has a project secret for. The sidebar renders either way, and it is what
+  // shows the signed-in console was reached.
```
IAM-marco: Comment guidance: the comment claims a 404 that nothing checks.

What changed: the comment now says the teams list call answers 404 instead of 401 since #1227.

Why here: the only assertion in this test is that the primary navigation is visible, which it was before this change too. The sibling test in this file does inspect requests, so the pattern exists.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/apps/console-e2e/src-embedded/embedded-console.spec.ts#L48-L56

What follows: issue #1227 asked to check whether this lane can drop its fails-closed expectation. Right now that question is answered in prose only. If the call starts answering 401 again, this test stays green.

Fix: either wait for the `/teams/query` response and assert its status, or trim the comment to what the test actually proves (the shell renders) and leave the status claim out.

Coverage: nothing in the e2e lanes asserts this call's status today.

internal/api/integration_test/team_session_test.go
```diff
@@ -0,0 +1,186 @@
...
+			api.QueryTeamsParams{ProjectID: api.ProjectID(console.ID)})
+		require.NoError(t, err)
+		listed, ok := resp.(*api.QueryTeamsResponse)
+		require.True(t, ok, helpers.MustMarshal(t, resp))
+		assert.True(t, teamListed(listed.Teams, claimerTeamID))
+
+		getResp, err := secret.GetTeam(t.Context(), api.GetTeamParams{TeamID: api.TeamID(claimerTeamID)})
+		require.NoError(t, err)
+		require.IsType(t, &api.TeamResponse{}, getResp, helpers.MustMarshal(t, getResp))
+	})
```
IAM-marco: Simplification: this subtest re-proves what `team_test.go` already covers.

What changed: a secret-authenticated client lists and gets the same team.

Why here: `TestGetTeam` and `TestQueryTeams` in `team_test.go` already run the secret path against both endpoints, and the secret path returns before the session hook runs in `HandleNextgenSession`, so this diff cannot alter it.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L55-L70

What follows: 16 lines that can only fail for reasons the older tests would also catch.

Fix: drop the subtest. One honest caveat: issue #1227 lists "a secret-authenticated call behaves exactly as before" as a required test, so if you want a named test that pins that sentence, keep it and ignore this note.

Coverage: covered by the existing team tests.

internal/api/integration_test/team_session_test.go
```diff
@@ -0,0 +1,186 @@
...
+		}{
+			{"create", http.MethodPost, "/teams?project_id=" + console.ID, `{"name":"x"}`},
+			{"update", http.MethodPatch, "/teams/" + claimerTeamID, `{"name":"x"}`},
+			{"delete", http.MethodDelete, "/teams/" + claimerTeamID, ""},
+		} {
+			t.Run(tc.name, func(t *testing.T) {
+				var body io.Reader
+				if tc.body != "" {
+					body = strings.NewReader(tc.body)
+				}
```
IAM-marco: Simplification: the body guard is not needed.

What changed: a nil-or-reader switch chooses the request body, and the `io` import exists only to name that variable.

Why here: `strings.NewReader("")` is a valid empty body. The server sees the same request either way.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L104-L107

What follows: four lines and an import for a distinction the server does not make.

Fix: pass `strings.NewReader(tc.body)` straight into `NewRequestWithContext` and drop the `io` import on line 6. Keep the Content-Type guard if you like, or set it unconditionally.

Coverage: no behaviour change.

internal/api/integration_test/team_session_test.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func teamListed(teams []api.TeamResponse, id string) bool {
+	for _, team := range teams {
+		if team.ID == id {
+			return true
+		}
+	}
+	return false
+}
```
IAM-marco: Simplification: the standard library has this loop.

What changed: a small helper walks the team list looking for an id.

Why here: Go 1.26 is in go.mod, so `slices.ContainsFunc` does the same thing in one expression.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L179-L186

What follows: eight lines for one line of intent.

Fix: `slices.ContainsFunc(teams, func(tm api.TeamResponse) bool { return tm.ID == id })`, either as the helper's body or inline at the three call sites.

Coverage: no behaviour change.

internal/api/integration_test/team_session_test.go
IAM-marco: Bug: this list check cannot fail.

What changed: the subtest lists the teams of a project the signed-in user is not a member of, then only asserts when the response is a successful team list.

Why here: this caller has no team in that project at all. For a caller with no team or grant in the project, the list gate in `internal/api/authz.go` (`requireProjectListAccess`) returns a not-found decision, which the handler answers as a 404, a `QueryTeamsNotFound` value. So the type check on line 172 is always false and the assert inside never runs.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L168-L176

What follows: the subtest is green as long as the request completes. If a future change let a stranger list a foreign project's teams with a 200, this test would still pass, which is the exact regression its name promises to catch.

Fix: require the response type directly, as the getTeam case three lines above does with `GetTeamNotFound`. If the intended contract is "either 404 or an empty list", assert both branches explicitly instead of skipping one.

Coverage: this is the only test of a session user listing a project they are not a member of, so nothing else would catch it.

internal/api/integration_test/team_session_test.go
IAM-marco, line 160: Bug (judgement call): pin the exact denial.

What changed: this subtest asserts only that the response is not a `TeamResponse`.

Why here: any error type satisfies that, including a 500 or a validation error. The foreign-project case just below requires `GetTeamNotFound` exactly, so the two neighbours hold the code to different standards.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L155-L166

What follows: a regression that swaps the intended refusal for a different error code stays invisible.

Fix: require the specific type this design intends, 403 permission denied or 404 not found. The caller is a member of the project here, so it may differ from the foreign case. Whichever it is, name it.

Coverage: no other test reads a session user's own team without a grant.

internal/api/integration_test/team_session_test.go
IAM-marco: Guideline note: this helper already exists.

What changed: three lines build a client and set the session cookie for a user.

Why here: the same package already has `sessionClientForUser` in `my_projects_test.go` (lines 21 to 28) doing exactly this, used seven times there. The same three lines appear again at lines 134 to 136 of this file.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L33-L35

What follows: two more copies to keep in step if the cookie name or client setup changes.

Fix: call `sessionClientForUser(t, claimerID)` here and `sessionClientForUser(t, strangerID)` in the second test.

Coverage: not a behaviour change, no test needed.

apps/console-e2e/src-embedded/embedded-console.spec.ts
IAM-marco: Comment guidance: the comment claims a 404 that nothing checks.

What changed: the comment now says the teams list call answers 404 instead of 401 since #1227.

Why here: the only assertion in this test is that the primary navigation is visible, which it was before this change too. The sibling test in this file does inspect requests, so the pattern exists.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/apps/console-e2e/src-embedded/embedded-console.spec.ts#L48-L56

What follows: issue #1227 asked to check whether this lane can drop its fails-closed expectation. Right now that question is answered in prose only. If the call starts answering 401 again, this test stays green.

Fix: either wait for the `/teams/query` response and assert its status, or trim the comment to what the test actually proves (the shell renders) and leave the status claim out.

Coverage: nothing in the e2e lanes asserts this call's status today.

internal/api/integration_test/team_session_test.go
IAM-marco, line 68: Simplification: this subtest re-proves what `team_test.go` already covers.

What changed: a secret-authenticated client lists and gets the same team.

Why here: `TestGetTeam` and `TestQueryTeams` in `team_test.go` already run the secret path against both endpoints, and the secret path returns before the session hook runs in `HandleNextgenSession`, so this diff cannot alter it.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L55-L70

What follows: 16 lines that can only fail for reasons the older tests would also catch.

Fix: drop the subtest. One honest caveat: issue #1227 lists "a secret-authenticated call behaves exactly as before" as a required test, so if you want a named test that pins that sentence, keep it and ignore this note.

Coverage: covered by the existing team tests.

internal/api/integration_test/team_session_test.go
IAM-marco: Simplification: the body guard is not needed.

What changed: a nil-or-reader switch chooses the request body, and the `io` import exists only to name that variable.

Why here: `strings.NewReader("")` is a valid empty body. The server sees the same request either way.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L104-L107

What follows: four lines and an import for a distinction the server does not make.

Fix: pass `strings.NewReader(tc.body)` straight into `NewRequestWithContext` and drop the `io` import on line 6. Keep the Content-Type guard if you like, or set it unconditionally.

Coverage: no behaviour change.

internal/api/integration_test/team_session_test.go
IAM-marco, line 183: Simplification: the standard library has this loop.

What changed: a small helper walks the team list looking for an id.

Why here: Go 1.26 is in go.mod, so `slices.ContainsFunc` does the same thing in one expression.
https://github.com/zitadel/nextgen/blob/f868e38372352c0a12c5d69e9bc79bc68e4d8f67/internal/api/integration_test/team_session_test.go#L179-L186

What follows: eight lines for one line of intent.

Fix: `slices.ContainsFunc(teams, func(tm api.TeamResponse) bool { return tm.ID == id })`, either as the helper's body or inline at the three call sites.

Coverage: no behaviour change.

IAM-marco (changes requested):
Two-axis review of the session-cookie change for team reads. The production change itself (two OpenAPI security blocks, two allowlist entries) matches issue #1227 and the way PR #1142 added grants. All findings are in the tests and comments.

**Code review:** 4 inline findings. Worst: the foreign-project list test can never fail. Two notes with no diff line to hang on: the new `SeedOwningTeam` helper duplicates two inline calls that still live in `claim_test.go` around line 246, and nothing checks that the OpenAPI security blocks and the Go allowlist stay in sync (both pre-existing, not introduced here).

**Ponytail (simplification only):** 3 inline findings, about 28 lines that could go. Worst: the secret-caller subtest re-proves what `team_test.go` already covers.

| Axis | Findings | Worst |
|---|---|---|
| Code review | 4 inline, 2 body notes | foreign-project list assertion can never fail |
| Ponytail | 3 inline | duplicate secret-caller subtest, 16 lines |

IAM-marco (approved):
All seven review findings are addressed at a809961c: the foreign-project list test now requires the 404 type, the own-team case pins 403 team.permission_denied, the session client helper is reused, the e2e lane asserts the teams query status, the body guard and the hand-rolled search are gone, and the claim test uses the new seed helper. The secret-caller subtest stays by choice, as the comment allowed. CI is green after a rerun; the one earlier failure was a handoff token expiring on the slow Spanner emulator, unrelated to this change.

PR #1241, opened 2026-09-16
feat: don't allow team to be deleted if it is owner of a project

### LINKED ISSUE #1231: Block deactivating a team that owns a project
**TL;DR:** you can deactivate the team that owns your claimed project, and the project becomes unmanageable. Block that.

Follow-up from the #769 discussion; implements the invariant recorded in ADR 054.

## Goal

A team that owns a project cannot be deactivated while it still owns it.

## Current state

`TeamService.Delete` (`internal/service/team.go:272-296`) deactivates the team and cascades to its memberships and lifecycle-owned users. It checks nothing about ownership. Claimed projects are owned by a team through the owning grant (a `project.team` row in `authz_assignments`, kept unique per project by the `authz_assignments_one_owning_team` index), so deactivating the claimer's personal team leaves the project owned by a dead team: nobody can manage it anymore.

In the #769 thread, an API-level "last admin cannot be removed" rule was proposed and countered with this guard instead. The guard is the right one: removing every granted admin is harmless (the owning team keeps the project manageable), while deactivating the owning team is the path that actually orphans a project.

- The guard lives in the service layer, inside the same transaction as the deactivation, so it cannot race a concurrent claim.
- The error is a plain, visible domain error following the existing `domain.Err*` conventions. A team you are allowed to deactivate is yours to see, so there is nothing to hide.
- There is deliberately no "last admin grant" rule: removing the last granted admin is legal and just returns the project to its post-claim state.

## Dependencies

- None.

## Test first (TDD)

Write service tests first: deactivating a team holding an active owning grant fails with the new error and changes nothing; deactivating a team with no owning grants behaves as today; deactivating a team whose owning grant is revoked or expired succeeds; deleting an already-deactivated team stays an idempotent no-op. Then implement.

## Tasks

- Add the owning-grant check inside the transaction in `TeamService.Delete` (`internal/service/team.go`).
- Add the error to `internal/domain` and to the OpenAPI error contract for `deleteTeam`.

## Reference

`docs/adrs/054-customer-collaboration-grants.md` (the invariant); `internal/domain/authz.go:315-331` (`NewClaimTeamAssignment`, the owning grant this checks for); the #769 thread (the last-admin proposal and its counter-proposal, which this implements).

wim07101993:
## Summary

Deleting a team that still owns a project is refused with `409 team.owns_project` instead of deactivating it and leaving the project unmanageable. Transfer or revoke the project ownership first.

## Validation

- `go test ./...`

## Release notes / changeset

- .changeset/block-deactivating-owning-team.md - don't allow team to be deleted if it is owner of a project (list @zitadel/server for server changes)

## Notes

closes: https://github.com/zitadel/nextgen/issues/1231


internal/service/team.go
```diff
@@ -278,6 +278,13 @@ func (s *TeamService) Delete(ctx context.Context, projectID, teamID string) erro
 		if !changed {
 			return nil
 		}
+		ownsAProject, err := tx.Statements().HasActiveOwningTeamGrant(ctx, teamID)
+		if err != nil {
+			return err
+		}
+		if ownsAProject {
+			return domain.ErrTeamOwnsProject()
```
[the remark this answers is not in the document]
wim07101993: @IAM-marco wdyt?

wim07101993:
@copilot can you solve the merge conflicts?

internal/service/team.go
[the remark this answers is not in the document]
IAM-marco: You're right. ADR 054 section 8 says deactivating an owning team is allowed and leaves its projects in `needs_owner`. This PR does the opposite, and the code comments that cite the ADR as support are wrong.

We are keeping the rejection and amending the ADR, not the other way round. Here is why.

The `needs_owner` path only works if someone can rescue an ownerless team afterwards. That rescue does not exist yet: nothing in the code shows `needs_owner`, the "final owner cannot be removed" rule is not built, and there is no platform-project admin endpoint that hands ownership to someone else. Until those exist, letting a team delete itself while it owns a project means that project can never be managed again. Refusing is the safe rule for now.

It also fits the ADR's own reasoning. Section 8 lets a suspended user leave a team ownerless because suspending a user is a security action that must not be blocked. Deleting a team is not a security action. It is the team's owner tidying up, so it belongs with "ordinary owner removal", which the ADR already refuses.

What this PR will change:

- Amend ADR 054 section 8 with a dated note: until the ownerless-team recovery is built (tracked in #1261), deleting a team that holds an active owning-team assignment is refused with `team.owns_project`. Ownership has to be transferred or revoked first. `needs_owner` currently only comes from suspending or deactivating the final owner user, which stays allowed. The note also says that in the API deletion and deactivation are the same operation (`DELETE /teams/{id}` tombstones the team), so the two words in the ADR and the API mean one thing. #1261 decides whether the refusal stays once recovery exists. README row for 054 becomes "Proposed (amended)".
- Fix the comments in the test and on `HasActiveOwningTeamGrant` that point at section 2 for the refusal. Section 2 only says how ownership ends. The section 2 reference for "owning grants have no expiry" is correct and stays.

The rest of section 8 (last-owner refusal, `needs_owner`, platform-admin rescue) is #1261, outside the claiming milestone.

internal/storage/dialect/spanner/authz_assignment.go
```diff
@@ -213,6 +221,22 @@ func (s authzAssignmentStatements) GetActiveOwningTeamGrant(ctx context.Context,
...
+		if qErr != nil {
+			return qErr
+		}
+		return row.Columns(&owns)
+	})
+	if err != nil {
+		return false, err
+	}
+	return owns, nil
+}
```
IAM-marco, line 240: Bug: this Spanner method reads the row by hand and never stops the iterator, unlike the function it calls its mirror.

What changed: the new `HasActiveOwningTeamGrant` calls `iter.Next()` and `row.Columns` itself and returns the raw error.

Why this spot matters: the function right above it, `GetActiveOwningTeamGrant`, goes through the shared `collectOneRow` helper (https://github.com/zitadel/nextgen/blob/e66500b5d729202ae9e9c28710f28a27d46eb591/internal/storage/dialect/spanner/authz_assignment.go#L207-L222), and so does `UserExists`, which has the exact same "one boolean from an EXISTS query" shape (https://github.com/zitadel/nextgen/blob/e66500b5d729202ae9e9c28710f28a27d46eb591/internal/storage/dialect/spanner/user.go#L367-L380). That helper defers `iter.Stop()` and rejects a second row.

What follows: the guard runs inside `TeamService.Delete`'s transaction, and the transactional `Query` in this dialect does not stop the iterator for you (https://github.com/zitadel/nextgen/blob/e66500b5d729202ae9e9c28710f28a27d46eb591/internal/storage/dialect/spanner/db.go#L94-L97), only the single-use client path does. So every team delete on Spanner, refused or allowed, leaves a row iterator open until the transaction ends. The raw error return also skips `wrapError`, so a transient Spanner failure comes back unclassified instead of as a typed database error the way the siblings return it.

Separate note: `ExistsResourceScopeElsewhere` has the same shape today, so the repo already has two competing patterns. This PR picked the weaker one.

Fix direction: use `collectOneRow` with a one-line scan closure, like `UserExists`, and wrap the error.

Coverage: the stmttest suite runs this against the Spanner emulator but only checks the boolean, so a leaked iterator or an unwrapped error would not fail any existing test.

internal/service/team_test.go
```diff
@@ -289,6 +290,36 @@ func TestTeamService_Delete(t *testing.T) {
...
+		{
+			// A revoked owning grant is not ownership; the guard reads active
+			// rows only, so this team deactivates like any other.
+			name:   "team whose owning grant was revoked deactivates",
+			teamID: "team_1",
+			setupStmt: func(s *servicemocks.MockAllStatements) {
+				s.EXPECT().DeactivateTeam(gomock.Any(), "proj_1", "team_1").Return(true, nil)
+				s.EXPECT().HasActiveOwningTeamGrant(gomock.Any(), "team_1").Return(false, nil)
+			},
+		},
```
IAM-marco: Simplification: this table case is byte-for-byte the same as the "ok" case above it.

The mock setup here (deactivate returns true, ownership check returns false, no error) is identical to the "ok" case (https://github.com/zitadel/nextgen/blob/e66500b5d729202ae9e9c28710f28a27d46eb591/internal/service/team_test.go#L263-L270). From the mock's point of view "the grant was revoked" and "there never was a grant" are the same answer, so this case runs the same code path and asserts the same nil error. The revoked-versus-active distinction is real, but it lives in the SQL, and the stmttest assertion "a revoked grant is not ownership" already covers it.

Fix direction: delete this case. Nothing replaces it.

internal/storage/stmttest/claim_test.go
```diff
@@ -383,5 +383,71 @@ func TestClaimStatements_OwningTeamGrant(t *testing.T) {
...
+			owned := ensureProject(t, d.stmts)
+
+			foreign := "team-foreign-" + uniqueSuffix(t)
+			require.NoError(t, d.stmts.CreateAuthzAssignment(t.Context(),
+				domain.NewClaimTeamAssignment(owned, foreign)))
+
+			owns, err := d.stmts.HasActiveOwningTeamGrant(t.Context(), foreign)
+			require.NoError(t, err)
+			assert.True(t, owns, "ownership of another project must still count")
+		})
```
IAM-marco: Simplification: this subtest cannot prove what its comment says, because the query has no project to scope by.

`HasActiveOwningTeamGrant` takes only a team id, so there is no per-project scoping for the extra "notional home" project to disprove. The subtest right above it (https://github.com/zitadel/nextgen/blob/e66500b5d729202ae9e9c28710f28a27d46eb591/internal/storage/stmttest/claim_test.go#L389-L413) already creates a project, grants ownership to a team, and asserts the lookup returns true. This one does the same thing with one unused project created first.

Fix direction: delete the subtest and move its one useful sentence (the lookup is keyed on the team because the row lives on the owned project) into the doc comment or the previous subtest's comment. Nothing else replaces it.

internal/domain/team.go
```diff
@@ -44,6 +44,21 @@ func ErrTeamPermissionDenied() Error {
...
+// which would leave that project without an owner.
+//
+// The message states the condition and does not tell the caller how to clear
+// it, because today nothing can: the grants API handles only viewer, editor
+// and admin rows, so an owning-team row cannot be revoked through it, and
+// there is no ownership-transfer endpoint. Promising a remedy the API does not
+// offer would be worse than saying nothing.
+//
+// Visible rather than hidden: a team the caller may deactivate is one they may
+// already see, so the reason gives away nothing.
```
IAM-marco: Simplification: an eleven-line doc comment on an error constructor whose siblings have none.

Every other `Err*` constructor in this file carries no comment at all (https://github.com/zitadel/nextgen/blob/e66500b5d729202ae9e9c28710f28a27d46eb591/internal/domain/team.go#L30-L46). The two rationales here (why the message names no remedy, why the reason is safe to show) are review-thread answers, and the first one is already written for the audience that needs it in the `deleteTeam` description in `api/openapi/endpoints/teams/by_id/methods.yaml`.

Fix direction: keep one line saying what the error means. If the "no remedy exists yet" note must live in code, one sentence pointing at #1261 is enough.

internal/service/statement.go
```diff
@@ -459,6 +459,17 @@ type AuthzAssignmentStatements interface {
...
+	// HasActiveOwningTeamGrant reports whether the team still owns a project.
+	// It is the mirror of GetActiveOwningTeamGrant: that one asks a project who
+	// owns it, this one asks a team what it owns, so the lookup is keyed on the
+	// team and spans every project. Ownership rows sit on the owned project,
+	// which for a claim is not the team's own project.
+	//
+	// Active means unrevoked. Expiry is not consulted because an owning-team
+	// grant cannot carry one: the authz_assignments CHECK rejects expires_at on
+	// (project, team) rows, since ADR 054 §2 ends ownership only by transfer or
+	// revocation.
```
IAM-marco: Simplification: a ten-line doc comment where the neighbouring statements get three or four.

`GetActiveOwningTeamGrant` a few lines up and `ListClaimedProjectIDs` below both get a short comment. This one re-derives from the schema why expiry is not consulted, which the CHECK constraint in the migrations already states. The second paragraph is also the one that cites ADR 054 section 2 as the reason for the refusal, which the Copilot thread already agreed to fix.

Fix direction: one sentence for what it returns, one for "keyed on the team because the owning row sits on the owned project", and keep the expiry sentence only as a pointer to the CHECK constraint.

IAM-marco (commented):
Two-axis review of this branch against main. The ADR 054 section 8 clash is not repeated here: it is already agreed on the Copilot thread (amend the ADR in this PR, follow-up is #1261). One reminder from that thread: the test case comment in `internal/service/team_test.go` and the doc comment on `HasActiveOwningTeamGrant` still cite ADR 054 section 2 as the reason for refusing. Section 2 only says how ownership ends, so those two citations should point at the amended section 8 when the ADR change lands.

## Code review

One finding. The Spanner version of `HasActiveOwningTeamGrant` reads the row by hand instead of using the shared one-row helper, so on the transactional path nothing stops the iterator and the error is not classified like its siblings. Details inline.

Checked and fine: the guard sits inside the same transaction as the deactivation and the Postgres helper rolls back on any returned error, so a refused delete really leaves the team, its memberships and its owned users untouched. The 409 mapping matches how the other packages map "current state refuses". The OpenAPI wiring follows the sibling endpoints. Expired owning grants need no test because all three dialects carry a CHECK constraint that forbids an expiry on an owning-team row, and an existing test already proves that constraint fires.

## Ponytail

Four cuts, all small: two duplicate test cases that exercise the exact same path as a neighbour, and two doc comments several times longer than every sibling in their file. Details inline.

## Tally

Code review: 1 finding. Worst: the Spanner iterator is never stopped on the transactional path.
Ponytail: 4 findings, about 40 lines removable. Worst: the "found across projects" subtest proves nothing the previous subtest does not.

This review is pending. Nothing is submitted until you submit it.

IAM-marco (changes requested):
Requesting changes for the items in my review above (https://github.com/zitadel/nextgen/pull/1241#pullrequestreview-5265456233) plus the agreed ADR 054 section 8 amendment from the Copilot thread. The Spanner iterator fix and the ADR amendment are the blocking ones; the four simplifications are cheap to take in the same pass.

internal/service/team.go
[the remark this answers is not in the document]
IAM-marco: Done in 4472632f: ADR 054 section 8 now carries a dated amendment saying the refusal is temporary until ownerless-team recovery (#1261) exists, the README row is "Proposed (amended)", and the comments that cited section 2 for the refusal now point at section 8.

internal/storage/dialect/spanner/authz_assignment.go
IAM-marco: Fixed in 4472632f: uses collectOneRow and wrapError, same shape as UserExists.

internal/service/team_test.go
IAM-marco: Removed in 4472632f.

internal/storage/stmttest/claim_test.go
IAM-marco: Removed in 4472632f; the keyed-on-the-team sentence moved into the previous subtest's comment.

internal/domain/team.go
IAM-marco: Trimmed to two lines in 4472632f, pointing at #1261.

internal/service/statement.go
IAM-marco: Trimmed in 4472632f; the section 2 reference now only backs the no-expiry sentence.

IAM-marco (approved):
Approved. The ADR 054 amendment, the Spanner iterator fix and the tidy-ups from my review are in 4472632f, and the follow-up for ownerless-team recovery is #1261.

PR #1245, opened 2026-09-16
docs: fixed id and revision id on every revisioned resource (ADR 063)

grvijayan:
## Summary

Records the resource versioning agreement from the review of #1217.

Every revisioned resource carries a fixed `id`, shared by all its revisions, and a `revision_id` per revision. Handles stay per kind and resolve through the kind's list filter. Get-by-id returns the newest revision, the list returns each resource once at its newest revision, and every kind has `GET /<kind>/{id}/revisions` and `GET /<kind>/revisions/{revision_id}`, so a pinned revision stays fetchable. Releases, auth attempts and user records pin `revision_id`; identity links reference a connection's `id`. A release pointer also records the resource `id`, so it becomes `{kind, handle, id, revision_id}`.

Connections (#1217) already have the two ids. Schemas, flow definitions and branding migrate when next touched, each in its own ticket, and before they can be pinned by a release. Each migration also switches the existence check at release creation from get-by-id to a lookup by `revision_id`, and that lookup records the resource `id` on the pointer.

If existing rows are carried over, they keep their current id as `revision_id`, so releases, users and flow steps stay resolvable, while get-by-id stops accepting those values. Whether they are carried at all, and where a fixed `id` then comes from, is an open question for the first migration ticket.

## Amendments

- **ADR 035**: the Revision column of the release table is the revision's `revision_id`, and the idp handle is `slug`. Header line plus inline notes at the release table and the pointer paragraph.
- **ADR 047 §2 and §5**: a schema's resource id is always dialect-minted; a declared `$id` stays a document property. Rows identified by a URI today keep that URI as their `revision_id`, because releases already pin it.
- Index rows updated for 035 and 047; new row for 063. The ADR is 063 because main took 062 for the variables ADR (#1145).

## Validation

- `node scripts/check-adrs-index.mjs`: ok, 63 records.
- `.redocly.lint-ignore.yaml`: the pinned `no-ambiguous-paths` entry from #1217 now cites ADR 063.


docs/adrs/035-configuration-environments.md
```diff
@@ -553,3 +554,13 @@ above already shows. Handle resolution is per kind, not a shared column read.
...
 uses, not the model.
+
+## Amendment (2026-09-16): fixed id and revision id
+
+[ADR 062](062-resource-revisions-fixed-id-and-revision-id.md) gives every
+revisioned resource a fixed `id` shared by all its revisions and a
+`revision_id` per revision. The [Releases](#releases) table above lists a
+resource id in its Revision column, which held while a kind allocated a new id
+per revision. Under ADR 062 that column is the revision's `revision_id`, and
+the idp row's handle is `slug`. The pointer tuple `{kind, handle, revision_id}`
```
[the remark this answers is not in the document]
grvijayan: Fixed in 9f054251. An inline note under the endpoint table now says a pinned revision is fetched through the per-kind read by `revision_id`, since get-by-id returns the newest revision.

docs/adrs/062-resource-revisions-fixed-id-and-revision-id.md
```diff
@@ -0,0 +1,165 @@
...
+### 1. Two ids on every revisioned resource
+
+Every revisioned resource carries two ids:
+
+- `id` is allocated once, when the resource is created, and is shared by
+  every revision of it.
+- `revision_id` is allocated per revision.
+
+Both are dialect-minted prefixed ids per ADR 047. Each kind registers two
+prefixes, one for the resource and one for the revision.
```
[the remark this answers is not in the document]
grvijayan: Fixed in 9f054251. Decision 1 now applies to newly allocated ids, and the schemas migration bullet carries the URI exception: rows identified by a URI today keep it as their `revision_id`, because releases already pin it.

docs/adrs/062-resource-revisions-fixed-id-and-revision-id.md
```diff
@@ -0,0 +1,165 @@
...
+
+### 5. References
+
+A resource can be referenced in three ways:
+
+- **By handle**, inside a release. Resources in a release reference each other
+  by handle, and a handle means the revision the release pins for it.
+  Unchanged from ADR 035.
+- **By `id`**, when the reference must outlive handle changes and revisions.
+  An identity link references a connection's `id`.
```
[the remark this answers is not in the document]
grvijayan: Fixed in 9f054251. The `id` bullet now says the reference must outlive revisions. Handles stay fixed, as section 3 says.

docs/adrs/063-resource-revisions-fixed-id-and-revision-id.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Decision
+
+### 1. Two ids on every revisioned resource
+
+Every revisioned resource carries two ids:
+
+- `id` is allocated once, when the resource is created, and is shared by
+  every revision of it.
+- `revision_id` is allocated per revision.
```
vitorbari, line 64: What is the format of the `revision_id`? (numeric, prefixed, uuid, etc?)

docs/adrs/062-resource-revisions-fixed-id-and-revision-id.md
```diff
@@ -0,0 +1,167 @@
...
+### 4. Reads
+
+Get-by-id returns the newest revision of the resource.
+
+A list takes a `revisions` parameter with two values:
+
+- `latest` returns the newest revision of each resource, one row per resource.
+- `all` returns every revision.
+
+Each kind picks its own default and states it in the operation's description.
```
wim07101993: why let each resource specify it's own default? Wouldn't it make sense to pick a default in this ADR, given that the ADR tries to generalize and add consistency across resources?

wim07101993 (commented):
Some ai comments:

Which id inherits the existing prefix isn't stated. §1 requires two prefixes per kind; §6 says prefixes "are not decided here." But the migration's back-compat story depends on the answer. It says URI-identified schema rows keep the URI as revision_id — by the same logic (releases already pin them) sch_*-identified rows must also become revision_ids, which means the fixed id needs a brand-new prefix. That's implied but only spelled out for the URI case. Stating it for sch_*/flowdef_*/brnd_* would close the gap.

The public API break isn't in Consequences. Migration says "An id that a release, an attempt or a user record holds stays resolvable," which covers internal holders. But an external client that stored a sch_* and calls GET /schemas/{id} is calling get-by-id with what becomes a revision id, and §4 redefines get-by-id to return "the newest revision of the resource." Consequences lists the registry growth and the per-revision reads but not this. Given ADR 035 tells release consumers to resolve pointers themselves, some of them will have stored these ids.

Two small inconsistencies

"Connections already have this shape" appears in both the PR body and the Migration section, but Consequences admits "Connections have no per-revision read at all; a follow-up to #1217 adds one" — while §4 makes that read mandatory. So no kind fully satisfies the ADR when it lands. Fine, but the Migration wording overstates it; "already have the two ids" would be accurate.

Also, §4 cites POST /idps/query as following the page-token rule "today," and §3 lists slug as the connections handle — but #1217 is still open and api/openapi/endpoints/idps/ doesn't exist. The present tense reads as shipped. Worth marking those as "as defined in #1217," which the Context table already does correctly.

docs/adrs/063-resource-revisions-fixed-id-and-revision-id.md
grvijayan: As defined in [047](https://github.com/zitadel/nextgen/blob/main/docs/adrs/047-dialect-id-generation.md#1-one-id-class-dialect-minted-prefix_opaque), it'd be a prefix + opaque ID (in case of idps, it'd be `idp_120...` for id and `idprev_345...` for revision_id). I've updated the ADR to specify it.

docs/adrs/062-resource-revisions-fixed-id-and-revision-id.md
grvijayan: Do we know what the default should be? For example, the list schemas command in the CLI lists all the revisions of a schema, but the console displays only the latest revisions. Which use case should we design for? Probably `latest` makes more sense to keep the fetch minimal instead of listing everything unless specifically asked for.

docs/adrs/062-resource-revisions-fixed-id-and-revision-id.md
```diff
@@ -0,0 +1,167 @@
...
+
+Each kind keeps its handle:
+
+- `objectType` for schemas
+- `name` for flow definitions
+- `slug` for connections
+- the constant `default` for branding, which has no naming field
+
+A handle is unique within a project and fixed for the life of the resource.
+It resolves through the kind's list filter.
```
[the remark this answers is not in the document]
grvijayan: Per kind. Reworded to "unique within a project for its kind".

docs/adrs/063-resource-revisions-fixed-id-and-revision-id.md
```diff
@@ -0,0 +1,167 @@
+# ADR 062: Resource Revisions: Fixed Id and Revision Id
+
+> **Status:** Proposed
```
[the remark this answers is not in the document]
grvijayan: ADR 052 is the precedent: Proposed, an unconditional "Amends" line, and "Amended by" notes in ADRs 009 and 020. This ADR follows that form. The status moves to Accepted on merge.

docs/adrs/062-resource-revisions-fixed-id-and-revision-id.md
wim07101993: yes, latest feels correct. However, it might also be usefull to have for each resource a `{resource}/revisions` route which does allow to query the revisions and a `{resource}/revisions/{id}` to get a specific revision. wdyt?

wim07101993: Or we would need a way to mark a revision as default as was mentioned in a discussion when versioning schemas.

grvijayan: get all revisions of a resource (by `slug`, for example), or get-by-revision_id should suffice, right? do we need an explicit endpoint? 🤔

grvijayan: > Or we would need a way to mark a revision as default as was mentioned in a discussion when versioning schemas.

could you please elaborate?

docs/adrs/062-resource-revisions-fixed-id-and-revision-id.md
```diff
@@ -0,0 +1,215 @@
...
+request, and the other way round. Today `GET /schemas` has both modes with
+`all` as its default (#957), and `POST /idps/query` in #1217 has `latest` as
+its default.
+
+Every kind also has a read by `revision_id`, so a pinned revision can be
+fetched after newer ones exist. Two callers depend on it: the release service
+validates each pointer by reading the revision it names, and ADR 035 tells
+release consumers to fetch content the same way. The read must exist before a
+kind can be pinned by a release. Whether it is a path or a list filter is
+decided per kind in its migration ticket.
```
vitorbari: I would change this to follow the new data model. There is no more correlations, it is a one to many relationship.

- `GET /<resource>` / `POST /<resource>/query` returns resources + latest revision. We can also explore ideas similar to the `expands` used to return related data. Maybe `expands=latest`, otherwise, only the top level data is returned.
- `GET /<resource>/<resource_id>` returns the resource + latest revision
- `GET /<resource>/<resource_id>/revisions`
- `GET /<resource>/<resource_id>/revisions/<rev_id>`

I think there is also the need to retrieve the revision without knowing the resource_id (because that is what we store in a release. There is not a straight-forward rest-line naming pattern, but something like:
`GET /<resource>/revisions/<rev_id>`

grvijayan: If we have explicit endpoints to list revisions, then the query endpoint doesn't need `revisions=latest|all`. The common list/query endpoint can return the latest revision by default. What do you think?

vitorbari: Yes, we can drop `revisions=latest|all`

grvijayan: done in https://github.com/zitadel/nextgen/pull/1245/commits/b3ce52ef0f18eec7e9b326b58acbde955de5068c

docs/adrs/062-resource-revisions-fixed-id-and-revision-id.md
grvijayan: we won't need the `revisions=all|latest` parameter anymore in the list/query endpoint and instead have explicit endpoints for the resource revisions, changes are in https://github.com/zitadel/nextgen/pull/1245/commits/b3ce52ef0f18eec7e9b326b58acbde955de5068c

wim07101993: Lets leave it as is. I think it is not needed for now, and would only make things more complicated

docs/adrs/062-resource-revisions-fixed-id-and-revision-id.md
wim07101993: Why not store both the resource and revision id on a release?

vitorbari: That is doable. It currently holds the "handle" (correlation ID) just because of flat table.

vitorbari: https://github.com/zitadel/nextgen/blob/main/internal/domain/release.go#L73
Now that resource ID and handle are 1 to 1, we can store the resource ID in the release.

docs/adrs/062-resource-revisions-fixed-id-and-revision-id.md
grvijayan: Only schemas and flow definition follow this pattern at the moment. It doesn't make sense to carry this on to new resources if this is anyway not needed. We could create low-prio follow-up tickets to fix schemas/flow definitions, though.

docs/adrs/062-resource-revisions-fixed-id-and-revision-id.md
grvijayan: added in https://github.com/zitadel/nextgen/pull/1245/commits/5d77ef795870c548919dd3aa7ad7811a4c25d5b3

docs/adrs/063-resource-revisions-fixed-id-and-revision-id.md
```diff
@@ -0,0 +1,226 @@
...
+
+A handle is unique within a project for its kind and fixed for the life of
+the resource. It resolves through the kind's list filter.
+
+### 4. Reads
+
+Every resource kind has four reads:
+
+- The list or query returns each resource once, at its newest revision. It
+  carries no history, so it takes no `revisions` parameter.
```
vitorbari, line 94: Please document if the _latest_ is tracked via a pointer in the resource table (parent) or via created_at in the revisions table (children).

vitorbari (approved):

vitorbari (approved):

PR #1246, opened 2026-09-16
feat(api): return the latest revision per name from listFlowDefinitions

vitorbari:
Applies the `revisions` filter from `GET /schemas` (#957) to `GET /flow_definitions`, as promised in https://github.com/zitadel/nextgen/pull/1222#discussion_r4026448704.

## Summary

`GET /flow_definitions` returns one row per revision. That is right for the CLI, which wants the full history, but wrong for the console directory, which wants one row per flow.

This adds the same `revisions` parameter schemas already have:

- `all` (default): every revision. The CLI keeps working unchanged.
- `latest`: the newest revision of each `name` — one row per flow. The console's **Login flows** directory now uses this.

The implementation follows #957 step for step: a unique index on `(project_id, name, created_at)` so "the latest" is well-defined, a `NOT EXISTS` sub-query to keep only the newest revision, page tokens that remember which mode they were issued in, and a 409 when two revisions of one flow are created at the same instant. The page-token helpers are reused from the schema service. `@zitadel/api-mock` mirrors the parameter.

## Differences from #957

Flow `name` is NOT NULL, unlike schema `objectType`, which removes two pieces of complexity:

- No special handling for rows without a grouping key (schemas' "orphan" rows and Spanner's `NULL_FILTERED` index).
- The 409 needs no follow-up query to identify which constraint fired: flow ids are server-minted, so the revision index is the only unique constraint a create can violate.

One caller stays on `all` on purpose: the engine's flow resolution filters by `status = active` and picks the newest match itself. The `latest` sub-query ignores other filters, so using it there would hide a flow whose newest revision is a draft.


vitorbari:
Re the suppressed Copilot remark about the console loader discarding `next_page_token`: deliberate. The flow directory has never paginated — that predates this PR — and with `revisions=latest` a project needs more than 20 distinct flows before a row goes missing. Adding a "Load more" like the schemas directory is a follow-up, not part of this change.

.changeset/flow-definitions-latest-revisions.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zitadel/server": minor
+---
+
+`GET /flow_definitions` takes a `revisions` parameter, matching `GET /schemas`: `all` (the default) keeps returning every revision, and `latest` returns the newest revision of each `name` — one row per flow. A `page_token` is bound to the mode it was issued in and is rejected by the other, and the console's login-flows directory now shows one row per flow rather than one per published revision.
```
grvijayan, line 5: Is there a reason to make `all` as the default?

grvijayan: context: https://github.com/zitadel/nextgen/pull/1245#discussion_r4026620737

vitorbari: This was driven by the actual data structure, which is a flat table. Taking only the latest, is an aggregation (not the default).

But that can change as part of #1245. I will add a comment there.

grvijayan: I guess we don't need this PR anymore

vitorbari: Not sure when exactly we are going to implement the changes proposed by #1245, so I would prefer to merge this PR to help with the CLI commands being added by #1222

grvijayan (approved):

Issue #1073, opened 2026-08-31
Storage - SSO state records


grvijayan:
## Context

The SSO ceremony relies on a single-use `state` record to bridge the initial submission and the provider callback, as the callback only carries the `code` and `state` parameters.

This ticket implements the foundational `state` record infrastructure. It covers the data shape, cryptographic minting, atomic consumption, and expiry.

The engine creates these records at submission in #1030 and consumes them at the callback in #1032.

## Scope

**1. Record Shape and Lookup**
*   **Lookup Key:** The `state` value sent to the provider. Store this value (or a hash of it, which prevents a database read from replaying a pending callback). The callback request carries no attempt reference and no flow cookie, so the record must be findable by this value alone.
*   **Payload Fields:** Per the [state record table](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/3-social-login-flow.md#the-state-record): flow/attempt ID, provider slug, connection revision ID, browser-binding nonce, PKCE verifier (if enabled), OIDC `nonce`, expiry (inheriting the attempt's TTL), and the return target.
**2. Minting & Limits**
*   **Generation:** `state` values must be generated using a CSPRNG with ≥128 bits of entropy. (Hardcoded placeholders like `state=sess_2_google` must never ship).
*   **Security Limits:** Record creation requires no authentication, but neither does any other flow interaction: creation caps and rate limits are a general flow-level protection, not an SSO-specific one, and are out of scope here. They belong to the platform's abuse-prevention design (#351, ADR in PR #472). The record TTL bounds growth in the meantime.

**3. Storage Architecture: State Records**

**Decision:** SSO state records will be stored in the `auth-attempts` database structure, specifically within the `checks` table ([`000008_checks.sql`](https://github.com/zitadel/nextgen/blob/3f3dd4312b52b90e3cb96d01826b488f55a1dcab/internal/storage/dialect/postgres/migration/sql/000008_checks.sql)). These rows belong to an attempt (or session) via foreign key and cascade-delete alongside it.

*   The table enforces one row per check type per attempt, so a repeat submission replaces the pending record instead of accumulating rows. Each row provides a `challenge_payload` slot, a `factor_payload` slot, and a `last_verified_at` marker.
[...]

Issue #1066, opened 2026-08-31
[GitHub][deferred] IdP engine - OAuth2 client construction and callback completion


grvijayan:
## Context

The engine's RP core (#1015) implements the OIDC half for Google; the OAuth2 half is deferred with GitHub, as the epic is Google-first. The schema still ships with the `oauth2` block and the API accepts oauth2 connections, so #1015 refuses an `oauth2` body with a typed error instead of building a client. This ticket lifts that refusal and completes the `oauth2` protocol block's slice of the RP core: client construction (#1015's half) and callback completion (#1027's half).

## Scope

*   Parse the `oauth2` block into the engine's view: `authorization_endpoint`, `token_endpoint`, `userinfo_endpoint`, the required `subject_claim` (OAuth2 has no `sub` guarantee), and the shared defaults (`client_secret_basic`, PKCE enabled).
*   Build the authorize URL; OAuth2 has no discovery, so construction makes no network request.
*   Complete the callback for an `oauth2` connection: no `id_token` step; claims are extracted from `userinfo` only; the subject is read from the required `subject_claim`.
*   Strategies (`supplementary_fetch`) run at the point #1027 defines; their semantics and implementations are #1029's.
*   Remove #1015's `oauth2` refusal.

## Acceptance Criteria

- [ ] An `oauth2` body loads with the engine defaults applied and produces a correct authorize URL with no network request.
- [ ] An `oauth2` body without `subject_claim` is rejected with a distinct error, re-checking what the schema enforces at write time.
- [ ] An `oauth2` connection completes a callback and returns the resolved identity, with claims from `userinfo` and the subject from `subject_claim`; both auth styles are asserted on the wire.

## Depends on

- #1015
- #1027

## Resources

*   [IdP resource model: The connection schema](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/1-resource-model.md#the-connection-schema)
*   [Social login flow: The Ceremony](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/3-social-login-flow.md#the-ceremony)

Issue #1030, opened 2026-08-28
Flow engine - SSO submission and `sso-redirect`


grvijayan:
## Context

Currently, the flow engine rejects all SSO submissions by returning `ErrFlowUnsupported` (`internal/domain/flow_state_machine.go`). This ticket replaces that placeholder behavior with the actual SSO ceremony initiation.

When a submission with `{action: "sso", sso_provider_id}` occurs, the engine will now validate the provider against the current step, pin the connection revision, create a single-use `state` record, set a browser-binding cookie, build the authorize URL, and return an `sso-redirect` step.

The `state` record infrastructure is handled in #1073, the callback consumption in #1032, and the URL generation in #1015.

## Scope

**1. Submission Dispatch**
*   **Validation:** A valid `{action: "sso", sso_provider_id: "<slug>"}` submission on a step carrying `sso_providers` replaces the `ErrFlowUnsupported` path. If the slug is absent from the step's `sso_providers`, it is rejected before any state record is created.
*   **Revision Pinning:** The slug is resolved to its current connection revision, and that exact ID is pinned to the record (the callback will never re-resolve the slug). Currently, this is the newest revision (#1002); once environments run releases, it will be the revision the environment's current release pins for the slug (ADR 035), and rolling back a release is how an older revision comes back into use.
*   **Unresolvable Slugs:** If a slug resolves to no connection, the attempt ends in the flow-level error surface (restart route) without building an authorize URL. Retrying on the same page will fail, and a fresh render drops the dead button per #1031. Reachable when a connection is deleted while a rendered page still offers it; what "deleted" means is #1013's decision.
*   **Environment Rules:** Submissions arriving under an `issuer_pattern` environment are refused before URL generation and routed to the flow-level error surface: the render never offers provider buttons there, so the submission comes from a stale or hand-built page.

**2. Creating the `state` Record**
*   **Execution:** The submission creates the record via #1073, passing the pinned connection revision ID, provider slug, browser-binding nonce, PKCE verifier (if enabled), OIDC `nonce`, and return target.
*   **Validation:** The return target, the destination the browser is sent to after callback processing, is captured at submission; how it is captured is open (see below). Its origin must equal the environment's issuer origin; a submission naming any other origin is rejected, and no record is created. The callback route reads the target from the record, never from callback input: an attacker-supplied target is an open redirect.

**3. The Binding Cookie**
[...]

Issue #1027, opened 2026-08-28
IdP engine - callback completion


grvijayan:
## Context

Following the authorization redirect built in #1015, this ticket implements the second half of the relying-party core: processing the provider's callback to yield a resolved external identity.

This covers steps 2 through 5 (and the first half of step 6) of the [Social login flow Callback Processing](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/3-social-login-flow.md#callback-processing):
*   Exchange the authorization code for tokens.
*   Validate the `id_token` (OIDC only).
*   Extract claims (from `id_token` or `userinfo`) and trigger any supplementary fetches.
*   Evaluate `verified_claims` rules.
*   Coerce the subject claim to a string.

**Boundaries:**
*   **Preceding step:** Looking up and consuming the `state` record (#1030).
*   **Succeeding step:** Identity lookup by `(connection, subject)` and outcome selection (#1035).
*   **Deferred:** The OAuth2 callback implementation is deferred alongside GitHub (#1066).

## Scope

**1. Inputs and Outputs**
*   **Inputs:** The parsed connection, authorization `code`, state-record expectations (`nonce`, PKCE verifier), the `redirect_uri` used at authorize, and the resolved `client_secret` (provided strictly for this execution; never persisted on any package type).
*   **Output:** The resolved external identity, consisting of the coerced subject, mapped claims, per-claim verification results, and the pinned connection revision.
*   *Constraint:* Provider tokens (Access, Refresh, ID) exist only during processing and must be dropped before returning.

**2. Code Exchange**
*   Executes against the connection's token endpoint via the shared egress client (#928).
*   Authenticates strictly per `token_endpoint_auth_method` (`client_secret_basic` in the header, `client_secret_post` in the body).
*   Sends the PKCE verifier (if originally used) and the exact `redirect_uri` from the authorize request.

**3. ID Token Validation (OIDC)**
*   **Verifier Configuration:** Built per-attempt using the fixed allowlist (`RS256`, `ES256`, `PS256`), the attempt's expected `nonce`, and engine-owned clock skew. `none` and HMAC algorithms are rejected at parse time.
*   **Assertions:** Validates signature, `iss`, `aud` (with `azp` for multiple audiences), `exp`, `iat`, and `nonce`. (A missing `nonce` echo fails the attempt if one was sent).
[...]

PR #1253, opened 2026-09-17
feat(idp): OIDC client construction and authorize redirect

### LINKED ISSUE #1015: IdP engine - OIDC client construction and authorize redirect
## Context

Currently, `nextgen` lacks an OIDC/OAuth2 relying party (RP) implementation. This ticket initiates that RP core as a pure, database-free package (`internal/idp`). It focuses on the construction phase: converting a pinned connection revision into a usable client and generating the authorize redirect to begin a sign-in.

By intentionally excluding database access, HTTP routes, and flow-machine wiring, this package creates clean, distinct integration seams for upcoming engine tickets (like callback processing and supplementary-fetch strategies).

This ticket implements the **OIDC half** (e.g., Google). The OAuth2 half (e.g., GitHub) is deferred to #1066.

## Scope

**1. Architecture & Dependencies**
*   **Mechanics:** Constructs the client through the `rp` package of `zitadel/oidc/v3`. The `id_token` verifier itself is built per attempt in #1027. This ticket fixes the engine's signing-algorithm allowlist and the parsed connection's key source, which #1027 consumes.
*   **Security:** The signing-algorithm allowlist is strictly fixed to `RS256`, `ES256`, and `PS256`. Tenant-authored discovery documents cannot override this.
*   **Egress:** All external network fetches (discovery, JWKS) must route through the egress policy established in #928 (blocking for #851 - **blocking to go live**).

**2. Connection Parsing (No DB Access)**
*   **Execution:** The caller reads the pinned revision from storage and passes the revision ID and body directly to the package.
*   **Secret:** `client_secret` is stored as a whole-value `${{ NAME }}` reference. Parsing keeps the reference unresolved and the authorize request needs no secret. The callback route resolves it through #1049's resolver and hands the value to the exchange (#1027).
*   **Engine Defaults:** Parses the stored revision into the engine's view, applying core defaults: `client_secret_basic` auth, PKCE enabled, and `sub` as the OIDC subject claim.
*   **Endpoint Resolution:** Each endpoint comes from its override when one is set. The engine fetches discovery only when an endpoint the current step needs has no override. The authorize step needs `authorization_endpoint`, so a connection that overrides it builds the authorize URL without a discovery fetch. When discovery is fetched, its `issuer` must equal the configured one. The callback step's endpoint set is defined in #1027.
*   **Validation Rejections:** Returns a typed error if the body cannot serve an attempt. This re-checks schema rules enforced at write time (#1003), as defense-in-depth (missing protocol block, OIDC scopes lacking `openid`, cleartext non-localhost endpoints).
*   **OAuth2 Deferred:** Explicitly refuses an `oauth2` body with a typed error. Because the schema still accepts OAuth2 connections, stored ones must fail cleanly here instead of attempting to build a client.

**3. Authorize Request**
*   **Inputs:** Parsed connection, `redirect_uri`, `state`, `nonce`, and a PKCE verifier.
*   **Outputs:** Authorize URL + expectations for the state-record seam (`state`, `nonce`, verifier, `redirect_uri`, revision ID, PKCE/nonce presence).
*   **Parameters:** The engine explicitly sets standard parameters (`client_id`, `response_type=code`, `scope`, etc.) on every call. `static_authorize_parameters` are appended only *after* filtering out reserved engine keys.
*   **Security:** PKCE strictly uses `S256` (no `plain` method allowed). OIDC always sends `nonce`. The package exposes helpers that generate `nonce` and the PKCE verifier from a CSPRNG with at least 128 bits of entropy. The `state` value is generated by the state record (#1073). The SSO submission (#1030) obtains all three, stores them in the record, and passes them to the authorize builder as inputs. The builder never generates its own.

**4. Error Handling**
*   **Typed Error Mapping:** Returns a single typed error. Its kind aligns with the *Failures and Recovery* table (provider denied, provider misconfigured, exchange/`userinfo` failure), plus one new kind: *discovery failure* (timeout, unparseable, issuer mismatch). It shares the user surface of *provider misconfigured*, a generic error on the originating step. It is a separate kind so server logs can tell an unreachable or malformed provider from a rejected client.
*   **Data Minimization:** User-facing text must be generic and devoid of tenant details. The wrapped cause is reserved exclusively for server logs.
*   **Security Constraint:** Error strings must **never** contain claim values, subjects, tokens, secrets, or URLs with query strings.

## Acceptance Criteria

- [ ] A body missing `token_endpoint_auth_method`, `pkce_enabled`, or `subject_claim` successfully loads with the engine defaults applied.
- [ ] A missing protocol block, OIDC scopes without `openid`, a cleartext non-localhost endpoint, and an `oauth2` body are each rejected with a distinct error.
- [ ] OIDC via discovery and OIDC with an `authorization_endpoint` override both produce a correct authorize URL. The override case performs no discovery fetch.
- [ ] Reserved keys within `static_authorize_parameters` are successfully dropped, ensuring the engine value is present in the URL.
- [ ] The engine's accepted signing algorithms are exactly `RS256`, `ES256`, and `PS256`. A fetched discovery document listing others (e.g., `HS256`) does not change the set.
- [ ] A discovery timeout, an unparseable document, and an issuer mismatch each correctly return the *discovery-failure* kind, with no URL query strings or secrets leaking in the error text.

## Out of scope

*   Caching discovery and JWKS documents with a bounded TTL (#1016, deferred). Until it ships, construction fetches discovery on every attempt start that needs it.

## Depends on

*   #1000
*   #928

## Resources

*   [Social login flow: The Ceremony](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/3-social-login-flow.md#the-ceremony) (state record, engine-owned parameters, entropy floor)
*   [Social login flow: Callback Processing](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/3-social-login-flow.md#callback-processing) (issuer check, allowlist)
*   [Social login flow: Server-Side Fetch Policy](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/3-social-login-flow.md#server-side-fetch-policy)
*   [Social login flow: Failures and Recovery](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/3-social-login-flow.md#failures-and-recovery)
*   [IdP resource model: The connection schema](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/1-resource-model.md#the-connection-schema) (endpoint overrides, TLS rule)
*   [IdP resource model: Immutability and Revisions](https://github.com/zitadel/nextgen/blob/main/docs/design/idp/1-resource-model.md#immutability-and-revisions)
*   [zitadel/oidc#944](https://github.com/zitadel/oidc/issues/944)

grvijayan:
## Summary

Closes #1015. Builds on the connection API from #1217, which introduced the domain file the new errors live in.

Adds `internal/idp`, the identity provider engine, as a pure package: it reads no storage and serves no routes. This PR ships the OIDC half of attempt start in three parts.

**1. Parsing a pinned revision.** `ParseConnection` takes the revision id and the raw document a caller read from storage and returns the engine's view with defaults applied: `client_secret_basic`, PKCE on, `sub` as the subject claim. The client secret stays an unresolved `${{ NAME }}` reference; the authorize request needs none and the callback resolves it per attempt. Four rules the schema already enforced at write time are re-checked as defense in depth, each with its own error code: a missing protocol block, OIDC scopes without `openid`, an endpoint that is not https and not on localhost, and an `oauth2` body, which the schema accepts but the engine cannot serve until #1066. One rule is the engine's own: the endpoint fields are either all empty or all set. A connection that names some of the endpoints it needs but not all is rejected, with the missing ones in the error details.

**2. Building the client.** `NewOIDCClient` builds a relying party from `zitadel/oidc/v3`'s `rp` package and picks the constructor by the connection's shape. With no endpoints set, the discovering constructor fetches the issuer's document through the http client the caller passes in, which must be the hardened egress client from ADR 061. The document must name the configured issuer, and the endpoints it exposes are checked against the same https pattern the schema enforces on stored ones. With every endpoint set, the static constructor takes them as given and no fetch happens; the engine builds the id_token verifier from the configured issuer and `jwks_uri`, since the static constructor has neither. Both modes hand the callback the same verifier: issuer, client id, key set, and the signing-algorithm allowlist, fixed to RS256, ES256, PS256. A discovery document never widens it.

**3. The authorize redirect.** `NewAuthorizeRedirect` takes the client plus state, nonce, and the PKCE verifier as inputs and returns the URL plus what the state record stores: state, nonce, verifier, redirect URI, and revision id. The engine sets every protocol parameter itself, PKCE is always S256, and static parameters from the connection are appended only after the reserved keys are dropped. The nonce and verifier helpers draw 256 bits from the CSPRNG.

## What this PR does not do

- **Nothing calls the package.** The SSO submission step in #1030 is the first caller, and the state record it writes is #1073.
- **No callback.** Code exchange, id_token verification, and claims extraction are #1027, which rebuilds the client through the same constructor and consumes the verifier and the allowlist. Nothing in this PR reads `SigningAlgorithms` outside construction.
- **No token endpoint auth.** `token_endpoint_auth_method` is parsed and defaulted here but read only by the callback in #1027, which sets the auth style on its own token config.
- **No OAuth2.** A stored `oauth2` body is refused with a typed error until #1066.
- **No discovery cache.** Construction fetches discovery on every attempt start in discovery mode until #1016.

## Implementation notes

- **Endpoint resolution deviates from the ticket.** The ticket resolves only `authorization_endpoint` at authorize, with overrides winning per field, and leaves the callback's endpoint set to #1027. This PR makes the endpoints all or nothing and resolves the full set at construction, because the callback cannot reuse the object built at authorize but should reuse the construction. One rule serves both steps, and a broken discovery document surfaces before the user is redirected instead of after they authenticated at the provider. Recorded as a comment on #1015.
- **Two library constructors, one per mode.** `NewRelyingPartyOIDC` always fetches the discovery document, so a provider without one needs `NewRelyingPartyOAuth`, and once every endpoint is configured there is nothing left to override. Discovery mode therefore uses the library's own discovery rather than reimplementing it. `golang.org/x/oauth2` becomes a direct dependency for the config type.
- **Schema descriptions follow the rule.** The OIDC block and its four endpoint fields described each endpoint as overriding discovery. They now say the endpoints are set together, for a provider without a .well-known document, and that a partial set is rejected. The generated json schema and the ogen comments change with them; no field, pattern, or constraint changes.
- **The discovered `jwks_uri` is not checked.** The library keeps it inside its key set and exposes no getter, so in discovery mode a cleartext or missing `jwks_uri` surfaces at the callback's first verification rather than at construction. The other discovered endpoints are re-checked. Closing the gap needs an endpoints getter on `rp.RelyingParty`, to be raised upstream.
- **Rejection messages stay literal.** The error schema generator reads each message as a string literal and silently skips anything else, so a message built from the field name has no schema in the API contract. The parse rejections therefore say "an endpoint is not https" and carry the field name in details for the client and in the log-only parent for the operator. Values and URLs never appear.
- **Dropping reserved keys is not optional.** `oauth2.Config.AuthCodeURL` applies extra parameters with `url.Values.Set`, so a static `state` or `nonce` passed through would overwrite the engine's and defeat the CSRF and token binding.
- **One error code per condition**, following ADR 030 and the json schema `fetch_*` precedent: `idp.discovery_failed`, `idp.protocol_block_missing`, `idp.scopes_missing_openid`, `idp.endpoint_cleartext`, `idp.endpoints_partial`, `idp.oauth2_unsupported`. The library includes the response body of a non-2xx discovery response in the discovery cause, so logs may carry tenant-served text; the user surface never does.

## Validation

- `go build ./...`, `go vet ./internal/idp/`, and `go test ./internal/idp/ ./internal/domain/`.

## Release notes

None. The package has no caller yet, so no behavior is user-visible. Empty changeset: the schema description change touches the json schema that `@zitadel/config` publishes under `meta-schemas/`, but nothing ships.


internal/idp/oidc.go
```diff
@@ -0,0 +1,178 @@
...
+// when there is none, and fails when neither names the endpoint.
+func overrideOrDiscovered(dst *string, discovered, name string) error {
+	if *dst != "" {
+		return nil
+	}
+	if discovered == "" {
+		return domain.ErrIDPDiscoveryFailed(fmt.Errorf("missing %s in discovery", name))
+	}
+	*dst = discovered
+	return nil
```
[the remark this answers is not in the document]
grvijayan: Fixed in 1bc8df3c. A discovered value now passes the same pattern the schema enforces on an override, and a document that names a cleartext or relative endpoint fails with `idp.discovery_failed` and a cause naming the field. Covered by a test row with an http authorization endpoint on a non-local host.

internal/idp/oidc.go
```diff
@@ -0,0 +1,178 @@
+// Package idp is the identity-provider engine: it turns a pinned connection
```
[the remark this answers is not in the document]
grvijayan: The changesets guard reports no changeset required for this diff, and the title guard accepts the type. Nothing here reaches a user: the package has no caller, and the five error schemas are not referenced by any endpoint, so the generated API and SDKs are unchanged. The release note comes with #1030, which wires the engine into the SSO step.

wim07101993:
In the issue you stated that we can use the zitadel/oidc pacakge. Why didn't you?

https://github.com/zitadel/nextgen/issues/1015#issuecomment-5700273282

grvijayan:
> In the issue you stated that we can use the zitadel/oidc pacakge. Why didn't you?
> 
> [#1015 (comment)](https://github.com/zitadel/nextgen/issues/1015#issuecomment-5700273282)

It does use zitadel/oidc ([example](https://github.com/zitadel/nextgen/pull/1253/changes#diff-94696c95617b99be18dee63395754df4dab2b07f2f4ed3b261221c6c8115a81fR13-R14)) 🤔 

The client is an `rp.RelyingParty` from `rp.NewRelyingPartyOAuth`, discovery is `client.Discover` and so on.

I deliberately did not use `NewRelyingPartyOIDC` (if this is your question) because it always does discovery, which we want to avoid in case there are endpoint overrides. Instead, [`rp.NewRelyingPartyOAuth`](https://github.com/zitadel/nextgen/pull/1253/changes#diff-94696c95617b99be18dee63395754df4dab2b07f2f4ed3b261221c6c8115a81fR104-R111) is used as it allows overriding endpoints.

internal/idp/authorize.go
```diff
@@ -0,0 +1,116 @@
...
+	"request_uri",
+	"client_secret",
+	"client_assertion",
+}
+
+// Authorize builds the authorize URL for one attempt. The engine sets every
+// protocol parameter itself: client_id, redirect_uri, response_type=code,
+// scope, state, nonce, and an S256 code challenge when PKCE is enabled. The
+// connection's static parameters follow, minus the reserved keys.
+func (c *OIDCClient) Authorize(req AuthorizeRequest) (AuthorizeRedirect, error) {
```
wim07101993: I find the name `Authorize` a bit misleading. The function creates a url to which the server has to redirect the user. But the function itself does not authorize, unless I am missing something.

wim07101993: Would it maybe be clearer to have a `NewAuthorizeRedirect` function which takes the `OIDCClient` as an argument and treat it as a constructor?

wim07101993: Why does the function take an `AuthorizeRequest`? As far as I can see, it only takes data which will be random generated anyway. Couldn't the `Authorize` function handle that as well?

internal/idp/oidc.go
```diff
@@ -0,0 +1,185 @@
...
+	// id_tokens the same way in both modes.
+	verifier := rp.NewIDTokenVerifier(conn.OIDC.Issuer, conn.OIDC.ClientID,
+		rp.NewRemoteKeySet(httpClient, conn.OIDC.JWKSURI),
+		rp.WithSupportedSigningAlgorithms(SigningAlgorithms()...))
+	return &OIDCClient{conn: conn, redirectURI: redirectURI, party: party, verifier: verifier, userinfo: conn.OIDC.UserinfoEndpoint}, nil
+}
+
+// validDiscovered fails when a discovery document omits a necessary endpoint or
+// names one that is not https.
+func validDiscovered(name, value string) error {
```
wim07101993: a nit, but maybe rename it to `validateEndpoint`? `validDiscovered` can mean anything to the reader

internal/idp/authorize.go
grvijayan: Just a design choice so the caller creates state, nonce, and the PKCE verifier, stores them in the state record first (the callback validates against this later), and then passes them in. This way the redirect URL is built from what was stored rather than the record being filled from the URL. State will anyway be created in the flow-engine (#1030), so it made sense to create nonce and the PKCE verifier together.

internal/idp/oidc.go
grvijayan: done in https://github.com/zitadel/nextgen/pull/1253/commits/c4fab3d4cdd30e7e13eae6e0147980142aad8a54

internal/idp/authorize.go
grvijayan: done in https://github.com/zitadel/nextgen/pull/1253/commits/c4fab3d4cdd30e7e13eae6e0147980142aad8a54

wim07101993 (approved):

PR #1256, opened 2026-09-18
fix(api): let an identifier proof name the attribute it identifies

mridang:
## Anything that is not a web page cannot log in

`POST /auth_attempts/{id}/challenges/{id}/verify` is how a client proves who a user is without rendering a login page — a CLI, a native app, a script. It takes a login name, but no way to say what that login name *is*. Is `alice@acme.com` an email? A username? Projects choose differently, so the server does not guess, and the request has no field to tell it. The lookup matches nobody and every identifier proof is rejected.

The login page never hits this, because it knows which box it drew and passes the attribute along with the value ([`flow_auth_attempt.go:50`](https://github.com/zitadel/nextgen/blob/main/internal/service/flow_auth_attempt.go#L50)). The REST handler builds the same struct and leaves that field empty ([`auth_attempt.go:137`](https://github.com/zitadel/nextgen/blob/main/internal/api/auth_attempt.go#L137)), so the query runs against attribute key `""`:

```go
// before — one half of a two-field struct
service.UserProof{ LoginName: p.GetLoginName() }

// after
service.UserProof{ LoginName: p.GetLoginName(), AttributeName: p.GetAttributeName() }
```

`IdentifierProof` gains `attribute_name` so a caller can fill it in.

## What the API looks like now

Signing a user in, end to end, with no browser and no rendered step:

```http
POST /auth_attempts
Authorization: Bearer <publishable key>
{ "project_id": "proj_platform" }
→ 201 { "attempt_id": "att_1", "state": "in_progress" }

POST /auth_attempts/att_1/challenges
{ "method": "identifier" }
→ 201 { "challenge_id": "ch_id", "method": "identifier", "state": "pending" }

POST /auth_attempts/att_1/challenges/ch_id/verify
{ "login_name": "alice@acme.com", "attribute_name": "email" }
→ 200 { "attempt_id": "att_1", "completed_factors": [ { "method": "identifier", … } ] }

POST /auth_attempts/att_1/challenges
{ "method": "password" }
→ 201 { "challenge_id": "ch_pw", "method": "password", "state": "pending" }

POST /auth_attempts/att_1/challenges/ch_pw/verify
{ "password": "correct-horse-battery-staple" }
→ 200 { "attempt_id": "att_1", "completed_factors": [ …, { "method": "password", … } ] }

POST /auth_attempts/att_1/handoff
→ 200 { "handoff_token": "handoff_zHzpUg…", "expires_at": "…" }
```

The handoff token is then exchanged at `POST /sessions/exchange` for a session, exactly as the login page does at the end of a flow. Only the third call changes in this PR: `attribute_name` is new.

## Open question, which is why this is a draft

The server takes the attribute from the caller. Review pointed out that this is more permissive than the login flow: the flow raises an identifier challenge only for the property the schema *designates* as the identifier (`x-identifier`), and [ADR 058 §5](https://github.com/zitadel/nextgen/blob/main/docs/adrs/058-user-identity-designation-and-references.md) explicitly retired the "any `x-unique` property can identify" rule. As written here, a project that designates `username` but also marks `phone` unique for data integrity would find `phone` usable as a login identifier over REST.

The alternative is for the server to resolve the designation itself and ignore the caller's choice. That matches the flow, removes the widening, and drops the request-contract change altogether — at the cost of wiring a schema lookup into the attempt service, which has none today. I would like an opinion from whoever owns auth before taking that on.

## A second, unrelated defect

Found while testing the first, and separable on request. The `completed_factors[].payload` union is discriminated on `method`, so the server writes that property into every payload — but the variants never declared it and `PasswordFactorPayload` set `additionalProperties: false`, so a generated client fails to decode a completed password factor:

```
decode PasswordFactorPayload: unexpected field "method"
```

The variants now declare the discriminator. Nothing on the wire changed; the schema simply now admits what the server already sent.

## Breaking change

`attribute_name` and `method` are required, so a client generated from the old spec no longer satisfies the new one. Nothing could have depended on either — the identifier proof resolved nobody and the password payload never decoded — but a generated client still needs regenerating. No caller of these endpoints exists in this repository: the console, the login surface and the SDKs all use `/flow`.

## Testing

`internal/api/integration_test/auth_attempt_password_test.go` drives the sequence above and asserts that a proof naming the wrong attribute resolves nobody. Run against Postgres with the rest of the integration suite, including the existing password-flow, passkey and registration tests that cover the login path this does not touch.


.changeset/auth-attempt-identifier-proof.md
```diff
@@ -0,0 +1,6 @@
+---
+"@zitadel/server": patch
+"@zitadel/api": patch
```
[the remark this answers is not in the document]
mridang: Agreed, and fixed in 44fb4937c: both packages are now a major bump, and the changeset spells out both breaks — a request built against the old schema omits `attribute_name`, and a decoder generated from it rejects the payload's `method`.

Worth saying what the mitigation is and is not: neither shape could have had a working consumer (the identifier proof resolved nobody, and a completed password factor failed to decode), but a client generated from the old spec still needs regenerating, which is exactly what a major communicates.

This PR is a draft pending a redesign that may remove the incompatibility altogether: review found that letting the caller name any `x-unique` attribute is more permissive than the login flow, which identifies only by the schema's designated `x-identifier` (ADR 058 §5 retired the 'any unique property can identify' rule). If the server derives the attribute from that designation instead of accepting it from the caller, the request contract stops changing and the widening goes away. I will re-version accordingly if that lands.

mridang:
Folded into #1225 as 7cdc3ecec, so the CLI can sign the local admin in through this API instead of driving the login flow. The description above still explains the change; the open question — whether the server should take the identifying attribute from the caller or resolve the schema's designated `x-identifier` itself — carries over to #1225's review.

api/openapi/components/checks/identifier-factor-payload.yaml
```diff
@@ -2,8 +2,12 @@ title: IdentifierFactorPayload
 type: object
 description: Identifier-specific factor metadata
 required:
+  - method
   - user_id
 properties:
+  method:
+    $ref: ../schemas/factor-method.yaml
+    description: Always `identifier`; the discriminator of the payload union.
```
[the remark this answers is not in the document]
mridang: Fixed in 0151ccd4c. Each variant's `method` is now a single-value enum (`[identifier]`, `[password]`, `[passkey]`) instead of the shared `FactorMethod` reference. I tried `const` first, but ogen accepts it and silently drops it: no validator is generated, so the schema would have claimed a constraint the server does not enforce. The single-value enum generates a dedicated type whose `Validate()` rejects any other method.

api/openapi/components/checks/passkey-factor-payload.yaml
```diff
@@ -2,9 +2,13 @@ title: PasskeyFactorPayload
 type: object
 description: Passkey-specific factor metadata including credential and verification details
 required:
+  - method
   - credential_id
   - user_verified
 properties:
+  method:
+    $ref: ../schemas/factor-method.yaml
+    description: Always `passkey`; the discriminator of the payload union.
```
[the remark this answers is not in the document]
mridang: Fixed in 0151ccd4c. Each variant's `method` is now a single-value enum (`[identifier]`, `[password]`, `[passkey]`) instead of the shared `FactorMethod` reference. I tried `const` first, but ogen accepts it and silently drops it: no validator is generated, so the schema would have claimed a constraint the server does not enforce. The single-value enum generates a dedicated type whose `Validate()` rejects any other method.

api/openapi/components/checks/password-factor-payload.yaml
```diff
@@ -1,4 +1,12 @@
...
+description: |
+  Password authentication carries no metadata beyond the discriminator the
+  `payload` union resolves on.
 additionalProperties: false
+required:
+  - method
+properties:
+  method:
+    $ref: ../schemas/factor-method.yaml
+    description: Always `password`; the discriminator of the payload union.
```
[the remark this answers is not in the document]
mridang: Fixed in 0151ccd4c. Each variant's `method` is now a single-value enum (`[identifier]`, `[password]`, `[passkey]`) instead of the shared `FactorMethod` reference. I tried `const` first, but ogen accepts it and silently drops it: no validator is generated, so the schema would have claimed a constraint the server does not enforce. The single-value enum generates a dedicated type whose `Validate()` rejects any other method.

api/openapi/endpoints/auth_attempts/by_id/challenges/by_id/identifier-proof.yaml
```diff
@@ -1,12 +1,23 @@
...
+  Proof for `identifier` method.
+
+  A project decides for itself which attributes identify a user — a login flow
+  step collects `email`, `username`, `phone`, or several of them — so the proof
+  names the attribute the value belongs to rather than the server assuming one.
+  The attribute must be registered unique for the project; a value matching only
+  a non-unique property resolves nobody.
 required:
   - login_name
+  - attribute_name
```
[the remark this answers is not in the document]
mridang: Agreed, and reworked in 177339c1a to follow ADR 058 §5. `attribute_name` is gone, so the request is back to `login_name` only and no longer changes. The service now resolves a bare login name against the designated identifier (`x-identifier`) of every user schema in the project. Each lookup is scoped to that schema's users and to uniquely registered values, and the name must identify exactly one user: none, or users of more than one schema, rejects the proof. There is no precedence. The flow path is unchanged, since it still names the property from the step it rendered.

The schema listing runs as a login resolve (`WithAuthzListUnrestricted`), the same way user-ref resolution lists these schemas. There is no signed-in caller yet whose grants could narrow it.

New integration coverage: identification through the designated property; a value held only in an undesignated unique property matching nobody; and one value designated by two schemas for different users being rejected. The rejections assert `att.proof_rejected` specifically. I also checked that removing the ambiguity check fails only the ambiguity test, and that ignoring the designation fails the other two. The changeset is now a patch, since the request contract no longer changes.

.changeset/auth-attempt-identifier-proof.md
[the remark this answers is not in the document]
mridang: Follow-up: this is back to a patch as of 177339c1a, and the reason for the major bump no longer applies. Following ADR 058 §5 removed `attribute_name`, so the request contract is identical to `main` again. What remains is a fix: the identifier proof now resolves instead of always rejecting, and the response schema admits the `method` discriminator the server always sent. Neither breaks a client generated from the old spec.

internal/service/auth_attempt.go
```diff
@@ -594,6 +601,82 @@ func (s *authAttemptService) buildChallenge(ctx context.Context, attempt *domain
...
+		if match != nil && match.ID != user.ID {
+			return nil, errAmbiguousIdentifier
+		}
+		match = user
+	}
+	if match == nil {
+		return nil, errNoIdentifierMatch
+	}
+	return match, nil
+}
```
IAM-marco: Simplification: this function re-implements a resolver the grant service already has.

What changed: resolveDesignatedIdentifier lists every user schema in the project, reads each schema's x-identifier, runs one GetUser per schema against uniquely registered values, and rejects on zero matches or on matches for different users.

Why this location matters: the grant service does the same job in resolveUserByIdentifier plus designatedIdentifierKeys:
https://github.com/zitadel/nextgen/blob/177339c1a8b533e523abe99b35c1c7f5f43e6529/internal/service/grant.go#L365-L438
Same listing helper (listUserSchemas), same designation reader (DesignatedIdentifier), same UniqueAttributesOnly lookup, same exactly-one-user rule. The only differences: the grant version groups schema URLs by property name, so it issues one query per distinct property instead of one per schema, and it only considers active users.

What follows: two copies of the ADR 058 §5 resolution rule now live in this package. The next change to that rule (the cross-key namespace the ADR lists under Revisit, or a status filter) has to be made twice, and the two can drift silently. They already differ on the active-user filter.

Kept separate: whether a login should also skip inactive users is its own decision (#553 tracks status enforcement). Worth deciding explicitly rather than inheriting one answer or the other from whichever copy gets edited.

Suggested direction: lift the grant version into one package-level helper (it reads no GrantService state) that takes the statements, the project id and the value, and returns the user or a not-found/ambiguous error. Call it from both services.

Coverage: the three new integration tests would keep passing after the extraction, so they are the safety net for it.

internal/service/auth_attempt.go
```diff
@@ -594,6 +601,82 @@ func (s *authAttemptService) buildChallenge(ctx context.Context, attempt *domain
 	}
 }
 
+var (
+	errNoIdentifierMatch   = errors.New("no user is identified by the login name")
+	errAmbiguousIdentifier = errors.New("the login name identifies users of more than one schema")
+)
```
IAM-marco: Simplification: two new sentinel errors that nothing tells apart.

What changed: errNoIdentifierMatch and errAmbiguousIdentifier are declared here and each returned once from resolveDesignatedIdentifier.

Why this location matters: the only caller is verify, which wraps whatever comes back in domain.ErrAuthAttemptProofRejected:
https://github.com/zitadel/nextgen/blob/177339c1a8b533e523abe99b35c1c7f5f43e6529/internal/service/auth_attempt.go#L690-L695
That error keeps the cause as a log-only parent (ADR 030), so the client sees the same att.proof_rejected either way, and no code path branches on which sentinel it was.

What follows: two names for one outcome. If the shared resolver from the comment above lands, the grant service's existing unresolved sentinel already covers both cases; if not, one sentinel is enough. If the goal is a distinguishable log line, a single error whose text names the reason does the same job.

Suggested direction: collapse to one sentinel, or drop both in favour of the shared helper's error.

internal/service/auth_attempt.go
```diff
@@ -594,6 +601,82 @@ func (s *authAttemptService) buildChallenge(ctx context.Context, attempt *domain
...
+// resolveIdentifier finds the user a login name identifies: by the named
+// property when the flow supplied one, otherwise by the project's designated
+// identifiers.
+func (s *authAttemptService) resolveIdentifier(ctx context.Context, projectID string, p UserProof) (*domain.User, error) {
+	if p.AttributeName != "" {
+		return s.users.GetByAttributes(ctx, projectID, []domain.Attribute{{
+			Key:   domain.AttributeKey(p.AttributeName),
+			Value: p.LoginName,
+		}})
+	}
```
IAM-marco, line 622: Spec note: the flow branch still looks users up project-wide, not within the bound schema.

What changed: resolveIdentifier now has two branches. The bare-name branch below scopes each lookup to one schema URL. This branch, taken when the flow supplied an attribute name, calls GetByAttributes, which filters by project and attribute only and never by schema:
https://github.com/zitadel/nextgen/blob/177339c1a8b533e523abe99b35c1c7f5f43e6529/internal/service/user.go#L866-L875

Why this location matters: ADR 058 §5 says an identifier lookup matches "on users of the schema in scope" and that this scoping "is what makes the flow path fully collision-proof". The PR description keeps the flow path out of scope, and this behaviour predates the PR, so I am not asking for the change here. But the two branches now sit side by side with different scoping, and the new UserProof doc comment reads as if the ADR's flow-path rule were already in force.

What follows concretely: a designated property must be project-unique, so one key cannot hold the same value twice under project scope. It can under team scope: another schema declares the same property name, undesignated, with x-unique set to team. A user there registers the value a legitimate user holds under the flow's schema. The flow lookup then finds two rows, GetUser returns a multiple-rows error, and the legitimate login is rejected. Narrow, but exactly the collision the ADR says the scoping prevents.

Suggested direction: open a follow-up that passes the bound schema URL through the flow path so this branch can scope the same way, and reference it from the doc comment. Or soften the doc comment so it does not promise what the flow path does not do yet.

Coverage: nothing exercises the flow path with a colliding team-unique value; the new tests only cover the bare-name branch.

.changeset/auth-attempt-identifier-proof.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+The auth-attempt API can now identify a user, so a client that renders no login step can sign one in. `POST /auth_attempts/{id}/challenges/{id}/verify` resolved an identifier proof against an attribute with an empty name, which matched nobody, so every proof was rejected. It now resolves the login name as ADR 058 §5 defines: against the designated identifier (`x-identifier`) of each user schema in the project, among that schema's users and uniquely registered values only, and it must identify exactly one user — none, or users of more than one schema, rejects the proof. The request is unchanged; the caller never names the property. The completed-factor payloads in an attempt response now also declare the `method` they are discriminated on, pinned per variant, which the server always sent but the schema forbade.
```
IAM-marco: Guideline note: the changeset summary cites an ADR section.

What changed: the summary says the login name is resolved "as ADR 058 §5 defines".

Why this location matters: the changeset summary is the release note, and the changeset README says release notes carry no ADR numbers, PR numbers or internal file paths; those belong in the PR description:
https://github.com/zitadel/nextgen/blob/177339c1a8b533e523abe99b35c1c7f5f43e6529/.changeset/README.md#L78-L90

What follows: the note ships to package consumers who cannot open the ADR, and the release tooling does not rewrite it.

Suggested direction: keep the behaviour description and drop the citation, for example "against the identifier each user schema designates". The PR description already carries the ADR link.

IAM-marco (changes requested):
Two-axis review of 177339c1: correctness (code review) and over-engineering (ponytail). Nothing here repeats what earlier threads already settled (attribute_name removed, method pinned per variant, changeset back to patch).

## Code review
- Spec note: the flow branch of resolveIdentifier still looks users up project-wide, while the new bare-name branch scopes to one schema. ADR 058 §5 asks for the scoping on both paths. Pre-existing, so a tracked follow-up is enough (inline comment).
- Guideline note: the changeset summary cites an ADR section; the changeset README says release notes carry none (inline comment).

## Ponytail
- Simplification: resolveDesignatedIdentifier re-implements the grant service's designated identifier resolver; one shared helper serves both (inline comment).
- Simplification: two new sentinel errors that no caller distinguishes (inline comment). About 55 lines removable in total.

## Notes that have no diff line to sit on
- The PR title and the first commit still describe the design that 177339c1 reverted (the caller naming the attribute). The squash-merge title will say the opposite of what the code does; please retitle before merging.
- ADR 058 §5 opens with "its current implementation looks the user up with an empty attribute key". After this PR that sentence is stale. A one-line touch-up (here or in a follow-up) keeps the ADR honest.
- Coverage: the bare-name branch talks to the statements directly, not through the UserLookup port that the unit tests in auth_attempt_test.go mock, so only the postgres/spanner integration lane exercises it. Fine as it is, just worth knowing when reading test output.

Tally: code review 2 findings (worst: unscoped flow branch); ponytail 2 findings (worst: duplicated resolver).

internal/service/auth_attempt.go
mridang: Done in db3c50ca7. The grant version is now the one shared helper, `resolveDesignatedUser` in `internal/service/identifier.go`, with `designatedIdentifierKeys` lifted out of `GrantService` alongside it. Both services call it, so the auth attempt now also gets the per-property grouping and the explicit multiple-rows handling. The status filter is an explicit argument rather than a shared default: grants pass `activeUsersOnly`, and login passes `anyUserStatus` to match the flow path, with a comment leaving that decision to status enforcement. As you predicted, the three auth-attempt tests are the safety net, and `TestGrantCreateLocators` joins them: I checked that ignoring the designation in the helper fails both the login tests and the grant locator test.

internal/service/auth_attempt.go
mridang: Done in db3c50ca7. Both sentinels are gone. The shared helper returns the existing `errIdentifierUnresolved` for a miss and for an ambiguous match, and records which one it was as a log line, as the grant path already did.

internal/service/auth_attempt.go
mridang: Softened in db3c50ca7 rather than changed, since the flow path's lookup predates this PR. The `UserProof` comment no longer implies the flow path is scoped, and `resolveIdentifier` now says plainly that it matches the property across the project, not within the bound schema as §5 asks, and that this is left for a follow-up. I have not opened the follow-up issue myself; that is for the author to decide.

.changeset/auth-attempt-identifier-proof.md
mridang: Done in db3c50ca7. The summary now reads "against the identifier each user schema designates (`x-identifier`)", with no ADR citation. The ADR link stays in the PR description.

mridang:
Thanks for the review. The four inline points are answered in their threads. On the notes without a diff line:

- **Title:** retitled to "fix(api): resolve identifier proofs by the project's designated identifiers", so the squash title matches the code. The first commit's message still describes the reverted design; the branch is squash-merged, so that message never reaches `main`.
- **ADR 058 §5:** the stale opening sentence is corrected in db3c50ca7. It now says the first implementation looked the user up with an empty attribute key and that it now resolves as the section describes.
- **Coverage:** noted. The designated-identifier path talks to the statements directly, so only the postgres/spanner lanes exercise it. Those lanes carry the three auth-attempt tests plus `TestGrantCreateLocators`, which now covers the same helper from the grant side.

internal/service/auth_attempt.go
IAM-marco: Follow-up filed as #1276: scope the login flow identifier lookup to the bound user schema (ADR 058 section 5).

mridang: Thanks. Nothing to change here: the flow path stays project-wide in this PR, and #1276 tracks scoping it.

IAM-marco (approved):
All four points from the earlier review are addressed: one shared designated-identifier resolver serves grants and login, the extra sentinels are gone, the changeset no longer cites the ADR, and the flow-path scoping gap is documented in the code and tracked in #1276. full-pr is green on the merged head 96fefa3a.

